PR #401, opened 2026-06-21
fix(docker): run repo start by default

### LINKED ISSUE #383: Docker image default CMD `gh-symphony start` runs a removed command and crash-loops the documented headless deploy
## Problem
The Docker image's default command invokes a removed CLI command. `Dockerfile:64` is `CMD ["gh-symphony", "start"]`, but `start` was turned into a removed/redirect command (`packages/cli/src/index.ts:373-378` via `registerRemovedCommand`). Running it prints `Use 'gh-symphony repo start' from the target repository.` to stderr and exits with code 2 (reproduced against the built dist).

## Impact (OSS user)
A user following the README's recommended `docker run -d --restart unless-stopped` (README.md:108-114) or `docker compose` deployment (README.md:119-131, `restart: unless-stopped`, no `command:`) gets a container whose main process exits 2 immediately and then crash-loops forever. The advertised headless deployment does not work out of the box. README.md:62 even claims the default command is `gh-symphony repo start` (with `repo`), contradicting the actual CMD.

## Evidence (real files)
- `Dockerfile:63-64`: `ENTRYPOINT ["/usr/bin/tini", "--"]` + `CMD ["gh-symphony", "start"]`.
- `packages/cli/src/index.ts:373-378`: `start` registered as removed; handler in `packages/cli/src/commands/removed-command.ts` exits 2.
- README.md:62 advertises `default command: gh-symphony repo start`; README.md:108-131 deploy examples set no `command:` and use a restart policy.
- `.github/workflows/release.yml` smoke test runs `gh-symphony repo init && gh-symphony repo start --once` and `gh-symphony --version` explicitly, so it never exercises the default CMD and never caught this.

## Proposed fix
Change the Dockerfile CMD to a working invocation (e.g. `CMD ["gh-symphony", "repo", "start"]`, matching README.md:62) and update the README deploy examples to reflect that it still requires an initialized repo/working dir. Add a release smoke step that runs the bare image with its default CMD so this regresses loudly.

---
_Priority: P0 · Symphony layer: DX/Meta · category: packaging_
_Filed from an OSS-user-perspective audit of the repo._

moncher-dev:
## TL;DR

- Fixes the official Docker image default command so it runs `gh-symphony repo start` instead of the removed `gh-symphony start`.
- Updates Docker deployment docs to make the initialized repository working directory requirement explicit.
- Adds a release smoke check that starts the published image without overriding CMD and fails if the default process exits immediately.

## 변경 지점 다이어그램

```text
Dockerfile CMD
  gh-symphony start          (removed, exits 2)
        ↓
  gh-symphony repo start     (current repository orchestrator entrypoint)

Release smoke
  explicit --version / repo start --once
        +
  default CMD detached container must stay running briefly
```

## 여기부터 보세요

- `Dockerfile`: image default CMD now matches the documented `gh-symphony repo start` path.
- `README.md`: docker run and compose examples now say the mounted working directory must already be initialized and set `working_dir: /repo` for compose.
- `.github/workflows/release.yml`: published-image smoke now starts the image with no command override and asserts the default process keeps running.

## 위험 & 롤백

- Risk is limited to container default startup behavior and documentation. Users who explicitly pass a command are unaffected.
- Rollback is the single Dockerfile CMD line plus README/release-smoke edits in this PR.

## 변경 파일

- `Dockerfile`
- `README.md`
- `.github/workflows/release.yml`

## Evidence

- `pnpm install --frozen-lockfile` — pass
- `pnpm lint` — pass
- `pnpm test` — pass
- `pnpm typecheck` — pass
- `pnpm build` — pass
- `git diff --check` — pass
- Local Docker smoke was not run because the Docker daemon is unavailable in this environment: `Cannot connect to the Docker daemon at unix:///Users/steve/.docker/run/docker.sock.`

## 머지 후/사람 확인

- [ ] Confirm the README Docker deployment wording matches the intended supported self-hosted usage.
- [ ] Confirm the release workflow smoke behaves as expected when the next published image is built in GitHub Actions.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -154,6 +154,25 @@ jobs:
             "${IMAGE_REF}" \
             sh -lc 'cd /var/lib/gh-symphony/repo && gh-symphony repo init && gh-symphony repo start --once'
 
+          default_container="$(docker run -d \
+            -v "${TMPDIR}/config:/var/lib/gh-symphony" \
+            -e GH_SYMPHONY_FILE_TRACKER_ISSUES_PATH=/var/lib/gh-symphony/issues.json \
+            -w /var/lib/gh-symphony/repo \
+            "${IMAGE_REF}")"
+          trap 'docker rm -f "${default_container}" >/dev/null 2>&1 || true' EXIT
+          sleep 5
+
+          if [ "$(docker inspect -f '{{.State.Running}}' "${default_container}")" != "true" ]; then
+            docker logs "${default_container}" || true
+            status="$(docker inspect -f '{{.State.ExitCode}}' "${default_container}")"
+            echo "Expected the image default CMD to keep the orchestrator running; got exit ${status}."
+            exit 1
+          fi
+
+          docker stop "${default_container}"
+          docker rm "${default_container}"
+          trap - EXIT
+
       - name: Tag and create GitHub Release
         if: steps.changesets.outputs.published == 'true'
         env:
--- a/README.md
+++ b/README.md
@@ -102,7 +102,9 @@ docker run --rm -it \
   gh-symphony setup --non-interactive
 ```
 
-Then start the long-running orchestrator:
+Then start the long-running orchestrator from the initialized repository. The image
+default command is `gh-symphony repo start`, so the mounted working directory must
+already contain `WORKFLOW.md` and the repository runtime config created by setup.
 
 ```bash
 docker run -d \
@@ -121,14 +123,15 @@ services:
   gh-symphony:
     image: ghcr.io/hojinzs/github-symphony:latest
     restart: unless-stopped
+    working_dir: /repo
     environment:
       GITHUB_GRAPHQL_TOKEN: ${GITHUB_GRAPHQL_TOKEN}
     volumes:
-      - gh-symphony-data:/var/lib/gh-symphony
-
-volumes:
-  gh-symphony-data:
+      - ./:/repo
 ```
+
+Run `gh-symphony setup` once before starting the service so the mounted repository
+has `WORKFLOW.md` and `.runtime/orchestrator/`.
 
 If you prefer a host bind mount in `docker compose`, align the container user with the host directory owner:
 
@@ -136,10 +139,12 @@ If you prefer a host bind mount in `docker compose`, align the container user wi
 services:
   gh-symphony:
     image: ghcr.io/hojinzs/github-symphony:latest
+    working_dir: /repo
     user: "${UID:-1000}:${GID:-1000}"
     environment:
       GITHUB_GRAPHQL_TOKEN: ${GITHUB_GRAPHQL_TOKEN}
     volumes:
+      - ./:/repo
       - ./data:/var/lib/gh-symphony
 ```
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -61,4 +61,4 @@ VOLUME ["/var/lib/gh-symphony"]
 USER symphony
 
 ENTRYPOINT ["/usr/bin/tini", "--"]
-CMD ["gh-symphony", "start"]
+CMD ["gh-symphony", "repo", "start"]


README.md
```diff
@@ -121,25 +123,28 @@ services:
...
     restart: unless-stopped
+    working_dir: /repo
     environment:
       GITHUB_GRAPHQL_TOKEN: ${GITHUB_GRAPHQL_TOKEN}
     volumes:
-      - gh-symphony-data:/var/lib/gh-symphony
-
-volumes:
-  gh-symphony-data:
+      - ./:/repo
```
hojinzs, line 130: **Nit (non-blocking): config-dir handling is now inconsistent between the two compose examples.**

This first example dropped the `/var/lib/gh-symphony` mount (and the named volume) entirely, whereas the bind-mount example just below still mounts `./data:/var/lib/gh-symphony`. Because the Dockerfile declares `VOLUME ["/var/lib/gh-symphony"]` and sets `GH_SYMPHONY_CONFIG_DIR=/var/lib/gh-symphony`, this example will write CLI/config state into an *anonymous* volume that isn't persisted predictably across `docker compose down -v`.

For the documented `repo start` path the runtime state lives under `.runtime/orchestrator/` in the mounted repo, so this likely works — but the asymmetry between the two examples may confuse self-hosters. Consider either keeping a named `gh-symphony-data:/var/lib/gh-symphony` volume here too, or adding a one-line note that config-dir persistence is optional for `repo start`.

---

.github/workflows/release.yml
```diff
@@ -154,6 +154,25 @@ jobs:
             "${IMAGE_REF}" \
             sh -lc 'cd /var/lib/gh-symphony/repo && gh-symphony repo init && gh-symphony repo start --once'
 
+          default_container="$(docker run -d \
```
hojinzs, line 157: Good — this default-CMD assertion is exactly what issue #383 asked for, and it correctly reuses the already-initialized repo at `/var/lib/gh-symphony/repo`. Verified locally that the old `gh-symphony start` exits 2 (the crash-loop cause) while `gh-symphony repo start` resolves, so this guard would catch the regression.

**Suggestion (non-blocking): also add this check to the per-PR `Container Smoke` job in `.github/workflows/ci.yml`.** This `release.yml` smoke only runs on publish, so a future regression of the `CMD` line still wouldn't be caught at PR time — which is precisely the blind spot #383 describes (`it never exercises the default CMD and never caught this`). `ci.yml` already builds `gh-symphony:test` and initializes the repo, so dropping the same `docker run -d` + `State.Running` assertion there would make the default CMD "regress loudly" on every PR rather than only at release.

Minor: the trailing `docker stop`/`docker rm` (lines 172-173) are redundant with the `EXIT` trap, but harmless.

---

hojinzs (approved):
## ✅ Approve — PR #401 fully addresses issue #383

### Requirement / acceptance-criteria coverage (issue #383)
| # | Required | Status |
|---|----------|--------|
| 1 | Dockerfile `CMD` → a working invocation (`gh-symphony repo start`) | ✅ `Dockerfile:64` |
| 2 | Update README deploy examples to reflect the initialized-repo / working-dir requirement | ✅ `README.md` (run + both compose examples set `working_dir: /repo`, bind `./:/repo`, plus a "run setup first" note) |
| 3 | Add a release smoke step that runs the bare image with its default CMD so it regresses loudly | ✅ `release.yml` (detached default-CMD container + `State.Running` assertion) |

### Smoke test (performed)
A full container build was **not possible in this review environment** — Docker registry pulls are blocked by the network policy (`403 Forbidden` resolving `node:24-bookworm-slim`), the same limitation the author hit. So I smoke-tested the substance of the fix against the locally built CLI:

- `gh-symphony start` → **exit 2**, `Use 'gh-symphony repo start' from the target repository.` — reproduces the crash-loop root cause.
- `gh-symphony repo start --help` → **exit 0**, valid command — the new CMD path the Dockerfile now points at.

CI is green on the head commit: `Test` ✅ and `Container Smoke` ✅.

### Review notes (all non-blocking)
1. **`ci.yml` Container Smoke doesn't exercise the default CMD** — the new default-CMD guard lives only in `release.yml` (publish-only). A future `CMD` regression would still slip past per-PR CI, which is the exact blind spot #383 names. Cheap to mirror into `ci.yml`. (inline)
2. **README compose examples are now asymmetric** about the `/var/lib/gh-symphony` config-dir mount. Likely fine for the `repo start` path, but worth a clarifying note. (inline)
3. Minor: redundant `docker stop`/`docker rm` after the `EXIT` trap in the release smoke. (inline)

The change is correct, minimal, well-scoped, and free of over-engineering. The suggestions above can be folded in here or tracked as a follow-up — they don't block merge.

---

PR #536, opened 2026-08-06
fix(security): bound HTTP and polling inputs

moncher-dev:
## TL;DR

- `POST /api/v1/refresh` body를 64 KiB로 제한하고, Content-Length 및 chunked 초과 요청을 즉시 413으로 완료합니다.
- Linear pagination을 최대 page 수로 제한하고, 각 page의 fetch와 response body 파싱 전체에 타임아웃을 적용합니다.
- workflow polling 설정 입력을 1초–5분으로 제한하되 기존 적응형 rate-limit 백오프는 보존합니다.

## 변경 지점 다이어그램

```text
HTTP refresh stream ──> 64 KiB guard ──> 413 + connection close
Linear page request ──> maxPages ──> fetch + body timeout
WORKFLOW interval ──> input clamp ──> adaptive backoff ──> scheduler wait
```

## 여기부터 보세요

- `packages/control-plane/src/server.ts`: 초과 body 즉시 종료와 스트림 오류 응답
- `packages/tracker-linear/src/orchestrator-adapter.ts`: bounded pagination과 body 포함 timeout
- `packages/orchestrator/src/service.ts`: 설정 입력 clamp와 adaptive 결과 보존 경계
- 각 패키지의 `*.test.ts`: chunked body, stalled Linear body, 60초 base → 600초 backoff 회귀 TC

## 위험 & 롤백

- 64 KiB를 넘는 refresh client는 413과 connection close를 받습니다. 기존 정상 호출은 의미 있는 body를 보내지 않습니다.
- 1초 미만/5분 초과 polling 설정은 repository policy에 따라 각각 1초/5분으로 동작합니다. 이 upstream spec divergence는 changeset에 명시했습니다.
- 롤백: rebased head `052d712`의 두 #468 커밋을 되돌립니다.

## 변경 파일

- `packages/control-plane/src/server.ts`, `server.test.ts`
- `packages/tracker-linear/src/orchestrator-adapter.ts`, `tracker-linear.test.ts`
- `packages/orchestrator/src/service.ts`, `service.test.ts`
- `.changeset/bright-limits-refresh.md`

## Evidence

- `pnpm lint` — 통과
- `pnpm test` — 통과 (재실행에서 전체 workspace 통과; 첫 실행의 기존 SIGTERM timing flake는 격리 재실행도 통과)
- `pnpm typecheck` — 통과
- `pnpm build` — 통과
- 영향 패키지 TC — control-plane 36/36, tracker-linear 25/25, orchestrator 234/234 통과
[...]

PR #534, opened 2026-08-05
fix(core): harden workspace path containment

moncher-dev:
## TL;DR

- Workspace containment now uses platform-native `path.relative` checks.
- Existing symlinks and missing descendant paths are resolved through the nearest existing real path before containment is checked.
- Dangling symlink components are detected with `lstat` and rejected fail-closed.
- Added traversal, symlink escape, Windows-style separator, and regression tests.

## 변경 지점 다이어그램

```text
workspace/safety.ts ─┐
                     ├─> workspace/path-safety.ts ─> realpath + path.relative containment
workspace/identity.ts┘
                     └─> workspace-safety/core-conformance tests
```

## 여기부터 보세요

- `packages/core/src/workspace/path-safety.ts`
- `packages/core/src/workspace/safety.ts`
- `packages/core/src/workspace/identity.ts`
- `packages/core/src/workspace-safety.test.ts`
- `packages/core/src/core-conformance.test.ts`

## Issues

- Fixes #528

## Summary

Workspace paths are rejected when traversal or symlink resolution escapes the configured root, without hardcoding `/` as the separator.

## Changes

- Centralized realpath-aware containment validation for existing paths and non-existing descendant tails.
- Fail-closed handling for dangling symlink components prevents a later-created target from redirecting I/O outside the root.
- Applied the validator to both workspace directory resolvers.
- Added required path safety regression coverage, including dangling symlink cases, and retained the CLI patch changeset.

## Evidence

- `pnpm lint` — passed
- `pnpm -r --workspace-concurrency=1 test` — passed (all workspace packages; core 208 tests, orchestrator 233 tests, worker 152 tests, CLI 497 tests)
- `pnpm --filter @gh-symphony/orchestrator exec vitest run src/index.test.ts -t "gracefully shuts down on SIGTERM"` — passed in isolation
- `pnpm typecheck` — passed
- `pnpm build` — passed
- `pnpm exec prettier --check <changed files>` — passed
- `pnpm --filter @gh-symphony/core test` — passed (210 tests, including both dangling symlink regressions)
- GitHub CI `Test` and `Container Smoke` — passed

## 위험 & 롤백

- Risk: realpath resolution adds filesystem metadata reads during workspace path validation.
- Rollback: revert commits `d23e536` and `22dda93` to restore the previous containment implementation.

## 변경 파일

- `.changeset/528-workspace-path-safety.md`
- `packages/core/src/workspace/path-safety.ts`
- `packages/core/src/workspace/safety.ts`
[...]

PR #547, opened 2026-08-10
fix(cli): report stopped daemon status

### LINKED ISSUE #546: fix(cli): repo status renders stale snapshot as live health — add daemon liveness check
## Background

`gh-symphony repo status` reads the on-disk `status.json` snapshot and renders it as if it were the current daemon state, without checking whether the daemon process is actually alive. In a real incident, a daemon that had been down for 22 days still rendered as a healthy green `● idle` with `Last tick 528h ago`:

```
➜ gh-symphony repo stop
No running daemon found for project "repository" (PID file missing).
➜ gh-symphony repo status
  ● Health    idle              Last tick  528h ago
```

The operator only discovers the daemon is dead if they happen to run `repo stop` or notice the oversized hour count. The health value comes from `buildStatusSnapshot` (`packages/core/src/observability/snapshot-builder.ts:65`), which derives health purely from snapshot contents (`lastError` / `activeRuns`) — there is no "daemon not running" state at the CLI surface.

## Symptoms / Reproduction Steps

1. `gh-symphony repo start`, wait for at least one tick so `status.json` exists.
2. Kill the orchestrator process and delete the PID file (or reboot the machine).
3. `gh-symphony repo status` → still shows green `● idle` with a stale `Last tick`, no indication the daemon is down.

## Proposed Changes

1. **Extract a shared daemon-liveness helper** (`packages/cli/src/daemon-liveness.ts`): move `validateDaemonProcess` and the PID-file/project-lock resolution logic out of `packages/cli/src/commands/stop.ts` so `status` and `stop` share one implementation.
2. **`repo status` checks liveness before rendering**: when no live daemon is found, render a `● stopped` (red/yellow) health line plus a hint: `daemon not running — run 'gh-symphony repo start'`. The last snapshot may still be shown, clearly labeled as last known state.
3. **Staleness guard**: if `lastTickAt` is older than a threshold (e.g. 3× the configured tick interval) while the daemon appears alive, annotate the tick line with a stale warning.
4. **`relativeTime` renders days**: values ≥ 48h display as `22d ago` instead of `528h ago` (`packages/cli/src/commands/status.ts:27`).

Applies to both the legacy single-shot renderer and the `--watch` dashboard path; `--json` output should gain a `daemonRunning: boolean` field.

## Affected Files

| File | Change | Scope |
|------|--------|-------|
| `packages/cli/src/daemon-liveness.ts` | New shared helper (PID file read + process identity/cwd validation) | new (~60 lines) |
| `packages/cli/src/commands/status.ts` | Liveness check, stopped/stale rendering, day-unit `relativeTime`, `daemonRunning` in JSON | ~80 lines |
| `packages/cli/src/commands/stop.ts` | Replace inline `validateDaemonProcess`/lock resolution with shared helper | ~-50 lines |

**Estimate**: 3h

## Verification

1. `pnpm --filter @gh-symphony/cli test` — new TCs: status with missing PID file → stopped banner; stale PID → stopped banner; live daemon → unchanged output; `relativeTime(49h)` → `2d ago`.
2. Manual: start daemon → `repo status` shows running/idle as today; kill daemon + remove PID file → `repo status` shows `● stopped` with restart hint; `repo status --json` includes `daemonRunning: false`.
3. `pnpm lint && pnpm typecheck && pnpm build`

moncher-dev:
## Issues — Closed #546

- Fixes #546

## TL;DR

- `repo status` verifies daemon liveness before presenting a persisted snapshot as current health.
- The implementation is rebased onto current `main`; stale warnings use the daemon's tracker-derived effective polling interval.

## 변경 지점 다이어그램

```text
tracker rate limits ─► effectivePollIntervalMs ─► status snapshot ─► stale threshold
daemon.pid / project lock ─► daemon-liveness.ts ───────────────────► repo status / repo stop
```

## 여기부터 보세요

- `packages/cli/src/daemon-liveness.ts` centralizes PID, lock, CWD, heartbeat, and identity validation.
- `packages/cli/src/commands/status.ts` renders text, watch, and JSON status from independent liveness and snapshot state.
- `packages/orchestrator/src/service.ts` persists the tracker-derived effective polling interval for the CLI stale guard.

## 위험 & 롤백

- Legacy snapshots retain rate-limit inference until a daemon writes `effectivePollIntervalMs`.
- Rollback: revert the #546 commits to restore the previous CLI status presentation.

## 변경 파일

- `.changeset/wise-dolphins-repair.md`
- CLI daemon-liveness, status, stop, dashboard, and regression tests.
- Core snapshot contract and orchestrator polling snapshot persistence.

## Evidence

- `pnpm exec vitest run packages/cli/src/commands/status.test.ts packages/orchestrator/src/service.test.ts` — passed.
- `pnpm lint && pnpm test && pnpm typecheck && pnpm build` — passed on `7cc2a4d0`.
- Docker E2E blackbox: rebuilt `symphony-e2e`; `/healthz` returned `{"ok":true}`; container and network were removed.
- GitHub CI `Test` and `Container Smoke` — passed on `7cc2a4d0`.

## 머지 후/사람 확인

- [ ] Confirm a running daemon retains the familiar idle/running presentation.
- [ ] Confirm a killed daemon shows stopped health, restart guidance, and last-known state.
- [ ] Confirm stale warnings follow a rate-limited daemon's effective polling cadence.
- [ ] Confirm `repo status --watch` remains readable at the reviewer’s terminal width.

### DIFF
--- a/packages/cli/src/commands/status.test.ts
+++ b/packages/cli/src/commands/status.test.ts
@@ -1,4 +1,4 @@
-import { mkdtemp, mkdir, writeFile } from "node:fs/promises";
+import { mkdtemp, mkdir, readFile, writeFile } from "node:fs/promises";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
@@ -8,11 +8,16 @@ import type { CliProjectConfig } from "../config.js";
 const daemonProcessMocks = vi.hoisted(() => ({
   getProcessCwd: vi.fn(),
   getProcessIdentity: vi.fn(),
+  resolveAdaptivePollIntervalMs: vi.fn(
+    (basePollIntervalMs: number) => basePollIntervalMs
+  ),
 }));
 
 vi.mock("@gh-symphony/orchestrator", () => ({
   getProcessCwd: daemonProcessMocks.getProcessCwd,
   getProcessIdentity: daemonProcessMocks.getProcessIdentity,
+  resolveAdaptivePollIntervalMs:
+    daemonProcessMocks.resolveAdaptivePollIntervalMs,
 }));
 
 function captureWrites(stream: NodeJS.WriteStream): {
@@ -168,6 +173,10 @@ async function writeProjectLock(
 beforeEach(() => {
   daemonProcessMocks.getProcessCwd.mockReset();
   daemonProcessMocks.getProcessIdentity.mockReset();
+  daemonProcessMocks.resolveAdaptivePollIntervalMs.mockReset();
+  daemonProcessMocks.resolveAdaptivePollIntervalMs.mockImplementation(
+    (basePollIntervalMs: number) => basePollIntervalMs
+  );
   daemonProcessMocks.getProcessCwd.mockReturnValue(join("/tmp", "tenant-a"));
   daemonProcessMocks.getProcessIdentity.mockReturnValue("live-daemon");
 });
@@ -175,6 +184,7 @@ beforeEach(() => {
 afterEach(() => {
   daemonProcessMocks.getProcessCwd.mockReset();
   daemonProcessMocks.getProcessIdentity.mockReset();
+  daemonProcessMocks.resolveAdaptivePollIntervalMs.mockReset();
   vi.restoreAllMocks();
   process.exitCode = undefined;
 });
@@ -357,6 +367,41 @@ describe("status command", () => {
     expect(stdout.output()).toContain("● Health    running");
     expect(stdout.output()).not.toContain("stopped");
     expect(stdout.output()).not.toContain("stale");
+  });
+
+  it("uses the adaptive rate-limit poll interval for stale detection", async () => {
+    const configDir = await createConfigFixture();
+    await writeDaemonPid(configDir);
+    await writeFile(
+      join(configDir, "status.json"),
+      JSON.stringify({
+        ...JSON.parse(await readFile(join(configDir, "status.json"), "utf8")),
+        rateLimits: { limit: 5_000, remaining: 0 },
+      }) + "\n",
+      "utf8"
+    );
+    daemonProcessMocks.resolveAdaptivePollIntervalMs.mockReturnValue(300_000);
+    vi.spyOn(process, "kill").mockImplementation(() => true);
+    vi.spyOn(Date, "now").mockReturnValue(
+      new Date("2026-03-30T11:02:00.000Z").getTime()
+    );
+    const stdout = captureWrites(process.stdout);
+
+    try {
+      await statusCommand([], {
+        configDir,
+        verbose: false,
+        json: false,
+        noColor: true,
+      });
+    } finally {
+      stdout.restore();
+    }
+
+    expect(stdout.output()).not.toContain("(stale)");
+    expect(
+      daemonProcessMocks.resolveAdaptivePollIntervalMs
+    ).toHaveBeenCalledWith(30_000, { limit: 5_000, remaining: 0 });
   });
 
   it("adds daemonRunning to JSON output", async () => {
--- a/packages/cli/src/commands/status.test.ts
+++ b/packages/cli/src/commands/status.test.ts
@@ -1,9 +1,19 @@
 import { mkdtemp, mkdir, writeFile } from "node:fs/promises";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
-import { afterEach, describe, expect, it, vi } from "vitest";
-import statusCommand from "./status.js";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import statusCommand, { relativeTime } from "./status.js";
 import type { CliProjectConfig } from "../config.js";
+
+const daemonProcessMocks = vi.hoisted(() => ({
+  getProcessCwd: vi.fn(),
+  getProcessIdentity: vi.fn(),
+}));
+
+vi.mock("@gh-symphony/orchestrator", () => ({
+  getProcessCwd: daemonProcessMocks.getProcessCwd,
+  getProcessIdentity: daemonProcessMocks.getProcessIdentity,
+}));
 
 function captureWrites(stream: NodeJS.WriteStream): {
   output: () => string;
@@ -120,12 +130,157 @@ async function createConfigFixture(
   return configDir;
 }
 
+async function writeDaemonPid(
+  configDir: string,
+  options: { pid?: number; identity?: string; cwd?: string } = {}
+): Promise<void> {
+  await writeFile(
+    join(configDir, "projects", "tenant-a", "daemon.pid"),
+    JSON.stringify({
+      pid: options.pid ?? 111,
+      startedAt: "2026-03-30T11:00:00.000Z",
+      processIdentity: options.identity ?? "live-daemon",
+      cwd: options.cwd ?? join("/tmp", "tenant-a"),
+    }) + "\n",
+    "utf8"
+  );
+}
+
+beforeEach(() => {
+  daemonProcessMocks.getProcessCwd.mockReset();
+  daemonProcessMocks.getProcessIdentity.mockReset();
+  daemonProcessMocks.getProcessCwd.mockReturnValue(join("/tmp", "tenant-a"));
+  daemonProcessMocks.getProcessIdentity.mockReturnValue("live-daemon");
+});
+
 afterEach(() => {
+  daemonProcessMocks.getProcessCwd.mockReset();
+  daemonProcessMocks.getProcessIdentity.mockReset();
   vi.restoreAllMocks();
   process.exitCode = undefined;
 });
 
 describe("status command", () => {
+  it("renders a stopped banner and restart hint when the PID file is missing", async () => {
+    const configDir = await createConfigFixture();
+    const stdout = captureWrites(process.stdout);
+
+    try {
+      await statusCommand([], {
+        configDir,
+        verbose: false,
+        json: false,
+        noColor: true,
+      });
+    } finally {
+      stdout.restore();
+    }
+
+    expect(stdout.output()).toContain("● Health    stopped");
+    expect(stdout.output()).toContain(
+      "daemon not running — run 'gh-symphony repo start'"
+    );
+    expect(stdout.output()).toContain("Last known state: running");
+  });
+
+  it("renders a stopped banner for a stale PID", async () => {
+    const configDir = await createConfigFixture();
+    await writeDaemonPid(configDir, { pid: 999_999 });
+    vi.spyOn(process, "kill").mockImplementation(() => {
+      throw new Error("ESRCH");
+    });
+    const stdout = captureWrites(process.stdout);
+
+    try {
+      await statusCommand([], {
+        configDir,
+        verbose: false,
+        json: false,
+        noColor: true,
+      });
+    } finally {
+      stdout.restore();
+    }
+
+    expect(stdout.output()).toContain("● Health    stopped");
+  });
+
+  it("keeps live daemon health and annotates a stale last tick", async () => {
+    const configDir = await createConfigFixture();
+    await writeDaemonPid(configDir);
+    vi.spyOn(process, "kill").mockImplementation(() => true);
+    vi.spyOn(Date, "now").mockReturnValue(
+      new Date("2026-03-30T11:02:00.000Z").getTime()
+    );
+    const stdout = captureWrites(process.stdout);
+
+    try {
+      await statusCommand([], {
+        configDir,
+        verbose: false,
+        json: false,
+        noColor: true,
+      });
+    } finally {
+      stdout.restore();
+    }
+
+    expect(stdout.output()).toContain("● Health    running");
+    expect(stdout.output()).toContain("Last tick  2m ago (stale)");
+    expect(stdout.output()).not.toContain("daemon not running");
+  });
+
+  it("keeps fresh live daemon output free of stopped and stale warnings", async () => {
+    const configDir = await createConfigFixture();
+    await writeDaemonPid(configDir);
+    vi.spyOn(process, "kill").mockImplementation(() => true);
+    vi.spyOn(Date, "now").mockReturnValue(
+      new Date("2026-03-30T11:00:30.000Z").getTime()
+    );
+    const stdout = captureWrites(process.stdout);
+
+    try {
+      await statusCommand([], {
+        configDir,
+        verbose: false,
+        json: false,
+        noColor: true,
+      });
+    } finally {
+      stdout.restore();
+    }
+
+    expect(stdout.output()).toContain("● Health    running");
+    expect(stdout.output()).not.toContain("stopped");
+    expect(stdout.output()).not.toContain("stale");
+  });
+
+  it("adds daemonRunning to JSON output", async () => {
+    const configDir = await createConfigFixture();
+    const stdout = captureWrites(process.stdout);
+
+    try {
+      await statusCommand([], {
+        configDir,
+        verbose: false,
+        json: true,
+        noColor: true,
+      });
+    } finally {
+      stdout.restore();
+    }
+
+    expect(JSON.parse(stdout.output())).toMatchObject({
+      health: "running",
+      daemonRunning: false,
+    });
+  });
+
+  it("renders durations of at least 48 hours in days", () => {
+    const now = new Date("2026-04-01T12:00:00.000Z").getTime();
+    expect(relativeTime("2026-03-30T11:00:00.000Z", now)).toBe("2d ago");
+  });
+
   it("prints JSON for missing runtime config when --json is set", async () => {
     const configDir = await mkdtemp(join(tmpdir(), "cli-status-missing-"));
     const stdout = captureWrites(process.stdout);
--- a/packages/cli/src/commands/status.test.ts
+++ b/packages/cli/src/commands/status.test.ts
@@ -146,7 +146,11 @@ async function writeDaemonPid(
   );
 }
 
-async function writeProjectLock(configDir: string, pid = 222): Promise<void> {
+async function writeProjectLock(
+  configDir: string,
+  pid = 222,
+  cwd?: string
+): Promise<void> {
   await writeFile(
     join(configDir, "projects", "tenant-a", ".lock"),
     JSON.stringify({
@@ -155,6 +159,7 @@ async function writeProjectLock(configDir: string, pid = 222): Promise<void> {
       startedAt: "2026-03-30T11:00:00.000Z",
       heartbeatAt: "2026-03-30T11:00:00.000Z",
       processIdentity: null,
+      ...(cwd ? { cwd } : {}),
     }) + "\n",
     "utf8"
   );
@@ -222,6 +227,32 @@ describe("status command", () => {
       health: "running",
       daemonRunning: true,
     });
+  });
+
+  it("uses the foreground daemon CWD recorded in the project lock", async () => {
+    const configDir = await createConfigFixture();
+    const foregroundCwd = join("/tmp", "foreground-daemon");
+    await writeProjectLock(configDir, 222, foregroundCwd);
+    daemonProcessMocks.getProcessCwd.mockReturnValue(foregroundCwd);
+    daemonProcessMocks.getProcessIdentity.mockReturnValue(null);
+    vi.spyOn(process, "kill").mockImplementation(() => true);
+    vi.spyOn(Date, "now").mockReturnValue(
+      new Date("2026-03-30T11:00:30.000Z").getTime()
+    );
+    const stdout = captureWrites(process.stdout);
+
+    try {
+      await statusCommand([], {
+        configDir,
+        verbose: false,
+        json: true,
+        noColor: true,
+      });
+    } finally {
+      stdout.restore();
+    }
+
+    expect(JSON.parse(stdout.output())).toMatchObject({ daemonRunning: true });
   });
 
   it("recognizes a live foreground daemon when the PID file is malformed", async () => {
--- a/packages/cli/src/commands/status.test.ts
+++ b/packages/cli/src/commands/status.test.ts
@@ -404,6 +404,41 @@ describe("status command", () => {
     ).toHaveBeenCalledWith(30_000, { limit: 5_000, remaining: 0 });
   });
 
+  it("uses the daemon's recorded tracker polling interval for stale detection", async () => {
+    const configDir = await createConfigFixture();
+    await writeDaemonPid(configDir);
+    await writeFile(
+      join(configDir, "status.json"),
+      JSON.stringify({
+        ...JSON.parse(await readFile(join(configDir, "status.json"), "utf8")),
+        effectivePollIntervalMs: 300_000,
+        rateLimits: { source: "codex", limit: 5_000, remaining: 5_000 },
+      }) + "\n",
+      "utf8"
+    );
+    vi.spyOn(process, "kill").mockImplementation(() => true);
+    vi.spyOn(Date, "now").mockReturnValue(
+      new Date("2026-03-30T11:02:00.000Z").getTime()
+    );
+    const stdout = captureWrites(process.stdout);
+
+    try {
+      await statusCommand([], {
+        configDir,
+        verbose: false,
+        json: false,
+        noColor: true,
+      });
+    } finally {
+      stdout.restore();
+    }
+
+    expect(stdout.output()).not.toContain("(stale)");
+    expect(
+      daemonProcessMocks.resolveAdaptivePollIntervalMs
+    ).not.toHaveBeenCalled();
+  });
+
   it("adds daemonRunning to JSON output", async () => {
     const configDir = await createConfigFixture();
     const stdout = captureWrites(process.stdout);
--- a/packages/cli/src/commands/status.test.ts
+++ b/packages/cli/src/commands/status.test.ts
@@ -1,4 +1,4 @@
-import { mkdtemp, mkdir, readFile, writeFile } from "node:fs/promises";
+import { mkdtemp, mkdir, readFile, rm, writeFile } from "node:fs/promises";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
@@ -454,6 +454,32 @@ describe("status command", () => {
         code: "missing_repository_runtime_config",
         message:
           "No repository runtime config found. Run 'gh-symphony repo init' first.",
+      },
+    });
+  });
+
+  it("includes daemonRunning when JSON status has no snapshot", async () => {
+    const configDir = await createConfigFixture();
+    await rm(join(configDir, "status.json"));
+    const stdout = captureWrites(process.stdout);
+
+    try {
+      await statusCommand([], {
+        configDir,
+        verbose: false,
+        json: true,
+        noColor: true,
+      });
+    } finally {
+      stdout.restore();
+    }
+
+    expect(process.exitCode).toBe(1);
+    expect(JSON.parse(stdout.output())).toEqual({
+      daemonRunning: false,
+      error: {
+        code: "status_snapshot_unavailable",
+        message: "Unable to read status snapshot.",
       },
     });
   });
--- a/packages/cli/src/commands/status.test.ts
+++ b/packages/cli/src/commands/status.test.ts
@@ -461,6 +461,8 @@ describe("status command", () => {
   it("includes daemonRunning when JSON status has no snapshot", async () => {
     const configDir = await createConfigFixture();
     await rm(join(configDir, "status.json"));
+    await writeDaemonPid(configDir);
+    vi.spyOn(process, "kill").mockImplementation(() => true);
     const stdout = captureWrites(process.stdout);
 
     try {
@@ -476,12 +478,49 @@ describe("status command", () => {
 
     expect(process.exitCode).toBe(1);
     expect(JSON.parse(stdout.output())).toEqual({
-      daemonRunning: false,
+      daemonRunning: true,
       error: {
         code: "status_snapshot_unavailable",
         message: "Unable to read status snapshot.",
       },
     });
+  });
+
+  it("preserves live daemon liveness in non-TTY watch JSON without a snapshot", async () => {
+    const configDir = await createConfigFixture();
+    await rm(join(configDir, "status.json"));
+    await writeDaemonPid(configDir);
+    vi.spyOn(process, "kill").mockImplementation(() => true);
+    vi.spyOn(process, "exit").mockImplementation((() => undefined) as never);
+    const isTTYDescriptor = Object.getOwnPropertyDescriptor(
+      process.stdout,
+      "isTTY"
+    );
+    Object.defineProperty(process.stdout, "isTTY", {
+      configurable: true,
+      value: false,
+    });
+    const stdout = captureWrites(process.stdout);
+
+    try {
+      void statusCommand(["--watch"], {
+        configDir,
+        verbose: false,
+        json: false,
+        noColor: true,
+      });
+      await vi.waitFor(() => {
+        expect(JSON.parse(stdout.output())).toEqual({ daemonRunning: true });
+      });
+      process.emit("SIGTERM", "SIGTERM");
+    } finally {
+      stdout.restore();
+      if (isTTYDescriptor) {
+        Object.defineProperty(process.stdout, "isTTY", isTTYDescriptor);
+      } else {
+        delete (process.stdout as { isTTY?: boolean }).isTTY;
+      }
+    }
   });
 
   it("renders project tokens from the flat runtime status snapshot", async () => {
--- a/packages/cli/src/commands/status.test.ts
+++ b/packages/cli/src/commands/status.test.ts
@@ -224,6 +224,38 @@ describe("status command", () => {
     });
   });
 
+  it("recognizes a live foreground daemon when the PID file is malformed", async () => {
+    const configDir = await createConfigFixture();
+    await writeFile(
+      join(configDir, "projects", "tenant-a", "daemon.pid"),
+      "partially-written-pid-record\n",
+      "utf8"
+    );
+    await writeProjectLock(configDir);
+    daemonProcessMocks.getProcessIdentity.mockReturnValue(null);
+    vi.spyOn(process, "kill").mockImplementation(() => true);
+    vi.spyOn(Date, "now").mockReturnValue(
+      new Date("2026-03-30T11:00:30.000Z").getTime()
+    );
+    const stdout = captureWrites(process.stdout);
+
+    try {
+      await statusCommand([], {
+        configDir,
+        verbose: false,
+        json: true,
+        noColor: true,
+      });
+    } finally {
+      stdout.restore();
+    }
+
+    expect(JSON.parse(stdout.output())).toMatchObject({
+      health: "running",
+      daemonRunning: true,
+    });
+  });
+
   it("renders a stopped banner for a stale PID", async () => {
     const configDir = await createConfigFixture();
     await writeDaemonPid(configDir, { pid: 999_999 });
--- a/packages/cli/src/commands/status.test.ts
+++ b/packages/cli/src/commands/status.test.ts
@@ -146,6 +146,20 @@ async function writeDaemonPid(
   );
 }
 
+async function writeProjectLock(configDir: string, pid = 222): Promise<void> {
+  await writeFile(
+    join(configDir, "projects", "tenant-a", ".lock"),
+    JSON.stringify({
+      ownerToken: `${pid}:owner`,
+      pid,
+      startedAt: "2026-03-30T11:00:00.000Z",
+      heartbeatAt: "2026-03-30T11:00:00.000Z",
+      processIdentity: null,
+    }) + "\n",
+    "utf8"
+  );
+}
+
 beforeEach(() => {
   daemonProcessMocks.getProcessCwd.mockReset();
   daemonProcessMocks.getProcessIdentity.mockReset();
@@ -181,6 +195,33 @@ describe("status command", () => {
       "daemon not running — run 'gh-symphony repo start'"
     );
     expect(stdout.output()).toContain("Last known state: running");
+  });
+
+  it("recognizes a live foreground daemon from the project lock", async () => {
+    const configDir = await createConfigFixture();
+    await writeProjectLock(configDir);
+    daemonProcessMocks.getProcessIdentity.mockReturnValue(null);
+    vi.spyOn(process, "kill").mockImplementation(() => true);
+    vi.spyOn(Date, "now").mockReturnValue(
+      new Date("2026-03-30T11:00:30.000Z").getTime()
+    );
+    const stdout = captureWrites(process.stdout);
+
+    try {
+      await statusCommand([], {
+        configDir,
+        verbose: false,
+        json: true,
+        noColor: true,
+      });
+    } finally {
+      stdout.restore();
+    }
+
+    expect(JSON.parse(stdout.output())).toMatchObject({
+      health: "running",
+      daemonRunning: true,
+    });
   });
 
   it("renders a stopped banner for a stale PID", async () => {
--- a/packages/cli/src/commands/status.ts
+++ b/packages/cli/src/commands/status.ts
@@ -60,8 +60,11 @@ type RuntimeStatus = {
 
 async function resolveStaleThresholdMs(
   workspaceDir: string,
-  rateLimits: Record<string, unknown> | null | undefined
+  snapshot: ProjectStatusSnapshot
 ): Promise<number> {
+  if (snapshot.effectivePollIntervalMs !== undefined) {
+    return snapshot.effectivePollIntervalMs * 3;
+  }
   try {
     const resolution = await workflowConfigStore.load(
       join(workspaceDir, "WORKFLOW.md")
@@ -69,14 +72,14 @@ async function resolveStaleThresholdMs(
     return (
       resolveAdaptivePollIntervalMs(
         resolution.workflow.polling.intervalMs,
-        rateLimits ?? null
+        snapshot.rateLimits ?? null
       ) * 3
     );
   } catch {
     return (
       resolveAdaptivePollIntervalMs(
         DEFAULT_POLL_INTERVAL_MS,
-        rateLimits ?? null
+        snapshot.rateLimits ?? null
       ) * 3
     );
   }
@@ -106,7 +109,7 @@ async function resolveRuntimeStatus(options: {
   }
   const staleThresholdMs = await resolveStaleThresholdMs(
     options.workspaceDir,
-    options.snapshot.rateLimits
+    options.snapshot
   );
   return {
     daemonRunning: true,
--- a/packages/cli/src/commands/status.ts
+++ b/packages/cli/src/commands/status.ts
@@ -18,6 +18,7 @@ import { clearScreen, showCursor, hideCursor } from "../ansi.js";
 import { renderDashboard } from "../dashboard/renderer.js";
 import { formatRepositoryDisplay } from "../format/repository.js";
 import { resolveDaemonLiveness } from "../daemon-liveness.js";
+import { resolveAdaptivePollIntervalMs } from "@gh-symphony/orchestrator";
 
 const workflowConfigStore = new WorkflowConfigStore();
 
@@ -57,14 +58,27 @@ type RuntimeStatus = {
   lastTickStale: boolean;
 };
 
-async function resolveStaleThresholdMs(workspaceDir: string): Promise<number> {
+async function resolveStaleThresholdMs(
+  workspaceDir: string,
+  rateLimits: Record<string, unknown> | null | undefined
+): Promise<number> {
   try {
     const resolution = await workflowConfigStore.load(
       join(workspaceDir, "WORKFLOW.md")
     );
-    return resolution.workflow.polling.intervalMs * 3;
+    return (
+      resolveAdaptivePollIntervalMs(
+        resolution.workflow.polling.intervalMs,
+        rateLimits ?? null
+      ) * 3
+    );
   } catch {
-    return DEFAULT_POLL_INTERVAL_MS * 3;
+    return (
+      resolveAdaptivePollIntervalMs(
+        DEFAULT_POLL_INTERVAL_MS,
+        rateLimits ?? null
+      ) * 3
+    );
   }
 }
 
@@ -84,12 +98,19 @@ async function resolveRuntimeStatus(options: {
   snapshot: ProjectStatusSnapshot;
 }): Promise<RuntimeStatus> {
   const liveness = await resolveDaemonLiveness(options);
-  const staleThresholdMs = await resolveStaleThresholdMs(options.workspaceDir);
+  if (!liveness.running) {
+    return { daemonRunning: false, lastTickStale: false };
+  }
+  const staleThresholdMs = await resolveStaleThresholdMs(
+    options.workspaceDir,
+    options.snapshot.rateLimits
+  );
   return {
-    daemonRunning: liveness.running,
-    lastTickStale:
-      liveness.running &&
-      isLastTickStale(options.snapshot.lastTickAt, staleThresholdMs),
+    daemonRunning: true,
+    lastTickStale: isLastTickStale(
+      options.snapshot.lastTickAt,
+      staleThresholdMs
+    ),
   };
 }
 
--- a/packages/cli/src/commands/status.ts
+++ b/packages/cli/src/commands/status.ts
@@ -95,10 +95,10 @@ async function resolveRuntimeStatus(options: {
   configDir: string;
   projectId: string;
   workspaceDir: string;
-  snapshot: ProjectStatusSnapshot;
+  snapshot: ProjectStatusSnapshot | null;
 }): Promise<RuntimeStatus> {
   const liveness = await resolveDaemonLiveness(options);
-  if (!liveness.running) {
+  if (!liveness.running || !options.snapshot) {
     return { daemonRunning: false, lastTickStale: false };
   }
   const staleThresholdMs = await resolveStaleThresholdMs(
@@ -354,18 +354,18 @@ const handler = async (
 
     const run = async () => {
       const snapshot = await readStatusSnapshot(runtimeRoot, projectId);
-      const runtimeStatus = snapshot
-        ? await resolveRuntimeStatus({
-            configDir: options.configDir,
-            projectId,
-            workspaceDir: projectConfig.workspaceDir,
-            snapshot,
-          })
-        : { daemonRunning: false, lastTickStale: false };
+      const runtimeStatus = await resolveRuntimeStatus({
+        configDir: options.configDir,
+        projectId,
+        workspaceDir: projectConfig.workspaceDir,
+        snapshot,
+      });
       if (options.json || !isTTY) {
         process.stdout.write(
           JSON.stringify(
-            withDaemonRunning(snapshot, runtimeStatus.daemonRunning),
+            snapshot
+              ? withDaemonRunning(snapshot, runtimeStatus.daemonRunning)
+              : { daemonRunning: runtimeStatus.daemonRunning },
             null,
             2
           ) + "\n"
@@ -446,11 +446,33 @@ const handler = async (
       );
     }
   } else {
-    writeCliError({
-      code: "status_snapshot_unavailable",
-      message: "Unable to read status snapshot.",
-      json: options.json,
+    const runtimeStatus = await resolveRuntimeStatus({
+      configDir: options.configDir,
+      projectId,
+      workspaceDir: projectConfig.workspaceDir,
+      snapshot: null,
     });
+    if (options.json) {
+      process.stdout.write(
+        JSON.stringify(
+          {
+            daemonRunning: runtimeStatus.daemonRunning,
+            error: {
+              code: "status_snapshot_unavailable",
+              message: "Unable to read status snapshot.",
+            },
+          },
+          null,
+          2
+        ) + "\n"
+      );
+      process.exitCode = 1;
+    } else {
+      writeCliError({
+        code: "status_snapshot_unavailable",
+        message: "Unable to read status snapshot.",
+      });
+    }
   }
 };
 
--- a/packages/cli/src/commands/status.ts
+++ b/packages/cli/src/commands/status.ts
@@ -98,8 +98,11 @@ async function resolveRuntimeStatus(options: {
   snapshot: ProjectStatusSnapshot | null;
 }): Promise<RuntimeStatus> {
   const liveness = await resolveDaemonLiveness(options);
-  if (!liveness.running || !options.snapshot) {
+  if (!liveness.running) {
     return { daemonRunning: false, lastTickStale: false };
+  }
+  if (!options.snapshot) {
+    return { daemonRunning: true, lastTickStale: false };
   }
   const staleThresholdMs = await resolveStaleThresholdMs(
     options.workspaceDir,
--- a/packages/cli/src/commands/status.ts
+++ b/packages/cli/src/commands/status.ts
@@ -1,5 +1,9 @@
 import type { GlobalOptions } from "../index.js";
-import type { ProjectStatusSnapshot } from "@gh-symphony/core";
+import {
+  DEFAULT_POLL_INTERVAL_MS,
+  WorkflowConfigStore,
+  type ProjectStatusSnapshot,
+} from "@gh-symphony/core";
 import { readFile } from "node:fs/promises";
 import { join } from "node:path";
 import { resolveRuntimeRoot } from "../orchestrator-runtime.js";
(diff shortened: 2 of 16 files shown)


packages/cli/src/daemon-liveness.ts
```diff
@@ -0,0 +1,210 @@
...
+        pidPath,
+        recordedPid: lockTarget.pid,
+      };
+    }
+    return { running: false, reason: "missing-pid", pidPath };
+  }
+
+  const pidRecord = parseDaemonPidRecord(pidContents);
+  if (!pidRecord) {
+    return {
```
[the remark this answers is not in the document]
moncher-dev: Resolved in the follow-up regression test and implementation: malformed PID files now fall back to a validated live project lock, including exact CWD and heartbeat/identity checks.

moncher-dev:
@codex review: the Ready rework fixes the malformed PID fallback finding in d113f6c; focused, full-suite, Docker, and CI evidence is refreshed above.

packages/cli/src/daemon-liveness.ts
```diff
@@ -0,0 +1,227 @@
...
+  }
+
+  const identity = getProcessIdentity(pid);
+  const identityMatches = recordedIdentity
+    ? identity === recordedIdentity
+    : identity
+      ? isLegacyOrchestratorIdentity(identity)
+      : isFreshProjectLockHeartbeat(heartbeatAt);
+  const cwd = getProcessCwd(pid);
+  if (!identityMatches || !cwd || resolve(cwd) !== expectedCwd) {
```
hojinzs, line 206: **P1 — a null-identity lock target can now be signalled by `repo stop`**

`validateDaemonProcess` (line 145) keeps `!identity` as a hard reject. This new `validateProjectLockProcess` drops it: when the lock records `processIdentity: null` **and** `getProcessIdentity(pid)` returns `null`, `identityMatches` falls through to `isFreshProjectLockHeartbeat(...)` and the function returns `{ pid, identity: null }`.

That is fine for `status` (read-only), but the same `DaemonLiveness` feeds `stop.ts`, and the pre-signal guard there compares against the recorded value:

```ts
// stop.ts:99
if (getProcessIdentity(pid) !== checkedIdentity || ...)   // null !== null → false → guard passes
```

so `process.kill(pid, "SIGKILL")` is delivered to a PID that was never identity-verified at all. The only remaining evidence is *PID alive + cwd matches + a lock heartbeat under 60s* — and PID reuse inside that 60s window is exactly what the identity check existed to catch. Before this PR `resolveProjectLockDaemon` routed through `validateDaemonProcess`, so a null identity always aborted.

This is reachable in practice on any image without `ps` (`getProcessIdentity` shells out to `ps` and returns `null` on non-zero exit, while `getProcessCwd` reads `/proc` and still succeeds). Your own test `"recognizes a live foreground daemon from the project lock"` sets up precisely this state.

Suggested fix — keep the heartbeat fallback for reporting, but refuse to signal an unverified target. Either narrow the type so `stop` must handle it:

```ts
if (liveness.target.identity === null) {
  process.stderr.write(
    `Refusing to signal PID ${liveness.target.pid}: process identity could not be verified.\n`
  );
  process.exitCode = 1;
  return;
}
```

or carry a `verified: boolean` on `DaemonProcessTarget` and gate the kill on it. `status` keeps reporting `daemonRunning: true`; `stop` stops trusting a heartbeat alone.

Minor, same function: `isFreshProjectLockHeartbeat` uses `Math.abs(ageMs)`, so a heartbeat up to 60s in the *future* also passes. If that clock-skew tolerance is deliberate it's worth a comment; otherwise drop the `Math.abs`.

---

packages/cli/src/daemon-liveness.ts
```diff
@@ -0,0 +1,227 @@
...
+  return (
+    Number.isFinite(heartbeatAtMs) &&
+    Math.abs(ageMs) <= PROJECT_LOCK_HEARTBEAT_MAX_AGE_MS
+  );
+}
+
+function isLegacyOrchestratorIdentity(identity: string | null): boolean {
+  return Boolean(
+    identity &&
+    /(?:gh-symphony|(?:^|\s)(?:\S*\/)?(?:dist\/)?index\.js)(?:\s|$)/.test(
```
hojinzs: **P2 — the legacy-identity regex was silently widened during the move**

This was supposed to be a lift-and-shift out of `stop.ts`, but the pattern gained a `(?:\S*\/)?` segment:

```diff
-/(?:gh-symphony|(?:^|\s)(?:dist\/)?index\.js)(?:\s|$)/
+/(?:gh-symphony|(?:^|\s)(?:\S*\/)?(?:dist\/)?index\.js)(?:\s|$)/
```

Verified behaviour change:

| command line | old | new |
|---|---|---|
| `node dist/index.js repo start` | ✅ | ✅ |
| `node /opt/x/dist/index.js repo start` | ❌ | ✅ |
| `node /home/me/unrelated-app/index.js repo start` | ❌ | ✅ |

The third row is the concern: any `index.js` under any path now qualifies as "a legacy orchestrator", and this predicate is one of the guards in front of `process.kill`. The absolute-path row is presumably what you were actually fixing — if so, please say so in the PR body, since neither the description nor the commit message mentions a heuristic change.

Also, `(?:dist\/)?` is now dead: `\S*\/` is greedy within a token and already consumes `.../dist/`. `(?:^|\s)(?:\S*\/)?index\.js(?:\s|$)` matches every case above identically. If the intent is only to allow absolute paths ending in `dist/index.js`, `(?:\S*\/)?dist\/index\.js` would be the tighter expression.

---

packages/cli/src/commands/status.ts
```diff
@@ -13,30 +17,89 @@ import { bold, dim, green, red, yellow, cyan, stripAnsi } from "../ansi.js";
...
+}
+
+async function resolveRuntimeStatus(options: {
+  configDir: string;
+  projectId: string;
+  workspaceDir: string;
+  snapshot: ProjectStatusSnapshot;
+}): Promise<RuntimeStatus> {
+  const liveness = await resolveDaemonLiveness(options);
+  const staleThresholdMs = await resolveStaleThresholdMs(options.workspaceDir);
```
hojinzs: **P3 — the threshold is computed even when it is guaranteed to be discarded**

`lastTickStale` is `liveness.running && isLastTickStale(...)`, so when the daemon is down `staleThresholdMs` is never read — yet `resolveStaleThresholdMs` still runs. It calls `WorkflowConfigStore.load`, which does `access` + `stat` + `readFile` + a SHA-256 of `WORKFLOW.md` on every call before its fingerprint cache can short-circuit.

In `--watch` this fires every 2s (`run()` → `resolveRuntimeStatus`), including on the stopped path — which is the exact scenario this PR exists to render. Cheap to avoid:

```ts
const liveness = await resolveDaemonLiveness(options);
if (!liveness.running) {
  return { daemonRunning: false, lastTickStale: false };
}
const staleThresholdMs = await resolveStaleThresholdMs(options.workspaceDir);
return {
  daemonRunning: true,
  lastTickStale: isLastTickStale(options.snapshot.lastTickAt, staleThresholdMs),
};
```

---

packages/cli/src/commands/status.ts
```diff
@@ -255,8 +331,22 @@ const handler = async (
...
     const run = async () => {
       const snapshot = await readStatusSnapshot(runtimeRoot, projectId);
+      const runtimeStatus = snapshot
+        ? await resolveRuntimeStatus({
+            configDir: options.configDir,
+            projectId,
+            workspaceDir: projectConfig.workspaceDir,
+            snapshot,
+          })
+        : { daemonRunning: false, lastTickStale: false };
```
hojinzs: **P3 — `daemonRunning` is absent from JSON in the one case where it is most informative**

When no snapshot exists, `runtimeStatus` is correctly defaulted to `{ daemonRunning: false }`, but `withDaemonRunning(null, false)` returns `null`, so `--watch --json` emits the literal `null`. The single-shot path (line 426) has the same gap — it writes a `status_snapshot_unavailable` CLI error with no liveness field.

"No snapshot" is the *never-started* daemon, i.e. the strongest possible `daemonRunning: false` signal, and a scripted consumer reading `.daemonRunning` gets `undefined` instead. Emitting `{ "daemonRunning": false }` (plus the error payload on the single-shot path) would make the field total, as the issue's `--json output should gain a daemonRunning: boolean` asks for.

Non-blocking — say the word if you'd rather track it separately, since the snapshot-missing path is arguably a different issue from the stale-snapshot one.

---

packages/cli/src/dashboard/renderer.ts
```diff
@@ -225,6 +230,27 @@ function buildSummaryLines(
...
+    const health = runtimeStatus.daemonRunning
+      ? primarySnapshot.health
+      : "stopped";
+    const healthDot =
+      health === "degraded" || health === "stopped"
+        ? c.red("\u25CF")
+        : c.green("\u25CF");
+    const staleLabel = runtimeStatus.lastTickStale ? c.yellow(" (stale)") : "";
+    lines.push(
+      `  ${healthDot} ${c.dim("Health")}  ${c.bold(health)}     ${c.dim("Last tick")}  ${runtimeStatus.lastTickLabel}${staleLabel}`
```
hojinzs, line 245: **P3 (non-blocking) — the stopped/stale banner is now implemented twice**

This block and `renderLegacyStatus` (`status.ts:143-161`) independently derive the same three things: `daemonRunning ? snapshot.health : "stopped"`, the red-vs-green dot, and the `(stale)` suffix — plus two separately-worded copies of the restart hint and the `Last known state:` line. `status.ts` routes the dot through `healthIcon()`; this path re-derives it inline with an equivalent but duplicated conditional.

They agree today. They're two edits away from disagreeing, and the layouts already differ (`status.ts` puts the hint and last-known-state on separate lines; here they share one). A small exported helper returning `{ health, isRed, staleSuffix }` — or just reusing `healthIcon` here — would collapse the shared half.

Verified both renderers locally at `d113f6c`, so this is purely a maintainability note, not a defect:

```
# legacy
  ● Health    stopped           Last tick  22d ago
  daemon not running — run 'gh-symphony repo start'
  Last known state: idle

# --watch, COLUMNS=100
  ● Health  stopped     Last tick  22d ago
  daemon not running — run 'gh-symphony repo start'  Last known state: idle
```

FWIW on the PR's "readable at reviewer's terminal width" checkbox: the new banner line is ~75 columns, versus the existing `Agents / Runtime / Tokens` line at ~74, so it doesn't make the dashboard's width assumption any worse than it already is.

---

hojinzs (changes requested):
## Review — completeness vs. #546, smoke test, code quality

Independently verified at `d113f6c` (local build + black-box CLI runs, not just re-reading the diff). **The feature works and every acceptance criterion in #546 is met.** Two items block merge; everything else is a non-blocking nit.

### 1. Requirements coverage — #546

| # | Requirement | Status | Evidence |
|---|---|---|---|
| 1 | Shared `daemon-liveness.ts`, `stop` + `status` on one implementation | ✅ | `stop.ts` net −114 lines, both callers go through `resolveDaemonLiveness` |
| 2 | `● stopped` + `daemon not running — run 'gh-symphony repo start'`, last snapshot labeled as last-known | ✅ | scenario 1 below |
| 3 | Staleness guard at 3× tick interval when daemon is alive | ✅ | scenario 3 below |
| 4 | `relativeTime` ≥ 48h renders days | ✅ | `528h ago` → `22d ago`; `relativeTime(49h)` → `2d ago` |
| — | Applies to legacy renderer **and** `--watch` | ✅ | scenario 6 below |
| — | `--json` gains `daemonRunning` | ⚠️ mostly | total on the snapshot path; missing when no snapshot exists — inline note on `status.ts:341` |

### 2. Smoke test — the #546 incident, reproduced and confirmed fixed

Built the CLI and drove `dist/index.js` against a fixture with a 22-day-old `status.json`. No preview environment applies to a CLI, so this is the local equivalent.

**Before-state reproduced / now fixed** — dead daemon, no PID file, no lock:
```
  ● Health    stopped           Last tick  22d ago
  daemon not running — run 'gh-symphony repo start'
  Last known state: idle
```
`--json` → `daemonRunning = false`. This is exactly the `● idle / 528h ago` case from the issue.

Remaining scenarios, all with a real spawned process whose cwd matches `workspaceDir`:

| # | Setup | Result |
|---|---|---|
| 2 | Valid PID file, live process, fresh tick | `● idle` green, no banner, `daemonRunning = true` |
| 3 | Live daemon, 22d-old tick | `● idle … Last tick 22d ago (stale)` — health preserved, staleness annotated |
| 4 | **Malformed PID file + live lock w/ identity** | `daemonRunning = true` — the Codex finding is genuinely fixed |
| 5 | Malformed PID + lock `processIdentity: null`, real identity present | `daemonRunning = false` — correctly rejected |
| 6 | `--watch` at `COLUMNS=100` | dashboard renders the stopped banner and stays inside the width |

Gates re-run from scratch on this branch: `pnpm lint` ✅ · `pnpm test` ✅ (508 CLI tests, 35 files) · `pnpm typecheck` ✅ · `pnpm build` ✅. The 14 `status.test.ts` cases map cleanly onto the issue's verification list.

### 3. Blocking

**a. Missing changeset.** `@gh-symphony/cli` is versioned by changesets and every comparable `fix(...)` PR on `main` ships one (#533, #538, #535, #536, #532, #534, #531, #527, #524 …). This PR adds none, so a user-facing CLI behaviour change would land with no version bump and no changelog entry. One file fixes it.

**b. `daemon-liveness.ts:201` — a null-identity target can reach `process.kill`.** `validateProjectLockProcess` drops the `!identity` reject that `validateDaemonProcess` still enforces, and `stop.ts:99` compares `getProcessIdentity(pid) !== checkedIdentity` — `null !== null` is false, so the guard passes and SIGKILL goes to a PID verified only by *alive + cwd + <60s heartbeat*. Pre-PR this path routed through `validateDaemonProcess` and always aborted. Reachable on any image without `ps`. Details and a suggested fix inline — note the fix is **not** "require identity" (that would break the foreground-lock detection you just added), but "let `status` report it while `stop` refuses to signal it".

### 4. Non-blocking

- `daemon-liveness.ts:222` — the legacy-identity regex was widened during the "lift-and-shift" (now matches *any* path ending in `index.js`); please confirm this is intentional and mention it in the PR body. The `(?:dist\/)?` group is also dead.
- `status.ts:87` — `resolveStaleThresholdMs` runs on the stopped path where its result is discarded; re-reads + hashes `WORKFLOW.md` every 2s under `--watch`.
- `status.ts:341` — `--watch --json` emits literal `null` when no snapshot exists, so `daemonRunning` is `undefined` in the never-started case.
- `renderer.ts:245` — the stopped/stale banner is now implemented twice; maintainability only, both outputs verified correct today.

### Verdict

Sound fix, well-scoped, real test coverage, and the reported incident is demonstrably gone. Address **(a)** and **(b)** — both small and mechanical — and I'll approve on the next pass. I'd rather not see the P3s expand this PR; take them as-is or spin them out.

---

packages/cli/src/daemon-liveness.ts
```diff
@@ -0,0 +1,227 @@
...
+  const pidPath = daemonPidPath(options.configDir, options.projectId);
+  let pidContents: string;
+  try {
+    pidContents = await readFile(pidPath, "utf8");
+  } catch {
+    const expectedCwd = resolve(options.workspaceDir);
+    const lockTarget = await resolveProjectLockDaemon(
+      options.configDir,
+      options.projectId,
+      expectedCwd
```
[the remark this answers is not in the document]
hojinzs: Independently verified — **this one is real, and it's a new false negative introduced by this PR.** Reproduced it end-to-end against the built CLI at `d113f6c`. Two identical live processes, lock file identical, the *only* difference being CWD:

```
A: foreground daemon, cwd == workspaceDir   → daemonRunning = true
B: foreground daemon, cwd == /elsewhere     → daemonRunning = false
   repo stop → No running daemon found for project "tenant-a" (PID file missing).
```

Root cause chain, confirmed in the source:

1. `start.ts:1341` writes `daemon.pid` **only on the detached path**, and records `cwd: process.cwd()` there. A foreground `repo start` never writes one — which is exactly why your project-lock fallback exists.
2. So foreground always lands in the `missing-pid` branch, where `expectedCwd = resolve(options.workspaceDir)` (line 41) — there is no `pidRecord.cwd` to prefer.
3. `ProjectLockRecord` (`orchestrator/src/lock.ts:17-23`) is `{ownerToken, pid, startedAt, heartbeatAt, processIdentity}` — **no `cwd` field**, so the lock cannot supply the real value either.
4. Nothing calls `process.chdir` in `start.ts` or anywhere in `packages/orchestrator/src`, so the foreground process keeps the caller's CWD.

Worth separating the two halves, because they aren't equally severe:

- **`repo stop`** already required a CWD match before this PR (`expectedCwd = resolve(pidRecord.cwd ?? projectConfig.workspaceDir)`), so its behaviour here is **pre-existing**, not a regression.
- **`repo status`** had no liveness check at all before this PR, so it could never say "stopped". It now reports a perfectly healthy foreground daemon as `● stopped` and advises `run 'gh-symphony repo start'` — advice that will fail on the existing lock. That's a **new** defect, and it's the mirror image of #546: the incident was "says running when dead", this is "says stopped when running".

Of the two fixes suggested, I'd take **persist the CWD in the lock record** over chdir-ing foreground starts — chdir would change orchestrator-relative path resolution for everything downstream, which is a much larger blast radius than adding a field. Add `cwd` to `ProjectLockRecord`, validate against it in `validateProjectLockProcess`, and fall back to `workspaceDir` when the field is absent so pre-existing locks still resolve.

Promoting this to a blocker on my review — details in the summary comment.

---

packages/cli/src/commands/status.ts
```diff
@@ -13,30 +17,89 @@ import { bold, dim, green, red, yellow, cyan, stripAnsi } from "../ansi.js";
...
+  daemonRunning: boolean;
+  lastTickStale: boolean;
+};
+
+async function resolveStaleThresholdMs(workspaceDir: string): Promise<number> {
+  try {
+    const resolution = await workflowConfigStore.load(
+      join(workspaceDir, "WORKFLOW.md")
+    );
+    return resolution.workflow.polling.intervalMs * 3;
```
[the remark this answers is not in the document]
hojinzs: Independently verified — **real, and the arithmetic checks out exactly.** Traced it through the orchestrator at `d113f6c`:

- `service.ts:88` — `MAX_ADAPTIVE_POLL_INTERVAL_MULTIPLIER = 10`
- `service.ts:4517-4539` — `resolveAdaptivePollIntervalMs` returns `base * 10` once the rate-limit ratio hits 0, and scales by `0.5 / ratio` below `ADAPTIVE_RATE_LIMIT_FULL_SPEED_RATIO`
- `service.ts:1681` — the result is stored via `projectPollIntervals.set(projectId, effectivePollIntervalMs)`
- `service.ts:3283` → `getEffectivePollIntervalMs()` — `waitForNextPoll` sleeps for exactly that value

So with the default `DEFAULT_POLL_INTERVAL_MS = 30_000`, a rate-limited-but-perfectly-healthy daemon legitimately sleeps **300s** between ticks, while `resolveStaleThresholdMs` computes `polling.intervalMs * 3` = **90s** off the *unadjusted* workflow value. The daemon gets branded `(stale)` for 210 of every 300 seconds — a 70% false-positive duty cycle in steady state.

This matters more than a normal threshold-tuning nit because of what the PR is for. #546 is about `repo status` telling the truth about health; a `(stale)` badge on a daemon that is behaving exactly as designed is a new lie in the opposite direction, and it fires precisely when the operator is already under rate-limit pressure and most needs to trust the display.

Cheapest correct fix, I think: the snapshot already carries `rateLimits` (`snapshot.rateLimits`, used in `renderer.ts:229`), so the CLI has the same inputs the orchestrator had. Export `resolveAdaptivePollIntervalMs` from `@gh-symphony/orchestrator` and derive the threshold from `resolveAdaptivePollIntervalMs(configured, snapshot.rateLimits) * 3`. That keeps one source of truth for the backoff curve instead of two constants drifting apart.

If you'd rather not widen the export surface in this PR, persisting the effective interval into the snapshot (the orchestrator already has it in hand at `service.ts:1681`) works too and is arguably cleaner long-term — but that touches the core snapshot type, so it's a bigger change than I'd want to force into this PR. A blunt `* MAX_ADAPTIVE_POLL_INTERVAL_MULTIPLIER` bound would also kill the false positive at the cost of making the stale warning much less sensitive; acceptable as a stopgap if you note the tradeoff.

Promoting to a blocker — details in the summary comment.

---

hojinzs:
## Review round 2 — revised blocking list

Codex's pass on `d113f6c` landed two findings I had missed. I verified both independently rather than taking them at face value, and **both are real**, so my blocking list grows from two items to four. No new commits since my last review, so everything below is still against `d113f6c`. CI remains green (`Test`, `Container Smoke`).

### Blocking

| # | Item | Source | Status |
|---|---|---|---|
| a | No changeset — `@gh-symphony/cli` would ship with no version bump or changelog entry | mine | unchanged |
| b | `daemon-liveness.ts:201` — null-identity lock target can reach `process.kill` in `repo stop` | mine | unchanged |
| c | `daemon-liveness.ts:45` — foreground daemon outside `workspaceDir` reported as `● stopped` | Codex, **verified + reproduced** | **new** |
| d | `status.ts:65` — staleness threshold ignores the adaptive poll interval | Codex, **verified** | **new** |

**(c)** and **(d)** both matter more than their P2 badges suggest, because they are the *same class of bug this PR exists to fix*, pointed the other way. #546 is "`repo status` lies about health"; (c) makes it say `● stopped` about a running daemon, and (d) makes it say `(stale)` about a daemon ticking exactly as designed. Shipping the fix with two fresh false readings in it undercuts the point.

Detail and suggested fixes are in the two threads — short version:

- **(c)** Reproduced with two identical live processes differing only in CWD: `daemonRunning` flips `true` → `false`. `daemon.pid` is written only on the detached path (`start.ts:1341`), `ProjectLockRecord` has no `cwd` field, and nothing chdirs — so foreground starts can only ever be validated against `workspaceDir`. Note the `repo stop` half is **pre-existing**; only the `repo status` false-`stopped` is a regression from this PR. Suggest adding `cwd` to the lock record with a `workspaceDir` fallback for old locks, rather than chdir-ing foreground starts.
- **(d)** With the default 30s interval, `resolveAdaptivePollIntervalMs` (`service.ts:4517`, multiplier capped at 10) lets a healthy rate-limited daemon sleep 300s, while the threshold is `configured × 3` = 90s — `(stale)` for ~70% of every cycle. The snapshot already carries `rateLimits`, so deriving the threshold from the same adaptive function keeps one source of truth.

### Unchanged from round 1

Requirements coverage against #546 is still complete, and the smoke tests still pass — the original incident (`● idle` / `528h ago` on a 22-day-dead daemon) is demonstrably fixed. Non-blocking P3s (widened identity regex, wasted `WORKFLOW.md` read per watch tick, `daemonRunning` absent when no snapshot, duplicated banner rendering) stand as written; still no need to expand this PR for them.

### Verdict

Still **Request changes** — 4 blockers, all small and mechanical, none requiring a redesign. Round 2 of a maximum 6. To keep this from turning into an open-ended loop: I'll approve as soon as (a)–(d) are addressed, and if the remaining rounds get consumed without them landing, my recommendation flips to closing this PR and re-cutting the change against a lock record that carries `cwd` from the start — that one missing field is the root of both (b) and (c), and building more liveness logic on top of it will keep producing findings like these.

---

hojinzs (commented):
## Blocker list correction — two more, and a smoke reproduction of one

Bookkeeping fix first: my summary above was submitted at `01:03:34`, but the two thread replies where I promoted the Codex findings landed at `01:07`. So those replies point at "details in the summary comment" and the summary doesn't mention them. Consolidating here so the blocker list is one thing, not three. **The `CHANGES_REQUESTED` state stands; the list is (a), (b), (c), (d) — not just (a) and (b).**

### (c) A live foreground daemon started outside the workspace reports `stopped`

This is the one I want to stress, because I reproduced it against the built CLI rather than reasoning about it. Two spawned processes, identical live state, identical lock files, the **only** difference being CWD:

```
B: cwd == workspaceDir   → "daemonRunning": true
                           ● Health    idle    Last tick  22d ago (stale)
                           repo stop → Sent SIGTERM to orchestrator (PID 22400). Daemon stopped.

C: cwd == /elsewhere     → "daemonRunning": false
                           ● Health    stopped   Last tick  22d ago
                           daemon not running — run 'gh-symphony repo start'
                           Last known state: idle
                           repo stop → No running daemon found for project "tenant-a" (PID file missing).
```

Case C is a healthy running daemon rendered with the red stopped banner, told to `repo start` (which will fail on the lock it already holds), and unreachable via `repo stop`. That is #546 inverted, produced by the fix for #546.

Worth being precise about which half is new, because it changes how much this PR owes: `repo stop` already required a CWD match before this PR, so its behaviour here is **pre-existing**. `repo status` had no liveness check at all, so it could never wrongly say "stopped" — that half is **new**, and it's the half the issue is about.

My scenario table in the summary above ran every case with `cwd == workspaceDir`, which is exactly why it came back clean. That was a gap in my testing, not in yours.

### (d) `(stale)` fires on healthy rate-limited daemons

`resolveStaleThresholdMs` uses the unadjusted `polling.intervalMs * 3` = 90s at the default, while `resolveAdaptivePollIntervalMs` (`service.ts:88`, `MAX_ADAPTIVE_POLL_INTERVAL_MULTIPLIER = 10`) legitimately stretches the sleep to 300s under rate-limit pressure. ~70% false-positive duty cycle in steady state, firing precisely when the operator is already under pressure and most needs the display to be true.

Note this one is visible in my case-B output above — `(stale)` on a daemon that is fine. There the tick really is 22 days old so the badge is correct, but it's the same code path.

### Revised list

| | Blocker | Size |
|---|---|---|
| a | Missing changeset (`@gh-symphony/cli`; every comparable `fix(...)` on `main` ships one — verified `.changeset/` is untouched on this branch) | 1 file |
| b | Null-identity lock target reaches `process.kill` via `stop.ts:99` | small |
| c | Foreground daemon outside workspace reports stopped | needs `cwd` on `ProjectLockRecord` |
| d | Staleness threshold ignores the adaptive interval | small |

### On scope, since this list grew

All four are "the new liveness signal is wrong", which is this PR's own subject — fixing them is finishing the change, not extending it. The four P3s stay non-blocking and I won't hold a re-review on them.

(c) is the only one that reaches outside `packages/cli` (`ProjectLockRecord` gains a field). **If that feels too wide, I'll take a narrower fix plus a follow-up issue** — say the word rather than spending a round on it. Same for (d): deriving the threshold from `resolveAdaptivePollIntervalMs` is the clean fix, but a blunt `* MAX_ADAPTIVE_POLL_INTERVAL_MULTIPLIER` bound with a comment noting the sensitivity tradeoff is acceptable if you'd rather not widen the export surface here.

Round 2 of 6. The core fix is good and I want to approve it — push (a)-(d) and re-request.

---

packages/cli/src/daemon-liveness.ts
moncher-dev: Resolved in `7ebb4c5`: lock-derived targets now carry `verified`, and `repo stop` refuses to signal a target whose identity could not be verified. Status can still report the fresh-heartbeat lock as running.

packages/cli/src/daemon-liveness.ts
moncher-dev: Resolved in `7ebb4c5`: the legacy fallback now accepts only `index.js` or a path ending in `dist/index.js`, so an unrelated absolute-path `index.js` no longer qualifies. The tighter expression preserves the detached CLI path while removing the widened kill guard.

packages/cli/src/commands/status.ts
moncher-dev: Resolved in `36c75f7`: liveness is resolved before the stale threshold, and stopped daemons return immediately without loading `WORKFLOW.md`. This keeps the watch stopped path free of the repeated workflow read.

packages/cli/src/commands/status.ts
moncher-dev: Resolved in `3cfdde7`: JSON output now includes `daemonRunning` when no snapshot is available, both for one-shot error output and the watch JSON path. A regression test covers the one-shot unavailable-snapshot payload.

packages/cli/src/dashboard/renderer.ts
moncher-dev: Tracked separately as #556; this rework keeps the reviewed renderer layouts unchanged while resolving the correctness blockers.

packages/cli/src/daemon-liveness.ts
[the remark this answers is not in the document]
moncher-dev: Resolved in `7ebb4c5`: project locks now persist `cwd` at acquisition, and lock fallback validates against it. Older locks without `cwd` retain the `workspaceDir` fallback. The new status regression covers a live foreground daemon outside the workspace directory.

packages/cli/src/commands/status.ts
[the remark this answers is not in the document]
moncher-dev: Resolved in `36c75f7`: `resolveAdaptivePollIntervalMs` is exported from the orchestrator and reused by CLI status with the snapshot's rate limits before applying the 3× stale threshold. The rate-limited regression verifies a 300s effective interval is not marked stale after two minutes.

moncher-dev:
@codex review: the current `3cfdde7` head addresses every prior blocking finding. Cycle 4 re-verified the resolved liveness, safe-stop, lock-CWD, adaptive-staleness, and changeset coverage; please review the current head.

hojinzs (approved):
## ✅ Approve — all four blockers verified fixed at `3cfdde7`

Re-verified independently: fresh install, build, full gates, and a black-box smoke matrix against the built CLI. Not a re-read of the diff — every fix below was exercised as a running process.

### Blockers

| | Item | Fix | Verified |
|---|---|---|---|
| a | Missing changeset | `.changeset/wise-dolphins-repair.md`, `@gh-symphony/cli: patch` | ✅ |
| b | Null-identity target reaching `process.kill` | `verified` flag on `DaemonProcessTarget`; `stop.ts` refuses | ✅ |
| c | Foreground daemon outside workspace read as stopped | `cwd` persisted on `ProjectLockRecord`, validated on fallback | ✅ |
| d | Staleness ignoring the adaptive interval | `resolveAdaptivePollIntervalMs` exported and reused | ✅ |

On (a): I checked whether `packages/orchestrator` also needed an entry, since its API changed (new export, `cwd` on the lock record) — it's `private: true`, so the single CLI patch is the correct scope. Nothing missing.

### Smoke matrix — the fixes, exercised

**(d)** The discriminating pair, same 120s-old tick:
```
low rate limits (limit 5000, remaining 0) → ● idle   Last tick 2m ago            ← no false (stale)
no rate limits                            → ● idle   Last tick 2m ago (stale)    ← still fires
```
The control case is the important half: the fix removes the false positive without quietly disabling the warning.

**(b)** Ran the CLI with `ps` unavailable so `getProcessIdentity` returns `null` — the only way to actually reach `verified: false`:
```
repo status → daemonRunning = true                                    ← read-only, still useful
repo stop   → Recovered orchestrator PID 28681 from the project lock…
              Refusing to signal PID 28681: process identity could not be verified.
              exit=1 · process ALIVE
```
Exactly the split I hoped for: `status` keeps reporting, `stop` stops trusting a heartbeat alone.

**(c)** Foreground daemon with cwd ≠ `workspaceDir` → `daemonRunning = true` (was `false`). Backward compatibility also holds — a pre-existing lock with no `cwd` field still resolves via the `workspaceDir` fallback, and correctly stays `false` when the daemon really is elsewhere. `renewProjectLock` spreads `{...existing}` and `parseProjectLock` now carries `cwd`, so the field survives every heartbeat rather than being dropped on the first renewal.

**Original #546 incident, still fixed:**
```
● Health    stopped           Last tick  21d ago
daemon not running — run 'gh-symphony repo start'
Last known state: idle
```

**P3 follow-through:** no-snapshot JSON now emits `{"daemonRunning": false, "error": {...}}`. I checked exit-code parity on the reworked error path since it stopped routing through `writeCliError` for JSON — both `--json` and text exit `1`, unchanged, and the JSON shape is a superset of the old one.

### Gates

`pnpm lint` ✅ · `pnpm typecheck` ✅ · `pnpm test` ✅ (35 CLI files, 511 tests — up from 508 at `d113f6c`) · `pnpm build` ✅. CI `Test` and `Container Smoke` green on `3cfdde7`.

### Closing

Round 3 of 6, and I'm approving rather than spending the remaining rounds. The two Codex findings were the valuable part of this review and they were real; the fixes went to the root cause — `cwd` on the lock record — rather than patching the symptom at the CLI, which is why (b) and (c) both closed with one structural change. Deferring the duplicated-banner cleanup to #556 is the right call and I won't hold the PR for it.

Nice work on the turnaround.

---

hojinzs (changes requested):
## Retracting my approval — Codex's finding on `status.ts:102` is real, and I missed it

**My approval above was premature. Please don't merge on it.** Codex found a live defect in `3cfdde7` — the commit that closed my own P3 — and I've reproduced it.

### Reproduced

Same live daemon (valid `daemon.pid`, matching identity and CWD), the only variable being whether `status.json` exists:

```
WITH snapshot    → daemonRunning=true
WITHOUT snapshot → daemonRunning=false   ← wrong, the daemon is alive

non-TTY watch, same state:
  { "daemonRunning": false }
```

`resolveRuntimeStatus` collapses two distinct conditions into one early return:

```ts
const liveness = await resolveDaemonLiveness(options);
if (!liveness.running || !options.snapshot) {
  return { daemonRunning: false, lastTickStale: false };   // discards a true liveness result
}
```

`liveness.running` is computed correctly and then thrown away whenever the snapshot is absent.

### Why this matters more than it looks

"Daemon is live but `status.json` doesn't exist yet" is not an exotic state — it's **every `repo status` run in the window between `repo start` and the first tick**, plus any transient unreadable snapshot. In that window a scripted consumer reading `.daemonRunning` is told the daemon is down. That is #546 inverted for the third time in this PR, and this instance was introduced by the fix for my own P3, which asked for the field to be present without my saying what it should contain when a daemon is alive and untickd.

**Scope, precisely:** JSON surfaces only — single-shot `--json` and the non-TTY watch path. The text path prints `Unable to read status snapshot.` and renders no health line, so it makes no false "stopped" claim. Both broken surfaces route through this one function, so one fix covers both.

### Fix

Codex's suggestion is right — separate the conditions and gate only the stale calculation on the snapshot:

```ts
const liveness = await resolveDaemonLiveness(options);
if (!liveness.running) {
  return { daemonRunning: false, lastTickStale: false };
}
if (!options.snapshot) {
  return { daemonRunning: true, lastTickStale: false };
}
const staleThresholdMs = await resolveStaleThresholdMs(
  options.workspaceDir,
  options.snapshot.rateLimits
);
return {
  daemonRunning: true,
  lastTickStale: isLastTickStale(options.snapshot.lastTickAt, staleThresholdMs),
};
```

Please add a regression test for **live daemon + absent snapshot → `daemonRunning: true`**. The existing `3cfdde7` test covers the unavailable-snapshot payload with no daemon running, where `false` is the correct answer — which is exactly why the suite stayed green.

### On my own miss

My smoke scenario 6 deleted `status.json` but had already killed the daemon, so `false` was correct in my run and the bug was invisible. I tested the field's *presence*, which is what I'd asked for, and not its *value* under a live daemon. The lesson is the same one that made the (d) check useful two rounds ago — a negative assertion needs a positive control — and I didn't apply it here. Credit to Codex for catching it.

### Everything else stands

Blockers (a)–(d) remain fixed and verified; nothing in my previous review is withdrawn beyond the approval itself. This is the only outstanding item.

Round 4 of 6. This is a handful of lines plus a test — push it and I'll re-verify and approve. If it lands clean, that's the last round I intend to spend here.

---

packages/cli/src/commands/status.ts
```diff
@@ -13,30 +17,110 @@ import { bold, dim, green, red, yellow, cyan, stripAnsi } from "../ansi.js";
...
+
+async function resolveRuntimeStatus(options: {
+  configDir: string;
+  projectId: string;
+  workspaceDir: string;
+  snapshot: ProjectStatusSnapshot | null;
+}): Promise<RuntimeStatus> {
+  const liveness = await resolveDaemonLiveness(options);
+  if (!liveness.running || !options.snapshot) {
+    return { daemonRunning: false, lastTickStale: false };
```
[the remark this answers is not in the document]
hojinzs: Confirmed and reproduced — **blocking**. Same live daemon (valid `daemon.pid`, matching identity and CWD), only `status.json` varying:

```
WITH snapshot    → daemonRunning=true
WITHOUT snapshot → daemonRunning=false   ← wrong
```

Your suggested shape is the fix — split the two conditions, gate only the stale calculation on the snapshot. Full write-up and a regression-test request in my review; I've retracted my approval, which was premature.

Scoping note for whoever picks this up: the defect is confined to the JSON surfaces (single-shot `--json` and non-TTY watch). The text path prints `Unable to read status snapshot.` with no health line, so it makes no false "stopped" claim. Both broken surfaces route through `resolveRuntimeStatus`, so one change covers them.

---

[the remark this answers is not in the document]
moncher-dev: Resolved in `53cdb7a`: `resolveRuntimeStatus` now preserves `liveness.running` when `status.json` is unavailable and only gates stale-tick calculation on a snapshot. Regression coverage verifies both one-shot JSON and non-TTY watch JSON report `daemonRunning: true` for a live daemon without a snapshot.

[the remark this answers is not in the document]
moncher-dev: Resolved in `53cdb7a`: `resolveRuntimeStatus` now preserves `liveness.running` when `status.json` is unavailable and only gates stale-tick calculation on a snapshot. Regression coverage verifies both one-shot JSON and non-TTY watch JSON report `daemonRunning: true` for a live daemon without a snapshot.

hojinzs (approved):
## ✅ Approve — `53cdb7a`, this time with the control I should have run the first time

The `status.ts:102` regression is fixed exactly as scoped: `!liveness.running` and `!options.snapshot` are now separate returns, so a live daemon with no snapshot keeps `daemonRunning: true` and only the stale calculation is gated on the snapshot.

### Verified both directions, both surfaces

The mistake that produced my premature approval was asserting only the negative case, so this time I ran the matched pair — same fixture, daemon alive vs. daemon killed, nothing else changed:

| | single-shot `--json` | non-TTY watch |
|---|---|---|
| **live daemon**, no `status.json` | `daemonRunning=true` (+ `status_snapshot_unavailable`, exit 1) | `{ "daemonRunning": true }` |
| **daemon killed**, no `status.json` | `daemonRunning=false` | `{ "daemonRunning": false }` |

The negative row is the one that makes the positive row mean anything — it confirms the fix preserves a true liveness result rather than hardcoding the field the other way.

### Regression coverage

Both surfaces are now pinned in the suite, and the update to the pre-existing test is the interesting part:

```diff
 it("includes daemonRunning when JSON status has no snapshot", async () => {
+  await writeDaemonPid(configDir);
+  vi.spyOn(process, "kill").mockImplementation(() => true);
-  daemonRunning: false,
+  daemonRunning: true,
```

That test previously asserted the bug as expected behaviour, which is precisely why CI stayed green through `3cfdde7`. Good catch fixing the assertion rather than adding a second test alongside it. The new non-TTY watch case covers the other surface.

Non-blocking observation: with that test converted to the live-daemon case, **no daemon + no snapshot → `daemonRunning: false`** no longer has a dedicated test. The `!liveness.running` branch is still exercised elsewhere (e.g. "adds daemonRunning to JSON output"), and I verified the behaviour by hand above, so this is a note for whenever someone next touches this function — not something to spend a round on.

### Gates

`pnpm lint` ✅ · `pnpm typecheck` ✅ (0 errors) · `pnpm test` ✅ (513 CLI tests, up from 511) · `pnpm build` ✅. CI `Test` and `Container Smoke` green on `53cdb7a`.

### Standing state

Blockers (a)–(d) remain verified from round 3 and nothing regressed. This was the only outstanding item. Renderer duplication stays deferred to #556.

Round 5 of 6, and this is my last. Good to merge.

For what it's worth, the review found six real defects across five rounds and Codex found half of them — including the one I'd approved over. Worth keeping both passes in the loop on changes like this.

---

packages/cli/src/commands/status.test.ts
```diff
@@ -154,6 +458,71 @@ describe("status command", () => {
...
+      void statusCommand(["--watch"], {
+        configDir,
+        verbose: false,
+        json: false,
+        noColor: true,
+      });
+      await vi.waitFor(() => {
+        expect(JSON.parse(stdout.output())).toEqual({ daemonRunning: true });
+      });
+      process.emit("SIGTERM", "SIGTERM");
```
hojinzs, line 550: 🧹 **이 TC는 정확히 맞는 케이스를 잡습니다. 다만 `process`에 리스너를 남기고 끝납니다 — 비블로킹 위생 지적입니다.**

새 TC 자체는 제가 요청한 그대로입니다: 살아있는 데몬 + 스냅샷 부재 → `daemonRunning: true`. `--watch` 경로가 단발 `--json`과 같은 `resolveRuntimeStatus`를 타므로 두 표면을 한 TC로 덮은 것도 맞습니다.

다만 `statusCommand(["--watch"])`는 `status.ts:412-422`에서 `SIGWINCH`/`SIGINT`/`SIGTERM` 리스너 세 개를 `process`에 등록하고, 이 TC는 그중 어느 것도 떼지 않습니다. `process.exit`가 스텁이라 `shutdown()`이 `clearInterval` 후 그냥 반환하고 — interval은 제대로 정리됩니다 — `await new Promise(() => {})`가 영원히 pending으로 남아 `void`한 프라미스도 회수되지 않습니다.

지금은 이 파일에서 `--watch`를 띄우는 TC가 하나뿐이라 실害가 없습니다. 문제는 두 번째가 생길 때인데, 그때 실패가 **이 TC가 아니라 나중 TC에서** 납니다:

- 남은 `SIGTERM` 리스너가 다음 `--watch` TC의 `process.emit("SIGTERM")`에 함께 반응합니다. 그 시점에 `process.exit` 스텁이 이미 `afterEach`의 `restoreAllMocks`로 풀려 있으면 **진짜 `process.exit(0)`가 호출되어 vitest 워커가 조용히 죽습니다.**
- `MaxListenersExceededWarning`도 11개째부터 뜹니다.

한 줄이면 닫힙니다:

```diff
     } finally {
+      process.removeAllListeners("SIGTERM");
+      process.removeAllListeners("SIGINT");
+      process.removeAllListeners("SIGWINCH");
       stdout.restore();
```

(`statusCommand`가 리스너를 스스로 정리하게 만드는 쪽이 더 낫지만, 그건 이 PR 범위 밖입니다. TC 쪽 한 줄로 충분합니다.)

**이 PR을 붙잡지 않겠습니다** — 지금 코드로 머지해도 되고, 다음에 이 파일을 만질 때 챙기셔도 됩니다.

---

hojinzs (approved):
## ✅ Approve — 라운드 4 블로커 수정 확인 (5/6 라운드, 최종)

`53cdb7a`에서 재검증했습니다. diff 재독이 아니라 **빌드된 `dist/index.js`를 실제 프로세스로 구동한 블랙박스 스모크**입니다.

### 라운드 4 블로커 — 해소

`resolveRuntimeStatus`가 두 조건을 분리했습니다 (`status.ts:100-106`). 제가 제안한 형태 그대로입니다.

**스모크 매트릭스** — 살아있는 자식 프로세스(cwd == `workspaceDir`, `ps` 기반 실제 identity를 `daemon.pid`에 기록)를 띄우고, **스냅샷 존재 여부만** 바꿔가며:

```
A) 살아있는 데몬 + status.json 없음   → daemonRunning: true    ← 라운드 4의 결함, 수정됨
B) 살아있는 데몬 + status.json 없음(text) → "Unable to read status snapshot." (health 라인 없음, 거짓 주장 없음)
C) 살아있는 데몬 + status.json 있음   → daemonRunning: true, ● idle
D) 죽은 데몬     + status.json 없음   → daemonRunning: false   ← positive control, 여전히 정확
E) 죽은 데몬     + status.json 있음   → ● stopped / "daemon not running — run 'gh-symphony repo start'" / Last known state: idle
```

**D가 이번 검증의 핵심입니다.** 라운드 4에서 제가 놓친 이유가 정확히 "부정 단정을 대조군 없이 확인했다"는 것이었으므로, 이번엔 A와 D를 같은 픽스처에서 **프로세스 생사만 바꿔** 돌렸습니다. 수정이 `daemonRunning`을 무조건 `true`로 만든 게 아니라 liveness 결과를 보존하는 것임을 확인했습니다. E로 원래 #546 사고도 여전히 고쳐져 있음을 재확인했습니다.

TC 두 건도 요청대로 들어왔고 — 단발 `--json`은 `false → true`로 기대값이 정정됐고, 비-TTY watch JSON은 신규 TC입니다 — 양쪽 표면이 한 함수를 타므로 커버리지가 맞습니다.

### 게이트

클린 install 후 전부 재실행: `pnpm lint` ✅ · `pnpm typecheck` ✅ · `pnpm test` ✅ (CLI 35 files / **513 tests**, `3cfdde7`의 511에서 +2) · `pnpm build` ✅. CI `Test`·`Container Smoke` 모두 green (`53cdb7a`).

### 요구사항 (#546) 최종

블로커 (a) changeset · (b) null-identity가 `process.kill` 도달 · (c) 워크스페이스 밖 foreground 데몬이 stopped로 표시 · (d) staleness가 adaptive interval 무시 — 라운드 3에서 확인한 상태 그대로 유지되고, 이번 라운드의 변경은 `resolveRuntimeStatus` 5줄과 TC뿐이라 회귀 위험이 없습니다. #546의 수용 조건은 라운드 1 표대로 전부 충족입니다.

### 남긴 것

인라인 1건 — 새 watch TC가 `process`에 시그널 리스너를 남깁니다. **비블로킹**이고, 두 번째 `--watch` TC가 생길 때 그 TC에서 터지는 종류입니다. 지금 고치든 다음에 고치든 상관없습니다.

배너 중복(#556)은 스핀아웃한 판단이 맞습니다. `daemon-liveness.ts:222` 정규식 확장과 `status.ts:87`의 불필요한 `resolveStaleThresholdMs` 호출은 P3로 그대로 두겠습니다 — 이 PR에서 더 키우지 않는 편이 낫습니다.

### 마무리

5라운드에 걸쳐 같은 축("새 liveness 신호가 틀린 값을 말한다")이 세 번 반복됐고 — 원래 #546, 라운드 2의 (c), 라운드 4의 JSON 경로 — 매번 근본 원인 쪽에서 닫혔습니다. 남은 라운드를 더 쓰지 않고 여기서 승인합니다. 머지하셔도 됩니다.

---

moncher-dev:
Land-return rework is complete on `bb806ac`: the feature branch is fresh against `main`, original #546 behavior remains covered, and full validation passes. @codex review

hojinzs:
## Approval carries to `bb806ac` — no re-review needed

Checked the delta rather than assuming, since a "rework" that re-cuts a branch is exactly the case where an approval can go stale under you.

**The branch was not rewritten.** `53cdb7a` — the commit I approved — is still an ancestor of the head:

```
git merge-base --is-ancestor 53cdb7a bb806ac  →  ancestor
git diff 53cdb7a..bb806ac --stat
  packages/orchestrator/src/index.test.ts | 17 ++++++-----------
```

One test-only commit on top, no production code touched. It swaps two hand-rolled `for (attempt < 20) await setTimeout(0)` polling loops for `vi.waitFor`. Strictly an improvement in the second case — the old loop would fall through silently after 20 attempts if `finishFirst` was never assigned, whereas `vi.waitFor` now actually asserts it. `packages/orchestrator/src/index.test.ts` passes locally (18 tests), and CI is green on `bb806ac`.

Small correction on the framing, since it changes whether anyone should expect a re-review: the branch is not "fresh against `main`" in the sense of having been re-cut. History is unchanged and `main` simply hasn't moved since this branch was created — its base is still the same commit it always was. Nothing was rebased or force-pushed. Worth being precise so a later reader doesn't go looking for a re-cut that didn't happen.

Everything from my `53cdb7a` approval stands unchanged: blockers (a)–(d) verified, the round-4 liveness regression fixed and confirmed with both controls, #556 deferred, and the `--watch` listener-hygiene note left as non-blocking.

That's round 6 of 6 — the ping-pong budget is spent and there's nothing outstanding. **Good to merge.** I won't add further review rounds; anything new from here is better as a follow-up issue than another cycle on this PR.

---

hojinzs:
## Codex's `status.ts:110` finding is real — correcting my round-3 "(d) verified"

Traced it rather than deferring, and it holds up. **`snapshot.rateLimits` is not the value the orchestrator used to compute its sleep.**

```
service.ts:1245   rateLimits = resolveProjectRateLimits(syncedActiveRuns, issues, trackerListLimits)
                  └─ resolveProjectRateLimits (4444-4464): returns latestRunRateLimits FIRST;
                     tracker limits are only the fallback argument
service.ts:1250   trackerRateLimits = resolveTrackerRateLimits(...)        ← tracker-only
service.ts:1658   resolveAdaptivePollIntervalMs(pollIntervalMs, trackerRateLimits)   ← the actual sleep
service.ts:1696   rateLimits = rateLimits ?? trackerRateLimits ?? ...      ← ?? no-ops; already set at 1245
service.ts:1708   buildProjectSnapshot({ ..., rateLimits })                → snapshot.rateLimits
```

Run `rateLimits` are populated from worker/Codex events (`event.rateLimits`, `workerInfo.rateLimits`), so whenever a run is active with Codex telemetry, `snapshot.rateLimits` carries **Codex quota**, not GitHub's.

The corroborating detail is three lines below: `dispatchRateLimits = trackerRateLimits ?? rateLimits` (1700). The codebase already treats these as semantically distinct and deliberately prefers the tracker value where tracker semantics are needed. `resolveStaleThresholdMs` is such a place, and it's getting the other one.

**Failure case:** GitHub throttled (remaining 0 → orchestrator sleeps 300s) while an active run reports healthy Codex quota → CLI computes `resolveAdaptivePollIntervalMs(30s, codexLimits)` = 30s, threshold 90s → `(stale)` on a daemon behaving exactly as designed. That is blocker (d), surviving in the active-run configuration.

### Where my verification fell short

My round-3 fixture set `rateLimits` to tracker-shaped low limits with `activeRuns: []`. That is precisely the configuration in which `resolveProjectRateLimits` falls through to the tracker fallback and the two values coincide — so the bug was invisible. I asserted "(d) verified ✅" on a fixture that couldn't distinguish the correct input from the incorrect one. Same failure mode as round 4: I varied the thing under test but not the thing that selects between code paths.

### Where this leaves the PR — your call, not another round from me

Being straight about the tradeoff rather than quietly re-blocking:

- By the standard I applied in round 2, this is the same defect class and would have been blocking.
- It is also **narrower than what shipped before**: `main` has no staleness signal at all, so #546's "dead daemon looks healthy" is still fixed and the common no-active-runs case is now correct. Net, this PR is an improvement even with the residual.
- The fix is small — pass tracker limits rather than `snapshot.rateLimits`, which likely means exposing `trackerRateLimits` on the snapshot separately (or persisting the effective interval, as Codex suggests and as I raised back in round 2 as the cleaner long-term option).

I've spent the 6-round budget and I'm not going to run another block-verify-approve cycle over a defect that doesn't make things worse than `main` — that's exactly the unbounded ping-pong this review was scoped to avoid. **My approval stands and I'm not withdrawing it.** But it now stands with (d) explicitly marked *partially* closed rather than closed, so please merge knowing that, and track the residual as a follow-up alongside #556. If you'd rather close it in this PR, it's a small change and I won't object to the extra commit.

Credit to Codex again — that's three of seven real findings, two of them on things I'd already signed off.

---

packages/cli/src/commands/status.ts
```diff
@@ -13,30 +17,113 @@ import { bold, dim, green, red, yellow, cyan, stripAnsi } from "../ansi.js";
...
+  if (!liveness.running) {
+    return { daemonRunning: false, lastTickStale: false };
+  }
+  if (!options.snapshot) {
+    return { daemonRunning: true, lastTickStale: false };
+  }
+  const staleThresholdMs = await resolveStaleThresholdMs(
+    options.workspaceDir,
+    options.snapshot.rateLimits
+  );
```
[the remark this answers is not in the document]
moncher-dev: Resolved in `1c6a93cf`: snapshots now carry `effectivePollIntervalMs`, calculated solely from the tracker rate limits that control the daemon scheduler. `repo status` uses that recorded interval for its 3× stale threshold, with the existing rate-limit inference retained only for legacy snapshots. CLI and orchestrator regression tests cover both persistence and consumption.

hojinzs:
## (d) now fully closed at `1c6a93c` — verified, approval stands unqualified

I invited this change, so I verified it rather than letting it ride on the standing approval. It's the option I'd flagged as the cleaner long-term fix back in round 2, and it's right: the snapshot now carries `effectivePollIntervalMs` — the exact value the daemon sleeps for (`service.ts:1658`, computed from `trackerRateLimits` alone) — instead of asking the CLI to re-derive it from a field that meant something else.

### Verified with the discriminator my earlier fixture lacked

The trap in every case below: `snapshot.rateLimits` carries **healthy Codex quota**, which is exactly what an active run reports and what previously poisoned the calculation. Tick 120s old throughout.

| snapshot | expected | result |
|---|---|---|
| `effectivePollIntervalMs: 300000` (tracker throttled) | not stale | `● idle  Last tick 2m ago` ✅ |
| `effectivePollIntervalMs: 30000` (tracker healthy) | stale | `● idle  Last tick 2m ago (stale)` ✅ |
| legacy, field absent | stale via old inference | `● idle  Last tick 2m ago (stale)` ✅ |

Row 2 is the one that matters and the one I skipped last time — it proves the threshold actually tracks the recorded interval rather than the fix having quietly disabled the warning. Row 3 confirms the legacy path still works, which matters because any snapshot written before this deploy won't have the field.

Two details I checked in the implementation: `buildProjectSnapshot` spreads the field conditionally (`...(effectivePollIntervalMs === undefined ? {} : {...})`), so old snapshots stay byte-clean rather than gaining an explicit `undefined`; and the CLI branches on `!== undefined` rather than truthiness, so a legitimately recorded `0` wouldn't silently fall through to inference.

### Gates

`pnpm lint` ✅ · `pnpm typecheck` ✅ (0 errors) · `pnpm test` ✅ (514 CLI tests) · `pnpm build` ✅.

### Final state

All four blockers closed, (d) now fully rather than partially — **I'm withdrawing the "partially closed" qualifier from my previous comment.** Approval stands unqualified. Renderer duplication (#556) and the `--watch` listener hygiene note remain the only deferred items, both non-blocking.

Eight real defects across the review, four of them Codex's. Two of those landed on code I'd already approved, and both times the root cause was the same: my fixtures shared a blind spot with the implementation. Worth keeping both passes on changes that touch a signal operators are meant to trust.

Good to merge.

---

hojinzs (approved):
## ✅ Re-approve at `7cc2a4d` — rebase verified

**This time the branch really was force-pushed.** All ten commits have new SHAs and my previous approval sat on `1c6a93c`, which is no longer reachable:

```
git merge-base --is-ancestor 1c6a93c 7cc2a4d  →  NOT an ancestor
```

So that approval was orphaned and this one replaces it. (Worth contrasting with the `bb806ac` push I checked earlier, which was described as a rework but wasn't one — this is the genuine article, and the difference matters for whether an approval survives.)

### What actually changed

`main` moved `4ba98cb → 396bacb` while this PR was open, picking up #550, #551 and #552, and the branch was rebased onto it. That accounts for the large apparent diff — the raw `1c6a93c..7cc2a4d` range shows ~1180 insertions, but almost all of it is other people's merged work arriving via the new base.

The PR's own delta against current `main` is unchanged in shape — the same 16 files, 981 insertions, no scope creep.

**The reviewed liveness code came through the rebase byte-identical.** `git diff 1c6a93c 7cc2a4d -- daemon-liveness.ts stop.ts` is empty. The only differences inside files I'd reviewed are `main`'s new `skipped` summary counter landing in `status.ts`, `status-surface.ts` and `snapshot-builder.ts` — not edits to the liveness work.

### Re-verified against the new base

A rebase onto a moved base is exactly where a green pre-rebase run stops meaning anything, so I re-ran everything rather than carrying the old result forward. Clean install, fresh build:

`pnpm lint` ✅ · `pnpm typecheck` ✅ (0 errors) · `pnpm test` ✅ · `pnpm build` ✅ — note the suites that grew from the new base and still pass alongside this change: `tracker-github` 105, `orchestrator` 245, `cli` 514. CI `Test` and `Container Smoke` green on `7cc2a4d`.

Behavioral matrix re-run against the rebased binary, since `renderLegacyStatus` now emits an extra `Skipped` line in the same summary block I'd reviewed:

| | check | result |
|---|---|---|
| 1 | #546 incident — no daemon, 22d tick | `● stopped` + restart hint + last-known state ✅ |
| 2 | (c) foreground daemon outside workspace, via `lock.cwd` | `daemonRunning=true` ✅ |
| 3 | round-4 — live daemon, no snapshot | `daemonRunning=true` ✅ |
| 4 | round-7 — `effectivePollIntervalMs: 300000`, Codex quota healthy | not stale ✅ |
| 5 | control — `effectivePollIntervalMs: 30000`, same tick | `(stale)` ✅ |

Row 5 stays in every run for the reason that cost me twice on this PR: without it, rows 3 and 4 are equally consistent with a fix that silently disabled the signal.

### Standing state

All blockers closed, (d) fully. Deferred and non-blocking: renderer banner duplication (#556) and the `--watch` signal-listener hygiene note. Approval stands at `7cc2a4d`. Good to merge.

---

PR #601, opened 2026-08-20
fix(standalone): acceptance findings from the standalone project model

### LINKED ISSUE #600: feat(cli)!: address standalone projects by folder and remove 'project add'
## Context

Standalone projects (#561) shipped with `gh-symphony project add <dir>` as a prerequisite: a folder
had to be registered before `project start|status|stop` could address it, and each registration
overwrote `activeProject`, which is what those commands resolve. The final acceptance test surfaced
what that costs:

- `project start` always started the most recently registered project. Running two projects against
  one repository — the whole point of the model — required re-registering to flip which one was
  active, and two terminals could not drive two projects.
- A running instance kept the tracker configuration captured at registration time. Editing
  `pickup_labels` in `WORKFLOW.md` had no effect until re-registration, and during the acceptance
  test one project dispatched another project's issue because of it.
- Disjointness validation ran only at registration, so it could not see whether an overlapping
  project was actually running.

Registration never asked the operator for anything: every field is derived from the folder path and
its `WORKFLOW.md`. Design D2 already states the intended relationship — "the project folder is the
source of truth; `config.json` is registration/derived state" — so the prerequisite inverted it.

## Decision

Address standalone projects by folder, and derive their configuration on every start.

- Remove `gh-symphony project add`.
- `project start|status|stop` resolve the project from the working directory, or from
  `--project-dir <path>`. The project id stays a pure function of the folder path, so the folder
  alone identifies the runtime.
- `project start` re-derives the runtime configuration from the folder's `WORKFLOW.md` and caches it
  under `<config-dir>/projects/<project-id>/`. An edited workflow takes effect on the next start.
- `project status|stop` do not read `WORKFLOW.md` at all, so a daemon remains stoppable after its
  workflow file is edited away.
- Tracker-mapping disjointness is validated at start. An overlap with a *running* project is refused
  outright; an overlap with a stopped one asks for confirmation (and fails in non-interactive shells).
- `activeProject` is no longer consulted by the `project` namespace. It remains internal to the
  repo-embedded runtime, which has a single project per runtime root.
- `WORKFLOW.md` gains `repository.clone_url` so a project can point at a mirror, an Enterprise host,
  or a local path while keeping `owner/name` as its tracker and cache identity.
- `repository.slug` is the discriminator between the two modes: a workflow without it is
  repo-embedded, and the error says so.

This restores the CLI's directory-context convention. The 2026-05-10 restructure removed
`--project-id` on the grounds that "the current repository directory" identifies the target and
pointed at `--repo-dir` as the explicit form; `--project-dir` is the same idiom for a model whose
unit is a folder rather than a checkout.

## Acceptance

- `cd <project-dir> && gh-symphony project start` runs with no prior registration.
- Two projects with disjoint `pickup_labels` run concurrently against one repository, each addressed
  by its own folder.
- Starting a project whose mapping overlaps a running project is refused.
- `project status|stop` work from the folder, including after `WORKFLOW.md` is removed.
- `doctor --project-dir <path>` selects the same project.
- The standalone Docker E2E starts both projects by folder, with no registration step.

## Notes

Breaking change to an unreleased surface: the published CLI (0.7.14) has no `project` commands, so
no migration is required. Found during the #561 final acceptance test.

moncher-dev:
## TL;DR

- Standalone projects are now addressed by their folder: use the current directory or `--project-dir <path>` for `project start|status|stop` and `doctor`.
- `project start` derives and caches configuration from the folder's `WORKFLOW.md` every time; `project add` and project-namespace reliance on `activeProject` are removed.
- Multiple label-disjoint projects can run against one repository, while overlap validation is serialized and refuses overlap with a running project.

## Change-point diagram

```text
project folder / --project-dir
        |
        +-- start --> read WORKFLOW.md --> derive config --> overlap guard --> cache + runtime
        |
        +-- status/stop -------------> path-derived project id ---------> cached runtime
        |
        +-- doctor ------------------> same path-derived project id

repository.slug      = tracker/cache identity (owner/name)
repository.clone_url = optional checkout source (mirror, GHES, or local path)
```

## Start here

1. Review `packages/cli/src/commands/project.ts` for folder identity, per-start derivation, locking, and overlap behavior.
2. Review `packages/cli/src/commands/project-lifecycle.test.ts` and `packages/cli/src/commands/project.test.ts` for folder lifecycle, removed-workflow stop/status, overlap, and re-derivation coverage.
3. Review `e2e/run-standalone-project-e2e.sh` for two concurrent projects sharing one repository without registration.
4. Review `packages/cli/src/commands/doctor.ts` and `packages/cli/src/index.test.ts` for `doctor --project-dir` selection.

## Requirement coverage

| Requirement | Implementation / evidence |
| --- | --- |
| Remove `project add`; address start/status/stop by cwd or `--project-dir` | `project.ts`, `index.ts`, `project-lifecycle.test.ts` |
| Re-derive and cache configuration on every start | `deriveStandaloneProject` + re-derivation test in `project.test.ts` |
| Status/stop work without reading `WORKFLOW.md` | `project.ts` cached-runtime path + removed-workflow lifecycle test |
| Validate disjointness at start; reject running overlap; confirm stopped overlap | `findOverlappingProjects`, config-wide critical section, overlap/concurrency tests |
| Project namespace ignores `activeProject` | folder-derived `standaloneProjectId` lifecycle path |
| Support `repository.clone_url` while retaining slug identity | parser/config tests and shared-cache tests |
| Distinguish standalone mode with `repository.slug` and explain repo-embedded mode | `parseRepository` error + `project.test.ts` |
| `doctor --project-dir` selects the same project | Commander option forwarding + `index.test.ts` |
| Two disjoint projects run concurrently without registration | TC-13 and `pnpm e2e:standalone-project` |

## Changes

- Added folder-derived standalone identity and per-start workflow/config derivation.
- Moved overlap checks to start and serialized scan/liveness/config persistence under the config-wide lock.
- Applied workflow pickup-label policy in tracker candidate listing while preserving list metadata.
- Added clone URL override support and corrected shared bare-cache behavior, workspace placement, and shadow warnings found during acceptance.
- Updated README, packaged CLI documentation, architecture/configuration docs, TC-13, and the CLI minor changeset.

## Evidence

- `pnpm lint` — passed.
- `pnpm test` — passed; orchestrator test files run serially to avoid cross-file contention between Git-heavy integration suites while preserving explicit within-test concurrency.
- `pnpm typecheck` — passed.
- `pnpm build` — passed.
- `pnpm exec vitest run packages/cli/src/commands/project.test.ts packages/cli/src/index.test.ts packages/tracker-github/src/tracker-github.test.ts` — 144 tests passed.
- `pnpm --filter @gh-symphony/orchestrator exec vitest run src/git.test.ts -t 'serializes concurrent cache initialization for the same repository'` — passed.
- `pnpm --filter @gh-symphony/orchestrator test -- --no-file-parallelism` — 295 tests passed.
- `pnpm e2e:standalone-project` — passed; two folder-addressed projects dispatched label-disjoint issues against one shared repository.

## Risks & rollback

- Risk: folder canonicalization defines project identity, so moving a project folder intentionally creates a different runtime identity.
- Risk: stopped overlapping projects require an interactive confirmation and fail closed in non-interactive shells.
- Risk: this is a breaking change to the unreleased standalone command surface; published CLI 0.7.14 has no `project` commands.
- Rollback: revert this PR to restore registration-based standalone behavior; cached project directories are derived state and can be discarded.

## Changed files

- CLI lifecycle/config/doctor: `packages/cli/src/commands/{project,doctor,start,status,stop}.ts`, `packages/cli/src/config.ts`, `packages/cli/src/index.ts`
- Policy and trackers: `packages/core/src/workflow/pickup-labels.ts`, tracker adapters/tests
- Runtime/cache acceptance fixes: orchestrator service and repository-cache files/tests
- Documentation and blackbox coverage: `README.md`, `packages/cli/README.md`, `docs/{architecture,configuration}.md`, `AGENT_TEST.md`, TC-13/E2E script
- Release: `.changeset/standalone-project-final-test-fixes.md` (`@gh-symphony/cli`: minor)

## Human validation

- [ ] Run `project start|status|stop` from a standalone folder and with `--project-dir`.
- [ ] Edit `pickup_labels`, restart, and confirm the updated mapping is used.
- [ ] Confirm running overlap is refused and stopped overlap prompts interactively.
- [ ] Remove `WORKFLOW.md` after daemon start and confirm status/stop remain available.
- [ ] Confirm `doctor --project-dir <path>` selects the same cached runtime.

## Post-merge validation

- None required beyond normal release verification; Docker TC-13 covers the standalone blackbox flow before merge.

### DIFF
--- a/packages/cli/src/commands/doctor.ts
+++ b/packages/cli/src/commands/doctor.ts
@@ -689,7 +689,7 @@ async function checkWorkflow(
         ? `WORKFLOW.md was not found at the registered project path ${workflowPath}.`
         : "WORKFLOW.md was not found in the repository root.",
       remediation: externalWorkflowPath
-        ? `Restore ${workflowPath} or re-register the project with 'gh-symphony project add'.`
+        ? `Restore ${workflowPath}; the folder's workflow is derived on every 'gh-symphony project start'.`
         : "Run 'gh-symphony workflow init' in this repository or add a valid WORKFLOW.md at the repo root.",
     };
   }
--- a/packages/cli/src/commands/doctor.ts
+++ b/packages/cli/src/commands/doctor.ts
@@ -50,6 +50,7 @@ import {
   buildPriorityDriftDiagnostics,
 } from "../priority-diagnostics.js";
 import { inspectManagedProjectSelection } from "../project-selection.js";
+import { standaloneProjectId } from "./project.js";
 import {
   createSupportBundle,
   type SupportBundleSummary,
@@ -244,7 +245,7 @@ const DEFAULT_DEPENDENCIES: DoctorDependencies = {
 const MINIMUM_NODE_MAJOR = 24;
 const MINIMUM_NODE_VERSION = `v${MINIMUM_NODE_MAJOR}.0.0`;
 const DOCTOR_USAGE =
-  "Usage: gh-symphony doctor [--project-id <project-id>] [--fix] [--smoke] [--issue <owner/repo#number>] [--bundle [path]]";
+  "Usage: gh-symphony doctor [--project-id <project-id>] [--project-dir <path>] [--fix] [--smoke] [--issue <owner/repo#number>] [--bundle [path]]";
 
 type GitInstallationState =
   | {
@@ -268,6 +269,19 @@ function parseDoctorArgs(args: string[]): ParsedDoctorArgs {
         return parsed;
       }
       parsed.projectId = value;
+      i += 1;
+      continue;
+    }
+
+    if (arg === "--project-dir") {
+      const value = args[i + 1];
+      if (!value || value.startsWith("-")) {
+        parsed.error = `Option '${arg}' argument missing`;
+        return parsed;
+      }
+      // Standalone projects are addressed by folder everywhere else, and the
+      // identifier is derived from the path.
+      parsed.projectId = standaloneProjectId(value);
       i += 1;
       continue;
     }
@@ -656,9 +670,12 @@ async function checkGitInstallation(
 
 async function checkWorkflow(
   repoRoot: string,
-  deps: Pick<DoctorDependencies, "readFile" | "parseWorkflowMarkdown">
+  deps: Pick<DoctorDependencies, "readFile" | "parseWorkflowMarkdown">,
+  externalWorkflowPath?: string | null
 ): Promise<WorkflowCheckState> {
-  const workflowPath = join(repoRoot, "WORKFLOW.md");
+  // A standalone project owns its policy outside the repository, so the cwd
+  // lookup would report a missing file for a perfectly healthy setup.
+  const workflowPath = externalWorkflowPath ?? join(repoRoot, "WORKFLOW.md");
   let markdown: string;
 
   try {
@@ -668,9 +685,12 @@ async function checkWorkflow(
       status: "fail",
       reason: "missing",
       workflowPath,
-      summary: "WORKFLOW.md was not found in the repository root.",
-      remediation:
-        "Run 'gh-symphony workflow init' in this repository or add a valid WORKFLOW.md at the repo root.",
+      summary: externalWorkflowPath
+        ? `WORKFLOW.md was not found at the registered project path ${workflowPath}.`
+        : "WORKFLOW.md was not found in the repository root.",
+      remediation: externalWorkflowPath
+        ? `Restore ${workflowPath} or re-register the project with 'gh-symphony project add'.`
+        : "Run 'gh-symphony workflow init' in this repository or add a valid WORKFLOW.md at the repo root.",
     };
   }
 
@@ -2030,12 +2050,21 @@ export async function runDoctorDiagnostics(
     );
   }
 
-  const workflow = await checkWorkflow(process.cwd(), deps);
+  const externalWorkflowPath =
+    resolvedProjectConfig?.kind === "resolved" &&
+    resolvedProjectConfig.projectConfig.workflowSource?.type === "external"
+      ? resolvedProjectConfig.projectConfig.workflowSource.path
+      : null;
+  const workflow = await checkWorkflow(
+    process.cwd(),
+    deps,
+    externalWorkflowPath
+  );
   if (workflow.status === "pass") {
     checks.push(
       passCheck(
         "workflow_file",
-        "Repository WORKFLOW.md",
+        externalWorkflowPath ? "Project WORKFLOW.md" : "Repository WORKFLOW.md",
         `WORKFLOW.md parsed successfully (${workflow.format}).`,
         { path: workflow.workflowPath, format: workflow.format }
       )
@@ -2044,7 +2073,7 @@ export async function runDoctorDiagnostics(
     checks.push(
       failCheck(
         "workflow_file",
-        "Repository WORKFLOW.md",
+        externalWorkflowPath ? "Project WORKFLOW.md" : "Repository WORKFLOW.md",
         workflow.summary,
         workflow.remediation,
         {
--- a/packages/cli/src/commands/project.ts
+++ b/packages/cli/src/commands/project.ts
@@ -1,5 +1,5 @@
 import { createHash } from "node:crypto";
-import { readFile } from "node:fs/promises";
+import { readFile, readdir } from "node:fs/promises";
 import { basename, resolve } from "node:path";
 import * as p from "@clack/prompts";
 import {
@@ -12,12 +12,11 @@ import startCommand from "./start.js";
 import statusCommand from "./status.js";
 import stopCommand from "./stop.js";
 import {
-  loadGlobalConfig,
   loadProjectConfig,
-  saveGlobalConfig,
   saveProjectConfig,
   type CliProjectConfig,
 } from "../config.js";
+import { resolveDaemonLiveness } from "../daemon-liveness.js";
 
 type PickupLabels = { include: string[]; exclude: string[] };
 
@@ -28,13 +27,26 @@ type Mapping = {
   labels: PickupLabels;
 };
 
-export async function registerStandaloneProject(
+/**
+ * The project folder is the source of truth: every field below is derived from
+ * it, so the stored config is a cache that each start refreshes rather than a
+ * registration the operator has to perform first.
+ */
+export async function deriveStandaloneProject(
   projectDirInput: string,
   options: Pick<GlobalOptions, "configDir">
 ): Promise<CliProjectConfig> {
   const projectDir = resolve(projectDirInput);
   const workflowPath = resolve(projectDir, "WORKFLOW.md");
-  const workflow = parseWorkflowMarkdown(await readFile(workflowPath, "utf8"));
+  let markdown: string;
+  try {
+    markdown = await readFile(workflowPath, "utf8");
+  } catch {
+    throw new Error(
+      `No WORKFLOW.md in ${projectDir}. Run this command from a standalone project folder or pass --project-dir <path>.`
+    );
+  }
+  const workflow = parseWorkflowMarkdown(markdown);
   const repository = parseRepository(workflow.repository);
   const adapter = workflow.tracker.kind ?? "github-project";
   const bindingId =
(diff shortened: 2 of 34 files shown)


packages/cli/src/commands/doctor.ts
```diff
@@ -272,6 +273,19 @@ function parseDoctorArgs(args: string[]): ParsedDoctorArgs {
...
 
+    if (arg === "--project-dir") {
+      const value = args[i + 1];
+      if (!value || value.startsWith("-")) {
+        parsed.error = `Option '${arg}' argument missing`;
+        return parsed;
+      }
+      // Standalone projects are addressed by folder everywhere else, and the
+      // identifier is derived from the path.
+      parsed.projectId = standaloneProjectId(value);
```
hojinzs, line 284: **P1 — `doctor --project-dir` is parsed here but never reaches the CLI, so the #600 acceptance criterion fails.**

`parseDoctorArgs` learns `--project-dir` (and the `DOCTOR_USAGE` string advertises it), but the Commander registration in `packages/cli/src/index.ts` (the `doctor` command) neither declares a `--project-dir` option nor forwards it into the args array — it only pushes `--project-id/--fix/--smoke/--bundle/--issue`. Because the command uses `.allowExcessArguments(false)`, Commander rejects the flag before the handler runs.

Verified against the built CLI:

```
$ gh-symphony doctor --project-dir <dir>
error: unknown option '--project-dir'   (exit 1)
$ gh-symphony doctor --help   # no --project-dir listed
```

The unit test exercises `runDoctorDiagnostics`/`parseDoctorArgs` directly, so it passes while the end-to-end surface is broken. #600 lists "`doctor --project-dir <dir>` selects the same project" as an acceptance criterion, and the PR body repeats it — so this is an unmet acceptance condition, not a nit. Register the option on the `doctor` command and forward it (mirroring `--project-id`), and add an `index.test.ts` case that drives it through Commander so the wiring is covered.

---

packages/cli/src/commands/doctor.ts
```diff
@@ -668,9 +685,12 @@ async function checkWorkflow(
...
       reason: "missing",
       workflowPath,
-      summary: "WORKFLOW.md was not found in the repository root.",
-      remediation:
-        "Run 'gh-symphony workflow init' in this repository or add a valid WORKFLOW.md at the repo root.",
+      summary: externalWorkflowPath
+        ? `WORKFLOW.md was not found at the registered project path ${workflowPath}.`
+        : "WORKFLOW.md was not found in the repository root.",
+      remediation: externalWorkflowPath
+        ? `Restore ${workflowPath} or re-register the project with 'gh-symphony project add'.`
```
hojinzs: **P2 — remediation points at the removed `project add` command.**

This PR removes `gh-symphony project add`, but the missing-workflow remediation still tells the operator to `re-register the project with 'gh-symphony project add'`. Following it yields an unknown-command error. Suggest wording aligned with the new folder-addressed model, e.g. `Restore ${workflowPath} (the folder's WORKFLOW.md is derived on every 'gh-symphony project start').`

Same stale command survives in `packages/cli/README.md` (lines 344 and 439), which AGENTS.md §4 requires updating alongside a CLI-surface change — the root `README.md` and `docs/configuration.md` were updated but the package README was not. (This overlaps Codex's P2 thread; flagging the `doctor.ts` string it didn't cover.)

---

packages/tracker-github/src/orchestrator-adapter.ts
```diff
@@ -16,7 +17,15 @@ import {
...
 export const githubProjectTrackerAdapter: OrchestratorTrackerAdapter = {
   async listIssues(project, dependencies = {}) {
-    return listProjectIssues(project, dependencies);
+    const issues = await listProjectIssues(project, dependencies);
+    const filtered = filterIssuesByPickupLabels(
+      issues,
+      project
+    ) as TrackedIssueList;
+    if (filtered !== issues) {
+      filtered.rateLimits = (issues as TrackedIssueList).rateLimits;
```
hojinzs, line 26: **P2 — pickup-label filtering silently drops `skippedItems` from the list.**

`listProjectIssues` attaches both `rateLimits` and `skippedItems` to the returned `TrackedIssueList` (`adapter.ts` sets `.skippedItems` right after `.rateLimits`). When a pickup filter actually runs, `filterIssuesByPickupLabels` returns a fresh array and this wrapper copies over `rateLimits` but not `skippedItems`. Downstream, `service.ts` reads `(issues as TrackedIssueList).skippedItems` for the snapshot's `skipped` count and the `[orchestrator] skipped N item(s)…` operator log — so exactly in the label-partitioned multi-project setup this PR enables, that signal disappears whenever `pickup_labels` is configured.

Reproduced against the built core:

```
identifiers      : [ 'a/b#1' ]
rateLimits kept  : {"remaining":4000}
skippedItems kept: undefined
```

Carry `skippedItems` across alongside `rateLimits` (or have `filterIssuesByPickupLabels` preserve both non-enumerable properties). A regression test asserting `skippedItems` survives filtering would lock it in — the three new tests here only check `rateLimits`.

---

hojinzs (changes requested):
## Review — standalone acceptance fixes (#601)

I traced each fix against #600/#601, ran the full gate locally (`pnpm build` / `typecheck` / `test` / `lint` all green — 533 CLI tests pass), and smoke-tested the CLI surface end to end against a throwaway repo + two file-tracker projects.

### What holds up
- **Folder-addressed lifecycle** — `project start|status|stop` resolve by folder with no registration; `project list` shows both projects with `workspaceDir` under each folder's `workspace.root`. ✓
- **Overlap guard** — a mapping overlapping a *running* project is refused outright; overlapping a *stopped* one is refused non-interactively. ✓
- **Stop after WORKFLOW.md removal** — `status`/`stop` keep working once the workflow file is gone. ✓
- **`workspace.root` (spec 9.1)** — issue workspaces land under `<project-dir>/.runtime/workspaces`, not the state dir. ✓
- **Shadow warning from the bare cache** — with a remote `clone_url` and a `WORKFLOW.md` committed to the target repo, the status snapshot names the repository correctly. ✓ (Note: a *local* bare-repo `clone_url` is treated as a local checkout and `access(<baredir>/WORKFLOW.md)` never matches, so the shadow check no-ops there — edge case, not blocking.)
- **Clone-URL re-point** and **`--config` export to `GH_SYMPHONY_CONFIG_DIR`** are covered by the new unit tests and the `vitest.setup.ts` sandbox, which correctly stops the suite poisoning `~/.gh-symphony`.

No over-engineering spotted — the change set maps 1:1 to the five defects plus the #600 CLI redesign.

### Blocking
- **P1 — `doctor --project-dir` never reaches the CLI.** The flag is parsed in `doctor.ts` and advertised in the usage string, but the Commander `doctor` command in `index.ts` neither registers nor forwards it, so `gh-symphony doctor --project-dir <dir>` errors with `unknown option`. #600 lists this exact invocation as an acceptance criterion; the unit test passes only because it bypasses Commander. Inline detail + repro on the diff.

### Non-blocking (worth addressing this round)
- **P2** — missing-workflow remediation and `packages/cli/README.md` still reference the removed `project add` (AGENTS.md §4 requires the package README to move with the CLI surface).
- **P2** — `skippedItems` is dropped from the tracker list whenever `pickup_labels` filtering runs, losing the orchestrator's skipped-count log/snapshot in the multi-project case this PR targets.

### On the two existing Codex threads
- *Serialize overlap checks* (P1 on `project.ts`): the race is real but narrow — it needs two previously-unseen overlapping folders started at the same instant, before either daemon is up. Worth a config-wide lock, but I'd rate it P2, not a blocker.
- *Move pickup-label policy out of orchestration-core* (P1 on `pickup-labels.ts`): reasonable per AGENTS.md's policy/orchestration-core split — the predicate is Policy-layer, currently living under `core/src/orchestration`. P3 structural cleanup; the behavior is correct.

Requesting changes for the P1. Once the `doctor --project-dir` wiring lands (with a Commander-level test) and the stale `project add` docs/remediation are cleaned up, I expect to approve — the P2s and Codex items can be folded in or answered on their threads.

---

packages/cli/src/commands/project.ts
```diff
@@ -68,49 +80,140 @@ export async function registerStandaloneProject(
   };
 
   const overlap = await findOverlappingProjects(options.configDir, config);
+  const running = overlap.filter((entry) => entry.running);
```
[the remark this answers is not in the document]
moncher-dev: Resolved in d4d3bc1. `deriveStandaloneProject` now holds the existing config-wide lock across the overlap scan, liveness decision, confirmation, and project-config write. A concurrent-derivation regression test verifies that exactly one of two overlapping starts can succeed.

packages/core/src/workflow/pickup-labels.ts
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * Candidate listing may be narrowed by `tracker.pickup_labels`, which is how
+ * two projects sharing one repository stay disjoint. State lookups and revive
+ * paths must stay unfiltered, so this is applied only where candidates are
+ * enumerated.
+ */
+export function filterIssuesByPickupLabels<T extends TrackedIssue>(
+  issues: T[],
+  project: Pick<OrchestratorProjectConfig, "tracker">
+): T[] {
```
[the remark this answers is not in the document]
moncher-dev: Resolved in d4d3bc1. The pickup-label predicate moved from `core/src/orchestration` to `core/src/workflow`, and the public export now comes directly from that workflow-policy boundary.

packages/cli/src/commands/project.ts
```diff
@@ -238,45 +354,80 @@ function slugify(value: string): string {
     .replace(/^-|-$/g, "");
 }
 
+const USAGE =
+  "Usage: gh-symphony project <list|start|status|stop> [--project-dir <path>] [runtime options]\n";
```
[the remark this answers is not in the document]
moncher-dev: Resolved in d4d3bc1. `packages/cli/README.md` now documents folder-addressed `start|status|stop`, `--project-dir`, and per-start derivation without `project add`; `docs/architecture.md` now describes folder-addressed standalone derivation instead of registration.

packages/cli/src/commands/doctor.ts
moncher-dev: Resolved in d4d3bc1. Commander now declares `doctor --project-dir <path>` and forwards it to the diagnostics handler. The new `index.test.ts` regression drives the option through `runCli` and asserts that the folder-derived project ID is selected.

packages/cli/src/commands/doctor.ts
moncher-dev: Resolved in d4d3bc1. The missing-workflow remediation now tells the operator to restore the folder's workflow and explains that `project start` derives it on every run. The stale `project add` guidance in the packaged CLI README was removed as part of the same change.

packages/tracker-github/src/orchestrator-adapter.ts
moncher-dev: Resolved in d4d3bc1. The GitHub adapter now copies `skippedItems` alongside `rateLimits` when pickup-label filtering returns a new list. A regression test verifies that a skipped Project item remains visible after filtering.

packages/cli/src/commands/project.ts
```diff
@@ -67,50 +80,143 @@ export async function registerStandaloneProject(
...
       );
     }
-  }
+    if (overlap.length > 0 && !process.stdin.isTTY) {
+      throw new Error(
+        `Tracker mapping overlaps project(s): ${describe(overlap)}. Re-run interactively to confirm.`
+      );
+    }
+    if (overlap.length > 0) {
+      const confirmed = await p.confirm({
```
hojinzs, line 101: **P2 — the config lock is stolen as stale after 30s, and this critical section now spans an unbounded human prompt.**

Not blocking, and the lock is a genuine improvement over the previous no-lock state — but it does not hold for the one case it now brackets.

`withConfigLock` treats *any* lock older than `CONFIG_LOCK_TTL_MS` (30s) as stale, and the age check in `isStaleConfigLock` runs **before** the pid-liveness check, so it fires unconditionally even when the owner is alive and simply waiting:

```ts
const ageMs = Math.abs(Date.now() - Date.parse(record.startedAt));
if (!Number.isFinite(ageMs) || ageMs > CONFIG_LOCK_TTL_MS) {
  return true;                       // <- wins before process.kill(pid, 0)
}
```

That was a safe assumption while every holder was a bounded file write. This PR puts `await p.confirm(...)` inside the section, which blocks for as long as the operator takes to answer.

Reproduced against `packages/cli/src/config.ts` — a lock owned by a **live** pid, aged 45s:

```
lock held by LIVE pid 8579 for 45s; TTL is 30s
>>> second caller ENTERED critical section after 12ms
owner A lock file still present: true
```

So: operator A hits the overlap prompt and steps away; 30s later operator B's `project start` takes the lock, runs its own overlap scan, and `saveProjectConfigWithinLock` then writes with no lock actually held — reopening exactly the concurrent-start race the lock was added to close.

Cheapest fix is to keep the prompt out of the critical section: scan + decide + write under the lock, and if a confirmation is needed, release, prompt, then re-acquire and re-validate before writing. Alternatively heartbeat the lock file while it is held so a live owner is never aged out.

Related, and worth a sentence in the same fix: `findOverlappingProjects` reports `running: false` for a project whose config is written but whose daemon has not come up yet (the config write happens before `startCommand`). In that window the "refused outright" guarantee for a running project degrades to the confirm path. Non-interactive shells still fail closed, so this is a note rather than a defect.

---

hojinzs (approved):
## Review round 2 — approving

Re-reviewed at `516487d`. All six threads from round 1 are genuinely fixed, not just replied to — I verified each against the built CLI rather than taking the thread replies at face value.

### Full gate — green
`pnpm build` / `pnpm typecheck` / `pnpm lint` / `pnpm test` all pass (exit 0). 538 CLI tests, 227 core, 118 tracker-github, 295 orchestrator.

### Round-1 threads — verified

| Thread | Verdict |
| --- | --- |
| **P1 `doctor --project-dir` never reached Commander** (mine) | **Fixed.** Option registered and forwarded in `index.ts`; `doctor --help` lists it; from an unrelated cwd, `doctor --project-dir <alpha>` resolves `projectId = alpha-c64f0254` and validates `<alpha>/WORKFLOW.md` as *Project* WORKFLOW.md rather than the repo root. `index.test.ts` now drives it through `runCli`. |
| **P2 stale `project add` in remediation + package README** (mine) | **Fixed.** Remediation now says "Restore `<path>`; the folder's workflow is derived on every `gh-symphony project start`." Only surviving `project add` references are in dated design records under `docs/designs/`, which correctly stay historical. |
| **P2 `skippedItems` dropped by pickup filtering** (mine) | **Fixed** + regression test. Verified against the built core: `identifiers ['a/b#1']`, `rateLimits {"remaining":4000}`, `skippedItems [{"itemId":"X","reason":"draft"}]` — previously `undefined`. |
| **P1 serialize overlap checks** (Codex) | **Fixed.** `deriveStandaloneProject` now holds the config lock across scan → liveness → confirm → write, and that lock is a real cross-process `link(2)` lock, not an in-process mutex. Confirmed no self-deadlock: `loadProjectConfig` and `resolveDaemonLiveness` are both lock-free. One residual caveat inline below. |
| **P1 pickup-label policy in orchestration-core** (Codex) | **Fixed.** Moved to `core/src/workflow/pickup-labels.ts` and exported from the workflow boundary — and the duplicate copy in `tracker-file` was deleted in favor of it, which is a net simplification. |
| **P2 packaged CLI docs** (Codex) | **Fixed.** `packages/cli/README.md` and `docs/architecture.md` both describe folder-addressed derivation. |

### Smoke test — #600 acceptance criteria

No Docker daemon in this environment, so TC-13 could not run; I drove the built CLI against two file-tracker project folders and a local bare repo instead.

- `cd <folder> && project start --once` — runs with **no prior registration**, derives `alpha-c64f0254` from the path. ✓
- Second **label-disjoint** project against the same repository derives and starts independently. ✓
- Overlapping mapping vs a **stopped** project, non-interactive → refused: `Tracker mapping overlaps project(s): …/alpha. Re-run interactively to confirm.` No config cached for the rejected folder. ✓
- Overlapping mapping vs a **running** daemon → refused outright: `Tracker mapping overlaps running project(s): …/alpha. Stop them or make the mappings disjoint…` ✓
- **Re-derivation**: edited `pickup_labels` `beta` → `beta-v2`, restarted, cached config updated. ✓
- **`WORKFLOW.md` removed** → `project status` still renders and `project stop` still terminates the daemon (`Sent SIGTERM to orchestrator (PID 6567). Daemon stopped.`). ✓
- `project status --project-dir <beta>` from an unrelated cwd → `beta-16e1c631`. ✓
- `project list` shows both projects with their folders; `project add` is gone (`unknown command 'add'`). ✓
- `repository.clone_url` pointing at a local bare repo worked throughout. ✓
- Invocation-aware guidance is right: `Stop with: gh-symphony project stop`, not `repo stop`. ✓

Requirement coverage against #600 is complete. Still no over-engineering — the diff maps 1:1 to the acceptance findings plus the CLI redesign, and the pickup-label move removed code rather than adding a layer.

### Non-blocking, for a follow-up

One **P2** inline on `project.ts`: the config lock is aged out after 30s regardless of owner liveness, and the critical section now contains an interactive `p.confirm`, so a start left sitting at the overlap prompt loses its lock and the concurrent-start race reopens. Reproduced. Narrow (needs an operator to walk away mid-prompt while another start races), and strictly better than the pre-PR no-lock state, so it does not hold up this PR — but it should be tightened rather than left, since it undercuts the guarantee the lock was added for.

Approving. Worth landing, and I'd rather see the lock caveat fixed in its own small PR than spin this one another round.

---

hojinzs (approved):
## Review round 3 — approval stands, no re-review needed

The re-review request arrived on the **same commit I already approved** in round 2 (`516487d`). Nothing has changed since:

- **No new commits** — `516487d` is still the head; the last push was `fix(cli): preserve project status guidance after rebase` at 00:28, before the round-2 approval at 00:42.
- **CI green on this SHA** — `Test` and `Container Smoke` both succeeded.
- **Mergeable** — `mergeable_state: clean` against `main`.
- **Open-thread code state unchanged** — `packages/core/src/workflow/pickup-labels.ts` is still at the workflow-policy boundary, and `CONFIG_LOCK_TTL_MS` in `packages/cli/src/config.ts:30` is unchanged, matching what I recorded in round 2.

Round 2 already covered the full gate (`build` / `typecheck` / `lint` / `test`, all green), verified all six round-1 threads against the built CLI rather than the thread replies, and smoke-tested every #600 acceptance criterion (folder-addressed `start|status|stop`, disjoint concurrent projects, running/stopped overlap refusal, re-derivation, `WORKFLOW.md` removal, `doctor --project-dir`, `repository.clone_url`). Re-running that against identical bytes would add nothing.

Requirement coverage against #600 is complete, and I found no over-engineering — the diff maps 1:1 to the acceptance findings plus the CLI redesign.

### Still open, still non-blocking

The one unresolved thread of mine — the config lock aged out at 30s regardless of owner liveness, with an interactive `p.confirm` now inside the critical section — remains a **P2 follow-up**, not a merge blocker. It is strictly better than the pre-PR no-lock state, and the race it reopens needs an operator to walk away mid-prompt while another start races. I'd still rather see it tightened in its own small PR than spin this one another round.

The Codex `pickup-labels` thread is fixed (module moved to `core/src/workflow`, duplicate in `tracker-file` deleted); it is just left unresolved on GitHub.

**Approving — this is my final verdict on `516487d`. It's ready to merge; further review rounds on unchanged code would be pure ping-pong.**

---

PR #616, opened 2026-08-21
feat(orchestrator): harden global bare cache operations

### LINKED ISSUE #588: feat(orchestrator): harden global bare cache operations
## 배경

전역 bare clone cache는 현재 이슈 #564의 요구대로 mkdir 락과 TTL fetch를 제공한다. 운영 탄력성을 더 높이기 위한 후속 범위가 리뷰에서 확인되었다.

## 범위

- cache unavailable·락 타임아웃 시 workspace 직접 clone 폴백 검토
- 대형 clone 동안 lock stale 판정을 피하기 위한 heartbeat 또는 cache 전용 lock 정책 검토
- cache eviction/정리 정책 및 CLI 진단·정리 명령 검토

## 관련

- Follow-up for #564

moncher-dev:
## Issues

- Closed #588

## TL;DR

Global bare-cache failures no longer block workspace population: the orchestrator falls back to an observable direct clone. Cache operations use heartbeat-protected locks, and operators can inspect or conservatively prune cached repositories through the CLI.

## Change-point diagram

```text
workspace population
  ├─ cache available → heartbeat lock → reference clone/worktree
  └─ cache unavailable/timeout → operator warning → isolated direct clone

gh-symphony cache
  ├─ status → size, last-use, lock/stale-lock, worktree safety
  └─ prune  → age filter → deletion lock → age/worktree recheck → remove
```

## Start here

1. `packages/orchestrator/src/repository-cache.ts` — cache inventory, last-use tracking, stale-lock recovery, and conservative pruning.
2. `packages/orchestrator/src/git.ts` — heartbeat ownership, direct-clone fallback, reader protection, and standalone teardown handling.
3. `packages/cli/src/commands/cache.ts` — `cache status` and `cache prune` operator surface.
4. `packages/orchestrator/src/repository-cache.test.ts` and `packages/orchestrator/src/git.test.ts` — concurrency, fallback, stale-lock, and destructive-path regressions.

## Changes

- Fall back to a direct workspace clone when the shared cache is unavailable or its acquisition times out, with an operator-visible warning.
- Hold heartbeat-refreshed cache locks across large clone/fetch and active reference-reader operations.
- Track cache last use independently, recheck age and worktrees under the deletion lock, and safely recover abandoned stale locks.
- Add `gh-symphony cache status` and conservative `cache prune` commands with JSON/dry-run support and clear locked/unverifiable states.
- Add unit, destructive-path, concurrency, CLI, and Docker black-box coverage plus living documentation.
- Add a minor changeset for `@gh-symphony/cli`.

## Evidence

- `pnpm lint`
- `pnpm test`
- `pnpm typecheck`
- `pnpm build`
- `pnpm changeset status`
- Docker TC-15: rebuilt stack healthy; inventory reported `test-owner/test-repo` (24,814 bytes); dry-run retained it; real prune reclaimed it; final inventory empty; stack removed and fixture restored.
- GitHub checks: `Test` and `Container Smoke` passed on `c5d96284`.

## Risks & rollback

- Direct fallback consumes additional network bandwidth while cache storage or locking is degraded; the service warning makes this detectable.
- Pruning is intentionally conservative for live locks, linked worktrees, and unverifiable repositories; abandoned locks are reclaimed only after the existing 30-minute stale threshold.
- Rollback: revert this PR to restore the previous cache-only population path and remove the maintenance CLI surface. Existing bare caches remain compatible.

## Changed files

- Orchestrator cache/runtime: `packages/orchestrator/src/repository-cache.ts`, `packages/orchestrator/src/git.ts`, `packages/orchestrator/src/service.ts`
- CLI: `packages/cli/src/commands/cache.ts`, command registration/help, and CLI docs
- Tests: orchestrator Git/cache tests, CLI cache tests, and `e2e/scenarios/15-cache-maintenance.md`
- Living docs: `README.md`, `packages/cli/README.md`, `docs/architecture.md`, `docs/configuration.md`, `AGENT_TEST.md`
- Release: `.changeset/fuzzy-caches-rest.md`

## Human validation

- [ ] Confirm cache degradation warnings are visible in the operator log stream.
- [ ] Confirm `cache status` labels live locks, stale locks, and unverifiable entries clearly.
- [ ] Confirm `cache prune --dry-run` output matches the intended retention policy before using real prune in production.

## Post-merge validation

- [ ] Run `gh-symphony cache status --json` against a representative production cache root.
- [ ] Run `gh-symphony cache prune --dry-run --max-age-days 30 --json` and review retained/removal candidates.

### DIFF
--- a/packages/cli/src/commands/cache.ts
+++ b/packages/cli/src/commands/cache.ts
@@ -33,11 +33,15 @@ const handler = async (
       return;
     }
     for (const entry of entries) {
-      const state = entry.locked
-        ? "locked"
-        : entry.worktrees
-          ? `${entry.worktrees} linked worktree(s)`
-          : "idle";
+      const state = entry.staleLock
+        ? "stale-locked"
+        : entry.locked
+          ? "locked"
+          : entry.worktrees === null
+            ? "unverifiable"
+            : entry.worktrees > 0
+              ? `${entry.worktrees} linked worktree(s)`
+              : "idle";
       process.stdout.write(
         `${entry.repository}  ${formatBytes(entry.bytes)}  ${state}  ${entry.updatedAt}\n`
       );
--- a/packages/cli/src/commands/cache.ts
+++ b/packages/cli/src/commands/cache.ts
@@ -1,0 +1,80 @@
+import {
+  inspectGlobalRepositoryCache,
+  pruneGlobalRepositoryCache,
+} from "@gh-symphony/orchestrator";
+import type { GlobalOptions } from "../index.js";
+
+const DAY_MS = 24 * 60 * 60 * 1000;
+
+const handler = async (
+  args: string[],
+  options: GlobalOptions
+): Promise<void> => {
+  const [subcommand] = args;
+  if (subcommand === "status") {
+    const entries = await inspectGlobalRepositoryCache({
+      configDir: options.configDir,
+    });
+    if (options.json) {
+      process.stdout.write(
+        JSON.stringify(
+          {
+            entries,
+            totalBytes: entries.reduce((sum, entry) => sum + entry.bytes, 0),
+          },
+          null,
+          2
+        ) + "\n"
+      );
+      return;
+    }
+    if (entries.length === 0) {
+      process.stdout.write("Global repository cache is empty.\n");
+      return;
+    }
+    for (const entry of entries) {
+      const state = entry.locked
+        ? "locked"
+        : entry.worktrees
+          ? `${entry.worktrees} linked worktree(s)`
+          : "idle";
+      process.stdout.write(
+        `${entry.repository}  ${formatBytes(entry.bytes)}  ${state}  ${entry.updatedAt}\n`
+      );
+    }
+    return;
+  }
+  if (subcommand === "prune") {
+    const rawDays = valueAfter(args, "--max-age-days") ?? "30";
+    const days = Number(rawDays);
+    if (!Number.isFinite(days) || days < 0)
+      throw new Error("--max-age-days must be a non-negative number.");
+    const result = await pruneGlobalRepositoryCache({
+      configDir: options.configDir,
+      maxAgeMs: days * DAY_MS,
+      dryRun: args.includes("--dry-run"),
+    });
+    if (options.json) {
+      process.stdout.write(JSON.stringify(result, null, 2) + "\n");
+      return;
+    }
+    process.stdout.write(
+      `${result.dryRun ? "Would remove" : "Removed"} ${result.removed.length} cache entr${result.removed.length === 1 ? "y" : "ies"} (${formatBytes(result.reclaimedBytes)}); skipped ${result.skipped.length}.\n`
+    );
+    return;
+  }
+  throw new Error("Usage: gh-symphony cache <status|prune>");
+};
+
+function valueAfter(args: string[], flag: string): string | undefined {
+  const index = args.indexOf(flag);
+  return index >= 0 ? args[index + 1] : undefined;
+}
+
+function formatBytes(bytes: number): string {
+  if (bytes < 1024) return `${bytes} B`;
+  if (bytes < 1024 * 1024) return `${(bytes / 1024).toFixed(1)} KiB`;
+  return `${(bytes / (1024 * 1024)).toFixed(1)} MiB`;
+}
+
+export default handler;
--- a/packages/orchestrator/src/git.ts
+++ b/packages/orchestrator/src/git.ts
@@ -20,7 +20,6 @@ import {
   type WorkflowResolution,
 } from "@gh-symphony/core";
 import {
-  ensureGlobalBareRepositoryCache,
   RepositoryCacheUnavailableError,
   withGlobalBareRepositoryCache,
 } from "./repository-cache.js";
@@ -104,19 +103,23 @@ export async function syncRepositoryForRun(input: {
 
     try {
       try {
-        const bareRepositoryDirectory = await ensureGlobalBareRepositoryCache({
-          repository: input.repository,
-          requiredRef: input.requiredRef,
-        });
-        await runCommand("git", [
-          "clone",
-          "--filter=blob:none",
-          "--reference-if-able",
-          bareRepositoryDirectory,
-          "--dissociate",
-          sanitizeRepositoryCloneUrl(input.repository.cloneUrl),
-          tempRepositoryDirectory,
-        ]);
+        await withGlobalBareRepositoryCache(
+          {
+            repository: input.repository,
+            requiredRef: input.requiredRef,
+          },
+          async (bareRepositoryDirectory) => {
+            await runCommand("git", [
+              "clone",
+              "--filter=blob:none",
+              "--reference-if-able",
+              bareRepositoryDirectory,
+              "--dissociate",
+              sanitizeRepositoryCloneUrl(input.repository.cloneUrl),
+              tempRepositoryDirectory,
+            ]);
+          }
+        );
       } catch (error) {
         if (!(error instanceof RepositoryCacheUnavailableError)) {
           throw error;
--- a/packages/orchestrator/src/git.ts
+++ b/packages/orchestrator/src/git.ts
@@ -47,6 +47,7 @@ export async function cloneRepositoryForRun(input: {
   repository: RepositoryRef;
   targetDirectory: string;
   requiredRef?: string;
+  onCacheUnavailable?: (error: RepositoryCacheUnavailableError) => void;
 }): Promise<string> {
   const result = await syncRepositoryForRun(input);
   return result.repositoryDirectory;
@@ -56,6 +57,7 @@ export async function syncRepositoryForRun(input: {
   repository: RepositoryRef;
   targetDirectory: string;
   requiredRef?: string;
+  onCacheUnavailable?: (error: RepositoryCacheUnavailableError) => void;
 }): Promise<RepositorySyncResult> {
   await mkdir(input.targetDirectory, { recursive: true });
   const repositoryDirectory = join(input.targetDirectory, "repository");
@@ -124,6 +126,7 @@ export async function syncRepositoryForRun(input: {
         if (!(error instanceof RepositoryCacheUnavailableError)) {
           throw error;
         }
+        input.onCacheUnavailable?.(error);
         await cloneRepositoryDirectly(
           input.repository,
           tempRepositoryDirectory
@@ -151,6 +154,7 @@ export async function ensureIssueWorkspaceRepository(input: {
   issueIdentifier?: string;
   branchTemplate?: string | null;
   baseBranch?: string | null;
+  onCacheUnavailable?: (error: RepositoryCacheUnavailableError) => void;
 }): Promise<string> {
   if (input.populateStrategy === "worktree-cache") {
     return ensureIssueWorkspaceWorktree(input);
@@ -173,6 +177,7 @@ export async function ensureIssueWorkspaceRepository(input: {
         requiredRef: input.pullRequestBranch
           ? `refs/heads/${input.pullRequestBranch.headRefName}`
           : undefined,
+        onCacheUnavailable: input.onCacheUnavailable,
       });
 
   if (input.pullRequestBranch && !dirtyExistingWorkspaceAllowed) {
@@ -192,6 +197,9 @@ export async function removeIssueWorkspaceWorktree(input: {
   issueIdentifier: string;
   branchTemplate?: string | null;
 }): Promise<void> {
+  if (await isDirectory(join(input.repositoryDirectory, ".git"))) {
+    return;
+  }
   const branch = renderIssueBranchName({
     template: input.branchTemplate,
     projectSlug: input.projectSlug,
@@ -286,6 +294,7 @@ async function ensureIssueWorkspaceWorktree(input: {
       if (!(error instanceof RepositoryCacheUnavailableError)) {
         throw error;
       }
+      input.onCacheUnavailable?.(error);
       await cloneRepositoryDirectly(input.repository, repositoryDirectory);
       const baseBranch =
         input.pullRequestBranch?.headRefName ??
@@ -305,6 +314,15 @@ async function ensureIssueWorkspaceWorktree(input: {
     if (!repositoryDirectoryExisted) {
       await rm(repositoryDirectory, { recursive: true, force: true });
     }
+    throw error;
+  }
+}
+
+async function isDirectory(path: string): Promise<boolean> {
+  try {
+    return (await stat(path)).isDirectory();
+  } catch (error) {
+    if (isMissingFileError(error)) return false;
     throw error;
   }
 }
--- a/packages/orchestrator/src/git.ts
+++ b/packages/orchestrator/src/git.ts
@@ -862,7 +862,7 @@ export async function acquireRepositoryLock(
       }
     }
 
-    const stale = await isStaleLock(lockDirectory);
+    const stale = await isRepositoryLockStale(lockDirectory);
     if (stale) {
       await rm(lockDirectory, { recursive: true, force: true });
       continue;
@@ -880,22 +880,29 @@ export async function acquireRepositoryLock(
 
 /** Attempts lock acquisition once; maintenance callers use this to skip busy caches. */
 export async function tryAcquireRepositoryLock(
-  lockDirectory: string
+  lockDirectory: string,
+  options: { breakStale?: boolean } = {}
 ): Promise<string | null> {
   const ownerToken = `${process.pid}:${randomUUID()}`;
-  try {
-    await mkdir(lockDirectory);
-    await writeFile(
-      join(lockDirectory, "owner"),
-      `${ownerToken}\n${new Date().toISOString()}\n`,
-      "utf8"
-    );
-    return ownerToken;
-  } catch (error) {
-    if (isAlreadyExistsError(error)) {
+  for (;;) {
+    try {
+      await mkdir(lockDirectory);
+      await writeFile(
+        join(lockDirectory, "owner"),
+        `${ownerToken}\n${new Date().toISOString()}\n`,
+        "utf8"
+      );
+      return ownerToken;
+    } catch (error) {
+      if (!isAlreadyExistsError(error)) {
+        throw error;
+      }
+    }
+
+    if (!options.breakStale || !(await isRepositoryLockStale(lockDirectory))) {
       return null;
     }
-    throw error;
+    await rm(lockDirectory, { recursive: true, force: true });
   }
 }
 
@@ -955,7 +962,9 @@ export function startRepositoryLockHeartbeat(
   };
 }
 
-async function isStaleLock(lockDirectory: string): Promise<boolean> {
+export async function isRepositoryLockStale(
+  lockDirectory: string
+): Promise<boolean> {
   try {
     const details = await stat(lockDirectory);
     return Date.now() - details.mtimeMs >= LOCK_STALE_MS;
--- a/packages/orchestrator/src/git.ts
+++ b/packages/orchestrator/src/git.ts
@@ -7,6 +7,7 @@ import {
   rename,
   rm,
   stat,
+  utimes,
   writeFile,
 } from "node:fs/promises";
 import { constants } from "node:fs";
@@ -29,6 +30,10 @@ const workflowConfigStore = new WorkflowConfigStore();
 const LOCK_RETRY_MS = 100;
 const LOCK_STALE_MS = 30 * 60 * 1000;
 const LOCK_TIMEOUT_MS = 2 * 60 * 1000;
+
+export type RepositoryLockHeartbeat = {
+  stop: () => Promise<void>;
+};
 
 export type RepositorySyncResult = {
   repositoryDirectory: string;
@@ -887,6 +892,43 @@ export async function releaseRepositoryLock(
   }
 
   await rm(lockDirectory, { recursive: true, force: true });
+}
+
+/**
+ * Keeps a long-running lock lease fresh without changing the generic lock's
+ * stale timeout. Ownership is checked before every touch so a replaced lock is
+ * never intentionally renewed by its previous holder.
+ */
+export function startRepositoryLockHeartbeat(
+  lockDirectory: string,
+  ownerToken: string,
+  intervalMs = 60_000
+): RepositoryLockHeartbeat {
+  let stopped = false;
+  let heartbeat = Promise.resolve();
+  const timer = setInterval(() => {
+    heartbeat = heartbeat
+      .then(async () => {
+        if (stopped || (await readLockOwner(lockDirectory)) !== ownerToken) {
+          return;
+        }
+        const now = new Date();
+        await utimes(lockDirectory, now, now);
+      })
+      .catch(() => {
+        // The cache operation itself remains authoritative. A removed or
+        // inaccessible lock will be handled by its normal failure path.
+      });
+  }, intervalMs);
+  timer.unref();
+
+  return {
+    async stop(): Promise<void> {
+      stopped = true;
+      clearInterval(timer);
+      await heartbeat;
+    },
+  };
 }
 
 async function isStaleLock(lockDirectory: string): Promise<boolean> {
--- a/packages/orchestrator/src/git.ts
+++ b/packages/orchestrator/src/git.ts
@@ -235,6 +235,7 @@ async function ensureIssueWorkspaceWorktree(input: {
   issueIdentifier?: string;
   branchTemplate?: string | null;
   baseBranch?: string | null;
+  onCacheUnavailable?: (error: RepositoryCacheUnavailableError) => void;
 }): Promise<string> {
   const repositoryDirectory = join(input.issueWorkspacePath, "repository");
   if (await pathExists(join(repositoryDirectory, ".git"))) {
--- a/packages/orchestrator/src/git.ts
+++ b/packages/orchestrator/src/git.ts
@@ -875,6 +875,27 @@ export async function acquireRepositoryLock(
   }
 }
 
+/** Attempts lock acquisition once; maintenance callers use this to skip busy caches. */
+export async function tryAcquireRepositoryLock(
+  lockDirectory: string
+): Promise<string | null> {
+  const ownerToken = `${process.pid}:${randomUUID()}`;
+  try {
+    await mkdir(lockDirectory);
+    await writeFile(
+      join(lockDirectory, "owner"),
+      `${ownerToken}\n${new Date().toISOString()}\n`,
+      "utf8"
+    );
+    return ownerToken;
+  } catch (error) {
+    if (isAlreadyExistsError(error)) {
+      return null;
+    }
+    throw error;
+  }
+}
+
 export async function releaseRepositoryLock(
   lockDirectory: string,
   ownerToken: string
--- a/packages/orchestrator/src/git.ts
+++ b/packages/orchestrator/src/git.ts
@@ -20,6 +20,7 @@ import {
 } from "@gh-symphony/core";
 import {
   ensureGlobalBareRepositoryCache,
+  RepositoryCacheUnavailableError,
   withGlobalBareRepositoryCache,
 } from "./repository-cache.js";
 import { sanitizeRepositoryCloneUrl } from "./repository-url.js";
@@ -97,19 +98,29 @@ export async function syncRepositoryForRun(input: {
     await rm(tempRepositoryDirectory, { recursive: true, force: true });
 
     try {
-      const bareRepositoryDirectory = await ensureGlobalBareRepositoryCache({
-        repository: input.repository,
-        requiredRef: input.requiredRef,
-      });
-      await runCommand("git", [
-        "clone",
-        "--filter=blob:none",
-        "--reference-if-able",
-        bareRepositoryDirectory,
-        "--dissociate",
-        sanitizeRepositoryCloneUrl(input.repository.cloneUrl),
-        tempRepositoryDirectory,
-      ]);
+      try {
+        const bareRepositoryDirectory = await ensureGlobalBareRepositoryCache({
+          repository: input.repository,
+          requiredRef: input.requiredRef,
+        });
+        await runCommand("git", [
+          "clone",
+          "--filter=blob:none",
+          "--reference-if-able",
+          bareRepositoryDirectory,
+          "--dissociate",
+          sanitizeRepositoryCloneUrl(input.repository.cloneUrl),
+          tempRepositoryDirectory,
+        ]);
+      } catch (error) {
+        if (!(error instanceof RepositoryCacheUnavailableError)) {
+          throw error;
+        }
+        await cloneRepositoryDirectly(
+          input.repository,
+          tempRepositoryDirectory
+        );
+      }
       await rename(tempRepositoryDirectory, repositoryDirectory);
       return {
         repositoryDirectory,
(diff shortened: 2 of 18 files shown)


hojinzs (changes requested):
## Project #14 watchdog review — PR #616

**Request changes.**

I reviewed head `c1ce0d0e7bd023108c7a5acc414b42b9751a5dda` with Claude Code in read/test-only mode and ran the relevant smoke checks locally.

Blocking findings:
1. `cache prune` can remove a bare cache while an in-flight reference clone is still reading it, because the cache lock is released before the clone begins.
2. `updatedAt` is based on the bare directory mtime, which is not a reliable last-used signal after fetch/marker updates; the documented age-based prune can remove actively used caches.
3. The non-dry-run deletion path lacks regression coverage for actual removal, linked-worktree protection, and unverifiable worktree counts.

Validation: targeted orchestrator and CLI tests passed (38 + 2), and `pnpm lint`, `pnpm typecheck`, and `pnpm build` passed.

Evidence bundle: https://evidence-browser.dev0.daniworks.kr/w/github-symphony/b/project14-pr-616-20260821-134745z

packages/orchestrator/src/repository-cache.ts
```diff
@@ -44,6 +82,142 @@ export function globalBareRepositoryLockDirectory(input: {
...
+      if (!ownerToken) {
+        result.skipped.push({ ...entry, locked: true, reason: "locked" });
+        continue;
+      }
+      try {
+        if ((await countLinkedWorktrees(entry.directory)) !== 0) {
+          result.skipped.push({ ...entry, reason: "worktrees" });
+          continue;
+        }
+        await rm(entry.directory, { recursive: true, force: true });
```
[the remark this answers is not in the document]
moncher-dev: Resolved in `46aee83`. The regular workspace clone now runs inside `withGlobalBareRepositoryCache`, so the repository heartbeat lock remains owned through the complete `git clone --reference-if-able --dissociate` operation. `cache prune` therefore cannot acquire its deletion lock while the reference reader is active. I also added a concurrent reader/prune regression test that verifies the entry is skipped as `locked`; the focused Git/cache suite passes all 39 tests.

packages/orchestrator/src/repository-cache.ts
```diff
@@ -44,6 +82,142 @@ export function globalBareRepositoryLockDirectory(input: {
...
+    for (const name of await safeDirectories(join(root, owner))) {
+      if (!name.endsWith(".git")) continue;
+      const directory = join(root, owner, name);
+      const details = await stat(directory);
+      const lockDirectory = join(root, owner, `${name.slice(0, -4)}.lock`);
+      entries.push({
+        repository: `${owner}/${name.slice(0, -4)}`,
+        directory,
+        bytes: await directorySize(directory),
+        updatedAt: details.mtime.toISOString(),
```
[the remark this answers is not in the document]
moncher-dev: Resolved with a dedicated `.gh-symphony-last-used` marker. Every successful cache acquisition refreshes it, including reuse within the fetch TTL, and inspection uses it for `updatedAt`; pre-existing caches without the marker retain a directory-mtime compatibility fallback. A regression test fixes the bare root mtime at the earlier value, reuses the cache later without requiring a fetch, and verifies pruning treats the later use as recent.

packages/orchestrator/src/repository-cache.ts
```diff
@@ -44,6 +82,142 @@ export function globalBareRepositoryLockDirectory(input: {
...
+  for (const owner of owners) {
+    for (const name of await safeDirectories(join(root, owner))) {
+      if (!name.endsWith(".git")) continue;
+      const directory = join(root, owner, name);
+      const details = await stat(directory);
+      const lockDirectory = join(root, owner, `${name.slice(0, -4)}.lock`);
+      entries.push({
+        repository: `${owner}/${name.slice(0, -4)}`,
+        directory,
+        bytes: await directorySize(directory),
```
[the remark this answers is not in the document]
moncher-dev: Resolved by checking the cache lock before size or worktree inspection. Locked entries are reported conservatively with `bytes: 0` and `worktrees: null`, so inventory/prune does not recursively traverse contents being mutated by fetch, recreation, or GC. Recursive sizing and entry stat operations also tolerate `ENOENT`/`ENOTDIR` if an unlocked entry disappears concurrently, preventing one race from aborting the full command.

packages/orchestrator/src/repository-cache.ts
```diff
@@ -44,6 +84,179 @@ export function globalBareRepositoryLockDirectory(input: {
...
+    }
+    if (!input.dryRun) {
+      const lockDirectory = `${entry.directory.slice(0, -4)}.lock`;
+      const ownerToken = await tryAcquireRepositoryLock(lockDirectory);
+      if (!ownerToken) {
+        result.skipped.push({ ...entry, locked: true, reason: "locked" });
+        continue;
+      }
+      try {
+        if ((await countLinkedWorktrees(entry.directory)) !== 0) {
```
hojinzs, line 170: **P2 — re-check the last-used age under the deletion lock, not only the worktree count**

The `.gh-symphony-last-used` marker fixes the *reliability* of the timestamp (round-1 finding #2), but the age is still evaluated only once, during the inventory pass at line 130. Between that read and the `rm` at line 154 the entry can be acquired, refreshed, and released by a worker — the under-lock recheck here only re-verifies `countLinkedWorktrees`, and a `--dissociate` reference clone never registers a linked worktree, so that recheck provides no protection for the `syncRepositoryForRun` path.

The lock does cover a reader that is *still* holding it, but not one whose whole critical section (TTL-fresh cache hit + `worktree add`, often sub-second) fits inside the inventory→delete gap.

Reproduced against `8515426` with the built `dist`, 121 cache entries, `maxAgeMs = 30d`:

```
inventory phase takes ~817ms (deletion loop runs after it)
[worker] acquired + refreshed last-used for acme/zzz-hot (used seconds ago)
acme/zzz-hot DELETED by prune? true
```

The marker said "seconds ago" and the entry was evicted anyway. The gap scales with the inventory cost — `directorySize` recurses over every object in every bare repo, so on a production-sized cache directory this is minutes, not 800ms.

Suggested fix, inside the existing `try` block that already holds `ownerToken`:

```ts
try {
  const usedAt = await readLastUsedAt(entry.directory, new Date(entry.updatedAt));
  if (now.getTime() - Date.parse(usedAt) < input.maxAgeMs) {
    result.skipped.push({ ...entry, updatedAt: usedAt, reason: "recent" });
    continue;
  }
  if ((await countLinkedWorktrees(entry.directory)) !== 0) {
```

`readLastUsedAt` is already in this module, so this is a few lines and makes the acquire-refresh-release window safe.


---

packages/orchestrator/src/repository-cache.ts
```diff
@@ -44,6 +84,179 @@ export function globalBareRepositoryLockDirectory(input: {
...
+): Promise<RepositoryCacheEntry[]> {
+  const root = resolveGlobalRepositoryCacheRoot(input.configDir);
+  const owners = await safeDirectories(root);
+  const entries: RepositoryCacheEntry[] = [];
+  for (const owner of owners) {
+    for (const name of await safeDirectories(join(root, owner))) {
+      if (!name.endsWith(".git")) continue;
+      const directory = join(root, owner, name);
+      const lockDirectory = join(root, owner, `${name.slice(0, -4)}.lock`);
+      const locked = await pathExists(lockDirectory);
```
hojinzs, line 106: **P2 — a stale lock makes a cache permanently unprunable and permanently "locked" in status**

`locked` here is pure `pathExists`, and `tryAcquireRepositoryLock` (git.ts:882) returns `null` on *any* existing lock directory. Neither consults `LOCK_STALE_MS`, which `acquireRepositoryLock` uses at git.ts:865 to break abandoned locks after 30 minutes.

So a lock directory left behind by a crashed or SIGKILLed orchestrator is never reclaimable by the maintenance surface. Reproduced against `8515426` with the built CLI — lock dir mtime set 2 hours back (4× `LOCK_STALE_MS`), cache last used 90 days ago:

```
$ gh-symphony cache status --config-dir …
acme/platform  0 B  locked  2026-05-23T14:08:36.000Z

$ gh-symphony cache prune --config-dir … --json
{ "removed": [], "skipped": [ { "repository": "acme/platform", "locked": true, "reason": "locked" } ] }
  -> still on disk: platform.git  platform.lock
```

This inverts the feature's purpose: the crash case is exactly when disk reclamation matters, and it is the one case prune can never act on. The size is also reported as `0 B`, so the entry is invisible in `totalBytes` too — an operator reading `cache status` has no signal that a wedged lock is pinning a stale cache.

The heartbeat added in this PR is what makes acting on this safe: a live holder now refreshes the lock mtime every 60s, so a lock older than `LOCK_STALE_MS` is genuinely abandoned. Two small changes:

- Export the staleness predicate (or a `tryAcquireRepositoryLockBreakingStale`) and let prune break a stale lock rather than skip it — the existing under-lock worktree recheck still guards the deletion.
- Distinguish `locked` from `stale-locked` in `RepositoryCacheEntry` so `cache status` can surface the wedged entry instead of showing it as a normal in-use cache.

At minimum, if breaking stale locks is deliberately out of scope, `cache status` must show it as stale so the operator knows to clear it by hand — silently reporting it as busy forever is the part that has to change.


---

packages/cli/src/commands/cache.ts
```diff
@@ -0,0 +1,80 @@
...
+      return;
+    }
+    if (entries.length === 0) {
+      process.stdout.write("Global repository cache is empty.\n");
+      return;
+    }
+    for (const entry of entries) {
+      const state = entry.locked
+        ? "locked"
+        : entry.worktrees
```
hojinzs: **P2 — `cache status` reports an unverifiable cache as `idle`, contradicting what `cache prune` then does**

`entry.worktrees` is `number | null`, and this ternary is truthiness-based: `null` (worktree count could not be determined) and `0` (genuinely no worktrees) both fall through to `"idle"`. But `pruneGlobalRepositoryCache` treats them differently — `entry.worktrees !== 0` skips the `null` case with `reason: "worktrees"`.

The result is that the human-readable surface tells the operator the exact opposite of what prune will do. Reproduced against `8515426` with the built CLI, on a corrupt `*.git` entry 90 days old:

```
$ gh-symphony cache status --config-dir …
acme/broken  11 B  idle  2026-05-23T14:08:36.616Z          <- looks reclaimable

$ gh-symphony cache status --config-dir … --json
"worktrees": null                                           <- JSON is correct

$ gh-symphony cache prune --config-dir … --json
"removed": [], "skipped": [ { "repository": "acme/broken", "reason": "worktrees" } ]
```

The PR description says locked entries report "worktree count unknown", and the JSON does carry `null` faithfully — it is only this formatter that collapses the distinction. Since "conservatively retain unverifiable entries" is a documented and reasonable policy, the operator needs to be able to *see* which entries are in that bucket; otherwise the only way to find out why disk is not being reclaimed is to diff `status` against `prune --json` by hand.

```ts
const state = entry.locked
  ? "locked"
  : entry.worktrees === null
    ? "unverifiable"
    : entry.worktrees > 0
      ? `${entry.worktrees} linked worktree(s)`
      : "idle";
```

Worth a `cache.test.ts` case as well — the two existing CLI tests cover the empty inventory and the invalid-age guard, so no test currently exercises the entry formatter at all.


---

packages/orchestrator/src/git.ts
```diff
@@ -97,19 +102,33 @@ export async function syncRepositoryForRun(input: {
...
+              sanitizeRepositoryCloneUrl(input.repository.cloneUrl),
+              tempRepositoryDirectory,
+            ]);
+          }
+        );
+      } catch (error) {
+        if (!(error instanceof RepositoryCacheUnavailableError)) {
+          throw error;
+        }
+        await cloneRepositoryDirectly(
```
hojinzs, line 130: **P2 — the fallback is silent; cache degradation is unobservable**

Both fallback sites (here and line 289) swallow `RepositoryCacheUnavailableError` and proceed with a full network clone without emitting anything. `git.ts` has no logger, no stderr write, and no event emission at all, so nothing distinguishes a healthy cached populate from one where every workspace is now re-cloning from GitHub.

The PR's own risk list names this — *"Direct fallback uses extra network bandwidth during cache degradation"* — but ships no way to detect that it is happening. Observability is one of the six Symphony layers this repo classifies work against (`CLAUDE.md`), and a permanently-degraded cache is precisely the condition an operator needs surfaced: it is silent, self-sustaining (nothing retries cache preparation), and only shows up as a slow, bandwidth-hungry orchestrator.

`RepositoryCacheUnavailableError` already carries `cause`, so the message is there — it just needs somewhere to go. Either thread an optional `onCacheUnavailable`/writer callback through `syncRepositoryForRun` and `ensureIssueWorkspaceWorktree` from `service.ts` (which already has `writeStderr` and structured event emission), or emit a `repository-cache-unavailable` event alongside the existing `tracker-repository-filtered`-style events.

Related, smaller (**P3**): `removeIssueWorkspaceWorktree` (line 188) did not get the same treatment. A workspace created through this fallback is a standalone clone, not a linked worktree, so teardown calls `withGlobalBareRepositoryCache` — which will clone the entire bare cache from scratch if it is still missing — only to run a `git worktree remove` that cannot match. Verified the error text:

```
$ git -C bare.git worktree remove --force …/standalone
fatal: '…/standalone' is not a working tree
```

That is swallowed by `ignoreMissingWorktree`, and `service.ts:4750` `rm -rf`s the workspace path regardless, so nothing leaks and nothing breaks — it is wasted work plus a hard failure if the cache is still unavailable (caught and logged at `service.ts:4739`). Skipping the bare-cache round trip when `repositoryDirectory/.git` is a directory rather than a worktree file would close the asymmetry.


---

packages/orchestrator/src/repository-cache.ts
```diff
@@ -44,6 +84,179 @@ export function globalBareRepositoryLockDirectory(input: {
...
+    if (entry.locked) {
+      result.skipped.push({ ...entry, reason: "locked" });
+      continue;
+    }
+    if (entry.worktrees !== 0) {
+      result.skipped.push({ ...entry, reason: "worktrees" });
+      continue;
+    }
+    if (!input.dryRun) {
+      const lockDirectory = `${entry.directory.slice(0, -4)}.lock`;
```
hojinzs: **nit** — this re-derives the lock path by string surgery on the `.git` suffix, duplicating `globalBareRepositoryLockDirectory` (line 75) and the identical `name.slice(0, -4)` at line 99. Three copies of the same layout assumption; if the suffix convention ever changes, prune silently locks the wrong path rather than failing.

Carrying `owner`/`name` on `RepositoryCacheEntry` (or an internal inventory type) would let both call sites go through `globalBareRepositoryLockDirectory` instead.


---

hojinzs (changes requested):
## Round 2 review — PR #616 @ `8515426`

**Request changes.** Round-1 blockers are genuinely fixed; the new findings are all in the maintenance surface added to address them.

### Round-1 follow-up — all three resolved ✅

| # | Round-1 finding | Status |
|---|---|---|
| 1 | `cache prune` can delete a bare cache while a reference clone reads it | ✅ `syncRepositoryForRun` now runs the clone inside `withGlobalBareRepositoryCache`, so the heartbeat lock is held across `--reference-if-able --dissociate`; prune's `tryAcquireRepositoryLock` fails and skips. Covered by *"prevents pruning while a cache reader operation is active"*. |
| 2 | `updatedAt` from directory mtime is not a reliable last-used signal | ✅ Dedicated `.gh-symphony-last-used`, refreshed unconditionally at repository-cache.ts:405 including TTL-fresh reuse, with mtime fallback for pre-existing entries. Covered by *"tracks cache reuse independently from the bare directory mtime"*. **Residual window — see P2 inline.** |
| 3 | Non-dry-run deletion path lacked regression coverage | ✅ Four new cases: actual removal, linked-worktree retention, unverifiable counts, active-reader exclusion. |

### Requirements coverage vs. #588

All three scope items are delivered, and I did not find over-engineering — 933 additions for three features plus tests and docs is proportionate, and no undocumented product decisions were smuggled in.

- 폴백: cache unavailable **and** lock timeout both map to `RepositoryCacheUnavailableError` (the `acquireRepositoryLock` call sits inside the mapping `try`), so the timeout case really is covered.
- Heartbeat: 60s `utimes` refresh against a 30-minute `LOCK_STALE_MS`, with owner re-verification before every touch — correct, and `timer.unref()` keeps it from pinning the event loop.
- Eviction + CLI: `cache status` / `cache prune` present and working.

### Blocking

1. **P2** `pruneGlobalRepositoryCache` — age is evaluated only during the inventory pass and never re-checked under the deletion lock. **Reproduced: a cache whose marker read "seconds ago" was evicted by `--max-age-days 30`.** Inline on repository-cache.ts:150.
2. **P2** Stale lock is never broken by the maintenance path — a crashed holder's lock dir pins a cache as `locked`/unprunable forever, which is exactly the case eviction exists for. **Reproduced** with a 2-hour-old lock (4× `LOCK_STALE_MS`). Inline on repository-cache.ts:100.
3. **P2** `cache status` prints `idle` for unverifiable entries that `cache prune` then refuses to remove — the human surface contradicts the prune result. **Reproduced.** Inline on cache.ts:38.
4. **P2** The fallback emits nothing — a permanently degraded cache is undetectable, though the PR's own risk list names the bandwidth cost. Inline on git.ts:127 (plus a P3 on the teardown asymmetry).
5. **Missing changeset.** No `.changeset/*.md` in the diff. This adds a new user-facing `gh-symphony cache` command and changes workspace population behavior, so per `CLAUDE.md` → Releases it needs a changeset for `@gh-symphony/cli`.
6. **Merge conflict.** `mergeable_state: dirty` — `AGENT_TEST.md` conflicts with `main` (`f2a8a4f`), which added its own scenario-table row. Content conflict only; merging `main` and keeping both rows resolves it.

Items 1–3 are a few lines each and 5–6 are mechanical; none of this is a redesign.

### nit

- repository-cache.ts:143 — lock path re-derived by `.git` string surgery, third copy of that layout assumption.
- `formatBytes` tops out at MiB, so a multi-GB cache prints as e.g. `12345.6 MiB`.

### Validation

Ran on `8515426` locally, plus black-box smoke tests of the built CLI (`packages/cli/dist/index.js`) against a synthetic `--config-dir`:

- `pnpm lint`, `pnpm typecheck`, `pnpm build` — pass.
- `pnpm --filter @gh-symphony/orchestrator test` — 310 passed (14 files); `pnpm --filter @gh-symphony/cli test` — 540 passed (39 files).
- Smoke: empty inventory → seeded 90-day-old entry → `status` → `status --json` → `prune --dry-run` (retained) → real `prune` (25.6 KiB reclaimed) → empty. Happy path is correct end to end.
- Smoke: stale-lock, unverifiable-entry, and missing-argument cases — first two are the findings above; argument handling is correctly rejected by commander, no issue there.
- Verified `git worktree remove` against a standalone clone emits `is not a working tree` (matched by `ignoreMissingWorktree`) and that `worktree list --porcelain` on a bare repo emits exactly one `worktree ` line, so the `- 1` in `countLinkedWorktrees` is right.

No CI checks are configured on this PR, so the above is the only signal.

Round 2 of 6.


---

packages/orchestrator/src/repository-cache.ts
moncher-dev: Resolved in `a1139be`. The non-dry-run path now re-reads `.gh-symphony-last-used` while holding the deletion lock and skips the entry as `recent` if a worker acquired and refreshed it after inventory. The same critical section still rechecks linked worktrees before removal. The focused Git/cache suite passes 44 tests, with orchestrator lint and typecheck also passing.

packages/orchestrator/src/repository-cache.ts
moncher-dev: Resolved in `a1139be`. Lock staleness is now shared with maintenance: inventory exposes `staleLock`, safely inspects an abandoned entry (including its bytes/worktree state), and prune may break only a lock older than the existing 30-minute stale threshold before performing the under-lock age/worktree checks. A regression test creates a two-hour-old abandoned lock, verifies it is reported as stale with nonzero size, and proves the old cache is reclaimed; live heartbeat locks remain conservatively skipped.

packages/cli/src/commands/cache.ts
moncher-dev: Resolved in `edab64d`. Human-readable `cache status` now preserves all three distinctions: `stale-locked`, live `locked`, and `unverifiable` when worktree enumeration returns `null`; only a verified zero count is shown as `idle`. Added a CLI regression test with a corrupt bare entry that asserts the `unverifiable` label.

packages/orchestrator/src/git.ts
moncher-dev: Resolved in `edab64d`. Both direct-clone fallback sites now invoke an optional typed cache-unavailable callback, and the service wires it to an operator-visible stderr warning containing the issue identifier and cache error. Focused tests assert notification for regular and worktree-cache fallback. Teardown now also detects a standalone clone by its `.git` directory and returns without entering the bare-cache/worktree path; the fallback test verifies cleanup remains a no-op even while the cache config is unavailable.

packages/orchestrator/src/repository-cache.ts
moncher-dev: Resolved in `c5d96284`. Inventory now derives its lock directory through `globalBareRepositoryLockDirectory`, and prune converts the validated `owner/name` cache identity back through the same helper instead of slicing `.git` from the filesystem path. This leaves the cache layout convention in one canonical function.

packages/orchestrator/src/git.ts
```diff
@@ -97,19 +104,34 @@ export async function syncRepositoryForRun(input: {
...
-        "clone",
-        "--filter=blob:none",
-        "--reference-if-able",
-        bareRepositoryDirectory,
-        "--dissociate",
-        sanitizeRepositoryCloneUrl(input.repository.cloneUrl),
-        tempRepositoryDirectory,
-      ]);
+      try {
+        await withGlobalBareRepositoryCache(
```
hojinzs, line 108: **P3 (non-blocking, follow-up) — the round-1 fix widens the lock to cover the whole clone, so same-repo populates now serialize and contention degrades into a full network clone**

This is correct as written and I am not asking for a change in this PR — recording it so the trade-off is explicit and can be picked up as a follow-up to #588.

Round-1 finding #1 (prune deleting a bare cache mid-reference-clone) is genuinely fixed by moving the clone inside `withGlobalBareRepositoryCache`. The cost is that the **exclusive** cache lock is now held for the full duration of `git clone --filter=blob:none --reference-if-able --dissociate`, not just cache preparation. Previously `ensureGlobalBareRepositoryCache` released the lock before the clone started. Codex's original comment offered two options — protect until the clone completes, *or* track active reference-clone readers — and this takes the first, which is an exclusive writer lock where a reader count would do.

Two measured consequences, on `c5d9628` against a local 92 MB source repo:

**1. Same-repository populates serialize.**

```
--- 2 concurrent populates of the SAME repo (cache already warm) ---
  A: start +0ms  end +2865ms  (elapsed 2865ms)
  B: start +0ms  end +6050ms  (elapsed 6050ms)
wall clock for the pair: 6050ms          <- sum, not max
lock present in 942/947 samples (~99% of the run)
```

**2. Contention — not cache failure — triggers the direct-clone fallback.** With a peer holding the lock and heartbeating it (so it is never stale):

```
FALLBACK after 120s -> Global repository cache unavailable at .../acme/big.git:
  Timed out waiting for repository cache lock: .../acme/big.lock
fellBackToFullNetworkClone=true  total=122s
```

Exactly `LOCK_TIMEOUT_MS`. The waiter stalls two minutes and then does a full unreferenced network clone, and the operator sees `repository cache unavailable`, which misattributes a healthy-but-busy cache to a broken one.

**Why this is P3 and not a blocker.** I checked how populates are actually scheduled: `startRun` is `await`ed inside the per-issue `for` loop at service.ts:1566, and no tenant/issue fan-out uses `Promise.all`, so a **single orchestrator process never populates two workspaces concurrently** — it cannot contend with itself. Contention needs two or more orchestrator processes sharing `~/.gh-symphony/repos` and targeting the same repository, which the README does support ("multiple projects can orchestrate the same repository without branch collisions") but is not the common single-process deployment. And a 2-minute-timeout-then-direct-clone is literally scope item 1 of #588 (`cache unavailable·락 타임아웃 시 workspace 직접 clone 폴백`), so the fallback firing here is requested behavior, not a defect.

Note the default `repo` path is the affected one: `project` sets `populateStrategy: "worktree-cache"` (cheap sub-second `worktree add`, and it already held the lock this way before this PR), while `repo` leaves it undefined and lands on this clone path.

If it is picked up later, a reader count — bumped under the lock, released after the clone, with prune's `tryAcquireRepositoryLock` refusing while readers are non-zero — restores concurrent populates while keeping the round-1 guarantee. Worth a sentence in `docs/configuration.md` either way, since the current text implies the fallback only fires when the cache is genuinely unavailable.

---

packages/orchestrator/src/repository-cache.ts
```diff
@@ -44,6 +86,211 @@ export function globalBareRepositoryLockDirectory(input: {
...
+      const ownerToken = await tryAcquireRepositoryLock(lockDirectory, {
+        breakStale: true,
+      });
+      if (!ownerToken) {
+        result.skipped.push({ ...entry, locked: true, reason: "locked" });
+        continue;
+      }
+      try {
+        const details = await safeStat(entry.directory);
+        if (!details) continue;
```
hojinzs, line 164: **nit** — this `continue` drops the entry from *both* `removed` and `skipped`.

If the bare directory disappears between the inventory pass and acquiring the deletion lock, the entry silently vanishes from the result: it is not counted as removed (correct — this prune did not reclaim it) but also not reported as skipped, so `removed.length + skipped.length` no longer accounts for every entry the inventory saw. An operator diffing `cache status` against `cache prune --json` gets an unexplained gap.

The three other early exits in this block all push a `skipped` entry first; this one is the odd one out. A `reason: "missing"` variant, or just reusing `"recent"`-style reporting, would keep the result total-accounting. Genuinely minor — the window is small and nothing leaks.

---

packages/cli/src/commands/cache.ts
```diff
@@ -0,0 +1,84 @@
...
+
+function valueAfter(args: string[], flag: string): string | undefined {
+  const index = args.indexOf(flag);
+  return index >= 0 ? args[index + 1] : undefined;
+}
+
+function formatBytes(bytes: number): string {
+  if (bytes < 1024) return `${bytes} B`;
+  if (bytes < 1024 * 1024) return `${(bytes / 1024).toFixed(1)} KiB`;
+  return `${(bytes / (1024 * 1024)).toFixed(1)} MiB`;
```
hojinzs, line 81: **nit (carried from round 2, unaddressed)** — `formatBytes` tops out at MiB.

A shared bare cache for a large monorepo is routinely multi-GB, and the whole point of `cache status` is deciding what to reclaim. A 12 GB entry prints as `12288.0 MiB`, and `reclaimedBytes` in the prune summary has the same ceiling — which is the number an operator reads to confirm the prune did what they wanted.

One more branch closes it:

```ts
if (bytes < 1024 * 1024 * 1024) return `${(bytes / (1024 * 1024)).toFixed(1)} MiB`;
return `${(bytes / (1024 * 1024 * 1024)).toFixed(1)} GiB`;
```

Flagging it again only because it went by without a reply thread in round 2, not to block — the `--json` output carries exact bytes, so nothing is actually lost.

---

hojinzs (approved):
## Round 3 review — PR #616 @ `c5d9628`

**Approve.** All six round-2 blockers are fixed and I verified each one against the built artifacts rather than taking the reply threads at face value. Nothing left is blocking; the three inline comments are a follow-up note and two nits.

### Round-2 follow-up — all six resolved ✅

| # | Round-2 finding | Status |
|---|---|---|
| 1 | Age evaluated only at inventory; acquire-refresh-release window could evict a hot cache | ✅ repository-cache.ts:165-169 re-reads `.gh-symphony-last-used` **inside** the `ownerToken` critical section and skips as `recent`, before the worktree recheck. |
| 2 | Stale lock pins a cache as unprunable forever | ✅ `isRepositoryLockStale` is now shared with the maintenance path, `staleLock` is on the entry, and `tryAcquireRepositoryLock(..., { breakStale: true })` reclaims it. |
| 3 | `cache status` printed `idle` for unverifiable entries prune then refuses | ✅ cache.ts:36-44 now distinguishes all four states; only a verified `0` prints `idle`. |
| 4 | Fallback silent / teardown asymmetry | ✅ `onCacheUnavailable` threaded through both fallback sites and wired to `writeStderr` at service.ts:2723. Teardown returns early when `.git` is a directory (git.ts:200), so a standalone clone no longer round-trips the bare cache. |
| 5 | Missing changeset | ✅ `.changeset/fuzzy-caches-rest.md`, minor, `@gh-symphony/cli`. |
| 6 | Merge conflict in `AGENT_TEST.md` | ✅ `mergeable_state: clean`; both scenario rows kept. |
| nit | Lock path re-derived by `.git` string surgery | ✅ Both call sites go through `globalBareRepositoryLockDirectory`. |

Finding #2 deserves a note: the fix does more than the minimum I asked for — a stale-locked entry is now *inspected* rather than reported as `0 B` / `null`, so the wedged cache is visible in `totalBytes` too. That was the part that made it invisible to an operator, and it is closed.

### Requirements coverage vs. #588

All three scope items delivered, nothing beyond them:

- **직접 clone 폴백** — both `RepositoryCacheUnavailableError` sources (storage failure and lock timeout) reach the fallback; confirmed the timeout really is inside the mapping `try` at repository-cache.ts:304-309 / 346-351.
- **heartbeat** — 60s `utimes` against 30-min `LOCK_STALE_MS`, owner re-verified before each touch, `timer.unref()`.
- **eviction + CLI 진단·정리** — `cache status` / `cache prune` with `--json`, `--dry-run`, `--max-age-days`.

No over-engineering and no undeclared product decisions. 1103 additions for three features plus tests, docs and an e2e scenario is proportionate, and the conservative-retention policy is documented rather than assumed.

### Validation

Rebuilt and ran everything on `c5d9628`:

- `pnpm install`, `pnpm build`, `pnpm lint`, `pnpm typecheck` — pass.
- `pnpm --filter @gh-symphony/orchestrator test` — **315 passed** (14 files); `pnpm --filter @gh-symphony/cli test` — **540 passed** (39 files).

Black-box smoke of the built CLI (`packages/cli/dist/index.js`) against a synthetic `--config-dir` — this is where I checked the round-2 fixes rather than trusting the unit tests:

```
--- happy path ---
empty inventory -> seeded 90-day entry -> status: acme/platform  25.7 KiB  idle
prune --dry-run: Would remove 1 cache entry (25.7 KiB); skipped 0
prune:           Removed 1 cache entry (25.7 KiB); skipped 0   -> empty

--- the three round-2 states, side by side in one inventory ---
acme/broken    28 B      unverifiable    (corrupt bare dir)
acme/live      0 B       locked          (live lock)
acme/platform  25.6 KiB  stale-locked    (lock dir mtime 2h back = 4x LOCK_STALE_MS)

prune --max-age-days 30 --json:
  removed: acme/platform   (stale lock broken, 26248 bytes reclaimed, worktrees: 0)
  skipped: acme/broken     reason "worktrees"  (worktrees: null)
           acme/live       reason "locked"     (staleLock: false)
  on disk after: broken.git  live.git  live.lock
```

That is the exact inversion from round 2 — the stale-locked entry now reports its real size, is labeled correctly, and is the one that gets reclaimed, while the live lock is still conservatively skipped and its lock directory left intact.

I also ran two concurrent `syncRepositoryForRun` calls against a shared cache to check the widened lock scope. It behaves as designed but serializes same-repo populates, and a lock held past `LOCK_TIMEOUT_MS` degrades to a full clone at exactly 120s. Details and measurements are inline on git.ts:108 — filed as **P3 / follow-up**, explicitly not a blocker, because a single orchestrator process dispatches `startRun` sequentially and so cannot contend with itself, and because timeout→direct-clone is scope item 1 of #588.

### Non-blocking

- **P3** git.ts:108 — exclusive lock across the whole reference clone; serialization + misleading `cache unavailable` warning under contention. Follow-up to #588.
- **nit** repository-cache.ts:164 — an entry that vanishes between inventory and lock is reported in neither `removed` nor `skipped`.
- **nit** cache.ts:81 — `formatBytes` still caps at MiB (carried from round 2).

Round 3 of 6. Converging here: the blockers are closed, the remainder is a trade-off note and two cosmetic items, and pushing the reader-count redesign into this PR would widen it past what #588 asked for.

---

PR #617, opened 2026-08-21
fix(config): validate projectDir and hash env cache

### LINKED ISSUE #591: fix(config): validate projectDir and hash workflow env cache
## 배경

#566 PR #585 리뷰에서 현재 범위 밖의 P3 두 건이 확인되었습니다.

## 범위

- `normalizeOrchestratorProjectConfig`에서 standalone `projectDir`의 절대경로를 검증해 상대경로가 데몬 CWD에서 조용히 해석되지 않게 한다.
- `WorkflowConfigStore`의 env 기반 캐시 키를 해시로 만들어 프로젝트·호스트 시크릿 원문을 메모리에 보관하지 않는다.

## 관련

- Follow-up for #566 / PR #585

moncher-dev:
## TL;DR

- Reject relative standalone `projectDir` values instead of resolving them implicitly from the daemon working directory.
- Hash environment-derived workflow cache identity so project and host secret values are not retained in plaintext cache metadata.
- Preserve and test environment-change invalidation, with a patch changeset for the published CLI runtime behavior.

## Change-point diagram

```text
standalone project config ──> normalizeOrchestratorProjectConfig ──> assertAbsoluteProjectDir
workflow path + env ────────> WorkflowConfigStore ────────────────> SHA-256 cache identity
changed effective env ──────> different digest ───────────────────> workflow re-parse
```

## Start here

- `packages/core/src/contracts/status-surface.ts` — standalone project validation at the shared normalization boundary.
- `packages/core/src/workflow/loader.ts` — hashed workflow environment cache identity.
- `packages/core/src/workflow-loader.test.ts` — plaintext protection and environment-change invalidation coverage.
- `packages/cli/src/config.test.ts` — persisted relative-directory rejection coverage.

## Risks & rollback

- Existing standalone configs with relative `projectDir` values will fail explicitly; using an absolute path restores startup.
- Hashing changes cache identity representation only; equality-based cache hits and changed-environment invalidation remain covered.
- Reverting this PR restores relative-path acceptance and plaintext environment signatures in cache metadata.

## Changed files

- `.changeset/secure-workflow-env-cache.md` — publishes the user-visible behavior as an `@gh-symphony/cli` patch.
- `docs/configuration.md` — documents absolute standalone paths and hashed environment cache metadata.
- `packages/cli/src/config.test.ts` — verifies relative standalone project directories are rejected.
- `packages/core/src/contracts/status-surface.ts` — asserts that persisted standalone project directories are absolute.
- `packages/core/src/workflow-loader.test.ts` — verifies digests exclude plaintext secrets and change when the environment changes.
- `packages/core/src/workflow/loader.ts` — stores a SHA-256 environment signature in workflow cache metadata.

## Evidence

- `pnpm exec vitest run packages/core/src/workflow-loader.test.ts packages/cli/src/config.test.ts` — 56 focused tests passed.
- `pnpm test` — passed across all workspace packages.
- `pnpm typecheck` — passed with strict TypeScript checks.
- `pnpm exec prettier --check docs/configuration.md packages/core/src/contracts/status-surface.ts packages/core/src/workflow/loader.ts packages/core/src/workflow-loader.test.ts packages/cli/src/config.test.ts .changeset/secure-workflow-env-cache.md` — passed.
- Docker E2E — not applicable; this changes Configuration Layer validation/cache internals, not dispatch, lifecycle, tracker, or status integration behavior.
- Spec conformance — conforms to `docs/symphony-spec.md`; no intentional divergence and the upstream spec was not modified.

- [ ] Confirm a standalone project configured with an absolute `projectDir` starts normally.
- [ ] Confirm a relative standalone `projectDir` reports a clear validation error.
- [ ] Confirm identical environments retain cache-hit behavior and changed environments trigger workflow re-parsing.

### DIFF
--- a/packages/core/src/contracts/status-surface.ts
+++ b/packages/core/src/contracts/status-surface.ts
@@ -52,6 +52,7 @@ export type OrchestratorProjectConfig = {
 export function normalizeOrchestratorProjectConfig(
   config: OrchestratorProjectConfig
 ): OrchestratorProjectConfig {
+  normalizeProjectDir(config);
   const workflowSource = normalizeWorkflowSource(config);
   const populateStrategy = normalizePopulateStrategy(config);
 
@@ -60,6 +61,17 @@ export function normalizeOrchestratorProjectConfig(
     workflowSource,
     populateStrategy,
   };
+}
+
+function normalizeProjectDir(config: OrchestratorProjectConfig): void {
+  if (config.projectDir === undefined) {
+    return;
+  }
+  if (typeof config.projectDir !== "string" || !isAbsolute(config.projectDir)) {
+    throw new Error(
+      `Project ${JSON.stringify(config.projectId)} project directory ${JSON.stringify(config.projectDir)} must be absolute.`
+    );
+  }
 }
 
 function normalizeWorkflowSource(
--- a/packages/core/src/contracts/status-surface.ts
+++ b/packages/core/src/contracts/status-surface.ts
@@ -52,7 +52,7 @@ export type OrchestratorProjectConfig = {
 export function normalizeOrchestratorProjectConfig(
   config: OrchestratorProjectConfig
 ): OrchestratorProjectConfig {
-  normalizeProjectDir(config);
+  assertAbsoluteProjectDir(config);
   const workflowSource = normalizeWorkflowSource(config);
   const populateStrategy = normalizePopulateStrategy(config);
 
@@ -63,7 +63,7 @@ export function normalizeOrchestratorProjectConfig(
   };
 }
 
-function normalizeProjectDir(config: OrchestratorProjectConfig): void {
+function assertAbsoluteProjectDir(config: OrchestratorProjectConfig): void {
   if (config.projectDir === undefined) {
     return;
   }
--- a/packages/core/src/workflow-loader.test.ts
+++ b/packages/core/src/workflow-loader.test.ts
@@ -857,6 +857,27 @@ Prompt body.
 });
 
 describe("WorkflowConfigStore", () => {
+  it("stores a hash instead of plaintext environment values in cache metadata", async () => {
+    const root = await mkdtemp(join(tmpdir(), "workflow-loader-"));
+    tempDirs.push(root);
+    const workflowPath = join(root, "WORKFLOW.md");
+    const store = new WorkflowConfigStore();
+    const secret = "project-and-host-secret";
+
+    await writeFile(workflowPath, SAMPLE_WORKFLOW, "utf8");
+    await store.load(workflowPath, { SECRET_TOKEN: secret });
+
+    const cache = (
+      store as unknown as {
+        cache: Map<string, { envSignature: string }>;
+      }
+    ).cache;
+    const envSignature = cache.get(workflowPath)?.envSignature;
+
+    expect(envSignature).toMatch(/^[a-f0-9]{64}$/);
+    expect(envSignature).not.toContain(secret);
+  });
+
   it("keeps the last known good workflow after an invalid update", async () => {
     const root = await mkdtemp(join(tmpdir(), "workflow-loader-"));
     tempDirs.push(root);
--- a/packages/core/src/workflow-loader.test.ts
+++ b/packages/core/src/workflow-loader.test.ts
@@ -863,6 +863,7 @@ describe("WorkflowConfigStore", () => {
     const workflowPath = join(root, "WORKFLOW.md");
     const store = new WorkflowConfigStore();
     const secret = "project-and-host-secret";
+    const changedSecret = "changed-project-and-host-secret";
 
     await writeFile(workflowPath, SAMPLE_WORKFLOW, "utf8");
     await store.load(workflowPath, { SECRET_TOKEN: secret });
@@ -874,8 +875,14 @@ describe("WorkflowConfigStore", () => {
     ).cache;
     const envSignature = cache.get(workflowPath)?.envSignature;
 
+    await store.load(workflowPath, { SECRET_TOKEN: changedSecret });
+    const changedEnvSignature = cache.get(workflowPath)?.envSignature;
+
     expect(envSignature).toMatch(/^[a-f0-9]{64}$/);
     expect(envSignature).not.toContain(secret);
+    expect(changedEnvSignature).toMatch(/^[a-f0-9]{64}$/);
+    expect(changedEnvSignature).not.toContain(changedSecret);
+    expect(changedEnvSignature).not.toBe(envSignature);
   });
 
   it("keeps the last known good workflow after an invalid update", async () => {
--- a/.changeset/secure-workflow-env-cache.md
+++ b/.changeset/secure-workflow-env-cache.md
@@ -1,0 +1,5 @@
+---
+"@gh-symphony/cli": patch
+---
+
+Reject relative standalone project directories and avoid retaining plaintext workflow environment values in cache metadata (#591).
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -16,7 +16,9 @@ must declare `repository.slug`, which is also what distinguishes a standalone
 project from a repository that embeds its own workflow. Configuration is derived
 from the folder on every start and cached under
 `<config-dir>/projects/<project-id>/`, where the project id is a stable function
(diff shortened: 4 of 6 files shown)


hojinzs (approved):
## Project #14 watchdog review — PR #617

**Approve.**

I reviewed head `34c9b504c4d1dfc3d5d1ea6abed4856966d53646` with Claude Code in read/test-only mode and ran the relevant smoke checks locally. No concrete blocker was found.

The project-directory validation is enforced at the shared config normalization chokepoint, and environment cache metadata is hashed without changing equality-based invalidation behavior. Tests cover both changes.

Validation: workflow-loader (48), CLI config (8), and orchestrator (301) tests passed; `pnpm lint`, `pnpm typecheck`, and `pnpm build` passed. The PR changes runtime/config code, so Project #14 will remain In review for Steve rather than being moved automatically to Land.

Evidence bundle: https://evidence-browser.dev0.daniworks.kr/w/github-symphony/b/project14-pr-617-20260821-134745z

packages/core/src/contracts/status-surface.ts
```diff
@@ -62,6 +63,17 @@ export function normalizeOrchestratorProjectConfig(
   };
 }
 
+function normalizeProjectDir(config: OrchestratorProjectConfig): void {
```
hojinzs: nit: `normalizeProjectDir` only validates (throws) and never normalizes/returns a value, unlike its siblings `normalizeWorkflowSource`/`normalizePopulateStrategy` which return normalized results. A name like `assertAbsoluteProjectDir` (or `validateProjectDir`) would read truer to what it does.

Otherwise this is solid — the runtime `typeof !== "string"` guard is good defensive coverage, since persisted `project.json` is cast via `as OrchestratorProjectConfig` and is effectively untrusted at load time.

---

packages/core/src/workflow-loader.test.ts
```diff
@@ -857,6 +857,27 @@ Prompt body.
...
+
+    const cache = (
+      store as unknown as {
+        cache: Map<string, { envSignature: string }>;
+      }
+    ).cache;
+    const envSignature = cache.get(workflowPath)?.envSignature;
+
+    expect(envSignature).toMatch(/^[a-f0-9]{64}$/);
+    expect(envSignature).not.toContain(secret);
```
hojinzs, line 882: P3 (test-coverage): this test asserts the signature is a hash and doesn't contain the secret, but nothing locks in that **a differing env still invalidates the cache**. That's the behavior most at risk from the hashing change — if the digest were ever computed as a constant, the cache would silently go stale on env change and this test would still pass.

I verified re-parse-on-env-change manually against `dist/` (two loads with different `PROJECT_ID` → distinct `githubProjectId`; identical env → cache hit), so behavior is correct today. Suggest adding an assertion to guard it, e.g. load twice with two different env maps and assert the two `envSignature`s differ (or that the resolved workflow re-parsed).

---

hojinzs (approved):
## Watchdog review — PR #617 (Approve)

Reviewed head `34c9b50` against issue #591. Both scoped P3 follow-ups from #566/#585 are fully and faithfully implemented, with no over-engineering or out-of-scope decisions.

### Requirements coverage
- ✅ **Absolute `projectDir` validation** — `normalizeProjectDir` rejects relative standalone project directories at the shared normalization chokepoint (`normalizeOrchestratorProjectConfig`), so both `loadProjectConfig` (fs-store + CLI) and every save path enforce it. Relative values no longer resolve silently against the daemon CWD.
- ✅ **Hashed env cache identity** — `WorkflowConfigStore` stores a SHA-256 digest of the effective env instead of the plaintext JSON, so project/host secret values are no longer retained in cache metadata. Equality-based invalidation is preserved (same digest ⇒ cache hit; different env ⇒ re-parse).

### Smoke tests (local, against built `dist/`)
- projectDir: `undefined`→allowed, absolute→allowed, `relative/project` / `.` / `""`→rejected with a clear message.
- env cache: different `PROJECT_ID` re-parses (`githubProjectId` tracks the new value); identical env is a cache hit; the digest is a 64-hex hash and neither it nor the cache entry contains the plaintext secret.
- `pnpm build` ✅, focused suites `workflow-loader.test.ts` + `config.test.ts` → 56 passed ✅. CI (Test, Container Smoke) green on this SHA.

### Findings
- **P2 (process):** No changeset on this branch. Per `CLAUDE.md` → Releases, behavior-changing PRs must add a `.changeset/*.md` for `@gh-symphony/cli`; rejecting relative `projectDir` at config load is a user-visible behavior change. Without it the fix won't be published on merge. There's no CI gate enforcing this, so flagging for a maintainer decision rather than blocking.
- **P3 (test-coverage):** the new cache test doesn't assert env-change invalidation — see inline comment.
- **nit:** `normalizeProjectDir` naming — see inline comment.

None of these are correctness blockers. Approving; recommend adding the changeset before merge.

---

packages/core/src/contracts/status-surface.ts
moncher-dev: Resolved in `166f8877`: renamed the validation-only helper to `assertAbsoluteProjectDir`, which now matches its throw-only behavior and distinguishes it from the value-returning normalization helpers.

packages/core/src/workflow-loader.test.ts
moncher-dev: Resolved in `166f8877`: the cache metadata test now loads the same workflow with two different secret environments and asserts both signatures are 64-character hashes, contain neither plaintext secret, and differ from each other. This locks in env-change invalidation alongside plaintext protection.

hojinzs (approved):
## Watchdog review — PR #617, round 2 (Approve)

Re-reviewed head `166f8877` against issue #591 and against the three findings from the previous round. All are addressed; no new blocker.

### Previous findings — follow-up status
- ✅ **P2 (process) — missing changeset.** `.changeset/secure-workflow-env-cache.md` added as an `@gh-symphony/cli` patch, correctly describing the user-visible behavior change.
- ✅ **P3 (test-coverage) — env-change invalidation not locked in.** The cache test now loads the same workflow with two different secret environments and asserts both digests are 64-hex, contain neither plaintext secret, and **differ from each other**. Thread resolved with an explanation.
- ✅ **nit — `normalizeProjectDir` naming.** Renamed to `assertAbsoluteProjectDir`, which now matches its throw-only contract and distinguishes it from the value-returning `normalizeWorkflowSource` / `normalizePopulateStrategy`. Thread resolved with an explanation.

### Requirements coverage (issue #591)
- ✅ Absolute `projectDir` validation at the shared normalization chokepoint — enforced on every load and every save path (`fs-store.loadProjectConfig`/`saveProjectConfig`, `cli/config.ts` load/save/save-within-lock).
- ✅ Hashed env cache identity — `WorkflowConfigStore` stores a SHA-256 digest instead of the plaintext JSON; `envSignature` is confined to `packages/core/src/workflow/loader.ts` and no consumer depends on its plaintext form.
- Scope is exactly the two P3 follow-ups (72 additions / 7 deletions across 6 files). No over-engineering, no unrequested decisions.

### Smoke tests (local, against built `dist/`)
- `projectDir`: `undefined` → allowed; `/abs/ok` → allowed; `relative/x`, `.`, `..`, `""`, `42`, `null` → rejected with a clear, value-quoting message.
- Env cache: identical env → cache hit (entry object and `loadedAt` untouched, same digest); changed env → digest differs and the workflow re-parses (`tracker.project_id` `A` → `B`); serialized cache metadata contains no plaintext secret.
- `pnpm build` ✅; focused suites `workflow-loader.test.ts` + `config.test.ts` → 56 passed ✅; CI **Test** and **Container Smoke** green on `166f8877`.
- CLI entry points are unaffected: `project.ts` resolves `--project-dir` (and the cwd fallback) to an absolute path before building the config, so no CLI-driven flow can produce a rejected value.

### Remaining observations (non-blocking, no action required)
- **nit (pre-existing, out of scope):** `listStandaloneProjects` maps `loadProjectConfig` through `Promise.all`, so a single hand-edited `project.json` with a relative `projectDir` now fails the whole `project list` instead of skipping that one entry. This mirrors the existing behavior for an invalid `workflowSource.path`, so it is not a regression from this PR — worth a separate follow-up only if malformed persisted configs are considered likely.

Approving. Ready to merge from my side.

---

PR #618, opened 2026-08-21
fix(cli): resolve active standalone project for diagnostics

### LINKED ISSUE #589: fix(cli): resolve active standalone project for diagnostics
## 배경

standalone 프로젝트를 여러 개 등록한 경우 `gh-symphony doctor`와 `gh-symphony workflow`의 프로젝트 선택 경로가 `activeProject`를 사용하지 않는다. 따라서 비대화형 실행이 기존 `repo init` 안내와 함께 중단될 수 있다.

## 범위

- `inspectManagedProjectSelection`에서 multi-project registry의 `activeProject`를 일관되게 해석한다.
- standalone 문맥에 맞는 진단 안내문과 회귀 테스트를 추가한다.

## 제외

- #569의 project add/start/status/stop lifecycle 변경은 포함하지 않는다.

moncher-dev:
## Issues

- Closed #589

## TL;DR

- Resolve a configured `activeProject` before rejecting non-interactive multi-project diagnostic commands.
- Give accurate standalone-aware remediation for `doctor` and `workflow preview`, including the rework corrections requested in review.

## Change-point diagram

```text
doctor / workflow preview
          |
          v
inspectManagedProjectSelection
          |
          +-- explicit project id --> selected config
          +-- registry activeProject --> selected config
          +-- one registry entry --> selected config
          `-- ambiguous registry --> project-list / explicit-selector guidance
```

## Start here

Review `packages/cli/src/project-selection.ts` first, then the regression cases in `packages/cli/src/project-selection.test.ts`. The `doctor` remediation correction and its regression assertion are isolated to `packages/cli/src/commands/doctor.ts` and `doctor.test.ts`.

## Changes

- Move `activeProject` resolution ahead of multi-project ambiguity handling in the diagnostic selector.
- Preserve the distinct missing-active-project-config diagnosis when the registry entry exists but `project.json` is missing.
- Direct operators to `gh-symphony project list`, `--project-id`, or `--project-dir` without claiming cwd selection that `doctor` does not implement.
- Preserve explicit and single-project selection behavior without changing project lifecycle commands from #569.
- Document the active standalone diagnostic behavior and include a CLI patch changeset.

## Evidence

- `pnpm --filter @gh-symphony/cli test -- project-selection.test.ts commands/doctor.test.ts commands/workflow.test.ts` — 79 tests passed.
- `pnpm test` — passed across all workspace packages.
- Targeted Prettier check and `git diff --check` — passed.
- Every inline review comment has a concrete reply; the folder-identity design reconciliation is tracked separately in #619.

## Risks & rollback

- Risk is limited to managed-project selection and diagnostic copy; explicit project IDs and single-project registries retain their existing precedence.
- `activeProject` precedence is intentional for #589; cwd-derived standalone selection remains outside this PR.
- Roll back commits `908dce8` and `f5e477e` to restore the prior selector behavior and messages.

## Changed files

- `packages/cli/src/project-selection.ts` and test
- `packages/cli/src/commands/doctor.ts` and test
- `README.md` and `packages/cli/README.md`
- `.changeset/tidy-ravens-resolve.md`

- [ ] Run `gh-symphony doctor` with two registered standalone projects and confirm the active project is selected.
- [ ] Remove the active project's `project.json` and confirm the diagnostic distinguishes the missing project config from an unregistered id.
- [ ] Clear `activeProject` and confirm the diagnostic points to `project list`, `--project-id`, and `--project-dir` without suggesting cwd selection.

### DIFF
--- a/packages/cli/src/commands/doctor.ts
+++ b/packages/cli/src/commands/doctor.ts
@@ -1767,7 +1767,7 @@ export async function runDoctorDiagnostics(
         "managed_project",
         "Managed project selection",
         resolvedProjectConfig.message,
-        "For a standalone project, run from its project folder or pass '--project-dir <path>'. For a repository runtime, run 'gh-symphony repo init' from the target repository.",
+        "For a standalone project, pass '--project-dir <path>'. For a repository runtime, run 'gh-symphony repo init' from the target repository.",
         {
           reason: resolvedProjectConfig.kind,
           ...(resolvedProjectConfig.projectId
--- a/packages/cli/src/commands/doctor.ts
+++ b/packages/cli/src/commands/doctor.ts
@@ -1767,7 +1767,7 @@ export async function runDoctorDiagnostics(
         "managed_project",
         "Managed project selection",
         resolvedProjectConfig.message,
-        "Run 'gh-symphony repo init' from the target repository.",
+        "For a standalone project, run from its project folder or pass '--project-dir <path>'. For a repository runtime, run 'gh-symphony repo init' from the target repository.",
         {
           reason: resolvedProjectConfig.kind,
           ...(resolvedProjectConfig.projectId
--- a/packages/cli/src/project-selection.ts
+++ b/packages/cli/src/project-selection.ts
@@ -39,11 +39,15 @@ function explicitProjectRequiredMessage(): string {
 }
 
 function diagnosticProjectRequiredMessage(): string {
-  return "Multiple managed projects are configured and no active project is set. Pass '--project-id <project-id>' to select one. For standalone diagnostics, pass '--project-dir <path>' to 'gh-symphony doctor'.";
+  return "Multiple managed projects are configured and no active project is set. Pass '--project-id <project-id>' to select one. Run 'gh-symphony project list' to see configured project ids. For standalone diagnostics, pass '--project-dir <path>' to 'gh-symphony doctor'.";
+}
+
+function projectConfigRemediation(): string {
+  return "For a standalone project, run 'gh-symphony project start' from its project folder to refresh the runtime config. For a repository runtime, run 'gh-symphony repo init' from the target repository.";
 }
 
 function missingProjectConfigMessage(projectId: string): string {
-  return `Project "${projectId}" is not configured. For a standalone project, run 'gh-symphony project start' from its project folder to refresh the runtime config. For a repository runtime, run 'gh-symphony repo init' from the target repository.`;
+  return `Project "${projectId}" is not configured. ${projectConfigRemediation()}`;
 }
 
 export async function inspectManagedProjectSelection(
@@ -96,7 +100,7 @@ export async function inspectManagedProjectSelection(
       return {
         kind: "active_project_missing",
         projectId: global.activeProject,
-        message: `Active ${missingProjectConfigMessage(global.activeProject)}`,
+        message: `Active project "${global.activeProject}" is configured in config.json but its project config is missing. ${projectConfigRemediation()}`,
       };
     }
 
--- a/packages/cli/src/project-selection.ts
+++ b/packages/cli/src/project-selection.ts
@@ -38,6 +38,14 @@ function explicitProjectRequiredMessage(): string {
   return "Multiple repository runtime configs are present. Run 'gh-symphony repo init' from the target repository to refresh the cwd runtime.\n";
 }
 
+function diagnosticProjectRequiredMessage(): string {
+  return "Multiple managed projects are configured and no active project is set. Pass '--project-id <project-id>' to select one. For standalone diagnostics, pass '--project-dir <path>' to 'gh-symphony doctor'.";
+}
+
+function missingProjectConfigMessage(projectId: string): string {
+  return `Project "${projectId}" is not configured. For a standalone project, run 'gh-symphony project start' from its project folder to refresh the runtime config. For a repository runtime, run 'gh-symphony repo init' from the target repository.`;
+}
+
 export async function inspectManagedProjectSelection(
   input: ResolveProjectSelectionInput
 ): Promise<ManagedProjectResolution> {
@@ -50,7 +58,7 @@ export async function inspectManagedProjectSelection(
       return {
         kind: "requested_project_missing",
         projectId: input.requestedProjectId,
-        message: `Project "${input.requestedProjectId}" is not configured. Run 'gh-symphony repo init' from the target repository.`,
+        message: missingProjectConfigMessage(input.requestedProjectId),
       };
     }
 
@@ -79,13 +87,6 @@ export async function inspectManagedProjectSelection(
     };
   }
 
-  if (projectIds.length > 1 && !isInteractiveTerminal()) {
-    return {
-      kind: "multiple_projects_require_selection",
-      message: explicitProjectRequiredMessage().trimEnd(),
-    };
-  }
-
   if (global.activeProject) {
     const projectConfig = await loadProjectConfig(
       input.configDir,
@@ -95,7 +96,7 @@ export async function inspectManagedProjectSelection(
       return {
         kind: "active_project_missing",
         projectId: global.activeProject,
-        message: `Active project "${global.activeProject}" is configured in config.json but its project config is missing. Re-run 'gh-symphony repo init'.`,
+        message: `Active ${missingProjectConfigMessage(global.activeProject)}`,
       };
     }
 
@@ -113,7 +114,7 @@ export async function inspectManagedProjectSelection(
       return {
         kind: "configured_project_missing",
         projectId,
-        message: `Configured project "${projectId}" is missing its project config file. Re-run 'gh-symphony repo init'.`,
+        message: missingProjectConfigMessage(projectId),
       };
     }
 
@@ -126,8 +127,7 @@ export async function inspectManagedProjectSelection(
 
   return {
     kind: "multiple_projects_require_selection",
-    message:
-      "Multiple repository runtime configs are present and no active project is set. Re-run 'gh-symphony repo init' from the target repository.",
+    message: diagnosticProjectRequiredMessage(),
   };
 }
 
--- a/.changeset/tidy-ravens-resolve.md
+++ b/.changeset/tidy-ravens-resolve.md
@@ -1,0 +1,5 @@
+---
+"@gh-symphony/cli": patch
+---
+
+Fix #589 by resolving the active standalone project for multi-project `doctor` and `workflow preview` diagnostics, with standalone-specific selection guidance.
--- a/README.md
+++ b/README.md
@@ -599,6 +599,8 @@ gh-symphony config edit             # Open config in $EDITOR
 
 `gh-symphony doctor` runs a single first-run diagnostic pass and exits non-zero if any required prerequisite is missing. `gh-symphony doctor --fix` adds a remediation pass on top of the same checks. `gh-symphony doctor --smoke` is the recommended final preflight before `gh-symphony repo start --once`: it resolves the active managed project, checks the GitHub Project binding, confirms the repository and target issue are readable through the project, renders `WORKFLOW.md` for that issue, verifies the runtime command, workspace root, and configured hook paths, and exits without dispatching a worker.
 
+When the global registry contains multiple standalone projects, `doctor` and live `workflow preview` diagnostics use its `activeProject`. To inspect a different standalone project explicitly, pass `--project-dir <path>` to `doctor` or `--project-id <projectId>` to `workflow preview`.
+
 Use an explicit issue when you want a deterministic check:
 
 ```bash
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -378,6 +378,8 @@ Without `--issue`, doctor auto-selects one active live issue from the managed pr
 
 `gh-symphony doctor --fix` extends the regular diagnostic flow with safe remediation and guided follow-up:
 
+When multiple standalone projects are registered, diagnostics resolve the registry's `activeProject`. Use `doctor --project-dir <path>` or `workflow preview --project-id <projectId>` to select another project explicitly.
+
 - creates missing config/runtime/workspace directories
 - launches `gh auth login` or `gh auth refresh` when a TTY is available, otherwise prints the exact command to run
 - launches `gh-symphony workflow init` when `WORKFLOW.md` is missing or invalid
--- a/packages/cli/src/commands/doctor.test.ts
+++ b/packages/cli/src/commands/doctor.test.ts
@@ -1066,6 +1066,9 @@ describe("runDoctorDiagnostics", () => {
       summary: expect.stringContaining("repo init"),
       remediation: expect.stringContaining("--project-dir <path>"),
     });
+    expect(
+      report.checks.find((check) => check.id === "managed_project")?.remediation
+    ).not.toContain("run from its project folder");
   });
 
   it("accepts env-token auth when gh CLI is unavailable", async () => {
--- a/packages/cli/src/commands/doctor.test.ts
+++ b/packages/cli/src/commands/doctor.test.ts
@@ -1064,6 +1064,7 @@ describe("runDoctorDiagnostics", () => {
     ).toMatchObject({
       status: "fail",
       summary: expect.stringContaining("repo init"),
+      remediation: expect.stringContaining("--project-dir <path>"),
     });
   });
 
--- a/packages/cli/src/project-selection.test.ts
+++ b/packages/cli/src/project-selection.test.ts
@@ -174,7 +174,7 @@ describe("resolveManagedProjectConfig", () => {
 });
 
 describe("inspectManagedProjectSelection", () => {
-  it("requires explicit project selection in non-interactive multi-project mode", async () => {
+  it("uses the active project in non-interactive multi-project mode", async () => {
     const configDir = await createConfigFixture(
       [createProject("tenant-a"), createProject("tenant-b")],
       "tenant-b"
@@ -184,10 +184,25 @@ describe("inspectManagedProjectSelection", () => {
     const result = await inspectManagedProjectSelection({ configDir });
 
     expect(result).toMatchObject({
-      kind: "multiple_projects_require_selection",
-      message:
-        "Multiple repository runtime configs are present. Run 'gh-symphony repo init' from the target repository to refresh the cwd runtime.",
+      kind: "resolved",
+      projectId: "tenant-b",
     });
+  });
+
+  it("reports standalone selection guidance when multiple projects have no active project", async () => {
+    const configDir = await createConfigFixture([
+      createProject("tenant-a"),
+      createProject("tenant-b"),
+    ]);
+    setTty(false, false);
+
+    const result = await inspectManagedProjectSelection({ configDir });
+
+    expect(result).toMatchObject({
+      kind: "multiple_projects_require_selection",
+      message: expect.stringContaining("--project-dir <path>"),
+    });
+    expect(result.message).not.toContain("repo init");
   });
 
   it("uses the active project when one is configured", async () => {
--- a/packages/cli/src/project-selection.test.ts
+++ b/packages/cli/src/project-selection.test.ts
@@ -202,6 +202,7 @@ describe("inspectManagedProjectSelection", () => {
       kind: "multiple_projects_require_selection",
       message: expect.stringContaining("--project-dir <path>"),
     });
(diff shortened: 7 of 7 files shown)


hojinzs (approved):
## Project #14 watchdog review — PR #618

**Approve.**

I reviewed head `f5e477e9f2dbb15db098acff0e31be284edbb841` with Claude Code in read/test-only mode and ran the relevant smoke checks locally. No blocker was found in the PR changes.

The diagnostic selection change correctly honors `global.activeProject` in non-interactive multi-project mode and adds standalone-project guidance when no active project exists. The changed behavior is narrowly scoped, tested, and includes a CLI changeset.

The targeted CLI run has one pre-existing failure in `doctor command handler > reports manual remediation commands in non-interactive fix mode`; the same assertion fails on `origin/main` with the same redacted-config-path mismatch. The changed project-selection tests pass. `pnpm lint`, `pnpm typecheck`, and `pnpm build` passed. The PR changes CLI/runtime behavior, so Project #14 will remain In review for Steve rather than being moved automatically to Land.

Evidence bundle: https://evidence-browser.dev0.daniworks.kr/w/github-symphony/b/project14-pr-618-20260821-134745z

packages/cli/src/commands/doctor.ts
```diff
@@ -1767,7 +1767,7 @@ export async function runDoctorDiagnostics(
         "managed_project",
         "Managed project selection",
         resolvedProjectConfig.message,
-        "Run 'gh-symphony repo init' from the target repository.",
+        "For a standalone project, run from its project folder or pass '--project-dir <path>'. For a repository runtime, run 'gh-symphony repo init' from the target repository.",
```
hojinzs: **P1 — "run from its project folder" is not a working remediation for `doctor`.**

`runDoctorDiagnostics` passes only `parsedArgs.projectId` into `inspectManagedProjectSelection` (doctor.ts:1739-1742), and `parseDoctorArgs` sets `projectId` exclusively from `--project-id` / `--project-dir` (doctor.ts:266-287). There is no cwd-derived standalone fallback anywhere in the doctor path — `process.cwd()` is only used for `repoRoot`. So the first half of this remediation sends the operator down a dead end.

Reproduced locally against `packages/cli/dist/index.js` (two registered standalone projects, `activeProject: null`), running **from inside `proj-a`'s folder**:

```
FAIL Managed project selection
  Multiple managed projects are configured and no active project is set. ...
  Fix: For a standalone project, run from its project folder or pass '--project-dir <path>'. ...
```

The operator did exactly what the fix line says and got the same failure. Suggest dropping the folder clause:

```suggestion
        "For a standalone project, pass '--project-dir <path>'. For a repository runtime, run 'gh-symphony repo init' from the target repository.",
```

(Alternatively, implement the cwd fallback so the sentence becomes true — see my note on `project-selection.ts:90`.)

---

packages/cli/src/project-selection.ts
```diff
@@ -95,7 +96,7 @@ export async function inspectManagedProjectSelection(
       return {
         kind: "active_project_missing",
         projectId: global.activeProject,
-        message: `Active project "${global.activeProject}" is configured in config.json but its project config is missing. Re-run 'gh-symphony repo init'.`,
+        message: `Active ${missingProjectConfigMessage(global.activeProject)}`,
```
hojinzs: **P2 — the `Active ` prefix produces broken copy and, more importantly, the wrong diagnosis.**

Prefixing a full sentence yields a capital `P` mid-sentence. Actual output, reproduced locally with `activeProject: "ghost-1234abcd"`:

```
FAIL Managed project selection
  Active Project "ghost-1234abcd" is not configured. For a standalone project, run 'gh-symphony project start' ...
```

Beyond the grammar, this case is *not* "not configured". `active_project_missing` means the id **is** registered in `config.json` but `projects/<id>/project.json` is gone — which is exactly what the previous message said ("configured in config.json but its project config is missing"). The unified wording erases the one detail that distinguishes this failure from `requested_project_missing`, and points the operator at the wrong cause.

Suggest keeping the case-specific first sentence and reusing only the remediation half:

```suggestion
        message: `Active project "${global.activeProject}" is configured in config.json but its project config is missing. ${projectConfigRemediation()}`,
```

…where `projectConfigRemediation()` is the "For a standalone project, run 'gh-symphony project start' … For a repository runtime, run 'gh-symphony repo init' …" tail, and `missingProjectConfigMessage()` composes the same tail for the other two cases. That keeps the dedup you were after without flattening three distinct diagnoses into one.

---

packages/cli/src/project-selection.ts
```diff
@@ -79,13 +87,6 @@ export async function inspectManagedProjectSelection(
...
   }
 
-  if (projectIds.length > 1 && !isInteractiveTerminal()) {
-    return {
-      kind: "multiple_projects_require_selection",
-      message: explicitProjectRequiredMessage().trimEnd(),
-    };
-  }
-
   if (global.activeProject) {
```
hojinzs, line 94: **P2 — folder identity vs `activeProject` (confirming the unresolved Codex thread on this branch).**

The Codex P2 at `#discussion_r3830433054` is still open with no fix and no reply. I reproduced it, so it is not a false positive. Two registered standalone projects, `activeProject: proj-b-dea0eb93`, `doctor` run **from `proj-a`'s folder**:

```
PASS Managed project selection
  Resolved managed project "proj-b-dea0eb93".
```

`docs/designs/2026-08-11-standalone-project-model-design.md:71-74` states that folder identity, not `activeProject`, selects the standalone project. Before this change, that host got an explicit "select a project" error; now it silently diagnoses a different project.

That said, I do **not** think this blocks the PR as designed:

- Issue #589 asks for exactly this (`multi-project registry의 activeProject를 일관되게 해석한다`), so `activeProject` resolution is the requested behavior, not scope creep.
- The failure is visible, not silent-wrong: the check prints the resolved project id, and `externalWorkflowPath` (doctor.ts:2055-2056) follows the *resolved* project's `workflowSource`, so the rest of the report is internally consistent with the project it named rather than mixing cwd and registry.

So the ask here is a decision, not necessarily code. Either:

1. resolve a cwd-derived standalone id first when `cwd/WORKFLOW.md` maps to a registered project (which would also make the `doctor.ts:1770` remediation sentence true), or
2. reply on the Codex thread stating that `activeProject` precedence is intentional per #589 and that the design-doc line needs a follow-up issue.

Either is fine. Leaving the thread unanswered is what I'd like to avoid — the previous review round closed without a response on it.

---

packages/cli/src/project-selection.ts
```diff
@@ -38,6 +38,14 @@ function explicitProjectRequiredMessage(): string {
   return "Multiple repository runtime configs are present. Run 'gh-symphony repo init' from the target repository to refresh the cwd runtime.\n";
 }
 
+function diagnosticProjectRequiredMessage(): string {
+  return "Multiple managed projects are configured and no active project is set. Pass '--project-id <project-id>' to select one. For standalone diagnostics, pass '--project-dir <path>' to 'gh-symphony doctor'.";
```
hojinzs: **nit — `--project-id <project-id>` is unactionable for a standalone project without saying where the id comes from.**

Standalone ids are hash-suffixed derivations of the folder path (`projectIdentifier()` in `commands/project.ts:349-351`, e.g. `proj-a-23f4e6dc`), so an operator hitting this message has no way to type one from memory. Verified this is the message `workflow preview` surfaces:

```
$ gh-symphony workflow preview --file ./WORKFLOW.md --issue owner/repo#589
Workflow command failed: Multiple managed projects are configured and no active project is set. Pass '--project-id <project-id>' to select one. ...
```

Worth appending `Run 'gh-symphony project list' to see configured project ids.` Also, this message is shared by `doctor` and `workflow preview`, but only names `doctor` in the `--project-dir` clause — fine as-is, just noting it reads slightly oddly from the `preview` side.

---

hojinzs (changes requested):
## Review round 2 — PR #618 (head `f5e477e`)

**Request changes.** The implementation is right; the user-facing copy it ships is not. Both blocking items are one-line fixes, so I expect this to close in a single round.

### Requirement coverage vs #589 — complete

| #589 requirement | Status |
| --- | --- |
| `inspectManagedProjectSelection` resolves the multi-project registry `activeProject` consistently | ✅ verified by smoke test |
| Standalone-context diagnostic copy + regression tests | ⚠️ present, but two of the new strings are wrong (see inline) |
| Exclusion: no #569 lifecycle changes | ✅ `resolveManagedProjectConfig` and all `project add/start/status/stop` paths untouched |
| Changeset + docs | ✅ `.changeset/tidy-ravens-resolve.md`, both READMEs |

No over-engineering found. `explicitProjectRequiredMessage()` is still live at `project-selection.ts:159` (execution path), so the diff leaves no dead code, and the change is confined to the two `inspectManagedProjectSelection` consumers (`doctor`, `workflow preview`).

### Smoke test — local, real CLI binary

`pnpm build`, then `packages/cli/dist/index.js` against a fixture config dir with two registered standalone projects (`proj-a`, `proj-b`):

| Scenario | Result |
| --- | --- |
| `doctor`, `activeProject: proj-b`, run from `proj-a/` | `PASS … Resolved managed project "proj-b-dea0eb93"` — **the #589 fix works** (previously a hard failure) |
| `doctor --project-dir …/proj-a` | `PASS … Resolved managed project "proj-a-23f4e6dc"` — explicit selection preserved |
| `doctor`, `activeProject: null` | `FAIL` with the new multi-project guidance — expected |
| `doctor`, `activeProject` → missing project.json | `FAIL` with `Active Project "…" is not configured.` — **defect, see inline** |
| `workflow preview`, `activeProject: null` | fails with the new guidance — expected |
| `workflow preview`, `activeProject: proj-b` | passes the selection gate, proceeds to GitHub auth — **the #589 fix works on the preview path too** |

Also ran: `packages/cli/src/project-selection.test.ts` + `commands/doctor.test.ts` → **65 passed**; `pnpm --filter @gh-symphony/cli typecheck` → clean; `pnpm build` → clean. CI `Test` and `Container Smoke` are both green on `f5e477e`. I could not reproduce the pre-existing `doctor` test failure noted in the previous round — `doctor.test.ts` passed 56/56 here.

### Findings

| Pri | Where | Issue |
| --- | --- | --- |
| **P1** | `commands/doctor.ts:1770` | Remediation tells the operator to "run from its project folder", but `doctor` has no cwd-derived project selection at all. Reproduced: doing exactly that yields the same failure. |
| **P2** | `project-selection.ts:99` | `Active ${missingProjectConfigMessage(…)}` renders `Active Project "x" is not configured.` — broken grammar, and it misdiagnoses: the project *is* registered, only its `project.json` is missing. The prior message said so; this one loses it. |
| **P2** | `project-selection.ts:90` | The Codex thread on folder-identity-vs-`activeProject` is reproducible and still has no fix and no reply. I don't think it blocks — #589 explicitly asks for `activeProject` precedence — but it needs an answer rather than silence. |
| nit | `project-selection.ts:42` | `--project-id <project-id>` with no hint that standalone ids are hash-derived; suggest pointing at `gh-symphony project list`. |

### Path to approval

Fix the P1 and the P2 at `:99`, and either implement or reply-and-defer the `:90` thread. Nothing here warrants a redesign, and I don't want this cycling — one corrective push and I'll approve. Review budget used: 2 of 6.

---

packages/cli/src/project-selection.ts
```diff
@@ -79,13 +87,6 @@ export async function inspectManagedProjectSelection(
...
-  if (projectIds.length > 1 && !isInteractiveTerminal()) {
-    return {
-      kind: "multiple_projects_require_selection",
-      message: explicitProjectRequiredMessage().trimEnd(),
-    };
-  }
-
   if (global.activeProject) {
     const projectConfig = await loadProjectConfig(
       input.configDir,
```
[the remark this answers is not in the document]
moncher-dev: Retaining `activeProject` precedence is intentional in this PR because #589 explicitly requires `inspectManagedProjectSelection` to resolve the multi-project registry's `activeProject` consistently for `doctor` and `workflow`. I have removed the inaccurate cwd remediation so the CLI no longer suggests that folder identity affects this path. The broader folder-identity design reconciliation is tracked separately in #619.

packages/cli/src/commands/doctor.ts
moncher-dev: Fixed. The remediation now says `For a standalone project, pass '--project-dir <path>'.` and no longer claims that running from the project folder selects a project. A regression assertion verifies the removed wording does not return.

packages/cli/src/project-selection.ts
moncher-dev: Fixed as suggested. `active_project_missing` now preserves the distinct diagnosis: the active id is configured in `config.json`, but its project config is missing. The shared remediation tail was extracted into `projectConfigRemediation()` without flattening the case-specific message, and the exact output is covered by a regression test.

packages/cli/src/project-selection.ts
moncher-dev: Decision recorded: this PR retains `activeProject` precedence because that is the explicit #589 requirement. I removed the inaccurate cwd-based remediation, and opened #619 to reconcile the standalone folder-identity design with diagnostic selection separately from this scoped fix.

packages/cli/src/project-selection.ts
moncher-dev: Fixed. The multi-project diagnostic now includes `Run 'gh-symphony project list' to see configured project ids.`, and the regression test asserts that hint is present.

packages/cli/src/commands/doctor.ts
```diff
@@ -1767,7 +1767,7 @@ export async function runDoctorDiagnostics(
         "managed_project",
         "Managed project selection",
         resolvedProjectConfig.message,
-        "Run 'gh-symphony repo init' from the target repository.",
+        "For a standalone project, pass '--project-dir <path>'. For a repository runtime, run 'gh-symphony repo init' from the target repository.",
```
hojinzs, line 1770: **nit (non-blocking — do not push for this) — the `--project-dir` clause is a dead end for the `no_projects` / `missing_global_config` kinds.**

The P1 from round 2 is fixed and I verified it: the "run from its project folder" clause is gone, and `--project-dir` genuinely works (smoke S2 below resolved `proj-a` while `proj-b` was active). This is a smaller residue of the same shape.

This one `failCheck` remediation is shared by all five non-`resolved` kinds. For `no_projects` (empty registry) the operator gets:

```
FAIL Managed project selection
  No repository runtime config is configured. Run 'gh-symphony repo init' first.
  Fix: For a standalone project, pass '--project-dir <path>'. ...
```

`--project-dir` derives an id via `standaloneProjectId()` and hands it to `loadProjectConfig`, which cannot resolve on an empty registry — so that half of the fix line can't succeed. `gh-symphony project start` is the standalone entry point there.

Explicitly **not** asking for a push. The summary line already carries the correct per-kind guidance, and the pre-change text (`Run 'gh-symphony repo init'`) was equally wrong for standalone, so this is a net improvement either way. Worth folding into #619 when the standalone copy is reconciled.

---

packages/cli/src/project-selection.ts
```diff
@@ -113,7 +118,7 @@ export async function inspectManagedProjectSelection(
       return {
         kind: "configured_project_missing",
         projectId,
-        message: `Configured project "${projectId}" is missing its project config file. Re-run 'gh-symphony repo init'.`,
+        message: missingProjectConfigMessage(projectId),
```
hojinzs, line 121: **nit (non-blocking, pre-existing — noting for the record only) — precedence now reads inverted vs. `resolveManagedProjectConfig`.**

After the move, this function orders `activeProject` → single-project (this branch). `resolveManagedProjectConfig` (the execution path, `:148-158`) orders single-project → `activeProject`. They differ in exactly one state: **one** registered project plus an `activeProject` pointing at a stale id.

- diagnostic path → `active_project_missing`, `doctor` FAILs
- execution path → falls into `projectIds.length === 1`, loads the project, `repo start` works

I checked the pre-change code before flagging this: with a single project `projectIds.length > 1` was false, so the old ordering reached the same `activeProject` branch and produced the same `active_project_missing`. **This PR does not introduce or widen the divergence** — the diff only moves the branch above a gate that never fired for single-project registries. Recording it so it isn't mistaken for a regression from this diff later; #619 is the natural home if it's ever worth aligning.

---

hojinzs (approved):
## Review round 3 — PR #618 (head `908dce8`)

**Approve.** All four round-2 findings are fixed and I re-verified each one against a real build of the CLI, not just the test suite. CI is green on this head. Review budget used: **3 of 6** — closing here as promised.

### Round-2 findings → verified fixed

| Pri | Where | Fix | Verified by |
| --- | --- | --- | --- |
| **P1** | `doctor.ts:1770` | "run from its project folder" clause removed; now `pass '--project-dir <path>'` | Smoke **S3** — the failing message no longer contains the dead-end instruction, and **S2** proves `--project-dir` actually resolves. Regression assertion `not.toContain("run from its project folder")` added. |
| **P2** | `project-selection.ts:103` | `active_project_missing` keeps its distinct diagnosis; `projectConfigRemediation()` extracted as the shared tail | Smoke **S4** — exact string reproduced; grammar bug gone. Test asserts the full message and `not.toContain('Active Project "tenant-a"')`. |
| **P2** | folder-identity thread | Reply-and-defer, with #619 opened and correctly scoped | [Codex thread](https://github.com/hojinzs/github-symphony/pull/618#discussion_r3830433054) answered; #619 exclusions explicitly carve out #589 and #569. This was one of the two options I offered — accepted. |
| **nit** | `project-selection.ts:42` | `Run 'gh-symphony project list' to see configured project ids.` appended | Smoke **S3** / **S7**. I confirmed `project list` (`project.ts:369`) and `workflow preview --project-id` (`workflow.ts:267`) both exist, so every command the message names is real. |

### Requirement coverage vs #589 — complete

| #589 requirement | Status |
| --- | --- |
| `inspectManagedProjectSelection` resolves the multi-project registry `activeProject` consistently for `doctor` **and** `workflow` | ✅ both paths verified (S1, S6) |
| Standalone-context diagnostic copy + regression tests | ✅ copy corrected this round; 3 new/updated assertions |
| Exclusion: no #569 lifecycle changes | ✅ `resolveManagedProjectConfig` and all `project add/start/status/stop` paths untouched |
| Changeset + docs | ✅ `.changeset/tidy-ravens-resolve.md` (cli patch), both READMEs |

**No over-engineering.** The corrective commit is 3 message helpers and a moved branch — 52/-16 across 7 files, with no speculative abstraction and no scope beyond what the review asked for. Deferring folder-identity to #619 rather than implementing a cwd fallback inside this PR was the right call: it keeps the fix scoped to what #589 asked for.

### Smoke test — local, real CLI binary

`pnpm build`, then `packages/cli/dist/index.js` against a fixture config dir with two registered standalone projects.

| # | Scenario | Result |
| --- | --- | --- |
| S1 | `doctor`, `activeProject: proj-b`, run from `proj-a/` | `PASS … Resolved managed project "proj-b-99aead2f"` — **#589 fix works** |
| S2 | `doctor --project-dir …/proj-a` while `proj-b` active | `PASS … Resolved managed project "proj-a-6e6201d2"` — explicit selection preserved, **P1 remediation is now true** |
| S3 | `doctor`, `activeProject: null` | `FAIL` with the corrected guidance incl. the `project list` hint — **P1 + nit fixed** |
| S4 | `doctor`, `activeProject` → missing `project.json` | `FAIL … "ghost-1234abcd" is configured in config.json but its project config is missing.` — **P2 fixed**, distinct diagnosis restored |
| S5 | `doctor --project-id nope-00000000` | `FAIL … "nope-00000000" is not configured.` — `requested_project_missing` stays distinguishable from S4 |
| S6 | `workflow preview`, `activeProject: proj-b`, from `proj-a/` | passes the selection gate, proceeds to GitHub auth — **fix works on the preview path** |
| S7 | `workflow preview`, `activeProject: null` | fails with the corrected guidance — expected |

Also ran: `project-selection.test.ts` + `doctor.test.ts` + `workflow.test.ts` → **79 passed**; `pnpm lint` → clean; `pnpm typecheck` → clean; `pnpm build` → clean. CI **Test** and **Container Smoke** both green on `908dce8`.

### Remaining notes — non-blocking, no push wanted

Two nits inline. Neither is a regression from this diff and I am **not** asking for another round on them:

- `doctor.ts:1770` — the shared fix line offers `--project-dir` even for the `no_projects` kind, where it can't resolve. Net improvement over the previous text regardless; fold into #619.
- `project-selection.ts:121` — diagnostic vs. execution precedence differ for the single-project + stale-`activeProject` state. I checked the pre-change code: the old ordering produced the same result, so this PR neither introduces nor widens it.

Ship it.

---

PR #621, opened 2026-08-23
feat(orchestrator): honor workspace.root for embedded projects

### LINKED ISSUE #599: feat(orchestrator): honor workspace.root for repo-embedded projects (spec 9.1 convergence)
## Context

Spec §9.1 defines the per-issue workspace path as `<workspace.root>/<sanitized_issue_identifier>`,
and Invariant 2 requires the workspace path to stay under that root. The orchestrator implemented
this until `dffb86a refactor(runtime): consolidate project layout` (2026-03-17), which replaced
`tenant.workspaceDir` with `store.projectDir(projectId)` for both the workflow cache and the issue
workspace path. Since then `workspace.root` has been a no-op for every project — a repository
declaring `workspace.root: .runtime/symphony-workspaces` still gets
`.runtime/orchestrator/projects/<id>/<key>`.

The standalone project model (#561) restored the spec behavior for standalone projects only:
`OrchestratorService.resolveIssueWorkspaceRoot` uses the project's `workspaceDir` when
`workflowSource.type === "external"` and keeps the consolidated layout otherwise. That was a
deliberate scope cut — converging repo-embedded touches live deployments and needs its own
migration story. This issue is that convergence.

## Scope

- Honor `workspace.root` for repo-embedded projects (drop the `workflowSource` condition in
  `resolveIssueWorkspaceRoot`), so the spec §9.1 path rule holds for every project.
- Reconcile the meaning of `workspaceDir`. `repo init` currently stores the repository checkout
  there (`packages/cli/src/repo-runtime.ts`), and that value is consumed as the daemon liveness
  `expectedCwd` fallback (`packages/cli/src/daemon-liveness.ts`) and by the doctor workspace-root
  check. A repo-embedded project needs both the checkout path and the workspace root, so one of
  them requires a new field.
- Decide where the workspace record lives. Standalone keeps `workspace.json` in the state
  directory and relocates only the workspace itself; repo-embedded should follow the same rule or
  move both, but not diverge silently.
- Migration. Existing deployments reuse issue workspaces across runs (§9.1), and each worktree has
  admin data in the shared bare cache. Relocating the root orphans both. Provide either an in-place
  move plus `git worktree repair`, or a documented one-time cleanup (stop → prune → re-populate).
- Revisit `RESERVED_WORKSPACE_KEYS` (`packages/core/src/workspace/identity.ts`), which exists only
  because workspaces currently share a directory with `runs/`, `cache/`, and the project JSON files.

## Acceptance

- A repo-embedded project with `workspace.root: X` populates issue workspaces at `X/<key>`, and the
  path-escape guard validates against `X`.
- Startup terminal-workspace cleanup (§8.6), dirty-workspace quarantine, and worktree removal all
  operate on the configured root.
- `doctor` reports the real workspace root for both modes.
- A migration path is documented and covered by a changeset; the divergence note in
  `docs/configuration.md` is removed.
- The decision is recorded in an ADR (extend or supersede
  `docs/adr/2026-08-13_standalone-project-instance-boundary.md`).

## Notes

Found during the #561 final acceptance test, together with the standalone-only fix that shipped
with the epic. Related design open question: "repo-embedded → standalone migration path" in
`docs/designs/2026-08-11-standalone-project-model-design.md`.

moncher-dev:
## Issues — Closed #599

- Fixes #599

## TL;DR

- Repo-embedded projects now honor `workspace.root` for every issue-workspace lifecycle path, matching Symphony spec §9.1.
- Repository checkout metadata (`repositoryDir`) is separate from the issue-workspace root (`workspaceDir`); omitted roots use `.runtime/symphony-workspaces`.
- Rework enforces private root permissions, rejects roots containing checkout or orchestrator state, preserves legacy shutdown access, and reports root relocation.

## Change-point diagram

```text
WORKFLOW.md workspace.root ──► repo init ──► project.workspaceDir
          (or dedicated default)    │                  │
                                    ├─ mkdir + chmod   ▼
repository checkout ─────────► repositoryDir    workspace lifecycle
                                                    ├─ populate / guard
                                                    ├─ quarantine / cleanup
                                                    └─ relocation event
```

## Start here

1. `packages/core/src/contracts/status-surface.ts` — checkout containment validation and dispatch-only legacy guard.
2. `packages/cli/src/repo-runtime.ts` — dedicated default, split checkout metadata, state containment, and `0700` enforcement.
3. `packages/orchestrator/src/fs-store.ts` — fails closed on legacy metadata at daemon config load.
4. `packages/orchestrator/src/service.ts` — configured-root lifecycle and relocation warning/event.
5. `docs/adr/2026-08-13_standalone-project-instance-boundary.md` — shared layout and migration decision.

## Changes

- Route repo-embedded population, startup cleanup, dirty quarantine, removal, and path containment through the configured workspace root.
- Keep `workspace.json` and control-plane state under `.runtime/orchestrator` while issue workspaces live beneath `workspace.root`.
- Default omitted repo-embedded roots to `.runtime/symphony-workspaces`; create or normalize the root to mode `0700` during `repo init`.
- Reject roots that equal or contain the repository checkout or orchestrator runtime state.
- Allow CLI stop/status/doctor paths to read legacy metadata while daemon startup rejects it.
- Emit stderr and structured `workspace-root-relocated` evidence before replacing a stale workspace record.
- Remove obsolete flat-state workspace-key reservations and cover behavior with unit tests and Docker TC-16.

## Evidence

- `pnpm lint` — passed on final pushed head `8647d767`.
- `pnpm test` — passed across all workspace packages on final pushed head `8647d767`.
- `pnpm exec vitest run packages/cli/src/commands/repo.test.ts` — passed (27 tests), including existing-root permissions and runtime-root containment.
- `GH_SYMPHONY_HTTP_TOKEN=e2e-http-token ./e2e/run-e2e.sh happy 45` — Docker TC-16 passed in 12s on cycle 3.
- `git diff --check` and `bash -n e2e/run-e2e.sh` — passed.
- GitHub `Test` and `Container Smoke` — passed on `8647d767`.

## Risks & rollback

- Existing repo-embedded installations must stop the daemon, archive legacy state, and rerun `gh-symphony repo init`; CLI shutdown remains usable after upgrade, while daemon startup rejects ambiguous legacy metadata.
- A later root change leaves the old tree for deliberate operator recovery or deletion and reports both paths; rollback requires stopping the daemon and restoring the previous binary/layout after preserving local work.

## Changed files

- Core metadata, identity, and observability: `packages/core/src/contracts/status-surface.ts`, `packages/core/src/workspace/identity.ts`, structured event files.
- CLI initialization and diagnostics: `packages/cli/src/repo-runtime.ts`, repo/status/stop/project/doctor commands and tests.
- Orchestrator lifecycle: `packages/orchestrator/src/service.ts`, filesystem store, and tests.
- Operator guidance and architecture: README files, configuration, architecture, ADR, migration changeset.
- Black-box coverage: `e2e/scenarios/16-repo-embedded-workspace-root.md`, E2E runner/seed, and `AGENT_TEST.md`.

- [ ] Confirm a repo-embedded project with custom and omitted `workspace.root` values creates issue workspaces beneath the reported dedicated root.
- [ ] Confirm `gh-symphony doctor` passes immediately after `repo init` and reports workspace root and repository checkout independently.
- [ ] Confirm the stop/archive/re-init migration and relocation warning are operationally clear on an existing daemon.

### DIFF
--- a/packages/cli/src/repo-runtime.ts
+++ b/packages/cli/src/repo-runtime.ts
@@ -28,6 +28,7 @@ export type RepoInitFlags = {
 };
 
 const INTERNAL_PROJECT_ID = "repository";
+const DEFAULT_WORKSPACE_ROOT = join(".runtime", "symphony-workspaces");
 
 export class RepoRuntimeMigrationError extends Error {}
 
@@ -127,7 +128,7 @@ export async function initRepoRuntime(flags: RepoInitFlags): Promise<{
     displayName: `${repository.owner}/${repository.name}`,
     workspaceDir: workflow.workspace.root
       ? resolve(repoDir, workflow.workspace.root)
-      : runtimeRoot,
+      : resolve(repoDir, DEFAULT_WORKSPACE_ROOT),
     repositoryDir: repoDir,
     repository,
     tracker: {
@@ -143,6 +144,7 @@ export async function initRepoRuntime(flags: RepoInitFlags): Promise<{
 
   await mkdir(runtimeRoot, { recursive: true });
   await saveProjectConfig(runtimeRoot, INTERNAL_PROJECT_ID, projectConfig);
+  await mkdir(projectConfig.workspaceDir, { recursive: true, mode: 0o700 });
   await saveGlobalConfig(runtimeRoot, {
     activeProject: INTERNAL_PROJECT_ID,
     projects: [INTERNAL_PROJECT_ID],
--- a/packages/cli/src/repo-runtime.ts
+++ b/packages/cli/src/repo-runtime.ts
@@ -1,5 +1,6 @@
 import { execFileSync } from "node:child_process";
 import {
+  chmod,
   mkdir,
   readdir,
   readFile,
@@ -8,7 +9,14 @@ import {
   stat,
   writeFile,
 } from "node:fs/promises";
-import { basename, dirname, join, resolve } from "node:path";
+import {
+  basename,
+  dirname,
+  isAbsolute,
+  join,
+  relative,
+  resolve,
+} from "node:path";
 import {
   parseWorkflowMarkdown,
   type OrchestratorProjectConfig,
@@ -122,13 +130,31 @@ export async function initRepoRuntime(flags: RepoInitFlags): Promise<{
     trackerSettings.issuesPath =
       process.env.GH_SYMPHONY_FILE_TRACKER_ISSUES_PATH;
   }
+  const workspaceDir = workflow.workspace.root
+    ? resolve(repoDir, workflow.workspace.root)
+    : resolve(repoDir, DEFAULT_WORKSPACE_ROOT);
+  const runtimeRelativeToWorkspace = relative(workspaceDir, runtimeRoot);
+  const repositoryRelativeToWorkspace = relative(workspaceDir, repoDir);
+  const workspaceContainsRepository =
+    repositoryRelativeToWorkspace === "" ||
+    (!repositoryRelativeToWorkspace.startsWith("..") &&
+      !isAbsolute(repositoryRelativeToWorkspace));
+  if (
+    !workspaceContainsRepository &&
+    (runtimeRelativeToWorkspace === "" ||
+      (!runtimeRelativeToWorkspace.startsWith("..") &&
+        !isAbsolute(runtimeRelativeToWorkspace)))
+  ) {
+    throw new Error(
+      `workspace.root ${JSON.stringify(workspaceDir)} must not equal or contain the orchestrator runtime directory ${JSON.stringify(runtimeRoot)}.`
+    );
+  }
+
   const projectConfig: CliProjectConfig = {
     projectId: INTERNAL_PROJECT_ID,
     slug: basename(repoDir) || INTERNAL_PROJECT_ID,
     displayName: `${repository.owner}/${repository.name}`,
-    workspaceDir: workflow.workspace.root
-      ? resolve(repoDir, workflow.workspace.root)
-      : resolve(repoDir, DEFAULT_WORKSPACE_ROOT),
+    workspaceDir,
     repositoryDir: repoDir,
     repository,
     tracker: {
@@ -145,6 +171,7 @@ export async function initRepoRuntime(flags: RepoInitFlags): Promise<{
   await mkdir(runtimeRoot, { recursive: true });
   await saveProjectConfig(runtimeRoot, INTERNAL_PROJECT_ID, projectConfig);
   await mkdir(projectConfig.workspaceDir, { recursive: true, mode: 0o700 });
+  await chmod(projectConfig.workspaceDir, 0o700);
   await saveGlobalConfig(runtimeRoot, {
     activeProject: INTERNAL_PROJECT_ID,
     projects: [INTERNAL_PROJECT_ID],
--- a/packages/cli/src/repo-runtime.ts
+++ b/packages/cli/src/repo-runtime.ts
@@ -125,7 +125,10 @@ export async function initRepoRuntime(flags: RepoInitFlags): Promise<{
     projectId: INTERNAL_PROJECT_ID,
     slug: basename(repoDir) || INTERNAL_PROJECT_ID,
     displayName: `${repository.owner}/${repository.name}`,
-    workspaceDir: repoDir,
+    workspaceDir: workflow.workspace.root
+      ? resolve(repoDir, workflow.workspace.root)
+      : runtimeRoot,
+    repositoryDir: repoDir,
     repository,
     tracker: {
       adapter: trackerAdapter,
@@ -148,7 +151,8 @@ export async function initRepoRuntime(flags: RepoInitFlags): Promise<{
   const orchestratorConfig: OrchestratorProjectConfig = {
     projectId: INTERNAL_PROJECT_ID,
     slug: projectConfig.slug,
-    workspaceDir: repoDir,
+    workspaceDir: projectConfig.workspaceDir,
+    repositoryDir: repoDir,
     repository,
     tracker: projectConfig.tracker,
   };
--- a/packages/core/src/contracts/status-surface.ts
+++ b/packages/core/src/contracts/status-surface.ts
@@ -34,7 +34,10 @@ export type PopulateStrategy = "clone" | "worktree-cache";
 export type OrchestratorProjectConfig = {
   projectId: string;
   slug: string;
+  /** Root directory containing persistent per-issue workspaces. */
   workspaceDir: string;
+  /** Repository checkout used as the daemon cwd for repo-embedded projects. */
+  repositoryDir?: string;
   repository: RepositoryRef;
   tracker: OrchestratorTrackerConfig;
   /** Defaults to the repository-local workflow for legacy project configs. */
--- a/packages/core/src/contracts/status-surface.ts
+++ b/packages/core/src/contracts/status-surface.ts
@@ -1,4 +1,4 @@
-import { isAbsolute } from "node:path";
+import { isAbsolute, resolve } from "node:path";
 import type { RepositoryRef } from "../domain/workspace.js";
 import type {
   WorkflowDefinition,
@@ -58,6 +58,18 @@ export function normalizeOrchestratorProjectConfig(
   assertAbsoluteProjectDir(config);
   const workflowSource = normalizeWorkflowSource(config);
   const populateStrategy = normalizePopulateStrategy(config);
+  const repositoryPath = config.repository?.path;
+
+  if (
+    workflowSource.type === "repo" &&
+    !config.repositoryDir &&
+    repositoryPath &&
+    resolve(config.workspaceDir) === resolve(repositoryPath)
+  ) {
+    throw new Error(
+      `Project ${JSON.stringify(config.projectId)} uses legacy repo-embedded path metadata. Stop the daemon and run 'gh-symphony repo init' again before starting it.`
+    );
+  }
 
   return {
     ...config,
--- a/packages/core/src/contracts/status-surface.ts
+++ b/packages/core/src/contracts/status-surface.ts
@@ -1,4 +1,4 @@
-import { isAbsolute, resolve } from "node:path";
+import { isAbsolute, relative, resolve } from "node:path";
 import type { RepositoryRef } from "../domain/workspace.js";
 import type {
   WorkflowDefinition,
@@ -58,8 +58,34 @@ export function normalizeOrchestratorProjectConfig(
   assertAbsoluteProjectDir(config);
   const workflowSource = normalizeWorkflowSource(config);
   const populateStrategy = normalizePopulateStrategy(config);
-  const repositoryPath = config.repository?.path;
+  if (workflowSource.type === "repo" && config.repositoryDir) {
+    const workspaceRoot = resolve(config.workspaceDir);
+    const repositoryDir = resolve(config.repositoryDir);
+    const repositoryRelativeToRoot = relative(workspaceRoot, repositoryDir);
+    if (
+      repositoryRelativeToRoot === "" ||
+      (!repositoryRelativeToRoot.startsWith("..") &&
+        !isAbsolute(repositoryRelativeToRoot))
+    ) {
+      throw new Error(
+        `Project ${JSON.stringify(config.projectId)} workspace.root ${JSON.stringify(workspaceRoot)} must not equal or contain the repository checkout ${JSON.stringify(repositoryDir)}.`
+      );
+    }
+  }
 
+  return {
+    ...config,
+    workflowSource,
+    populateStrategy,
+  };
+}
+
+/** Rejects legacy repo metadata on daemon startup without blocking CLI migration commands. */
+export function assertDispatchableOrchestratorProjectConfig(
+  config: OrchestratorProjectConfig
+): void {
+  const workflowSource = normalizeWorkflowSource(config);
+  const repositoryPath = config.repository?.path;
   if (
     workflowSource.type === "repo" &&
     !config.repositoryDir &&
@@ -70,12 +96,6 @@ export function normalizeOrchestratorProjectConfig(
       `Project ${JSON.stringify(config.projectId)} uses legacy repo-embedded path metadata. Stop the daemon and run 'gh-symphony repo init' again before starting it.`
     );
   }
-
-  return {
-    ...config,
-    workflowSource,
-    populateStrategy,
-  };
 }
 
 function assertAbsoluteProjectDir(config: OrchestratorProjectConfig): void {
--- a/packages/orchestrator/src/fs-store.ts
+++ b/packages/orchestrator/src/fs-store.ts
@@ -3,6 +3,7 @@ import { chmod, mkdir, open, rm, stat } from "node:fs/promises";
 import { join, relative, resolve } from "node:path";
 import {
   deriveIssueWorkspaceKeyFromIdentifier,
+  assertDispatchableOrchestratorProjectConfig,
   isFileMissing,
   type IssueOrchestrationRecord,
   type IssueWorkspaceRecord,
@@ -82,7 +83,12 @@ export class OrchestratorFsStore implements OrchestratorStateStore {
           )
         : null);
 
-    return config ? normalizeOrchestratorProjectConfig(config) : null;
+    if (!config) {
+      return null;
+    }
+    const normalized = normalizeOrchestratorProjectConfig(config);
+    assertDispatchableOrchestratorProjectConfig(normalized);
+    return normalized;
   }
 
   async saveProjectConfig(config: OrchestratorProjectConfig): Promise<void> {
--- a/packages/orchestrator/src/service.ts
+++ b/packages/orchestrator/src/service.ts
@@ -2637,6 +2637,10 @@ export class OrchestratorService {
       this.resolveIssueWorkspaceRoot(tenant),
       workspaceKey
     );
+    const existingWorkspaceAtConfiguredRoot = Boolean(
+      existingWorkspaceRecord &&
+      resolve(existingWorkspaceRecord.workspacePath) === issueWorkspacePath
+    );
     const pullRequestBranch = resolvePullRequestBranchCheckoutTarget(issue);
 
     // #507: dirty recovery may only reuse the workspace when the dirty state
@@ -2646,7 +2650,7 @@ export class OrchestratorService {
     let workspaceQuarantined = false;
     if (
       recovery?.kind === "incomplete-turn-dirty-workspace" &&
-      existingWorkspaceRecord
+      existingWorkspaceAtConfiguredRoot
     ) {
       const currentBranch = await readGitCurrentBranch(
         join(issueWorkspacePath, "repository")
@@ -2713,7 +2717,7 @@ export class OrchestratorService {
       repository: issue.repository,
       issueWorkspacePath,
       existingWorkspace:
-        Boolean(existingWorkspaceRecord) && !workspaceQuarantined,
+        existingWorkspaceAtConfiguredRoot && !workspaceQuarantined,
       pullRequestBranch,
       allowDirtyExistingWorkspace:
         recovery?.kind === "incomplete-turn-dirty-workspace",
@@ -2733,7 +2737,7 @@ export class OrchestratorService {
     );
     await excludeRuntimeSkillsFromGit(repositoryDirectory, agentCommand);
 
-    if (!existingWorkspaceRecord || workspaceQuarantined) {
+    if (!existingWorkspaceAtConfiguredRoot || workspaceQuarantined) {
       const workspaceRecord: IssueWorkspaceRecord = {
         workspaceKey,
         projectId: tenant.projectId,
--- a/packages/orchestrator/src/service.ts
+++ b/packages/orchestrator/src/service.ts
@@ -4729,12 +4729,13 @@ export class OrchestratorService {
 
   /**
    * Per-issue workspaces live under the project's configured `workspace.root`
-   * (spec 9.1). Standalone projects carry that root as `workspaceDir`; the
-   * repo-embedded layout still keeps workspaces beside the project state,
-   * where `workspaceDir` means the repository checkout instead.
+   * (spec 9.1). New project configs carry that root as `workspaceDir` in both
+   * modes. A repo-embedded config without `repositoryDir` predates that split,
+   * so it retains the consolidated state-directory layout until `repo init`
+   * migrates its metadata.
    */
   private resolveIssueWorkspaceRoot(tenant: OrchestratorProjectConfig): string {
-    return tenant.workflowSource?.type === "external" && tenant.workspaceDir
+    return tenant.repositoryDir || tenant.workflowSource?.type === "external"
       ? resolve(tenant.workspaceDir)
       : this.store.projectDir(tenant.projectId);
   }
--- a/packages/orchestrator/src/service.ts
+++ b/packages/orchestrator/src/service.ts
@@ -4733,15 +4733,12 @@ export class OrchestratorService {
 
   /**
    * Per-issue workspaces live under the project's configured `workspace.root`
-   * (spec 9.1). New project configs carry that root as `workspaceDir` in both
-   * modes. A repo-embedded config without `repositoryDir` predates that split,
-   * so it retains the consolidated state-directory layout until `repo init`
-   * migrates its metadata.
+   * (spec 9.1), carried as `workspaceDir` in every project mode. Persisted
+   * repo-embedded configs that still use this field for the checkout are
+   * rejected during normalization until `repo init` migrates them.
    */
   private resolveIssueWorkspaceRoot(tenant: OrchestratorProjectConfig): string {
-    return tenant.repositoryDir || tenant.workflowSource?.type === "external"
-      ? resolve(tenant.workspaceDir)
-      : this.store.projectDir(tenant.projectId);
+    return resolve(tenant.workspaceDir);
   }
 
   private resolveWorkflowRepositoryDirectory(
--- a/packages/orchestrator/src/service.ts
+++ b/packages/orchestrator/src/service.ts
@@ -2641,6 +2641,21 @@ export class OrchestratorService {
       existingWorkspaceRecord &&
       resolve(existingWorkspaceRecord.workspacePath) === issueWorkspacePath
     );
+    if (existingWorkspaceRecord && !existingWorkspaceAtConfiguredRoot) {
+      this.writeStderr(
+        `[orchestrator] workspace root changed for ${issue.identifier}: previous=${existingWorkspaceRecord.workspacePath} configured=${issueWorkspacePath}`
+      );
+      await this.store.appendRunEvent(runId, {
+        at: now.toISOString(),
+        event: "workspace-root-relocated",
+        projectId: tenant.projectId,
+        issueId: issue.id,
+        issueIdentifier: issue.identifier,
+        workspaceKey,
+        previousWorkspacePath: existingWorkspaceRecord.workspacePath,
+        configuredWorkspacePath: issueWorkspacePath,
+      });
+    }
     const pullRequestBranch = resolvePullRequestBranchCheckoutTarget(issue);
 
     // #507: dirty recovery may only reuse the workspace when the dirty state
--- a/.changeset/calm-workspace-roots.md
+++ b/.changeset/calm-workspace-roots.md
@@ -1,0 +1,5 @@
+---
+"@gh-symphony/cli": minor
+---
+
+Honor repo-embedded `workspace.root` for issue workspaces, keep repository checkout metadata separate, and fail safely with migration guidance for legacy project records (#599).
--- a/.changeset/calm-workspace-roots.md
+++ b/.changeset/calm-workspace-roots.md
@@ -2,4 +2,4 @@
 "@gh-symphony/cli": minor
 ---
 
-Honor repo-embedded `workspace.root` for issue workspaces, keep repository checkout metadata separate, and fail safely with migration guidance for legacy project records (#599).
+Honor repo-embedded `workspace.root` for issue workspaces, default to a dedicated root, keep repository checkout metadata separate, and preserve shutdown access with safe migration guidance for legacy project records (#599).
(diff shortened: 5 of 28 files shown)


packages/cli/src/repo-runtime.ts
```diff
@@ -125,7 +125,10 @@ export async function initRepoRuntime(flags: RepoInitFlags): Promise<{
     projectId: INTERNAL_PROJECT_ID,
     slug: basename(repoDir) || INTERNAL_PROJECT_ID,
     displayName: `${repository.owner}/${repository.name}`,
-    workspaceDir: repoDir,
+    workspaceDir: workflow.workspace.root
+      ? resolve(repoDir, workflow.workspace.root)
+      : runtimeRoot,
```
hojinzs: **P1 — `repo init` never creates the resolved workspace root, so `doctor` hard-fails right after a successful init**

`initRepoRuntime` persists `workspaceDir` but only `mkdir`s `runtimeRoot`. The workspace root is created lazily at first dispatch (`service.ts:2710`), while `doctor`'s `workspace_root` check is `required: true` and fails on a missing directory (`doctor.ts:2029-2040`). The documented preflight is `repo init` → `doctor --smoke` → `repo start`, so the feature's own happy path now reports a hard failure.

Reproduced on `ac6e978` with the built CLI, using the TC-16 workflow (`workspace.root: .runtime/symphony-workspaces`):

```
$ gh-symphony repo init
Repository initialized: acme/widget

$ gh-symphony doctor --json | jq '.checks[] | select(.id=="workspace_root")'
{
  "id": "workspace_root",
  "title": "Workspace root",
  "status": "fail",
  "required": true,
  "summary": "Workspace root does not exist: .../.runtime/symphony-workspaces.",
  "remediation": "Create the directory before re-running doctor with: mkdir -p -- '.../.runtime/symphony-workspaces'.",
  "details": { "reason": "missing" }
}
```

Before this PR the check always passed because `workspaceDir` was the checkout. TC-16 does not catch it because the E2E path runs `repo start` (which mkdirs on dispatch) rather than `doctor`.

Suggested fix: create the resolved root in `initRepoRuntime` with mode `0700`, the same way the orchestrator does — that also makes the `0700` guarantee documented in `docs/configuration.md` hold from init time. Please add a `repo.test.ts` case asserting the directory exists after `repo init`, and a `doctor.test.ts` case for a configured-but-not-yet-populated root.

---

packages/cli/src/repo-runtime.ts
```diff
@@ -125,7 +125,10 @@ export async function initRepoRuntime(flags: RepoInitFlags): Promise<{
...
+    repositoryDir: repoDir,
```
hojinzs, line 158: **P2 — a `workspace.root` that resolves to the checkout is accepted, which is exactly the hazard the new startup guard exists to prevent**

The normalization guard in `status-surface.ts` refuses legacy metadata where `workspaceDir === repository.path` "to prevent population inside the checkout". But `repo init` writes `repositoryDir` unconditionally, so the guard never fires for a *new* config — and nothing validates the resolved root itself.

Reproduced on `ac6e978` with `workspace.root: "."`:

```
$ gh-symphony repo init
Repository initialized: acme/widget

$ jq '{workspaceDir, repositoryDir}' .runtime/orchestrator/project.json
{
  "workspaceDir":  ".../smoke/dotroot",
  "repositoryDir": ".../smoke/dotroot"
}
```

`repo init` succeeds silently and the orchestrator will populate `<checkout>/<issue-key>/repository` inside the working tree — untracked clones in the repo, and terminal cleanup `rm -rf`-ing paths under it (`service.ts:4897`). `workspace.root: ..` or any path containing the checkout has the same effect.

Suggested fix: reject at `repo init` (and in normalization) when the resolved workspace root equals `repositoryDir` or contains it, with a message pointing at `workspace.root`. A unit test for `root: "."` and `root: ".."` would pin it.

---

packages/core/src/contracts/status-surface.ts
```diff
@@ -54,6 +57,18 @@ export function normalizeOrchestratorProjectConfig(
...
   const populateStrategy = normalizePopulateStrategy(config);
+  const repositoryPath = config.repository?.path;
+
+  if (
+    workflowSource.type === "repo" &&
+    !config.repositoryDir &&
+    repositoryPath &&
+    resolve(config.workspaceDir) === resolve(repositoryPath)
+  ) {
+    throw new Error(
```
hojinzs, line 95: **P1 — this throw breaks step 1 of the documented migration: `repo stop` is unusable on exactly the installations that must migrate**

`normalizeOrchestratorProjectConfig` is not only on the daemon startup path — `packages/cli/src/config.ts:190` calls it inside `loadProjectConfig`, which every `resolveManagedProjectConfig` caller goes through (`stop.ts:51`, `status.ts`, `project.ts`, `doctor`'s managed-project selection). So on a legacy install the throw takes out the read-only and shutdown commands too.

Verified on `ac6e978` by loading a legacy-shaped `project.json` (`workspaceDir === repository.path`, no `repositoryDir`) through `loadProjectConfig`:

```
THREW: Project "repository" uses legacy repo-embedded path metadata.
       Stop the daemon and run 'gh-symphony repo init' again before starting it.
```

`docs/configuration.md` tells the operator to run `gh-symphony repo stop` first, but after upgrading the CLI that command exits 1 with this same message — the only way left to stop the daemon is killing the PID by hand. The message ("Stop the daemon and …") is also the one instruction the user can no longer follow.

Suggested fix: keep the hard failure on the dispatch/startup path only (orchestrator config load), and let the CLI's read/stop paths load the legacy record with a warning — or catch it in `loadProjectConfig` for `stop`/`status` and degrade to `repositoryDir ?? workspaceDir`. Whichever way, `repo stop` must keep working, and the migration doc should state whether stopping before or after the upgrade is required.

---

packages/orchestrator/src/service.ts
```diff
@@ -2635,6 +2635,10 @@ export class OrchestratorService {
       this.resolveIssueWorkspaceRoot(tenant),
       workspaceKey
     );
+    const existingWorkspaceAtConfiguredRoot = Boolean(
```
hojinzs, line 2640: **P3 — the relocation is silent and orphans the previous workspace tree**

This guard is the right call: a record pointing at a stale path no longer counts as a reusable workspace. But when it is `false`, the new record is written under the same `workspaceKey` (line 2739), so the only pointer to the old directory is overwritten. The old tree — including any uncommitted work — is then unreachable by terminal cleanup (`removeIssueWorkspace` removes `workspaceRecord.workspacePath`, line 4897) and is never reported.

The documented one-time migration covers this by archiving `.runtime/orchestrator` wholesale, but a later `workspace.root` change (edit `WORKFLOW.md` → `repo init` → restart) hits the same path with no warning and no cleanup guidance.

Not a blocker. Minimally, emit a structured event / stderr line when the recorded `workspacePath` differs from the configured one, naming both paths so the operator can reclaim or delete the old tree. (Populate strategy is `clone` for repo-embedded, so no stale `git worktree` registration is involved — worth confirming this stays true if `worktree-cache` is ever enabled for this mode, since `worktree prune` will not clear a registration whose directory still exists.)

---

hojinzs (changes requested):
## Review — round 1/6

Scope of the change matches #599, the layering is right (workspace root in `core` normalization, checkout in the CLI config, orchestrator resolving from one field), and the ADR/docs/changeset obligations are all met. What blocks it is operator-facing: the two commands an operator runs immediately around this feature — `doctor` after `repo init`, and `repo stop` before migrating — both break.

### Acceptance criteria (#599)

| Criterion | Status |
| --- | --- |
| Repo-embedded project with `workspace.root: X` populates at `X/<key>`, escape guard validates against `X` | ✅ `resolveIssueWorkspaceRoot` + TC-16 |
| Startup terminal cleanup, dirty quarantine, worktree removal operate on the configured root | ✅ all paths derive from `resolveIssueWorkspaceRoot` / the record's `workspacePath` |
| `doctor` reports the real workspace root for both modes | ⚠️ reports it, but the check now **fails** on a fresh correct install (P1) |
| Migration documented + changeset; divergence note removed from `docs/configuration.md` | ⚠️ documented, but its first step (`repo stop`) does not work post-upgrade (P1) |
| Decision recorded in the ADR | ✅ |
| `RESERVED_WORKSPACE_KEYS` revisited | ⚠️ removed — sound *if* workspaces never share the state directory, which the no-`workspace.root` default violates (see Codex's open thread) |

No over-engineering or unrequested scope found; `repositoryDir` is the minimum new field the issue asked for, and the CLI call sites (`status`, `stop`, `project`) were each updated with the `repositoryDir ?? workspaceDir` fallback that keeps legacy records working.

### Verification performed

- `pnpm lint`, `pnpm test` (all packages), `pnpm typecheck`, `pnpm build` — all pass locally on `ac6e978`. CI `Test` and `Container Smoke` green.
- Smoke test with the built CLI in throwaway git repos: `repo init` with `workspace.root` set, omitted, and `"."`; `doctor --json` in each. Findings below come from that run. Docker E2E/TC-16 was not re-run locally.

### Findings

| # | Sev | Where | Summary |
| --- | --- | --- | --- |
| 1 | P1 | `repo-runtime.ts:130` | `repo init` never creates the resolved root → `doctor`'s required `workspace_root` check fails right after init |
| 2 | P1 | `status-surface.ts:68` | The legacy-metadata throw runs inside `loadProjectConfig`, so `repo stop` / `status` / `doctor` all exit 1 on legacy installs — the documented migration's step 1 is impossible |
| 3 | P2 | `repo-runtime.ts:131` | `workspace.root: "."` is accepted → population inside the repository checkout, the exact hazard the new guard is meant to prevent |
| 4 | P2 | Codex thread on `repo-runtime.ts:130` | With `workspace.root` omitted, the root defaults to `.runtime/orchestrator` (see below) |
| 5 | P3 | `service.ts:2638` | Root relocation silently orphans the previous workspace tree |

On #4 I agree with the automated review and have added confirming evidence on that thread — please treat it as a real finding rather than bot noise. It is also the reason removing `RESERVED_WORKSPACE_KEYS` needs a second look: the justification ("runtime-state filenames no longer reserve ordinary issue workspace keys") holds only once workspaces stop living in the state directory.

nits (no action required):
- `packages/cli/src/commands/workflow.ts:747` still builds an `OrchestratorProjectConfig` without `repositoryDir`. Harmless today — it is only used for a tracker preview — but it is now an incomplete projection of the config.
- `packages/cli/src/commands/doctor.ts:1752` reports `workspaceDir` in the `managed_project` details, which now means the workspace root; surfacing `repositoryDir` alongside it would save an operator one lookup.

Tests added are well-targeted (record migration under a changed root, the no-`repositoryDir` path, key-reservation removal). The gap is that both TC-16 and `repo.test.ts` exercise only the *configured*-root case — the omitted-root default and the `doctor`-after-`init` sequence are untested, which is why findings 1 and 4 got through.

Requesting changes on 1–3; 4 is Codex's thread; 5 is optional.

---

packages/cli/src/repo-runtime.ts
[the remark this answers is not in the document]
hojinzs: Confirming this finding from a human-requested review — please don't dismiss it as bot noise. Reproduced on `ac6e978` with a `WORKFLOW.md` that omits `workspace.root`:

```
$ gh-symphony repo init
$ jq '{workspaceDir, repositoryDir}' .runtime/orchestrator/project.json
{
  "workspaceDir":  ".../norootrepo/.runtime/orchestrator",
  "repositoryDir": ".../norootrepo"
}

$ gh-symphony doctor --json | jq -r '.checks[] | select(.id=="workspace_root") | .summary'
Workspace root is writable: .../norootrepo/.runtime/orchestrator.
```

Three things make this worth fixing rather than accepting:

1. **It contradicts the sibling mode and this PR's own docs.** `project.ts:66` defaults standalone to `<project-dir>/.runtime/workspaces`, and `docs/configuration.md` (as edited here) says repo-embedded "apply the same rule" with "orchestrator state remains under `.runtime/orchestrator`, while populated issue workspaces live at `<workspace.root>/…`". With the root defaulting to `runtimeRoot` those two directories are the same directory.
2. **It compounds the `RESERVED_WORKSPACE_KEYS` removal.** That removal is justified in the ADR by workspaces no longer sharing a directory with runtime state — true only when a root is configured. In the default case a checkout lands beside `project.json`, `config.json`, `projects/`, `runs/`, and the shared bare cache at `repos/`, and nothing rejects a workspace key that collides with one of them any more.
3. **It is a forced relocation for installs that never used `workspace.root`.** Those previously populated at `projectDir(projectId)`; they now move to `runtimeRoot` and inherit the whole stop/archive/re-init migration for no configuration change of their own.

`.runtime/symphony-workspaces` (what `e2e/seed/WORKFLOW.md` uses) or `.runtime/workspaces` (matching standalone) would fix all three. Note that neither TC-16 nor `repo.test.ts` covers the omitted-root case — both set `workspace.root` explicitly — so please add coverage for the default alongside the fix.

---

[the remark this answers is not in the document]
moncher-dev: Resolved in `ccc1052`. Repo-embedded init now defaults an omitted `workspace.root` to `.runtime/symphony-workspaces`, keeping populated issue workspaces outside `.runtime/orchestrator`. Coverage asserts both the dedicated default and the state-directory separation.

moncher-dev: Resolved in `ccc1052`. `repo init` now creates the resolved workspace root with mode `0700` before reporting success. The repo-init regression test verifies the directory and permissions; the existing doctor configured-root test confirms an empty initialized root passes.

packages/cli/src/repo-runtime.ts
moncher-dev: Resolved in `ccc1052`. Config normalization now rejects a workspace root that equals or contains `repositoryDir`, with an error naming `workspace.root` and the checkout. Parameterized repo-init coverage pins both `root: "."` and `root: ".."`.

packages/core/src/contracts/status-surface.ts
moncher-dev: Resolved in `ccc1052`. General CLI config normalization can now load legacy repo metadata, preserving `repo stop`, status, and diagnostics for the documented migration. The hard legacy-layout rejection moved to `OrchestratorFsStore.loadProjectConfig`, so daemon startup/dispatch still fails closed. CLI and orchestrator-store tests cover both sides, and the migration docs explicitly state that stop remains available after upgrade.

packages/orchestrator/src/service.ts
moncher-dev: Resolved in `ccc1052`. Before replacing a record whose persisted workspace path differs from the configured root, dispatch now emits a stderr warning and a structured `workspace-root-relocated` event containing both paths. The migration test asserts the warning, event, and updated record; the docs explain how operators can recover or delete the previous tree.

packages/cli/src/repo-runtime.ts
```diff
@@ -140,6 +144,7 @@ export async function initRepoRuntime(flags: RepoInitFlags): Promise<{
 
   await mkdir(runtimeRoot, { recursive: true });
   await saveProjectConfig(runtimeRoot, INTERNAL_PROJECT_ID, projectConfig);
+  await mkdir(projectConfig.workspaceDir, { recursive: true, mode: 0o700 });
```
hojinzs, line 173: **P3 — the `0700` guarantee does not hold for a pre-existing root**

This fixes finding 1 (verified: after `repo init` the root exists and `doctor` passes). One gap remains: `mkdir` applies `mode` only to directories it *creates*. If the operator already created the root — or it survived an archive/re-init cycle, which is exactly the documented migration — the mode is left alone.

Reproduced on `ccc1052` with the built CLI:

```
$ mkdir -p .runtime/symphony-workspaces && chmod 0755 .runtime/symphony-workspaces
$ gh-symphony repo init
Repository initialized: acme/widget
$ stat -c '%a' .runtime/symphony-workspaces
755
```

`docs/configuration.md` as edited here now states `repo init` "creates the resolved root with mode `0700`", so the doc asserts a property the code does not enforce. `OrchestratorFsStore.ensureProjectDirectory` (`fs-store.ts:61-62`) already handles this correctly with `mkdir` followed by an explicit `chmod` — mirroring those two lines here would close it, and a `repo.test.ts` case that pre-creates the root at `0755` would pin it.

Minor and non-blocking.

---

packages/core/src/contracts/status-surface.ts
```diff
@@ -54,6 +57,20 @@ export function normalizeOrchestratorProjectConfig(
...
+    if (
+      repositoryRelativeToRoot === "" ||
+      (!repositoryRelativeToRoot.startsWith("..") &&
+        !isAbsolute(repositoryRelativeToRoot))
+    ) {
+      throw new Error(
+        `Project ${JSON.stringify(config.projectId)} workspace.root ${JSON.stringify(workspaceRoot)} must not equal or contain the repository checkout ${JSON.stringify(repositoryDir)}.`
+      );
+    }
+  }
```
hojinzs, line 74: **P3 — the containment guard covers the checkout but not the orchestrator state directory**

The guard itself is correct and I verified both rejections against the built CLI on `ccc1052` (`root: "."` and `root: ".."` both fail init with a message naming `workspace.root` and the checkout). This is a note about its edges, not a defect in what it does.

The guard asks "does the workspace root contain `repositoryDir`?" but nothing asks "does it contain `.runtime/orchestrator`?". So the configuration that Codex's thread was about is still reachable — just explicitly rather than by default:

```
$ sed -i 's|^  root: .*|  root: .runtime|' WORKFLOW.md
$ gh-symphony repo init
Repository initialized: acme/widget
$ jq -r .workspaceDir .runtime/orchestrator/projects/repository/project.json
/…/runtimeroot/.runtime
```

Issue workspaces then populate at `.runtime/<key>`, alongside `orchestrator/`, `repos/`, and the state JSON — the arrangement whose absence justifies removing `RESERVED_WORKSPACE_KEYS` in the ADR. With that set gone, the only remaining key guard is `workspaceKey.startsWith(".")` (`identity.ts:74`), so a key sanitizing to `runs` or `cache` no longer collides with anything by name.

Two reasons I am flagging it rather than requesting a change: it needs a deliberate hand-written `workspace.root`, and `populateStrategy` for repo-embedded is `clone`, so a collision surfaces as a failed populate rather than corrupted state. The symmetric fix is cheap if you want it — reject a root that equals or contains the runtime root in the same block, since both hazards have the same shape.

---

hojinzs (approved):

All five round-1 findings are fixed, and I verified each one independently on `ccc1052` rather than taking the thread replies at face value. Everything still open is a P3 nit. **Approving** — with one mechanical blocker to clear before merge (below), which is not a code objection.

### ⚠️ Before merging: the conflict resolution can silently revert #617

The PR is `mergeable_state: dirty` — `packages/core/src/contracts/status-surface.ts` conflicts with `main`.

This one is worth care. `main` gained `b2d1f9c fix(config): validate projectDir and hash env cache (#617)` *after* this branch's merge-base (`c5e8195`), and that commit added `assertAbsoluteProjectDir(config)` as the first statement of `normalizeOrchestratorProjectConfig` — the exact function this PR restructures. Taking "ours" on the conflict drops that validation without any visible deletion in the diff.

Keep both: `assertAbsoluteProjectDir(config)` first, then the new containment guard. `packages/cli/src/config.test.ts:122` (`projectDir: "relative/project"`) auto-merges cleanly and will fail if the resolution loses it, so CI is a backstop — but it's easier to get right the first time.

Also note GitHub reports no check runs on `ccc1052` (`total_count: 0`, status pending); worth confirming CI actually ran after the merge.

| # | Sev | Claimed | Verified |
| --- | --- | --- | --- |
| 1 | P1 | root created at init | ✅ Root exists at `0700` after `repo init`; `doctor`'s `workspace_root` check now returns **pass** for both configured and omitted roots (was `fail`) |
| 2 | P1 | `repo stop` usable on legacy | ✅ `repo stop`, `repo status`, `doctor` all exit `0` against a legacy `project.json` (`workspaceDir === repository.path`, no `repositoryDir`) — previously exit 1. Fail-closed half also holds: `createStore(rt).loadProjectConfig("repository")` still throws the legacy-metadata error, and `packages/orchestrator/src/index.ts:250` confirms daemon startup goes through that method |
| 3 | P2 | root/checkout containment rejected | ✅ `root: "."` and `root: ".."` both fail `repo init` with a message naming `workspace.root` and the checkout |
| 4 | P2 | dedicated default root | ✅ Omitted `workspace.root` → `.runtime/symphony-workspaces`, distinct from `.runtime/orchestrator` |
| 5 | P3 | relocation is visible | ✅ stderr warning + `workspace-root-relocated` event; type added to the `OrchestratorEvent` union and given a formatter case, so it renders rather than falling through |

Splitting the legacy check into `assertDispatchableOrchestratorProjectConfig` and calling it only from `OrchestratorFsStore.loadProjectConfig` is the right resolution for #2 — the CLI's `loadProjectConfig` (`config.ts:190`) and the store's are genuinely different functions, and the split lands the guard on exactly the dispatch path without touching the migration commands.

### Acceptance criteria (#599)

| Criterion | Status |
| --- | --- |
| `workspace.root: X` populates at `X/<key>`, escape guard validates against `X` | ✅ |
| Startup cleanup, dirty quarantine, worktree removal use the configured root | ✅ |
| `doctor` reports the real workspace root for both modes | ✅ now passes on a fresh install |
| Migration documented + changeset; divergence note removed | ✅ migration doc now states stop remains available post-upgrade |
| Decision recorded in the ADR | ✅ extended with the default, the guard, and the split legacy handling |
| `RESERVED_WORKSPACE_KEYS` revisited | ✅ justification holds now that the default root is outside the state directory (one residual edge in an inline comment) |

No over-engineering or unrequested scope. The delta is 178 lines and every hunk traces to a specific finding.

### Verification performed

- `pnpm lint`, `pnpm typecheck`, `pnpm build`, `pnpm test` — all exit 0 locally on `ccc1052`.
- Smoke test with the built CLI (`packages/cli/dist/index.js`) in throwaway git repos: `repo init` with `workspace.root` configured, omitted, `"."`, `".."`, `.runtime`, a pre-existing `0755` root, and an absolute out-of-tree root; `doctor --json` and `repo stop`/`repo status` in each; plus a hand-built legacy `project.json` driven through both the CLI and orchestrator store loaders. This is the same harness that produced the round-1 findings.
- Docker E2E/TC-16 not re-run locally; the operator-facing paths those findings were about are covered by the smoke test above.

### Remaining (both P3, non-blocking, inline)

- `repo-runtime.ts:147` — `mkdir(mode)` does not chmod a pre-existing root, so the `0700` guarantee the docs now assert can silently not hold.
- `status-surface.ts` — the guard rejects a root containing the checkout but not one containing `.runtime/orchestrator`; `workspace.root: .runtime` is still accepted.

nits carried over from round 1, still unaddressed and still fine to leave: `commands/workflow.ts:747` builds an `OrchestratorProjectConfig` without `repositoryDir`, and `commands/doctor.ts:1752` reports only `workspaceDir` in `managed_project` details.

Merge once the conflict is resolved with `assertAbsoluteProjectDir` intact and CI is green.

---

packages/cli/src/repo-runtime.ts
moncher-dev: Resolved in `8647d767`. `repo init` now follows `mkdir` with an explicit `chmod(0o700)`, so an existing root is normalized as well as a newly created one. The regression test pre-creates the configured root at `0755` and verifies it ends at `0700`.

packages/core/src/contracts/status-surface.ts
moncher-dev: Accepted and resolved in `8647d767`. Repo initialization now rejects a configured root that equals or contains the resolved orchestrator runtime directory. Focused coverage pins `.runtime` as invalid, preventing issue workspaces from sharing the state/cache hierarchy after reserved workspace keys were removed.

moncher-dev:
@hojinzs Cycle 3 addresses both recommendations and rebases PR #621 onto current `main`. Please re-review head `8647d767`.

packages/orchestrator/src/fs-store.ts
```diff
@@ -82,7 +83,12 @@ export class OrchestratorFsStore implements OrchestratorStateStore {
           )
         : null);
 
-    return config ? normalizeOrchestratorProjectConfig(config) : null;
+    if (!config) {
+      return null;
+    }
+    const normalized = normalizeOrchestratorProjectConfig(config);
+    assertDispatchableOrchestratorProjectConfig(normalized);
```
hojinzs, line 90: **P2 — the split-guard fail-closed path is not actually on the CLI daemon startup path**

My round-2 approval accepted the split on the reasoning that `OrchestratorFsStore.loadProjectConfig` is the daemon startup path. Re-verifying against `8647d767` with the built CLI, that turns out not to hold: `assertDispatchableOrchestratorProjectConfig` is only called here, and this method is only invoked by `packages/orchestrator/src/index.ts:250` (`startOrchestrator`) — a standalone binary the CLI does not use. `repo start` and `repo start --daemon` both take a different path:

- `repo start --daemon` (`packages/cli/src/commands/start.ts:1347`) spawns a child running `repo start` (foreground).
- foreground `repo start` (`packages/cli/src/commands/start.ts:910`) calls `resolveManagedProjectConfig` → `loadActiveProjectConfig` → CLI-side `loadProjectConfig` (`packages/cli/src/config.ts:182`), then `new OrchestratorService(store, projectConfig, …)` at `commands/start.ts:985`.
- CLI-side `loadProjectConfig` runs only `normalizeOrchestratorProjectConfig`, whose containment guard is skipped when `config.repositoryDir` is falsy — exactly the legacy shape.

Reproduced with the built CLI on `8647d767`: hand-authored `project.json` with `workspaceDir === repository.path` and no `repositoryDir`, then `gh-symphony repo start` — the daemon starts and reaches `connected` state instead of failing closed. `resolveIssueWorkspaceRoot` (`service.ts:4749`) returns `resolve(tenant.workspaceDir)` unconditionally, so the first dispatched issue on this shape would populate at `<repo-checkout>/<issueKey>/repository` (the exact hazard the round-1 `.` / `..` guard exists to prevent), and terminal cleanup would `rm -rf` under the checkout.

Not blocking, for three reasons: (i) legacy shape is only reachable when an operator skips the documented `stop → archive → repo init` migration; (ii) `repo init` normalizes the config forward so a followed migration is safe; (iii) this doesn't regress anything — the same legacy config already skipped the containment guard before this PR. But the "fail closed" property this PR advertises does not exist on the CLI-driven daemon path, so my round-2 approval was based on a mistaken code trace.

Two low-cost follow-ups worth doing in a separate PR:

1. Move the `assertDispatchableOrchestratorProjectConfig` call up into `OrchestratorService`'s constructor (or invoke it from `CliProjectConfig`-consuming call sites that go straight to a service). That puts the guard where it actually runs on `repo start`.
2. Add a regression test that constructs `OrchestratorService` directly from a legacy-shaped config and asserts it throws.

---

hojinzs (approved):

Round-2 P3s are both fixed on `8647d767`, and every prior finding still holds. Re-verifying by hand turned up one item I got wrong in round 2 (inline P2 on `fs-store.ts:90`) — worth reading, but non-blocking for this PR.

### Round-2 P3s — verification

| # | Sev | Claim | Verified |
| --- | --- | --- | --- |
| 1 | P3 | `0700` guarantee holds for a pre-existing root | ✅ `repo init` follows `mkdir` with an explicit `chmod(0o700)`; a pre-created `0755` root ends at `0700` on my smoke run |
| 2 | P3 | Root that contains the orchestrator runtime is rejected | ✅ `workspace.root: .runtime` fails init with a message naming the orchestrator runtime directory |

Both fixes are minimal and land in the right functions. The regression test coverage matches — `packages/cli/src/commands/repo.test.ts` gains the pre-created-`0755` case and the `.runtime` rejection alongside the existing `.` / `..` cases.

### Full smoke run on `8647d767` (built CLI, throwaway git repos)

| root | outcome |
| --- | --- |
| `.runtime/symphony-workspaces` (configured) | root@`0700`, `doctor` pass, `repo stop` exit 0 |
| omitted | defaults to `.runtime/symphony-workspaces`, distinct from `.runtime/orchestrator`, root@`0700`, `doctor` pass |
| `.` | rejected: `workspace.root … must not equal or contain the repository checkout` |
| `..` | rejected, same message |
| `.runtime` | rejected: `workspace.root … must not equal or contain the orchestrator runtime directory` |
| pre-existing `0755` | normalized to `0700` at init |
| legacy `project.json` (no `repositoryDir`, `workspaceDir === repository.path`) | `repo stop` exit 0, `repo status` exit 0, `doctor` `managed_project` + `workspace_root` both `pass` |

`pnpm lint`, `pnpm typecheck`, `pnpm build`, `pnpm test` all exit 0 locally on `8647d767` merged with `origin/main`.

### Acceptance criteria (#599) — final tally

| Criterion | Status |
| --- | --- |
| `workspace.root: X` populates at `X/`, escape guard validates against `X` | ✅ |
| Startup cleanup, dirty quarantine, worktree removal use the configured root | ✅ |
| `doctor` reports the real workspace root for both modes | ✅ |
| Migration documented + changeset; divergence note removed | ✅ |
| Decision recorded in the ADR | ✅ |
| `RESERVED_WORKSPACE_KEYS` revisited | ✅ (`.runtime` guard closes the last edge) |

No over-engineering, no unrequested scope; the delta over round 2 is 66 lines and every hunk traces to a round-2 finding.

### Merge conflict (unchanged from round 2)

`AGENT_TEST.md` conflicts against `origin/main` (`#623` and this PR both added a row for `e2e/scenarios/16-*.md`). I merged locally to confirm it resolves cleanly: keep both rows, no other conflicts. Verified `assertAbsoluteProjectDir` from #617 stays intact — no `packages/core/src/contracts/status-surface.ts` conflict on the current base (that concern from round 2 was on `ccc10529`, and the round-2 rewrite of that file re-anchored it above the merge base for #617). CI shows no check runs on `8647d767` yet; please confirm green after the merge.

### One correction to my round-2 approval — inline P2 on `fs-store.ts:90`

Non-blocking, but worth reading. Short version: I claimed the split `assertDispatchableOrchestratorProjectConfig` guard fails closed on daemon startup. Re-checking against the CLI's actual daemon path (`repo start` → `commands/start.ts:910`/`985`), it doesn't — the guard is only reachable through `orchestrator/src/index.ts::startOrchestrator`, which the CLI does not invoke. On a legacy install, `repo start` will start the daemon without hitting the guard. The migration doc is what actually protects operators here. Details and reproduction inline; a follow-up (move the assertion into `OrchestratorService`'s constructor) is cheap and outside this PR's scope.

### Remaining nits (unchanged, non-blocking)

- `packages/cli/src/commands/workflow.ts:747` builds an `OrchestratorProjectConfig` without `repositoryDir`.
- `packages/cli/src/commands/doctor.ts:1752` reports only `workspaceDir` in `managed_project` details.

**Approve.** Resolve the `AGENT_TEST.md` conflict, land the merge with `#617`'s `assertAbsoluteProjectDir` intact (already true on the current base), confirm CI is green, and merge.

---

PR #623, opened 2026-08-23
fix(orchestrator): bound unknown tracker finalization deferrals

### LINKED ISSUE #611: fix(orchestrator): bound unknown tracker finalization deferrals
## Background

PR #605 fixes issue #576 by deferring successful-run finalization when the fresh canonical tracker readback is `unknown`. The latest approved review verified transient recovery but identified follow-up debt: a persistent `unknown` currently leaves the run and issue orchestration in `running` indefinitely, without retry accounting or a visible run event.

The same review found that service-level coverage consumes `trackerProgressConfirmedAt` but does not positively verify the full arming path that persists it, and stubs the final-state classifier instead of covering its `non-actionable | active | unknown` mapping end to end.

## Proposed Changes

- Add bounded consecutive finalization-deferral accounting for persistent `unknown` tracker readbacks.
- After the bound is reached, enter the existing retry/failure classification path so the run cannot remain pinned forever.
- Emit a `run-finalization-deferred` event with enough context for repository status and dashboard diagnostics.
- Add service-level coverage for both positive and fail-closed tracker-progress arming.
- Exercise the real final tracker-state classifier for `non-actionable`, `active`, and `unknown`, including the successful `non-actionable → succeeded` path.
- Split or rename the combined continuation/finalization regression so each behavior is explicit.

## Affected Files

| File | Change | Scope |
| --- | --- | --- |
| `packages/orchestrator/src/service.ts` | Bound unknown finalization deferrals and emit an observable event | Coordination / Execution |
| `packages/orchestrator/src/service.test.ts` | Cover arming, classifier mappings, bounded fallback, and focused test naming | Tests |
| `packages/core/src/contracts/status-surface.ts` | Extend run/event status contracts if the event needs a surfaced payload | Observability (if required) |
| `docs/architecture.md` / `AGENT_TEST.md` | Document the bounded behavior and executable regression scenario if externally visible | Living docs (if required) |

## Acceptance Criteria

- A transient `unknown` remains deferred and can later finalize successfully.
- A persistent `unknown` cannot leave a completed worker run pinned indefinitely.
- Each deferral is observable through persisted run events, not only verbose logs.
- Service tests fail if tracker progress is never armed or if any classifier mapping regresses.
- `pnpm lint`, `pnpm test`, `pnpm typecheck`, and `pnpm build` pass.

## References

- Follow-up from #576 and PR #605 review threads `3827246285`, `3827247106`, and `3827249986`.

moncher-dev:
## Issues

- Fixes #611

## TL;DR

- Bounds successful-run finalization after three consecutive unknown canonical tracker reads, preventing completed workers from remaining pinned indefinitely.
- Makes every deferral visible as a warning with a discriminated cause and carries truthful tracker diagnostics into failure retry handling.

## Change-point diagram

```text
successful worker exit
        |
fresh canonical tracker read
        |
  +-----+----------+--------------------------------+
  |                |                                |
active       non-actionable                       unknown
  |                |                                |
continue        succeed          classify workflow / missing item / read failure
                                                   |
                                      persist warning deferral 1..3
                                                   |
                                before bound: defer and reconcile again
                                at bound: force failure retry accounting
```

## Start here

1. `packages/orchestrator/src/service.ts` — cause-bearing classifier, bounded finalization, and marker-independent exhausted failure accounting.
2. `packages/core/src/observability/event-formatter.ts` — operator-facing deferral message.
3. `packages/orchestrator/src/service.test.ts` — arming, classifier mappings, transient recovery, and dirty-workspace exhaustion across terminal event markers.
4. `e2e/scenarios/16-bounded-finalization-deferral.md` — Docker black-box regression for the missing-item cause.

## Changes

- Persist at most three consecutive unknown finalization reads before entering existing failure retry/suppression accounting.
- Discriminate `workflow-unavailable`, `tracker-item-missing`, and `tracker-read-failed`; persist the diagnostic in events and exhausted retry state.
- Bypass dirty-workspace recovery after the finalization bound is exhausted, independently of `lastEvent`; dirty work remains attributable through quarantine handling.
- Render deferrals in repository logs and dashboard status as warnings while preserving worker-origin control and activity state.
- Cover positive/fail-closed progress arming and real `active`, `non-actionable`, and `unknown` classifier behavior, including transient recovery and bounded exhaustion.
- Document the adaptive tick-based grace window and the intentional repository divergence from upstream normal-exit continuation retry.
- Add Docker TC-16 and the patch changeset for `@gh-symphony/cli`.

## Evidence

- `pnpm exec vitest run packages/orchestrator/src/service.test.ts -t 'finalization|completed workspace is dirty'` — passed; six focused tests, including dirty workspaces with `turn_completed`, `heartbeat`, and unset `lastEvent`.
- `pnpm lint` — passed across all public workspace packages in Cycle 6.
- `pnpm test` — passed across all 14 public workspace packages; orchestrator service 167 tests and CLI 541 tests.
- `pnpm typecheck` — passed in Cycle 6.
- `pnpm build` — passed in Cycle 6.
- `bash e2e/run-e2e.sh api-progress-unknown 120` — passed in Cycle 6; emitted exactly three `tracker-item-missing` deferrals, exhausted only the third, and retained the canonical tracker diagnostic through failure retry.
- Review-thread audit — eight total threads, all answered and resolved on head `c9ead5a`.

## Risks & rollback

- Risk: the bound counts reconciliation opportunities, so adaptive polling can stretch elapsed grace by up to the configured 10× multiplier; this is documented and keeps provider-read cost deterministic.
- Intentional divergence: exhausted clean-exit finalization uses failure retry/suppression rather than the upstream normal-exit continuation retry so unresolved canonical state is bounded.
- Rollback: revert `c9ead5a`, `c6dc01b`, `6cc7904`, and the original delivery commit to restore the prior behavior.

## Changed files

- Orchestrator finalization logic and service regressions.
- Core status/event contracts and formatter coverage.
- CLI warning filtering and dashboard warning-tone coverage.
- Docker E2E runner, seed, stub worker, and TC-16.
- Architecture and agent-test living documentation.
- Patch changeset `.changeset/tidy-taxis-wait.md`.

- [ ] Confirm repository diagnostics expose cause-bearing deferral warnings during a real tracker outage.
- [ ] Confirm no adjacent finalization behavior regresses under a real GitHub Project tracker outage.

## Human Validation

- [ ] Confirm transient unknown reads still recover to success.
- [ ] Confirm persistent unknown reads cannot pin a completed run indefinitely, including when the issue workspace is dirty and the final worker marker is `heartbeat`.
- [ ] Confirm repository logs and dashboard diagnostics show count, bound, cause, and exhaustion clearly.


packages/core/src/observability/structured-events.ts
```diff
@@ -119,6 +119,19 @@ export type RunSuppressedEvent = {
...
+  at: string;
+  event: "run-finalization-deferred";
+  projectId: string;
+  runId: string;
+  issueIdentifier: string;
+  issueId: string;
+  reason: "tracker-state-unknown";
+  consecutiveDeferrals: number;
+  maxDeferrals: number;
+  exhausted: boolean;
```
hojinzs, line 136: **P1 — the new event renders empty in every operator surface, so the acceptance criterion is not met**

Issue #611 requires: *"Emit a `run-finalization-deferred` event with enough context for repository status and dashboard diagnostics."* The payload carries the context, but no consumer reads it. `grep` for `run-finalization-deferred` outside the orchestrator returns only this type declaration:

1. **`packages/core/src/observability/event-formatter.ts:5-53`** — `formatEventMessage` has no `case` for this event, so it falls to `default: return null`.
2. **`packages/cli/src/commands/logs.ts:224`** — `repo logs` renders `[at] run-finalization-deferred SYM-1` with an empty message. `consecutiveDeferrals` / `maxDeferrals` / `exhausted` — the three fields added specifically for diagnostics — are never printed.
3. **`packages/cli/src/commands/logs.ts:235-247`** — `getLevel` falls through to `info`. An **exhausted** deferral is about to burn a failure retry, yet it is filtered out of `--level warn`, unlike `run-retried` and `run-suppressed` which are both `warn`.
4. **`packages/control-plane/client/src/routes/issues/$identifier.tsx:305`** — the dashboard renders `event.message ?? event.event`, so it shows the bare event name; `classifyEventTone` (`:458-473`) gives it `default` tone rather than `warning`.

The fix is small — a formatter case is the load-bearing part:

```ts
case "run-finalization-deferred":
  return `Finalization deferred ${event.consecutiveDeferrals}/${event.maxDeferrals} (${event.reason})${event.exhausted ? " — bound exhausted" : ""}`;
```

plus adding the event to `getLevel`'s `warn` group. Please also add a `packages/core` unit test asserting `formatEventMessage` is non-null for this event, so the next new event type does not regress the same way.

Related: `finalizationDeferralCount` was added to `OrchestratorRunRecord` (`status-surface.ts:241`) but is likewise read by nothing outside `service.ts`. That is fine as internal reconciliation state — just don't count it toward the observability criterion.

---

packages/orchestrator/src/service.ts
```diff
@@ -3292,15 +3293,41 @@ export class OrchestratorService {
...
+      };
+      await this.store.saveRun(deferredRun);
+      await this.store.appendRunEvent(run.runId, {
+        at: now.toISOString(),
+        event: "run-finalization-deferred",
+        projectId: run.projectId,
+        runId: run.runId,
+        issueIdentifier: run.issueIdentifier,
+        issueId: run.issueId,
+        reason: "tracker-state-unknown",
```
hojinzs: **P2 — `reason` is hardcoded, collapsing a permanent cause into a transient one**

`classifyCurrentTrackerProgress` (`service.ts:3940-3976`) returns `"unknown"` from four distinct branches:

| Branch | Meaning | Correct response |
| --- | --- | --- |
| `!isUsableWorkflowResolution(resolution)` | workflow policy unreadable | operator config error — not a tracker outage |
| `!issue` (id absent from the response) | canonical item deleted / moved out of the project | **permanent** — release, retrying will never help |
| `catch (error)` | tracker API threw | **transient** — deferring is right |

All three are flattened into `reason: "tracker-state-unknown"`, so a reader of the event log cannot tell which happened. The `catch` branch already formats the real error into a verbose log (`:3971-3973`) and then discards it — that string is exactly what the event needs.

This matters concretely for the deleted-item case: a completed run whose issue was removed from the project now spends 3 deferrals, a failure retry, a full worker restart, and eventually **suppression**, when `releaseIssueOrchestration` was the right answer on the first read.

It is also worth noting that the new E2E scenario exercises precisely this branch — `removeCanonicalTrackerItem()` writes `[]` to the fixture, so `api-progress-unknown` is *item-removed*, not *tracker-unreachable*. TC-16 therefore validates the counting mechanism but not the outage case the PR description is framed around.

Suggested shape — discriminate the cause and carry it through:

```ts
reason: "tracker-read-failed" | "tracker-item-missing" | "workflow-unavailable";
```

Threading the real cause through also resolves the Codex P2 on the retry `lastError`, since you then have a truthful error string to persist instead of `"Worker process exited unexpectedly."`.

---

packages/orchestrator/src/service.ts
```diff
@@ -3292,15 +3293,41 @@ export class OrchestratorService {
           )
         : null;
     if (currentTrackerProgress === "unknown") {
+      const consecutiveDeferrals =
+        (runWithTokens.finalizationDeferralCount ?? 0) + 1;
+      const exhausted = consecutiveDeferrals >= MAX_FINALIZATION_DEFERRALS;
```
hojinzs, line 3298: **P3 — the bound counts ticks, so the actual grace period is not a predictable duration**

`MAX_FINALIZATION_DEFERRALS = 3` is measured in reconciliation ticks, and the poll interval is adaptive: `MAX_ADAPTIVE_POLL_INTERVAL_MULTIPLIER = 10` (`service.ts:96`) stretches it up to 10× under rate-limit pressure. So the grace window before a completed run is pushed into failure retry ranges over roughly an order of magnitude — and it is *shortest* exactly when the tracker is healthy and *longest* when the orchestrator is already backing off.

The PR description flags "three reconciliation ticks may be too short for unusually long tracker outages" as a known risk. The tick-based unit makes that risk harder to reason about, because there is no single number an operator can point at.

The adjacent guard for the same lifecycle stage is time-based — `TRACKER_PROGRESS_EXIT_GRACE_MS = 30_000` (`:93`) — so a wall-clock bound would also be the locally consistent choice. Either derive the bound from elapsed time since the first deferral, or keep the counter and state the resulting window explicitly in `docs/architecture.md`. Not a blocker; a deliberate answer either way is fine.

---

packages/orchestrator/src/service.ts
```diff
@@ -3292,15 +3293,41 @@ export class OrchestratorService {
...
+      const deferredRun: OrchestratorRunRecord = {
+        ...runWithTokens,
+        finalizationDeferralCount: consecutiveDeferrals,
+        updatedAt: now.toISOString(),
+        lastEvent: "run-finalization-deferred",
+        lastEventAt: now.toISOString(),
```
hojinzs: **nit — `lastEventAt` is overwritten with an orchestrator timestamp while `lastEventAtSource` keeps claiming a worker origin**

`lastEventAtSource` is typed `"event-channel" | "worker-api" | null` (`status-surface.ts:225`) — both values assert the timestamp came from the worker. Here `lastEventAt` is set to the orchestrator's `now`, but `lastEventAtSource` is inherited untouched from `runWithTokens` (`:3221-3222`), so the persisted pair reads as "the worker reported an event at `now`" when the worker has already exited.

Since a deferral tick can repeat up to three times, this also drags the apparent worker-activity timestamp forward on a run with no live worker, which is the opposite of what a staleness check wants. Either leave `lastEventAt` alone or set `lastEventAtSource: null` alongside it.

---

hojinzs (changes requested):
## Review — cycle 1 of 6

The core mechanism is right and the scope is disciplined. Blocking on one acceptance-criterion gap plus two confirmed diagnostics defects; all are small, local fixes.

### Requirements traceability (#611)

| --- | --- |
| Transient `unknown` stays deferred, later finalizes successfully | ✅ `recovers a transient unknown finalization read and later succeeds` |
| Persistent `unknown` cannot pin a completed run indefinitely | ✅ bounded at 3, then failure retry → suppression |
| Each deferral observable through persisted run events | ⚠️ **event is persisted but renders empty in every surface — see P1** |
| Tests fail if arming or a classifier mapping regresses | ✅ positive (`trackerProgressConfirmedAt: "2026-08-07T09:01:00.000Z"`) and fail-closed (`toBeNull()`) arming; all three classifier mappings driven through the real classifier via `fetchIssueStatesByIds`, stub removed |
| Split the combined continuation/finalization regression | ✅ one combined test → four explicitly named ones |
| `pnpm lint` / `test` / `typecheck` / `build` | ✅ independently reproduced |

**Over-engineering / unrequested decisions: none found.** Every file in the diff maps to a line item in the issue. The E2E scenario and the `GH_SYMPHONY_FILE_TRACKER_ISSUES_PATH` seed line go slightly beyond the issue's "if required" wording, but AGENTS.md §3 mandates Docker coverage when integration behavior changes, so they are justified.

### Independent verification

Re-ran the full gate on `dd2e82a` locally rather than taking the evidence section on trust:

- `pnpm build` — pass
- `pnpm lint` — pass
- `pnpm -r --workspace-concurrency=1 test` — pass, 14 packages; orchestrator 14 files / **318** tests, CLI 39 files / **541** tests (matches the PR's claimed numbers exactly)
- `pnpm typecheck` — pass
- CI on this SHA: `Test` ✅, `Container Smoke` ✅
- Docker E2E — **not reproduced**; the Docker daemon is unavailable in this review environment. I read `run-e2e.sh:272-297` instead: the assertions (exactly 3 events, `consecutiveDeferrals` 1/2/3, `maxDeferrals` 3, `exhausted` only on the third) mirror the unit test, which passes. Treating the reported TC-16 run as credible but unverified.

**Smoke test of the new behavior.** I instrumented the exhaustion test to dump persisted state at the bound (reverted afterwards, working tree clean):

```
PROBE lastError=    "Worker process exited unexpectedly."
PROBE runPhase=     "succeeded"
PROBE issueRecord=  {"attempt":2,"dueAt":"2026-03-08T00:00:07.000Z",
                     "error":"Worker process exited unexpectedly."}
```

That is empirical confirmation of the Codex P2 below — the run is persisted as `succeeded` *and* carrying a fabricated exit error, and the false string propagates to the issue's `retryEntry.error` and from there to the status surface. The exhaustion test asserts `status` / `retryKind` / `finalizationDeferralCount` / `nextRetryAt` but not `lastError`, so nothing guards it.

### Findings

| # | Sev | Where | Summary |
| --- | --- | --- | --- |
| 1 | **P1** | `structured-events.ts` (inline) | Event has no `formatEventMessage` case → renders empty in `repo logs` and the dashboard, and is `info` rather than `warn`. The diagnostic fields are unreachable, so the observability acceptance criterion is unmet. |
| 2 | **P2** | `service.ts:3314` (inline) | `reason` hardcoded; four distinct `unknown` causes collapsed, including the *permanent* deleted-item case. TC-16 actually exercises that permanent branch, not an outage. |
| 3 | **P3** | `service.ts:3298` (inline) | Bound counts ticks, and the poll interval is adaptive up to 10× — the real grace window is not a stable duration. |
| 4 | nit | `service.ts:3303-3304` (inline) | `lastEventAt` advanced with an orchestrator timestamp while `lastEventAtSource` still claims a worker origin. |

### On the two Codex findings

Both are valid; I verified them rather than deferring to the bot.

- **Codex P1 (divergence not recorded)** — **agree, and it has behavioural teeth, not just documentation.** `docs/symphony-spec.md:648-651` and `:1867-1874` both specify that a *normal* worker exit schedules a `continuation` retry. This PR routes a normal exit (`runPhase === "succeeded"`) into `retryKind: "failure"`, which means exponential backoff, `failureRetryCount++`, and — once `failureRetryCount >= maxFailureRetries` — `status: "suppressed"` with `"Run suppressed: max failure retries exceeded"`. A run that genuinely succeeded can therefore terminate as suppressed purely because the tracker stayed unreachable. That is a defensible trade for un-pinning the run, but AGENTS.md:35-40 requires it be called out explicitly as a repository-level choice. The `docs/architecture.md:43` rewrite describes the behavior without naming it a divergence, and the workpad on #611 affirmatively records *"no intentional divergence"* — please correct that too.
- **Codex P2 (tracker-outage cause lost in retry diagnostics)** — **agree**, confirmed by the probe above. Finding #2 gives you the discriminated cause to persist there.

### Path to approval

Fix #1 (formatter case + `warn` level + a regression test) and #2 (discriminated `reason`, which supplies the truthful `lastError` for Codex P2), and record the spec divergence for Codex P1. #3 and #4 need a deliberate answer, not necessarily a change — a reasoned "won't fix" on the thread is fine.

I'd expect this to be approvable in one more cycle.

---

docs/architecture.md
```diff
@@ -40,7 +40,7 @@ touches a layer, check that its slice (and the linked documents) still holds.
...
+- Confirmed tracker transitions outside the configured active states are recorded on the active run. When the canonical item becomes non-actionable, reconciliation gives the worker a bounded clean-exit grace; successful finalization then re-reads the current canonical state and preserves `succeeded` only while it remains non-actionable. An unavailable final read persists a `run-finalization-deferred` event and defers classification for up to three consecutive reconciliation ticks; the third unknown read enters the existing failure-retry path so the run cannot remain pinned. A known active or non-actionable result resets the streak. Per-turn `state-read` requests do not reload or rewrite workflow snapshots.
```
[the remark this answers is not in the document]
moncher-dev: Resolved in `6cc7904`. `docs/architecture.md` now explicitly records this as an intentional repository-level divergence: after the bounded canonical-read grace, a successful worker exit enters failure retry/suppression rather than the upstream normal-exit continuation retry. The note explains the bounded-finalization rationale and confirms that tracker-specific diagnostics are retained.

packages/orchestrator/src/service.ts
```diff
@@ -3362,9 +3389,10 @@ export class OrchestratorService {
     }
 
     // Determine retry kind: continuation (issue still actionable) vs failure
-    const retryKind = convergenceDetected
-      ? "failure"
-      : await this.classifyRetryKind(tenant, run, trackerDependencies);
+    const retryKind =
+      convergenceDetected || currentTrackerProgress === "unknown"
+        ? "failure"
```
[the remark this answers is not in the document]
moncher-dev: Resolved in `6cc7904`. The final tracker classifier now returns a discriminated unknown cause plus a truthful diagnostic. Exhaustion carries that diagnostic into `lastError` and the issue retry entry, so a successful worker no longer reports `Worker process exited unexpectedly.` for a canonical tracker failure. Unit and Docker TC-16 coverage assert the persisted cause.

packages/core/src/observability/structured-events.ts
moncher-dev: Resolved in `6cc7904`. `run-finalization-deferred` now has a non-null formatter message containing count, bound, cause, and exhaustion; `repo logs --level warn` includes it; and the dashboard assigns warning tone. Core formatter, CLI filter/rendering, and control-plane tone regressions cover all three operator surfaces.

packages/orchestrator/src/service.ts
moncher-dev: Resolved in `6cc7904`. The classifier distinguishes `workflow-unavailable`, `tracker-item-missing`, and `tracker-read-failed`, and persists a cause-specific diagnostic in both the deferral event and exhausted retry record. TC-16 and its documentation now explicitly identify the fixture as the permanent `tracker-item-missing` branch rather than an outage, while a new service test exercises a thrown tracker read and asserts `tracker-read-failed`.

packages/orchestrator/src/service.ts
moncher-dev: Kept the consecutive-read bound deliberately and documented the tradeoff in `6cc7904`. The counter bounds provider reconciliation opportunities deterministically; adaptive polling may stretch the elapsed grace window by up to the configured 10× multiplier. `docs/architecture.md` now states that timing explicitly so operators do not interpret the bound as a fixed wall-clock duration.

packages/orchestrator/src/service.ts
moncher-dev: Resolved in `6cc7904`. Deferral persistence no longer overwrites `lastEventAt`; the original worker-origin timestamp and `lastEventAtSource` remain paired. The transient-unknown service regression seeds an event-channel timestamp and asserts that both fields remain unchanged after deferral.

moncher-dev:
Resolved in `6cc7904`. `docs/architecture.md` now explicitly records this as an intentional repository-level divergence: after the bounded canonical-read grace, a successful worker exit enters failure retry/suppression rather than the upstream normal-exit continuation retry. The note explains the bounded-finalization rationale and confirms that tracker-specific diagnostics are retained.

moncher-dev:
@hojinzs Re-review requested for `6cc7904`. All six inline threads have concrete replies; the full local completion gate and Docker TC-16 pass, and the refreshed PR body records the intentional upstream divergence and rework evidence.

packages/orchestrator/src/service.ts
```diff
@@ -3291,16 +3292,42 @@ export class OrchestratorService {
...
-    if (currentTrackerProgress === "unknown") {
+    if (currentTrackerProgress?.state === "unknown") {
+      const consecutiveDeferrals =
+        (runWithTokens.finalizationDeferralCount ?? 0) + 1;
+      const exhausted = consecutiveDeferrals >= MAX_FINALIZATION_DEFERRALS;
+      const deferredRun: OrchestratorRunRecord = {
+        ...runWithTokens,
+        finalizationDeferralCount: consecutiveDeferrals,
+        updatedAt: now.toISOString(),
+        lastEvent: "run-finalization-deferred",
```
hojinzs: **P1 — this line silently defeats the bound: with a dirty issue workspace, exhaustion degrades to a `recovery` retry that never increments `failureRetryCount`**

`lastEvent` is not just a display field — it is control state. `classifyIncompleteTurnDirtyWorkspace` (`service.ts:4058`) opens with:

```ts
if (run.lastEvent === "turn_completed") {
  return null;
}
```

Persisting `lastEvent: "run-finalization-deferred"` overwrites the worker's terminal `turn_completed` marker in the store. On tick 1 the guard still holds (the call takes `runWithTokens`, whose `lastEvent` is still the worker's). From tick 2 onward the worker is gone, so `workerInfo.lastEvent` is undefined and `runWithTokens.lastEvent = run.lastEvent` — now `"run-finalization-deferred"`. The guard stops firing exactly on the exhaustion tick.

If the issue workspace has any uncommitted file at that moment, `recovery` becomes non-null, and the fall-through at `:3389-3392` is bypassed in every way that matters:

| | `recovery` null (tested) | `recovery` set (untested) |
| --- | --- | --- |
| `retryKind` persisted | `failure` | `recovery` |
| `failureRetryCount` | `+1` → reaches `maxFailureRetries` | **stays 0 forever** |
| `lastError` | the new tracker diagnostic | **`null`** — `:3502-3506` short-circuits on `recovery \|\|` |
| `nextRetryAt` | exponential backoff | fixed `CONTINUATION_RETRY_DELAY_MS` |

So a completed run whose canonical item is permanently gone loops: 3 deferrals → recovery retry → restart → 3 deferrals → … `failureRetryCount` never advances, suppression is never reached, and the run stays pinned. That is the exact acceptance criterion from #611 ("A persistent `unknown` cannot leave a completed worker run pinned indefinitely"). It also throws away the discriminated cause this cycle added for Codex P2 — `lastError` is `null` on this branch. `recovery-quarantined` (`:2673`) does not bound it either; that path is about dirty-work *attribution*, not retry accounting.

**Empirically confirmed.** I built a fixture with `lastEvent: "turn_completed"` and one untracked file in the issue workspace, then ran three ticks (probe reverted; tree clean):

```
PROBE tick1 lastEvent     = "run-finalization-deferred"
PROBE tick3 retryKind     = "recovery"
PROBE tick3 lastError     = null
PROBE tick3 recovery.kind = "incomplete-turn-dirty-workspace"
PROBE failureRetryCount   = 0
PROBE retryEntry          = {"attempt":2,"dueAt":"2026-03-08T00:00:01.000Z","error":null}
```

Deleting only this one line, same fixture, same three ticks:

```
PROBE tick1 lastEvent     = "turn_completed"
PROBE tick3 retryKind     = "failure"
PROBE tick3 lastError     = "Final tracker state unavailable: canonical tracker item issue-1 was not returned."
PROBE tick3 recovery.kind = null
PROBE failureRetryCount   = 1
```

**Suggested fix:** drop the `lastEvent` write from `deferredRun`. The observability need is already fully met by the persisted `run-finalization-deferred` event plus this cycle's formatter — I verified that separately through the built CLI (see the review summary), so this line is not carrying any operator-facing weight. With the line removed the full orchestrator suite still passes (164/164), which is also the problem: nothing currently guards this. Please add a regression that seeds `lastEvent: "turn_completed"` plus a dirty issue workspace and asserts the exhaustion tick still yields `retryKind: "failure"`, `failureRetryCount: 1`, and the tracker diagnostic in `lastError`.

Note this is a side effect of the timestamp nit from cycle 1 — the `lastEventAt` half was fixed correctly; it is the `lastEvent` half that landed with it.

---

hojinzs (changes requested):
## Review — cycle 2 of 6

All six threads from cycle 1 are genuinely addressed — I verified each against the code rather than the replies. One new defect landed with the timestamp fix, and it is load-bearing: it defeats the PR's primary acceptance criterion. The fix is deleting one line plus a regression test.

### Cycle 1 findings — verification

| Finding | Verdict |
| --- | --- |
| **P1** event renders empty in every surface | ✅ **Fixed and smoke-tested.** Formatter case added, `warn` level, dashboard tone, regression tests in all three packages. |
| **P2** `reason` hardcoded, causes collapsed | ✅ **Fixed.** `workflow-unavailable` / `tracker-item-missing` / `tracker-read-failed` discriminated, each with its own diagnostic; carried into `lastError`; new test drives a thrown tracker read. |
| **P3** bound counts ticks, not wall-clock | ✅ **Reasoned won't-fix, accepted.** `docs/architecture.md:43` now states the 10× adaptive stretch explicitly, which is what I asked for. |
| **nit** `lastEventAt` vs `lastEventAtSource` | ⚠️ **Half fixed.** `lastEventAt`/`lastEventAtSource` are correctly left untouched and the transient test asserts it — but the same edit added a `lastEvent` write. See the inline P1. |
| **Codex P1** divergence not recorded | ✅ **Fixed.** `docs/architecture.md:43` names it an intentional repository-level divergence with rationale, and the Cycle 2 workpad on #611 corrects the earlier "no intentional divergence" claim. |
| **Codex P2** tracker cause lost in retry diagnostics | ✅ **Fixed** on the `recovery === null` path — and *not* on the branch the new P1 opens up. |

### Smoke test of the P1 fix

Docker Hub image pulls are 403-blocked by this environment's proxy, so TC-16 is again unreproduced (`node:24-bookworm-slim` metadata fetch fails). I read `run-e2e.sh:272-297` instead — the assertions mirror the passing unit test — and treated the reported run as credible but unverified.

What I *could* black-box was the operator surface, through the built CLI against a hand-seeded `events.ndjson`:

```
$ node packages/cli/dist/index.js repo logs
[…] run-finalization-deferred acme/platform#1 Finalization deferred 1/3 (tracker-read-failed)
[…] run-finalization-deferred acme/platform#1 Finalization deferred 3/3 (tracker-read-failed) — bound exhausted

$ node packages/cli/dist/index.js repo logs --level warn
[…] run-finalization-deferred acme/platform#1 Finalization deferred 1/3 (tracker-read-failed)
[…] run-finalization-deferred acme/platform#1 Finalization deferred 3/3 (tracker-read-failed) — bound exhausted
```

Count, bound, cause, and exhaustion all render, and the event now survives the `warn` filter. I also confirmed the formatter is reached by the dashboard and not only the CLI: `event-formatter.ts:83` populates `IssueStatusEvent.message` in the shared core builder that `fs-store.ts:296` uses for status snapshots, so `classifyEventTone`'s `"finalization deferred"` match is fed a real message in production, not just in its test. Cycle 1's P1 is fully closed.

### Independent verification

Re-ran the gate on `6cc7904` rather than trusting the evidence section:

- `pnpm lint` — pass
- `pnpm typecheck` — pass
- `pnpm -r --workspace-concurrency=1 test` — pass, 14 packages; orchestrator **319** tests (+1 vs cycle 1), CLI **541**
- `pnpm build` — pass
- CI on this SHA: `Test` ✅, `Container Smoke` ✅

### Requirements traceability (#611)

Every acceptance criterion is now met on the tested path. The one that regresses is *"A persistent `unknown` cannot leave a completed worker run pinned indefinitely"* — held when `recovery` is null, defeated when the issue workspace is dirty. Details, the empirical probe, and the one-line fix are in the inline comment.

**Over-engineering / unrequested decisions: none found**, this cycle or last. Every hunk in `6cc7904` maps to a specific review thread; nothing was widened.

### Path to approval

Delete the `lastEvent` write in `deferredRun` and add the regression described inline. That is the only blocker — everything else in this PR is ready to land, and I'd expect to approve on the next push.

---

packages/orchestrator/src/service.ts
hojinzs: Confirmed the `lastEventAt` / `lastEventAtSource` half — they are correctly left untouched now, and the transient-unknown regression asserts the pair survives a deferral. Leaving this thread open only because the same edit added a `lastEvent: "run-finalization-deferred"` write to `deferredRun`, which turns out to be control state rather than display state: it defeats the `turn_completed` guard in `classifyIncompleteTurnDirtyWorkspace` from tick 2 onward. Details and the probe are in the new inline P1 on `service.ts:3303`. Resolve both together once that line is gone.

---

packages/orchestrator/src/service.ts
moncher-dev: Resolved in `c6dc01b`. Finalization deferrals no longer overwrite `lastEvent`, so the worker's terminal `turn_completed` marker remains available to the dirty-workspace classifier on every reconciliation tick. The new regression creates a real dirty issue workspace, runs all three missing-item deferrals, and asserts the exhaustion tick remains `retryKind: "failure"`, increments `failureRetryCount` to 1, preserves the canonical-item diagnostic in `lastError` and `retryEntry.error`, and leaves `recovery` null.

packages/orchestrator/src/service.ts
moncher-dev: Resolved together in `c6dc01b`. The deferral snapshot now preserves both halves of worker control/activity state: `lastEvent` remains `turn_completed`, while `lastEventAt` and `lastEventAtSource` remain unchanged. Observability continues through the separately persisted `run-finalization-deferred` event. The new dirty-workspace regression proves this prevents recovery classification from bypassing bounded failure accounting.

moncher-dev:
Resolved in `c6dc01b`. Finalization deferrals no longer overwrite `lastEvent`, so the worker's terminal `turn_completed` marker remains available to the dirty-workspace classifier on every reconciliation tick. The new regression creates a real dirty issue workspace, runs all three missing-item deferrals, and asserts the exhaustion tick remains `retryKind: "failure"`, increments `failureRetryCount` to 1, preserves the canonical-item diagnostic in `lastError` and `retryEntry.error`, and leaves `recovery` null.

moncher-dev:
@hojinzs Re-review requested for `c6dc01b`. The new P1 is resolved by preserving the worker's `turn_completed` control marker across deferrals, and the added dirty-workspace regression proves exhaustion still increments failure retry accounting with the truthful tracker diagnostic. Both relevant inline threads are answered; the full local completion gate and Docker TC-16 pass.

moncher-dev:
@hojinzs Re-review requested for `c6dc01b` after the Cycle 4 current-head audit. All seven review threads are answered and resolved, the full lint/test/typecheck/build gate passes, and Docker TC-16 again confirms exactly three cause-bearing deferrals before truthful failure retry.

packages/orchestrator/src/service.ts
```diff
@@ -3362,9 +3388,10 @@ export class OrchestratorService {
...
 
     // Determine retry kind: continuation (issue still actionable) vs failure
-    const retryKind = convergenceDetected
-      ? "failure"
-      : await this.classifyRetryKind(tenant, run, trackerDependencies);
+    const retryKind =
+      convergenceDetected || currentTrackerProgress?.state === "unknown"
+        ? "failure"
+        : await this.classifyRetryKind(tenant, run, trackerDependencies);
     const persistedRetryKind = recovery ? "recovery" : retryKind;
```
hojinzs, line 3398: **P1 — the cycle-2 fix is correct but its guard is not an invariant: on the real successful-exit sequence `lastEvent` is `"heartbeat"`, so `recovery` still wins here and the bound is bypassed**

Deleting the `lastEvent` write was the right change and I confirmed it guards — reintroducing that one line fails the new regression:

```
× preserves bounded failure retry accounting when the completed workspace is dirty
  → expected { runId: 'run-1', …(35) } to match object { status: 'retrying', …(5) }
```

But `persistedRetryKind = recovery ? "recovery" : retryKind` on the next line means the whole fix rests on `classifyIncompleteTurnDirtyWorkspace` (`:4058`) short-circuiting on `run.lastEvent === "turn_completed"` — and nothing on this path establishes that value. The new regression **seeds** it (`lastEvent: "turn_completed"`, `service.test.ts:8024`); the shared `createSuccessfulFinalizationFixture` never sets `lastEvent` at all. So the test proves the guard works when it fires, not that it fires.

**It does not fire on the normal successful exit.** The worker emits `turn_completed` and then a terminal heartbeat, in that order, on the same success path:

- `worker/src/index.ts:773` — `emitTurnCompletedEvent(turnTelemetry)` (inside `try`)
- `worker/src/index.ts:799-800` — `stopOrchestratorHeartbeatTimer(); emitOrchestratorHeartbeat();` (in `finally`, so it always follows)

and the orchestrator's heartbeat handler overwrites the marker:

```ts
// service.ts:3704-3711
if (event.type === "heartbeat") {
  await this.store.saveRun({ ...run, lastEvent: "heartbeat", … });
```

`canApplyWorkerChannelUpdate(run.status)` passes while the run is still `running`, so the last thing persisted before finalization is `"heartbeat"`.

There is a second door on the same path. The **only** writer of `trackerProgressConfirmedAt` — the precondition for reaching finalization at all (`:3287-3288`) — sets `lastEvent` in the same object literal, and never to `turn_completed`:

```ts
// service.ts:625-636
lastEvent:
  input.request.type === "transition-request"
    ? "tracker-transition"
    : "tracker-state-read",
…
trackerProgressConfirmedAt: recordConfirmedTrackerProgress ? nowIso : (latestRun.trackerProgressConfirmedAt ?? null),
```

Arming the finalization path and disarming the guard are literally the same statement; only a subsequent best-effort channel event can restore `turn_completed`, and the terminal heartbeat then clobbers it again.

**Empirically confirmed.** Same fixture and same three ticks as the new regression, changing only the seeded marker to the production-realistic `"heartbeat"` (probe reverted; tree clean):

```
PROBE status            = retrying
PROBE retryKind         = recovery      ← not "failure"
PROBE lastError         = null          ← discriminated cause discarded
PROBE recovery.kind     = incomplete-turn-dirty-workspace
PROBE deferralCount     = 0
PROBE failureRetryCount = 0             ← never advances
PROBE retryEntry        = {"attempt":2,"dueAt":"2026-03-08T00:00:01.000Z","error":null}
```

Identical with `lastEvent: "turn_started"`. That is exactly the cycle-2 table — `failureRetryCount` pinned at 0, suppression unreachable, `lastError` null — reached through a different door: 3 deferrals → recovery retry → restart → 3 deferrals → … So #611's primary criterion (*"A persistent `unknown` cannot leave a completed worker run pinned indefinitely"*) still fails whenever the issue workspace is dirty, which is the common case for a run that could not finalize.

**Suggested fix** — don't route control flow through a telemetry marker. Once the bound is exhausted the run must take failure accounting regardless of workspace state, and we only reach past the `unknown` block when `exhausted` is true, so at `:3348`:

```ts
const finalizationExhausted = currentTrackerProgress?.state === "unknown";
const recovery = finalizationExhausted
  ? null
  : await this.classifyIncompleteTurnDirtyWorkspace(tenant, runWithTokens, now);
```

Dirty work is not lost — `recovery-quarantined` (`:2673`) still attributes it, and cycle 2 already settled that failure retry/suppression is the intended terminal path for unresolved canonical state.

Please also re-point the regression at the real sequence rather than the seeded ideal: drop the `lastEvent: "turn_completed"` seed (or set `"heartbeat"`) so the test exercises what production actually persists. With the fix above it should pass under `turn_completed`, `heartbeat`, and an unset marker alike — a parameterised case over those three would keep this door shut for good.

---

hojinzs (changes requested):
## Review — cycle 3 of 6

The cycle-2 blocker was fixed exactly as asked, and I verified the new regression genuinely guards it. Blocking on one thing: the guard that fix relies on isn't an invariant, so the same defect is still reachable through the normal successful-exit sequence. One inline P1, ~3 lines plus a test seed change.

### Cycle 2 finding — verification

| Finding | Verdict |
| --- | --- |
| **P1** `lastEvent` write defeats the `turn_completed` guard | ⚠️ **Line correctly deleted, regression genuinely guards it — but the guard it restores does not fire in production. See inline P1.** |

I mutation-tested rather than reading the diff: reintroducing the single deleted line in `deferredRun` fails the new test —

```
× preserves bounded failure retry accounting when the completed workspace is dirty
  → expected { runId: 'run-1', …(35) } to match object { status: 'retrying', …(5) }
```

— and it passes with the line gone. That is a real regression, not a decorative one. The problem is what it asserts against: the test seeds `lastEvent: "turn_completed"`, and the worker's own shutdown path overwrites that marker with `"heartbeat"` before finalization ever runs. Details, both code paths, and the probe are inline.

### Independent verification

Re-ran the full gate on `c6dc01b` rather than trusting the evidence section:

- `pnpm build` — pass
- `pnpm lint` — pass
- `pnpm typecheck` — pass
- `pnpm -r --workspace-concurrency=1 test` — pass, 14 packages; orchestrator 14 files / **320** tests, CLI 39 files / **541** (matches the PR's claimed numbers exactly)
- CI on this SHA: `Test` ✅, `Container Smoke` ✅
- `git diff --check` / worktree clean — confirmed

### Smoke test

Docker Hub image pulls remain 403-blocked by this environment's proxy, so TC-16 is again unreproduced. I read `run-e2e.sh:272-297` instead — the assertions mirror the passing unit test — and treat the reported run as credible but unverified.

The operator surface I could black-box, re-run against the built CLI on this head with a hand-seeded `events.ndjson`:

```
$ node packages/cli/dist/index.js repo logs
[…] run-finalization-deferred acme/platform#1 Finalization deferred 1/3 (tracker-item-missing)
[…] run-finalization-deferred acme/platform#1 Finalization deferred 3/3 (tracker-item-missing) — bound exhausted

$ node packages/cli/dist/index.js repo logs --level warn
[…] run-finalization-deferred acme/platform#1 Finalization deferred 1/3 (tracker-item-missing)
[…] run-finalization-deferred acme/platform#1 Finalization deferred 3/3 (tracker-item-missing) — bound exhausted
```

Count, bound, cause and exhaustion all render and survive the `warn` filter. Cycle 1's P1 stays closed.

### Requirements traceability (#611)

| --- | --- |
| Transient `unknown` stays deferred, later finalizes successfully | ✅ |
| Persistent `unknown` cannot pin a completed run indefinitely | ⚠️ **holds only when `lastEvent === "turn_completed"`; fails on the real exit sequence with a dirty workspace — see inline P1** |
| Each deferral observable through persisted run events | ✅ smoke-tested above |
| Tests fail if arming or a classifier mapping regresses | ✅ |
| Split the combined continuation/finalization regression | ✅ |
| `pnpm lint` / `test` / `typecheck` / `build` | ✅ independently reproduced |

**Over-engineering / unrequested decisions: none found**, in any of the three cycles. `c6dc01b` is a one-line deletion plus one regression — nothing was widened.

### Path to approval

Make exhausted finalization stop depending on the `lastEvent` marker (the 3-line change inline), and re-seed the regression to the marker production actually persists. That is the only blocker; everything else in this PR is ready to land, and this is the last thread I have open on it — I intend to approve on the next push.

---

packages/orchestrator/src/service.ts
moncher-dev: Resolved in `c9ead5a`. Exhausted unknown finalization now bypasses dirty-workspace recovery directly, so failure retry accounting no longer depends on the worker's final telemetry marker. The regression is parameterized across `turn_completed`, production-realistic `heartbeat`, and an unset marker; all variants assert `retryKind: "failure"`, `failureRetryCount: 1`, the canonical-item diagnostic, and `recovery: null`. The focused finalization suite, full workspace gate, and Docker TC-16 pass.

packages/orchestrator/src/service.ts
```diff
@@ -3319,11 +3345,14 @@ export class OrchestratorService {
       };
     }
 
-    const recovery = await this.classifyIncompleteTurnDirtyWorkspace(
-      tenant,
-      runWithTokens,
-      now
-    );
+    const finalizationExhausted = currentTrackerProgress?.state === "unknown";
+    const recovery = finalizationExhausted
```
hojinzs, line 3349: **P2 (follow-up, not blocking) — the bypass works, but "dirty work remains attributable through quarantine handling" is not what actually happens**

The cycle-3 fix is correct and I confirmed it: exhaustion now yields `retryKind: "failure"` with the tracker diagnostic under `heartbeat`, `turn_completed`, and an unset marker alike. #611's primary criterion holds.

What I got wrong in my cycle-3 suggestion — and what carried into the PR description ("dirty work remains attributable through quarantine handling") — is the claim about `recovery-quarantined`. That path is gated on the recovery object this line now forces to `null`:

```ts
// service.ts:2646
if (recovery?.kind === "incomplete-turn-dirty-workspace" && existingWorkspaceRecord) {
```

and `resolveRetryRunRecoveryContext` (`:4143-4146`) short-circuits on the same field, so the restart carries `recovery: null` → no quarantine, and `allowDirtyExistingWorkspace: false`.

**What happens instead**, probed on the `tracker-read-failed` cause (fixture from the new regression + `lastEvent: "heartbeat"`, three ticks to exhaustion, then `nextRetryAt` made due with the outage ongoing so `resolveRetryRestartAction` lands in its `catch → "restart"`; probe reverted, tree clean):

```
PROBE tick3 {"status":"retrying","retryKind":"failure",
             "lastError":"Final tracker state unavailable: tracker read failed: tracker timeout",
             "recovery":null}                                    ← the fix, working

PROBE tick4 threw = Issue workspace repository at …/repository was preserved because it has
                    uncommitted changes: ?? completed-turn-output.txt. …
PROBE tick4 run   = {"status":"failed","lastError":"Superseded by recovered run."}
PROBE tick4 issue = [{"state":"released","failureRetryCount":1,"retryEntry":null,…}]
PROBE dirty file still present = true
```

Two consequences, both bounded:

1. `syncExistingIssueWorkspaceRepository` (`git.ts:595-600`) throws rather than quarantining. `restartRun` rethrows, and the `for (const run of activeRuns)` loop at `:1208-1214` has no `try`, so the remaining reconciliation for that tick is skipped. The daemon loop catches it (`:966-973`) and only writes to stderr — no `run-failed`, no `recovery-quarantined`.
2. The terminal record reads `lastError: "Superseded by recovered run."` — nothing superseded it, and the cause this cycle worked to make truthful is discarded exactly where an operator looks last.

**Why I'm not blocking on it:** no data is lost (the workspace is preserved on disk, which is the point of that error), and it self-heals. The issue goes to `released`, and the next dispatch takes the *caught* candidate path (`:1587-1610`), which records `Worker spawn failed: Issue workspace … has uncommitted changes` with real failure-retry accounting and eventual suppression. So the run is bounded and the operator does get a truthful message — one tick later, through a different door than the description claims. Item 1 is also pre-existing structure in `restartRun`; this PR only routes a new scenario into it.

Please at minimum correct the description and the divergence note so they describe the preserved-workspace error rather than quarantine. If you want the smaller real fix, keeping the classification for workspace handling while decoupling retry accounting from it does both jobs:

```ts
const recovery = await this.classifyIncompleteTurnDirtyWorkspace(tenant, runWithTokens, now);
const finalizationExhausted = currentTrackerProgress?.state === "unknown";
const persistedRetryKind = recovery && !finalizationExhausted ? "recovery" : retryKind;
```

with `failureRetryCount` / suppression / `lastError` keyed on `finalizationExhausted || !recovery` rather than `!recovery`. Your call — documenting the actual behavior is enough for me.

---

packages/control-plane/client/src/routes/issues/$identifier.tsx
```diff
@@ -466,7 +466,7 @@ export function classifyEventTone(event: IssueStatusEvent): EventTone {
     return "success";
   }
 
-  if (text.includes("convergence")) {
+  if (text.includes("convergence") || text.includes("finalization deferred")) {
```
hojinzs, line 469: **nit — this matches the formatter's prose, not the event, so a reword silently drops the warning tone**

`text` is `` `${event.event} ${event.message ?? ""}` `` lowercased, so the event slug `run-finalization-deferred` is already in it. The literal here is `"finalization deferred"` with a space, which only ever matches the message half — i.e. the tone is coupled to the exact wording `formatEventMessage` emits.

The paired test can't catch that, because it hardcodes the message string rather than deriving it:

```ts
// issueDetail.test.tsx:129
message: "Finalization deferred 2/3 (tracker-read-failed)",
```

Change the formatter to `"Finalization bound exhausted…"` and this test still passes while the dashboard quietly renders `default` tone.

Matching the hyphenated slug instead is robust and needs no other change:

```ts
if (text.includes("convergence") || text.includes("finalization-deferred")) {
```

(The existing `"convergence"` match has the same shape but happens to be substring-safe in both halves, so it isn't affected.)

---

packages/orchestrator/src/service.ts
```diff
@@ -3362,9 +3391,10 @@ export class OrchestratorService {
     }
 
     // Determine retry kind: continuation (issue still actionable) vs failure
-    const retryKind = convergenceDetected
-      ? "failure"
-      : await this.classifyRetryKind(tenant, run, trackerDependencies);
+    const retryKind =
+      convergenceDetected || currentTrackerProgress?.state === "unknown"
```
hojinzs, line 3395: **nit — `finalizationExhausted` is named once and then re-derived twice**

The same predicate appears three times: `:3348` (named), here, and `:3506`. All three are load-bearing and all three depend on a non-obvious invariant — reaching this point with `state === "unknown"` implies exhaustion only because the non-exhausted branch returned ~50 lines earlier at `:3325`.

Reusing the binding in all three places makes the invariant readable and keeps the three uses from drifting apart:

```ts
convergenceDetected || finalizationExhausted
```

---

hojinzs (approved):
## Approve — cycle 4

Every P1 from the previous three cycles is genuinely fixed, and I verified the two that mattered rather than taking the thread replies at face value. Remaining findings are one follow-up and two nits; none of them justifies a fifth round.

### Requirements vs. #611

| --- | --- |
| Transient `unknown` defers, then finalizes successfully | ✅ `recovers a transient unknown finalization read and later succeeds` |
| Persistent `unknown` cannot leave a completed run pinned | ✅ verified below, across all three terminal markers |
| Each deferral observable through persisted run events | ✅ verified end-to-end through the built CLI |
| Tests fail if arming regresses or a classifier mapping breaks | ✅ positive arming at `service.test.ts:703`, fail-closed at `:554`; the `classifyCurrentTrackerProgress` stub is gone and all three mappings are exercised for real |
| `lint` / `test` / `typecheck` / `build` pass | ✅ reproduced locally |

No over-engineering or unrequested scope: the changeset is a patch on `@gh-symphony/cli`, the cause discriminator and the bypass both trace to review findings, and the layer split holds — tracker-specific strings stay in the orchestrator, core only gains the event type and the formatter case.

### Verification

**Cycle-3 P1 (the `heartbeat` door) — confirmed closed.** Probed the exhaustion tick on the real successful-exit marker, not the seeded ideal:

```
PROBE tick3 {"status":"retrying","retryKind":"failure",
             "lastError":"Final tracker state unavailable: tracker read failed: tracker timeout",
             "recovery":null}
```

Matches under `turn_completed`, `heartbeat`, and unset — the parameterised regression is pointed at what production actually persists, and the fix no longer routes control flow through a telemetry marker.

**Cycle-1 P1 (empty operator surfaces) — confirmed closed** against built artifacts rather than unit mocks. `formatEventMessage` from `packages/core/dist`:

```
"Finalization deferred 1/3 (tracker-read-failed)"
"Finalization deferred 3/3 (workflow-unavailable) — bound exhausted"
```

and `repo logs` from `packages/cli/dist` over a real `events.ndjson`:

```
--- --level warn ---
[…00:01:00.000Z] run-finalization-deferred acme/repo#1 Finalization deferred 1/3 (tracker-read-failed)
[…00:03:00.000Z] run-finalization-deferred acme/repo#1 Finalization deferred 3/3 (tracker-item-missing) — bound exhausted
--- --level info ---
[…00:00:00.000Z] run-started acme/repo#1
```

Count, bound, cause and exhaustion all render, and the exhausted deferral is no longer filtered out of `--level warn`.

**Full gate, reproduced locally on `c9ead5a`:** `pnpm lint` ✅ · `pnpm build` ✅ · `pnpm test` ✅ (14 packages; orchestrator 322, CLI 541, core 228) · `pnpm typecheck` ✅.

**Smoke test:** done at the two operator surfaces above. TC-16 could not be executed here — the Docker daemon is unavailable in this environment — so the E2E claim rests on your Cycle 6 run. The unit-level probe covers the same branch more directly, so I don't consider that a gap.

### Findings

- **P2 (follow-up)** — `service.ts:3349`. The bypass is right, but "dirty work remains attributable through quarantine handling" isn't what the code does: quarantine is gated on the very `recovery` field this path nulls, so the restart tick throws the preserved-workspace error instead. Bounded and self-healing — the next dispatch takes the *caught* candidate path with real failure accounting — so it's a description/doc correction plus an optional small fix, not a blocker. Details and the probe are inline.
- **nit** — `$identifier.tsx:469`: the tone match keys on the formatter's prose instead of the event slug, and the paired test hardcodes the message so it can't catch a reword.
- **nit** — `service.ts:3395`: `finalizationExhausted` is bound once and re-derived twice.

Approving on the understanding that the P2's description/divergence-note correction lands before merge; the two nits are yours to take or leave.

---

packages/orchestrator/src/service.ts
hojinzs: @moncher-dev Land 시 관찰을 위해 후속 이슈 등록

PR #636, opened 2026-08-26
fix(cli): preserve standalone lock through confirmation

### LINKED ISSUE #607: fix(cli): keep standalone start serialization across confirmation
## Context

PR #601 serializes standalone-project overlap validation and config persistence with the shared config lock. The critical section can now include an interactive overlap confirmation, but `withConfigLock` considers any lock older than 30 seconds stale before checking whether its owner PID is still alive.

## Problem

If an operator waits at the confirmation prompt for more than 30 seconds, a concurrent `project start` can steal the still-live lock and enter the same critical section. This can reopen the concurrent-start race that the serialization is intended to close. A newly cached project whose daemon has not started yet can also appear stopped during the overlap scan.

## Acceptance

- A live lock owner waiting at an interactive confirmation cannot be displaced solely because 30 seconds elapsed.
- After confirmation, overlap state is revalidated under an exclusively held lock before config persistence.
- Add a regression test covering a live, aged lock and concurrent standalone-project starts.

## Origin

Follow-up to #600 / PR #601. This was explicitly classified as non-blocking during approval and is kept out of the approved PR's scope.

moncher-dev:
## TL;DR

- Keep a positively verified live standalone-start config lock exclusively held beyond 30 seconds while an operator waits at confirmation.
- Retain TTL recovery for expired locks whose ownership cannot be verified, and refresh only daemon liveness after confirmation.

## Change-point diagram

```text
project start A ── acquire config lock ── overlap prompt ── confirm ── refresh daemon liveness ── persist
                                      │
project start B ── sees aged verified-live lock ┴─ waits; cannot enter the critical section

unverifiable aged lock ── TTL recovery ── acquire
```

## Start here

- `packages/cli/src/config.ts` — distinguishes verified-live locks from inconclusive ownership and preserves recovery.
- `packages/cli/src/commands/project.ts` — refreshes liveness over the locked overlap set after confirmation.
- `packages/cli/src/config.test.ts` and `packages/cli/src/commands/project.test.ts` — recovery and concurrent-start regressions.

## Risks & rollback

- A lock is retained beyond its TTL only when its PID and process identity both match; unavailable identity retains the prior expiration fallback.
- Reverting this PR restores the former age-first lock behavior.

## Changed files

- `packages/cli/src/config.ts`
- `packages/cli/src/config.test.ts`
- `packages/cli/src/commands/project.ts`
- `packages/cli/src/commands/project.test.ts`
- `.changeset/standalone-lock-confirmation.md`

## Evidence

- `pnpm --filter @gh-symphony/cli test -- --run src/config.test.ts src/commands/project.test.ts` — 21 passed
- `pnpm test` — pass
- `pnpm lint` — pass
- `pnpm typecheck` — pass
- `pnpm build` — pass
- `pnpm exec changeset status --since=origin/main` — `@gh-symphony/cli` patch only
- Docker E2E not run: this changes local CLI lock handling only; it does not alter orchestrator dispatch, worker lifecycle, tracker adapters, or status APIs.

## Issues — Closed #607

Fixes #607

- [ ] Confirm an operator can wait at an interactive standalone overlap confirmation while a second start remains blocked.
- [ ] Confirm expired locks with unverifiable ownership are recovered rather than wedging subsequent CLI commands.

### DIFF
--- a/packages/cli/src/commands/project.test.ts
+++ b/packages/cli/src/commands/project.test.ts
@@ -1,7 +1,15 @@
-import { mkdtemp, mkdir, writeFile } from "node:fs/promises";
+import { mkdtemp, mkdir, readFile, writeFile } from "node:fs/promises";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 import { describe, expect, it, vi } from "vitest";
+
+const { confirmMock } = vi.hoisted(() => ({ confirmMock: vi.fn() }));
+
+vi.mock("@clack/prompts", async (importOriginal) => {
+  const actual = await importOriginal<typeof import("@clack/prompts")>();
+  return { ...actual, confirm: confirmMock };
+});
+
 import { loadProjectConfig } from "../config.js";
 import { deriveStandaloneProject, standaloneProjectId } from "./project.js";
 import projectCommand from "./project.js";
@@ -157,6 +165,64 @@ describe("deriveStandaloneProject", () => {
     expect(
       results.filter((result) => result.status === "rejected")
     ).toHaveLength(1);
+  });
+
+  it("keeps an aged live lock through overlap confirmation", async () => {
+    const configDir = await mkdtemp(join(tmpdir(), "cli-standalone-config-"));
+    const existing = await mkdtemp(join(tmpdir(), "cli-standalone-project-"));
+    const first = await mkdtemp(join(tmpdir(), "cli-standalone-project-"));
+    const second = await mkdtemp(join(tmpdir(), "cli-standalone-project-"));
+    await Promise.all(
+      [existing, first, second].map((projectDir) =>
+        writeFile(join(projectDir, "WORKFLOW.md"), workflow, "utf8")
+      )
+    );
+    await deriveStandaloneProject(existing, { configDir });
+
+    const originalIsTTY = Object.getOwnPropertyDescriptor(
+      process.stdin,
+      "isTTY"
+    );
+    Object.defineProperty(process.stdin, "isTTY", {
+      configurable: true,
+      value: true,
+    });
+    let resolveConfirmation!: (value: boolean) => void;
+    const confirmation = new Promise<boolean>((resolve) => {
+      resolveConfirmation = resolve;
+    });
+    confirmMock
+      .mockImplementationOnce(() => confirmation)
+      .mockResolvedValue(false);
+
+    try {
+      const firstStart = deriveStandaloneProject(first, { configDir });
+      await vi.waitFor(() => expect(confirmMock).toHaveBeenCalledTimes(1));
+
+      const lockPath = join(configDir, ".config.lock");
+      const lock = JSON.parse(await readFile(lockPath, "utf8")) as {
+        startedAt: string;
+      };
+      lock.startedAt = new Date(Date.now() - 31_000).toISOString();
+      await writeFile(lockPath, JSON.stringify(lock), "utf8");
+
+      const secondStart = deriveStandaloneProject(second, { configDir });
+      await new Promise((resolve) => setTimeout(resolve, 75));
+      expect(confirmMock).toHaveBeenCalledTimes(1);
+
+      resolveConfirmation(true);
+      await expect(firstStart).resolves.toBeDefined();
+      await expect(secondStart).rejects.toThrow(
+        "Standalone project start cancelled"
+      );
+    } finally {
+      confirmMock.mockReset();
+      if (originalIsTTY) {
+        Object.defineProperty(process.stdin, "isTTY", originalIsTTY);
+      } else {
+        delete (process.stdin as { isTTY?: boolean }).isTTY;
+      }
+    }
   });
 
   it("preserves Linear label filters and normalizes overlapping states", async () => {
--- a/packages/cli/src/commands/project.ts
+++ b/packages/cli/src/commands/project.ts
@@ -109,6 +109,22 @@ export async function deriveStandaloneProject(
           "Standalone project start cancelled because tracker mappings overlap."
         );
       }
+
+      // Confirmation can take an arbitrary amount of time. The config lock
+      // remains held while waiting, but re-check before persisting so the
+      // decision is based on the state that exists at commit time.
+      const confirmedOverlap = await findOverlappingProjects(
+        options.configDir,
+        config
+      );
+      const runningAfterConfirmation = confirmedOverlap.filter(
+        (entry) => entry.running
+      );
+      if (runningAfterConfirmation.length > 0) {
+        throw new Error(
+          `Tracker mapping overlaps running project(s): ${describe(runningAfterConfirmation)}. Stop them or make the mappings disjoint with tracker.pickup_labels.`
+        );
+      }
     }
 
     await saveProjectConfigWithinLock(options.configDir, projectId, config);
--- a/packages/cli/src/commands/project.ts
+++ b/packages/cli/src/commands/project.ts
@@ -110,15 +110,12 @@ export async function deriveStandaloneProject(
         );
       }
 
-      // Confirmation can take an arbitrary amount of time. The config lock
-      // remains held while waiting, but re-check before persisting so the
-      // decision is based on the state that exists at commit time.
-      const confirmedOverlap = await findOverlappingProjects(
+      // Project mappings cannot change while the config lock is held, but a
+      // daemon can start while the operator is deciding. Refresh only that
+      // liveness state before persisting the confirmed configuration.
+      const runningAfterConfirmation = await runningOverlaps(
         options.configDir,
-        config
-      );
-      const runningAfterConfirmation = confirmedOverlap.filter(
-        (entry) => entry.running
+        overlap
       );
       if (runningAfterConfirmation.length > 0) {
         throw new Error(
@@ -199,6 +196,7 @@ type OverlappingProject = {
   projectId: string;
   label: string;
   running: boolean;
+  workspaceDir: string;
 };
 
 /**
@@ -230,9 +228,29 @@ async function findOverlappingProjects(
       projectId: existing.projectId,
       label: existing.projectDir ?? existing.projectId,
       running: liveness.running,
+      workspaceDir: existing.repositoryDir ?? existing.workspaceDir,
     });
   }
   return overlaps;
+}
+
+async function runningOverlaps(
+  configDir: string,
+  overlaps: OverlappingProject[]
+): Promise<OverlappingProject[]> {
+  const refreshed = await Promise.all(
+    overlaps.map(async (overlap) => ({
+      ...overlap,
+      running: (
+        await resolveDaemonLiveness({
+          configDir,
+          projectId: overlap.projectId,
+          workspaceDir: overlap.workspaceDir,
+        })
+      ).running,
+    }))
+  );
+  return refreshed.filter((overlap) => overlap.running);
 }
 
 function mappingFor(config: CliProjectConfig): Mapping {
--- a/packages/cli/src/config.test.ts
+++ b/packages/cli/src/config.test.ts
@@ -19,6 +19,7 @@ import {
   saveGlobalConfig,
   saveProjectConfig,
   updateGlobalConfig,
+  withConfigLock,
 } from "./config.js";
 
 const originalCwd = process.cwd();
@@ -191,6 +192,48 @@ describe("config persistence", () => {
       projects: ["project-1"],
     });
     expect(await readdir(configDir)).toEqual(["config.json"]);
+  });
+
+  it("reclaims an aged lock whose live owner cannot be identified", async () => {
+    const configDir = await mkdtemp(join(tmpdir(), "cli-config-lock-"));
+    const lockPath = join(configDir, ".config.lock");
+    await writeFile(
+      lockPath,
+      JSON.stringify({
+        ownerToken: "unverifiable-owner",
+        pid: process.pid,
+        startedAt: new Date(Date.now() - 31_000).toISOString(),
+        processIdentity: null,
+      }),
+      "utf8"
+    );
+    const expired = new Date(Date.now() - 31_000);
+    await utimes(lockPath, expired, expired);
+
+    await expect(
+      withConfigLock(configDir, async () => "acquired")
+    ).resolves.toBe("acquired");
+  });
+
+  it("reclaims an aged lock whose owner process is gone", async () => {
+    const configDir = await mkdtemp(join(tmpdir(), "cli-config-lock-"));
+    const lockPath = join(configDir, ".config.lock");
+    await writeFile(
+      lockPath,
+      JSON.stringify({
+        ownerToken: "dead-owner",
+        pid: 999_999_999,
+        startedAt: new Date(Date.now() - 31_000).toISOString(),
+        processIdentity: null,
+      }),
+      "utf8"
+    );
+    const expired = new Date(Date.now() - 31_000);
+    await utimes(lockPath, expired, expired);
+
+    await expect(
+      withConfigLock(configDir, async () => "acquired")
+    ).resolves.toBe("acquired");
   });
 
   it("reads structured and legacy daemon PID records", () => {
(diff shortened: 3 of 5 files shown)


packages/cli/src/config.ts
```diff
@@ -403,8 +403,7 @@ async function isStaleConfigLock(path: string): Promise<boolean> {
     ) {
       return isExpiredConfigLockFile(path);
     }
-    const ageMs = Math.abs(Date.now() - Date.parse(record.startedAt));
-    if (!Number.isFinite(ageMs) || ageMs > CONFIG_LOCK_TTL_MS) {
+    if (!Number.isFinite(Date.parse(record.startedAt))) {
```
hojinzs, line 406: **P1 — removing the TTL leaves no reclamation path when ownership cannot be *disproved*, so a recycled PID wedges the config directory permanently.**

The TTL was the only backstop for the cases where the remaining checks are *inconclusive* rather than negative. After this change, two reachable paths return `false` (not stale) forever:

1. **`processIdentity` is null/absent.** `getProcessIdentity()` shells out to `ps` and returns `null` whenever `ps` is missing or exits non-zero (`packages/orchestrator/src/lock.ts:412`). On such a host the lock record is written with `processIdentity: null`, and the guard below short-circuits on the falsy `record.processIdentity` — so the identity check can never mark it stale. Same for any lock file written before `processIdentity` existed.
2. **`process.kill(pid, 0)` throws `EPERM`.** If the PID was recycled by a process owned by another user, the catch only returns `true` for `ESRCH`, so `EPERM` → not stale.

In both cases the lock survives its owner's death indefinitely, and every `withConfigLock` caller (`saveProjectConfig`, `project start`, `project stop`, …) fails forever with `Timed out waiting for config lock` and no hint that `.config.lock` needs manual removal.

Reproduced on this head vs. the base commit — an aged lock naming a live-but-unrelated PID with `processIdentity: null`:

| | outcome | elapsed |
|---|---|---|
| `0e6139a` (base) | lock reclaimed, acquired | 11 ms |
| `c2cf6d6` (this PR) | `Timed out waiting for config lock`, lock file still present | 42 983 ms, and on every subsequent attempt |

Suggested shape: skip the TTL only when liveness is *positively confirmed*, and fall back to the age check when it is inconclusive — roughly, keep the identity comparison as the authority, and when `!record.processIdentity || !actualIdentity` (or `kill` threw something other than `ESRCH`), defer to `isExpiredConfigLockFile(path)` instead of returning `false`. That preserves the issue's requirement (a verified-live owner is never displaced at 30 s) without giving up self-healing.

---

packages/cli/src/commands/project.test.ts
```diff
@@ -159,6 +167,64 @@ describe("deriveStandaloneProject", () => {
...
+      await vi.waitFor(() => expect(confirmMock).toHaveBeenCalledTimes(1));
+
+      const lockPath = join(configDir, ".config.lock");
+      const lock = JSON.parse(await readFile(lockPath, "utf8")) as {
+        startedAt: string;
+      };
+      lock.startedAt = new Date(Date.now() - 31_000).toISOString();
+      await writeFile(lockPath, JSON.stringify(lock), "utf8");
+
+      const secondStart = deriveStandaloneProject(second, { configDir });
```
hojinzs, line 209: **P2 — this test exercises the one case that was already trivially safe, and none of the cases whose semantics the diff actually changed.**

Both "starts" run in the same process, so the lock record's `pid` is `process.pid`: `process.kill` always succeeds and `getProcessIdentity(process.pid)` always matches the value written at acquisition. The test therefore only covers *identity positively verified* — the branch where retaining the lock is uncontroversial. The branches this PR changed behaviour for are the inconclusive ones (`processIdentity: null`, `EPERM`, `ps` unavailable), and the reclamation path that used to be TTL-driven; neither is covered here or in `config.test.ts` (`config.test.ts:177` only covers the *partial/unparseable file* path, which still goes through `isExpiredConfigLockFile`).

Two cases worth adding, both cheap since they only need a hand-written lock file:

- an aged lock whose owner PID is **dead** is still reclaimed (guards against the TTL removal silently disabling reclamation altogether);
- an aged lock whose owner PID is **alive but unverifiable** (`processIdentity: null`) — this is the P1 above, and a test here would have caught it.

`await new Promise((resolve) => setTimeout(resolve, 75))` as the proof that the second start has not entered the critical section is also timing-dependent; `vi.waitFor` on a positive signal plus a `confirmMock` call-count assertion would be sturdier under a loaded CI runner. nit.

---

packages/cli/src/commands/project.ts
```diff
@@ -109,6 +109,22 @@ export async function deriveStandaloneProject(
           "Standalone project start cancelled because tracker mappings overlap."
         );
       }
+
+      // Confirmation can take an arbitrary amount of time. The config lock
+      // remains held while waiting, but re-check before persisting so the
+      // decision is based on the state that exists at commit time.
+      const confirmedOverlap = await findOverlappingProjects(
```
hojinzs: **P3 — the re-scan re-reads state that provably cannot have changed, which obscures what is actually being re-validated.**

Every writer of `projects/<id>/config.json` goes through `withConfigLock` (`config.ts:160`, `config.ts:201`, and `saveProjectConfigWithinLock` below), so while this lock is held the *set* of standalone projects and their tracker mappings are invariant. The only thing that can change across the confirmation prompt is daemon **liveness**, since `resolveDaemonLiveness` reads PID/state files that the config lock does not guard — which is exactly the failure mode issue #607 describes.

Re-running the full `findOverlappingProjects` (re-`readdir`, re-parse and re-match every project config) and then discarding everything but `.running` works, but it reads as "the overlap set might have changed", which is a stronger and less accurate claim than the one the comment above makes. Re-checking liveness over the already-computed `overlap` array would say the same thing in one line and document the invariant:

```ts
const runningAfterConfirmation = (
  await Promise.all(overlap.map(async (entry) => ({ entry, running: await isRunning(entry) })))
).filter(({ running }) => running).map(({ entry }) => entry);
```

Not blocking either way — the current form is correct, just louder than it needs to be.

---

hojinzs (changes requested):
## Review — round 1 of 6

The scope is tight and matches issue #607: no over-engineering, no unrequested decisions, changeset present and correctly scoped to `@gh-symphony/cli` as a patch. One blocking correctness issue in the lock change.

### Requirements vs. issue #607

|---|---|
| A live lock owner at an interactive confirmation cannot be displaced solely because 30 s elapsed | ✅ Met |
| After confirmation, overlap state is revalidated under an exclusively held lock before persistence | ✅ Met (`project.ts:116-127`) |
| Regression test covering a live, aged lock and concurrent standalone starts | ⚠️ Met in letter; the test only exercises the same-process case — see inline |

The issue's Problem section also notes *"a newly cached project whose daemon has not started yet can also appear stopped during the overlap scan."* The post-confirmation re-scan narrows that window but cannot close it — a daemon that comes up between the re-scan and `saveProjectConfigWithinLock` is still missed. That isn't in Acceptance and I'm not asking for it here; noting it so it isn't assumed fixed.

### Findings

- **P1** — `config.ts:406`: dropping the age check removes the only reclamation path for locks whose ownership is *inconclusive* rather than disproved (`processIdentity: null` when `ps` is unavailable; `EPERM` from a recycled PID owned by another user). A recycled PID then wedges the config directory permanently. Verified against the base commit: base reclaims the lock in 11 ms, this head times out after ~43 s and leaves the lock in place, on every attempt. Details and a suggested shape inline.
- **P2** — `project.test.ts:209`: the new test covers the branch that was already safe and none of the branches whose semantics changed. Inline.
- **P2** — `config.ts:32-33` (outside the diff, so noting it here): `CONFIG_LOCK_RETRY_LIMIT = ceil(CONFIG_LOCK_TTL_MS / CONFIG_LOCK_RETRY_MS) + 50` was sized on the premise "wait slightly past the TTL, then take the expired lock." That premise no longer holds for live locks, so the waiter's budget is now arbitrary: measured at ~43 s, after which a concurrent command hard-fails with `Timed out waiting for config lock at "…"`. Failing is the right call over racing, but the PR's post-merge checklist says the second start *"remains blocked"* — it doesn't, it dies after ~43 s. Worth either decoupling the waiter budget from the TTL or naming the holding PID in the timeout message, so an operator can distinguish "someone is sitting at a prompt" from "the lock is wedged" (which, per P1, is currently unrecoverable without deleting `.config.lock` by hand).
- **P3** — `project.ts:116`: the re-scan re-reads config state that cannot change under the held lock; only liveness can. Inline. Non-blocking.

### Smoke test

No API or UI surface in this change and no preview environment applies, so this was exercised locally against the PR head:

- `pnpm --filter @gh-symphony/cli test -- --run src/commands/project.test.ts src/config.test.ts` — 19 passed (2 files), including the new regression test.
- Black-box repro of the lock path through the public `withConfigLock` entrypoint, run against both `c2cf6d6` and the base `0e6139a` — this is the P1 evidence above.

I agree with skipping the Docker E2E suite: the change touches local CLI lock handling only and does not reach orchestrator dispatch, worker lifecycle, tracker adapters, or the status APIs.

### Verdict

Requesting changes on the P1 only. Everything else is advisory — if you disagree with P2/P3, a reasoned reply on the thread is a fine resolution and I won't re-raise them.

---

packages/cli/src/config.ts
moncher-dev: Resolved in 7c5205a. `isStaleConfigLock` now keeps a lock beyond the TTL only when both the recorded and current process identities positively match. `ESRCH` still reclaims immediately; missing identity, `ps` failure, and non-`ESRCH` liveness errors fall back to the lock-file TTL, preserving recovery from orphaned or recycled-PID locks. Added an aged, live-but-unverifiable-owner regression test.

packages/cli/src/commands/project.test.ts
moncher-dev: Addressed in 7c5205a. The original test remains the end-to-end acceptance regression for a verified-live owner waiting at confirmation; focused `config.test.ts` coverage now exercises the changed lock semantics directly: an aged live owner with `processIdentity: null` is reclaimed, and an aged lock whose PID is gone is reclaimed. This covers the inconclusive and dead-owner recovery paths without relying on platform-specific `ps` failures or `EPERM` simulation.

packages/cli/src/commands/project.ts
moncher-dev: Resolved in 7c5205a. Post-confirmation validation now refreshes daemon liveness over the already-computed overlap set rather than re-reading project configs and mappings. The comment now explicitly states the config-lock invariant.

packages/cli/src/config.test.ts
```diff
@@ -193,6 +194,48 @@ describe("config persistence", () => {
...
+      JSON.stringify({
+        ownerToken: "dead-owner",
+        pid: 999_999_999,
+        startedAt: new Date(Date.now() - 31_000).toISOString(),
+        processIdentity: null,
+      }),
+      "utf8"
+    );
+    const expired = new Date(Date.now() - 31_000);
+    await utimes(lockPath, expired, expired);
```
hojinzs, line 232: **nit** — this test ages the lock file, so it passes through the TTL fallback whether or not the `ESRCH` branch still works.

`utimes(lockPath, -31s)` on line 231-232 means `isExpiredConfigLockFile(path)` returns `true` regardless. If `processIsGone` were accidentally inverted or dropped, `processIsGone || (await isExpiredConfigLockFile(path))` would still be `true` and this test would still pass — so it asserts the recovery *outcome* but not the dead-owner *path* that gives it its name. Dropping the two `utimes` lines (keeping `startedAt` fresh too) would pin the `ESRCH` branch specifically, and the companion `processIdentity: null` test above already covers the TTL fallback.

I verified the branch does work, so this is a durability point about the test rather than a defect: a fresh (non-aged) lock naming a PID that has already exited is reclaimed in **11 ms**, which can only be the `ESRCH` path.

For the record, `process.kill(999_999_999, 0)` does throw `ESRCH` on Linux (I checked, since an out-of-range PID could plausibly have produced a different errno) — so the PID choice itself is fine.

---

hojinzs (approved):

The P1 from round 1 is genuinely fixed, not just described as fixed. Verified by re-running the same black-box repro against this head. Approving.

| # | Finding | Status |
|---|---|---|
| P1 | TTL removal left no reclamation path for *inconclusive* ownership | ✅ **Fixed and verified** — `config.ts:412` / `config.ts:421-431` |
| P2 | New test covered only the already-safe branch | ✅ Addressed — `config.test.ts:197,218` cover the two changed branches; one nit inline |
| P2 | `CONFIG_LOCK_RETRY_LIMIT` sized on a premise that no longer holds | ⚠️ Not addressed — advisory, see below; **not blocking** |
| P3 | Re-scan re-read state that cannot change under the lock | ✅ Addressed — `runningOverlaps()` now refreshes liveness over the already-computed `overlap` set |

The `isStaleConfigLock` shape now matches what I suggested: identity comparison is the authority when both sides are present, and every inconclusive outcome (`processIdentity: null`, `ps` unavailable, non-`ESRCH` liveness error) defers to `isExpiredConfigLockFile(path)` rather than returning "not stale" forever.

### Smoke test

No API or UI surface here and no preview environment applies, so this was exercised locally against the PR head (`7c5205a`) through the public `withConfigLock` entrypoint — the same harness used for the round-1 P1 evidence:

| Scenario | Expected | Result |
|---|---|---|
| S1 — aged lock, live **unrelated** PID, `processIdentity: null` | reclaimed | ✅ **209 ms** (was: timeout at ~43 s on `c2cf6d6`) |
| S2 — aged lock, live unrelated PID, **mismatched** identity | reclaimed | ✅ 17 ms |
| S3 — **fresh** lock, owner PID already exited | reclaimed via `ESRCH` | ✅ 11 ms |
| S4 — aged lock owned by **this verified-live** process | **not** reclaimed | ✅ held; waiter timed out at 44.9 s, lock file intact |

S1 is the exact case that regressed in round 1; it now behaves like the base commit. S4 confirms the fix did not undo the issue's actual requirement — a verified-live owner still cannot be displaced at 30 s.

### Requirements vs. issue #607

|---|---|
| A live lock owner at an interactive confirmation cannot be displaced solely because 30 s elapsed | ✅ Met (S4) |
| After confirmation, overlap state is revalidated under an exclusively held lock before persistence | ✅ Met (`project.ts:113-127`) |
| Regression test covering a live, aged lock and concurrent standalone starts | ✅ Met (`project.test.ts:170`, plus the two focused `config.test.ts` cases) |

Scope is still tight — no over-engineering, no unrequested decisions, changeset present and correctly scoped to `@gh-symphony/cli` as a patch.

### Local validation

`pnpm lint` · `pnpm typecheck` · `pnpm build` — pass. `pnpm test` — pass, all 14 packages (`tracker-file` fails to collect on a clean checkout until `pnpm build` has run; pre-existing ordering quirk, unrelated to this PR). `prettier --check` on all five changed files — clean. CI on `7c5205a`: Test ✅, Container Smoke ✅.

I agree with skipping the Docker E2E suite: this touches local CLI lock handling only and does not reach orchestrator dispatch, worker lifecycle, tracker adapters, or the status APIs.

### Non-blocking, for a follow-up issue rather than this PR

Two loose ends I am explicitly **not** holding the PR for — both are outside #607's Acceptance and fixing either here would widen the scope:

1. **Waiter budget and timeout message.** `CONFIG_LOCK_RETRY_LIMIT` is still derived from `CONFIG_LOCK_TTL_MS`, a premise that no longer holds now that a verified-live lock outlives its TTL. Measured at 44.9 s (S4), after which the second command hard-fails with `Timed out waiting for config lock at "…"`. Failing beats racing, but the message names only the path — naming the holding PID would let an operator tell "someone is at a prompt" from "the lock is wedged". Worth its own issue.
2. **PR body accuracy.** The post-merge checklist says a second start *"remains blocked"*. Per S4 it does not remain blocked — it dies after ~45 s. Worth rewording so the human validator checks for the right thing.

### Verdict

**Approve.** The blocking issue is resolved and independently verified; everything remaining is advisory. The inline nit needs no action before merge.

---

PR #633, opened 2026-08-24
fix(cli): expose standalone project start options

moncher-dev:
## Issues — Closed #625

- Fixes #625

## TL;DR

Standalone project orchestration is now visible and typo-safe: root help lists `project` lifecycle commands, and `project start --help` declares every supported runtime option, including the assignment safety gate.

## Change-point diagram

```text
gh-symphony project start
        │
        ▼
declared Commander options
   ├── complete --help output
   ├── typo/unknown-option rejection
   └── normalized forwarding ──> shared start parser ──> unchanged runtime
```

## Start here

1. `packages/cli/src/index.ts` declares and forwards standalone start options.
2. `packages/cli/src/commands/help.ts` adds the standalone root-help section.
3. `packages/cli/src/index.test.ts` covers option discovery and typo rejection.
4. GitHub and Linear adapter tests cover the revised legacy-setting guidance.

## Changes

- Declare `--daemon`, `--once`, `--assigned-only`, `--bind-all`, `--http`, `--web`, `--log-level`, and `--project-dir` on `project start`.
- Reject unknown standalone start options instead of silently forwarding misspellings.
- Add `project list|start|status|stop` to grouped top-level help.
- Point legacy `tracker.settings.assignedOnly` warnings to both valid CLI invocations.
- Update root/package CLI documentation and the architecture layer mapping.
- Add the required `@gh-symphony/cli` patch changeset.

## Evidence

- `pnpm lint` — passed across all workspace packages.
- `pnpm test` — passed across all workspace packages.
- `pnpm typecheck` — passed across all workspace packages.
- `pnpm build` — passed across all workspace packages.
- Focused CLI/tracker run — 177 tests passed and 2 root-help snapshots updated.
- Docker E2E: N/A; this changes CLI declaration/help and warning copy without changing dispatch, worker lifecycle, tracker filtering, or status APIs.

## Risks & rollback

- Risk is limited to Commander registration and help text. Declared values are reconstructed into the existing shared start parser, preserving runtime semantics.
- Roll back the two branch commits; there is no persisted configuration or runtime-data migration.

## Changed files

- `.changeset/quiet-project-help.md`
- `packages/cli/src/index.ts`, grouped help, tests, and snapshots
- GitHub/Linear tracker warning copy and tests
- `README.md`, `packages/cli/README.md`, and `docs/architecture.md`

## Post-merge / human validation

- [ ] Confirm packaged `gh-symphony --help` shows the standalone project section.
[...]

PR #620, opened 2026-08-23
fix: align blocker semantics with Symphony spec

moncher-dev:
## TL;DR

- Derive Linear `blockedBy` from inverse `blocks` relations instead of source-side relations.
- Enable blocker gating for the built-in `Todo` default and derive omitted custom-workflow gating from the first configured active state, while preserving explicit `[]` as a documented opt-out.
- Keep planning disabled unless explicitly configured, so blocker gating does not change execution phases.

## Change-point diagram

```text
Linear inverseRelations(type: blocks)
                │
                ▼
       TrackedIssue.blockedBy
                │
                ▼
active_states[0] + unresolved blocker ──> dispatch suppressed
                │
                └─ planning_states remains independent (default: [])
```

## Start here

- `packages/tracker-linear/src/orchestrator-adapter.ts` — GraphQL inverse relation query and blocker normalization.
- `packages/core/src/workflow/lifecycle.ts` and `parser.ts` — built-in `Todo` default, first-active-state parser derivation, and independent planning fallback.
- `packages/cli/src/mapping/smart-defaults.ts` — shared generated-workflow first-active-state inference.
- `packages/tracker-linear/src/tracker-linear.test.ts` — inverse/source direction regression TC.
- `packages/core/src/workflow-loader.test.ts` — omitted custom-state blocker default and planning-isolation regression TC.

## Changes

- Query `inverseRelations { nodes { type issue { ... } } }` and map only inverse relations with `type: blocks`.
- Default core blocker checks to `Todo`; workflows with custom active states derive omitted blocker checks from their first configured active state.
- Preserve explicit `blocker_check_states: []` as an intentional repository-level spec divergence.
- Keep omitted blocker settings independent from planning/human-review; Linear init and examples retain `planning_states: []`.
- Update generated workflow schema guidance, both CLI-facing READMEs, configuration reference, Linear example, and the spec-gap report.
- Add the required `@gh-symphony/cli` patch changeset.

## Evidence

- `pnpm lint` — pass across all public workspace packages on `d81b56b`.
- `pnpm test` — pass across all public workspace packages on `d81b56b`.
- `pnpm typecheck` — pass across all public workspace packages on `d81b56b`.
- `pnpm build` — pass across all public workspace packages on `d81b56b`.
- Focused Cycle 3 regressions — core workflow loader 48 tests and CLI workflow init 37 tests passed.
[...]

PR #637, opened 2026-08-26
fix(linear): pass MCP auth credentials

### LINKED ISSUE #634: Linear tracker: workers cannot write to the tracker — MCP server never receives auth, and the LINEAR_API_KEY path sends a malformed header (regression of #314)
## Summary

On a Linear-tracked standalone project, the built-in `linear_graphql` tool fails on every call, so workers cannot transition issues, write workpads, or post PR links. Because the issue stays in an active state, the orchestrator re-dispatches it forever — I watched one issue burn four attempts producing nothing.

Two independent defects cause this. **Fixing either one alone does not restore the tool.** They are separable if you prefer two issues; I filed them together because the symptom, the diagnosis and the regression history are shared.

This is a regression against #314 ("Phase 2: worker linear_graphql tool and runtime auth bridge"), whose acceptance criteria included:

> - Runtime attaches the auth header without exposing the token to the model.
> - MVP supports `LINEAR_API_KEY` fallback.
> - Worker can perform status transitions, workpad/comment writes, and PR-link comments through the tool.

None of the three hold on 0.10.1. The failure was masked until yesterday: #613 meant `dist/mcp-server.js` was never published, so the servers could not start at all. Once #613 was fixed in 0.10.1 and the servers finally started, the defects below surfaced.

## Defect A — the MCP server entry declares no auth env, and codex passes children only the declared env

The runtime→worker→codex chain is intact. Measured on a live run (`ps -E`):

```
worker-entry.js  (PID 31420)   LINEAR_API_KEY=…  LINEAR_AUTHORIZATION=…  LINEAR_GRAPHQL_URL=…
codex app-server (PID 31458)   LINEAR_API_KEY=…  LINEAR_AUTHORIZATION=…  LINEAR_GRAPHQL_TOOL_NAME=…
```

`packages/worker/src/index.ts:571-572` and `packages/runtime-codex/src/runtime.ts:598-613` both do their part.

The break is one step further down. The entry declares only the URL:

```ts
// createLinearGraphQLMcpServerEntry
return {
  command: "node",
  args: [resolveLinearGraphQLMcpServerEntryPoint(), "--server", "linear"],
  env: { LINEAR_GRAPHQL_URL: linearGraphqlUrl },   // no LINEAR_API_KEY / LINEAR_AUTHORIZATION
};
```

while the server reads them from its own process env:

```ts
// packages/tool-linear-graphql/src/mcp-server.ts:101-103
apiKey: process.env.LINEAR_API_KEY,
authorizationHeader: process.env.LINEAR_AUTHORIZATION,
```

codex spawns MCP servers with **only** the declared env — children do not inherit. Two Linear MCP servers under the same codex parent (31458), one built-in and one I declared myself in the project `.mcp.json`:

```
PID 31518  built-in linear_graphql   LINEAR_GRAPHQL_URL          → auth missing, every call fails
PID 31519  my sidecar  linear_api    LINEAR_AUTHORIZATION        → works
```

The worker's own report from a failing run:

> 블로커: 세 번째 우선 시도에서도 Linear 런타임 인증이 설정되지 않았습니다. 이슈 요구사항 조회, 워크패드 기록 및 `Blocked` 상태 전환을 수행할 수 없습니다.
> ("Linear runtime auth is still not configured on the third attempt. Cannot query issue requirements, write the workpad, or transition to `Blocked`.")

This looks like fallout from extracting the tool into a separate MCP server process: the server still reads `process.env` as if it were in-process, but the spawn contract changed.

**Fix**: add the resolved credentials to the entry's `env`, e.g.

```ts
env: {
  LINEAR_GRAPHQL_URL: linearGraphqlUrl,
  ...(options.linearAuthorization ? { LINEAR_AUTHORIZATION: options.linearAuthorization } : {}),
  ...(options.linearApiKey ? { LINEAR_API_KEY: options.linearApiKey } : {}),
},
```

`createLinearGraphQLToolDefinition` / `createSymphonyMcpServers` would need to thread the values through from `symphonyTokenEnv`, which already carries them.

## Defect B — `LINEAR_API_KEY` is sent as a Bearer token, which Linear rejects

Even with the key delivered, the API-key path 400s:

```ts
// packages/tool-linear-graphql/src/tool.ts
if (config.apiKey) return `Bearer ${config.apiKey}`;
```

Reproduction — same server, same key, only the env var differs:

```
LINEAR_API_KEY=lin_api_…      → {"errors":[{"message":"It looks like you're trying to use an API key
                                   as a Bearer token. Remove the Bearer prefix from the Authorization
                                   header.","extensions":{"type":"invalid…"}}]}   HTTP 400

LINEAR_AUTHORIZATION=lin_api_… → {"data":{"issue":{"identifier":"DEV-2","state":{"name":"Backlog"}}}}
```

Linear personal API keys go into `Authorization` verbatim; `Bearer` is for OAuth access tokens. So the documented `LINEAR_API_KEY` fallback from #314 cannot work as written, and only the raw-header escape hatch functions.

**Fix**: send the key unprefixed, or detect the `lin_api_` prefix and skip `Bearer`. A test asserting the exact header value for an API key would pin this.

## Impact

With both defects present, a Linear-tracked worker has no way to write to the tracker — the MCP tool is the only sanctioned path (spec §11.5 puts ticket writes on the agent, and `requestState` is not implemented for the Linear adapter, so the worker `tracker-state` API answers `tracker_state_requests_unsupported`). The issue therefore never leaves its active state and the orchestrator re-dispatches it indefinitely.

## Workaround (confirmed working)

A project-level `.mcp.json` registering the same bundled server under a **different name** with auth declared explicitly. It must be a different name: `composeMcpServers` merges `{...layers, ...builtins}`, so built-ins win and `linear_graphql` cannot be overridden.

```json
{
  "mcpServers": {
    "linear_api": {
      "command": "node",
      "args": ["<npm root -g>/@gh-symphony/cli/dist/mcp-server.js", "--server", "linear"],
      "env": { "LINEAR_AUTHORIZATION": "$LINEAR_AUTHORIZATION" }
    }
  }
}
```

with `LINEAR_AUTHORIZATION` (raw key, no `Bearer`) in the project `.env`. Note `.mcp.json` env values must be `$VAR` references — literals are rejected, which is a good design.

After applying this, the same issue that had failed four dispatches completed on attempt 1 with no retries: `Todo → In Progress` (+ workpad comment) → PR opened → `In Review`, in 4m10s.

## Environment

- OS: macOS 15 (Darwin 25.5.0), Node.js v24.18.0
- `@gh-symphony/cli` 0.10.1 (global install)
- Standalone project folder, Linear tracker, `codex-app-server` runtime

moncher-dev:
## TL;DR

Built-in Linear GraphQL MCP servers now receive resolved credentials across the orchestrator-to-worker and runtime-to-MCP boundaries. `LINEAR_API_KEY` is sent as the raw Authorization value, never with an invalid `Bearer` prefix.

## Change-point diagram

`LINEAR_API_KEY` / `LINEAR_AUTHORIZATION` → `linearTrackerAdapter.buildWorkerEnvironment` → worker runtime configuration → `linear_graphql` MCP server environment → normalized raw `Authorization` header

## Start here

- `packages/tracker-linear/src/orchestrator-adapter.ts` explicitly forwards normalized Linear credentials through the worker environment, avoiding the inherited-environment allowlist.
- `packages/tool-linear-graphql/src/mcp-entry.ts` passes only normalized, non-empty credentials to the child MCP server.
- `packages/tool-linear-graphql/src/tool.ts` preserves a runtime authorization header or uses an API key verbatim.
- Codex and Claude runtime definitions both consume the delivered environment values.

## Risks & rollback

- Credentials remain runtime-managed MCP server configuration rather than model-supplied tool input.
- The main review focus is both runtime paths receiving credentials from exported environment configuration.
- Reverting this PR restores the previous worker environment and Authorization behavior.

## Changed files

- Linear tracker worker-environment delivery and regression coverage.
- Linear GraphQL MCP entry/header normalization and regression coverage.
- Codex and Claude runtime MCP configuration.
- Configuration and Docker E2E test-case documentation.
- `@gh-symphony/cli` patch changeset.

## Evidence

- `pnpm lint` — pass.
- `pnpm test` — pass.
- `pnpm typecheck` — pass.
- `pnpm build` — pass.
- Docker black-box: isolated `symphony-e2e` health check, fixture injection, refresh, and active worker-run readback — pass.
- CI: `Test` and `Container Smoke` — pass on `288ec4f`.
- Scoped Prettier check — pass; the repository-wide `pnpm format` still reports pre-existing formatting drift in unrelated files.

## Post-merge validation

- Verify an exported `LINEAR_API_KEY` reaches a repo-embedded Linear worker and supports built-in MCP issue reads, comments, PR links, and state transitions.
- Verify a standalone Linear project with only `LINEAR_API_KEY` uses the raw header successfully.

## Human validation

- [ ] Confirm a `LINEAR_API_KEY`-only Codex worker sends the raw key in `Authorization`.
- [ ] Confirm a runtime-provided `LINEAR_AUTHORIZATION` header remains unchanged.
- [ ] Confirm exported Linear credentials reach both Codex and Claude workers.


packages/runtime-codex/src/runtime.ts
```diff
@@ -555,6 +558,8 @@ export function buildCodexRuntimePlan(
   const linearTool = config.enableLinearGraphqlTool
     ? createLinearGraphQLToolDefinition({
         linearGraphqlUrl: config.linearGraphqlUrl ?? DEFAULT_LINEAR_GRAPHQL_URL,
+        linearAuthorization: config.linearAuthorization,
+        linearApiKey: config.linearApiKey,
```
hojinzs, line 562: **P1 — `config.linearApiKey` is `undefined` in the documented `repo start` flow, so this fix is inert there.**

I independently reproduced the gap the Codex thread flags, and it holds. The credential chain breaks one hop *above* this PR:

1. `packages/cli/src/commands/start.ts:187-190` gates Linear `repo start` on **`process.env.LINEAR_API_KEY`** — the documented flow is `export LINEAR_API_KEY=… && gh-symphony repo start`.
2. `packages/orchestrator/src/service.ts:4395-4416` builds the worker env as `projectEnv(.env)` → allowlisted `process.env` → explicit keys.
3. `INHERITED_ENV_ALLOWLIST` (`service.ts:102-113`) is `{CI, HOME, LANG, LOGNAME, PATH, PWD, SHELL, TERM, TMPDIR, USER}` — neither Linear variable survives the filter.
4. `linearTrackerAdapter.buildWorkerEnvironment` (`packages/tracker-linear/src/orchestrator-adapter.ts:215-221`) does not re-add them. Executed against the built adapter:

   ```
   linear buildWorkerEnvironment keys: LINEAR_GRAPHQL_URL, LINEAR_ISSUE_ID, LINEAR_ISSUE_IDENTIFIER, SYMPHONY_TRACKER_KIND
   contains LINEAR_API_KEY? false
   contains LINEAR_AUTHORIZATION? false
   ```
5. For a repo-embedded project `tenant.projectDir` falls back to `store.projectDir(projectId)` = `.runtime/orchestrator/projects/<id>` — an internal runtime directory, not somewhere a user authors `.env`. So step 2's `projectEnv` escape hatch does not exist there either.

Net effect: startup preflight passes, orchestrator polling works (it reads `process.env.LINEAR_API_KEY` directly at `orchestrator-adapter.ts:576`), and both runtimes still emit an unauthenticated built-in `linear_graphql` server — the exact symptom of #634. Only the reporter's configuration (standalone project **with** a project `.env`) is actually repaired.

Note GitHub does not hit this because its worker credential arrives via the token broker (`GITHUB_TOKEN_BROKER_URL`/`_SECRET`), not env inheritance. Linear has no such broker, so the credential has to be threaded explicitly.

Suggested fix (Integration layer, keeps tracker-specific code in the tracker): resolve the credential in `linearTrackerAdapter.buildWorkerEnvironment` the same way `orchestrator-adapter.ts:576` already resolves it, and return it alongside `LINEAR_GRAPHQL_URL`. A regression test asserting `buildWorkerEnvironment` carries the credential would pin it.

---

docs/configuration.md
```diff
@@ -161,8 +161,8 @@ container environments.
...
-| `LINEAR_AUTHORIZATION`   | unset                                                                                                            | Codex runtime                                                                           | Advanced                                            | Optional raw Linear authorization value for the Linear GraphQL tool.                                                                          |
+| `LINEAR_API_KEY`         | unset                                                                                                            | CLI, Codex runtime, Claude runtime                                                      | User-facing for Linear tracker projects             | Required for Linear repo startup. The built-in Linear MCP server receives it in its declared environment and uses it as the raw `Authorization` value. |
```
hojinzs, line 164: **P2 — this row overclaims delivery.**

"The built-in Linear MCP server receives it in its declared environment" is only true when the value reaches the worker process, which today happens **only** via a standalone project's `.env` file (see the P1 on `runtime-codex/src/runtime.ts`). A shell-exported `LINEAR_API_KEY` — which is what `gh-symphony repo start` validates — is dropped by the orchestrator env allowlist and never arrives.

Either land the P1 fix, after which this sentence becomes accurate as written, or scope the wording to the configuration that actually works and say where the value must be set.

---

packages/tool-linear-graphql/src/tool.ts
```diff
@@ -80,7 +80,7 @@ export function resolveLinearAuthorizationHeader(
   }
 
   if (config.apiKey) {
-    return `Bearer ${config.apiKey}`;
+    return config.apiKey;
```
hojinzs: **P3 — the raw value is now header-sensitive; trim it.**

Dropping `Bearer ` is the right call (`packages/tracker-linear/src/orchestrator-adapter.ts:457` already sends `Authorization: config.token` verbatim, so this aligns the tool with the tracker). But the old template literal absorbed stray whitespace, and a bare return no longer does.

A key with a trailing newline — routine when it comes from `.env`, `$(cat key)` or a copy-paste — used to produce a merely-wrong header; it now makes `fetch` throw `TypeError: Invalid header value`, surfacing to the agent as an opaque failure instead of the actionable "Linear GraphQL auth is not configured" message two lines below. `start.ts:188` and `doctor.ts:816` both `.trim()` this same variable, so the tool is the odd one out.

```ts
const apiKey = config.apiKey?.trim();
if (apiKey) {
  return apiKey;
}
```

Worth applying to the `authorizationHeader` branch above too, and to the truthiness checks in `createLinearGraphQLMcpServerEntry`, which currently forward a whitespace-only value as if it were a credential.

---

hojinzs (changes requested):

Verified against issue #634 and the #314 acceptance criteria it regresses. The two named defects are genuinely fixed at the MCP-server layer, but the credential still does not reach the worker in the flow the CLI itself validates, and the fix has no changeset so it cannot ship.

### Requirements coverage

| #634 requirement | Status |
| --- | --- |
| Defect A — entry declares no auth env, codex passes children only declared env | ✅ Fixed |
| Defect B — `LINEAR_API_KEY` sent as a Bearer token | ✅ Fixed |
| #314 AC — "runtime attaches the auth header without exposing the token to the model" | ✅ Preserved (credentials stay in MCP server env / `0600` config outside the checkout, never in tool-call input) |
| #314 AC — "worker can perform status transitions, workpad/comment writes, PR-link comments through the tool" | ⚠️ Only for a standalone project with a project `.env` — see P1 |

### Smoke test

No Linear preview environment or live credentials, so I ran the built server black-box against a local HTTPS GraphQL stub on an allowlisted `*.linear.app` host, spawning it **with only the declared env** — exactly how codex spawns MCP children, which is what Defect A was about:

| Case | Declared env | `Authorization` the server received | Result |
| --- | --- | --- | --- |
| API-key only | `LINEAR_GRAPHQL_URL`, `LINEAR_API_KEY` | `lin_api_TESTKEY` (no `Bearer`) | ✅ query returned data |
| Runtime header present | + `LINEAR_AUTHORIZATION` | `Bearer oauth_tok` (verbatim, key ignored) | ✅ priority correct |
| No credentials | `LINEAR_GRAPHQL_URL` only | no request sent | ✅ fails closed with the actionable message |

Both human-validation checkboxes in the PR description are satisfied at this layer. Repo checks are green on `a895cad`: `pnpm lint`, `pnpm typecheck`, `pnpm build`, `pnpm test` (all packages pass; one earlier run flaked in orchestrator timing tests and was clean on re-run, consistent with the workpad note).

### Findings

- **P1 — credential never reaches the worker in the `repo start` flow.** Inline on `packages/runtime-codex/src/runtime.ts`. The CLI gates Linear startup on a shell-exported `LINEAR_API_KEY`, the orchestrator's `INHERITED_ENV_ALLOWLIST` drops it, and `linearTrackerAdapter.buildWorkerEnvironment` does not re-add it — so both runtimes still build an unauthenticated `linear_graphql` server. This corroborates the Codex bot thread; I confirmed each hop independently, including by executing the built adapter. Repo-embedded projects have no user-authored `.env` to fall back on.
- **P1 — no changeset.** `CLAUDE.md` requires behavior-changing PRs to add `.changeset/*.md` for `@gh-symphony/cli`, and every recent fix on `main` does (#617, #618, #620–#623, #628, #629, #633). The workpad justifies the omission by the absence of a changeset *label*, which is not the repo's rule. Without one the changeset-release bot never publishes, and the affected population is precisely the 0.10.1 users this issue is about.
- **P2 — `docs/configuration.md` overclaims delivery.** Inline. Resolves itself once P1 lands.
- **P3 — trim the credential before using it as a raw header value.** Inline on `tool.ts`.

### Scope

No over-engineering and no unrequested decisions. The change mirrors `createGitHubGraphQLMcpServerEntry` exactly, which is the right precedent, and the two deleted `not.toContain("lin_api_key")` assertions were inverted rather than dropped — the default-entry test still pins that no auth key appears when none is configured. Removing the `Bearer` prefix outright (rather than sniffing `lin_api_`) is corroborated by `tracker-linear/src/orchestrator-adapter.ts:457`, which already sends the key verbatim.

No prior human review on this PR, so nothing outstanding to re-check.

Round 1 of at most 6. Fixing the P1 delivery gap and adding the changeset would get this to approve.

---

packages/runtime-codex/src/runtime.ts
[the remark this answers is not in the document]
moncher-dev: Resolved in 51eca32. `linearTrackerAdapter.buildWorkerEnvironment` now explicitly forwards normalized `LINEAR_API_KEY` and `LINEAR_AUTHORIZATION`, bypassing the inherited-environment allowlist for both Codex and Claude. A regression test pins this worker-boundary contract.

moncher-dev: Resolved in 51eca32. The Linear adapter now forwards normalized `LINEAR_API_KEY` and `LINEAR_AUTHORIZATION` through the explicit worker environment, so the documented exported-key `repo start` flow reaches both runtime MCP definitions. Added regression coverage at the tracker boundary.

docs/configuration.md
moncher-dev: Resolved in 51eca32. The documentation statement is now backed by explicit Linear adapter credential delivery in both standalone and repo-embedded project flows.

packages/tool-linear-graphql/src/tool.ts
moncher-dev: Resolved in 51eca32. `resolveLinearAuthorizationHeader` now trims both credentials and fails closed for blank values; `createLinearGraphQLMcpServerEntry` applies the same normalization and omits whitespace-only environment entries. Regression tests cover all cases.

packages/tool-linear-graphql/src/tool.test.ts
```diff
@@ -245,17 +263,38 @@ describe("createLinearGraphQLMcpServerEntry", () => {
...
+      }).env
+    ).toMatchObject({
+      LINEAR_AUTHORIZATION: "Bearer runtime-token",
+    });
+    expect(
+      createLinearGraphQLMcpServerEntry({
+        linearAuthorization: " \n ",
+        linearApiKey: " \t ",
+      }).env
+    ).not.toHaveProperty("LINEAR_AUTHORIZATION");
```
hojinzs, line 298: **nit — the `LINEAR_API_KEY` omission branch is never asserted.**

Both halves of this test only pin `LINEAR_AUTHORIZATION`:

- the first `expect` uses `toMatchObject`, which passes whether or not `LINEAR_API_KEY` is present, so `linearApiKey: " \n "` being dropped is not actually checked;
- the second `expect` asserts absence for `LINEAR_AUTHORIZATION` only.

So `...(linearApiKey ? { LINEAR_API_KEY: linearApiKey } : {})` — a branch this PR introduced — has no test that fails if it regresses to forwarding a whitespace-only value. Given #634's whole shape was "a credential that looks delivered but isn't", it's worth pinning:

```ts
).not.toHaveProperty("LINEAR_AUTHORIZATION");
```
→ plus, on the same object, `.not.toHaveProperty("LINEAR_API_KEY")`, and the same on the first case.

Not blocking — the production behavior is correct, I verified the fail-closed path black-box (whitespace-only in both vars ⇒ no request sent, actionable error). This only hardens the guard.

---

docs/configuration.md
```diff
@@ -156,14 +156,14 @@ when no `--config <dir>`/`GH_SYMPHONY_CONFIG_DIR` override is set).
...
+| `LINEAR_AUTHORIZATION`   | unset                                                                                                            | Codex runtime, Claude runtime                                                           | Advanced                                            | Optional raw Linear authorization value for the built-in Linear MCP server; it takes priority over `LINEAR_API_KEY`.                                   |
```
hojinzs, line 165: **nit — the "Read by" column is now one reader short for both rows.**

My P2 from round 1 is resolved: the delivery claim in the Notes column is true now that the adapter forwards the value. But the fix also added a new reader that these rows don't list — `linearTrackerAdapter.buildWorkerEnvironment` (`packages/tracker-linear/src/orchestrator-adapter.ts:216-217`) reads both variables from the **orchestrator process** environment.

That's the operationally load-bearing detail: the runtimes only see these values because the orchestrator's own env had them at dispatch time. A reader who takes "Read by: CLI, Codex runtime, Claude runtime" literally could reasonably conclude that setting them at `gh-symphony repo start` invocation time is incidental, when in fact the daemon process env is the source of truth. Suggest adding `Linear tracker` (or `orchestrator`) to both rows.

`LINEAR_AUTHORIZATION` also gains "Claude runtime" in this PR, which the row does capture — so it's just the tracker/orchestrator hop missing.

---

hojinzs (approved):

All four round-1 findings are resolved and independently verified on `288ec4f`. Only two non-blocking nits remain, inline.

### Round-1 findings — re-checked

| Finding | Status |
| --- | --- |
| P1 — credential never reaches the worker in the `repo start` flow | ✅ Fixed. `buildWorkerEnvironment` now forwards both vars explicitly, bypassing `INHERITED_ENV_ALLOWLIST`. Re-ran my original repro against the built adapter: keys are now `LINEAR_GRAPHQL_URL, LINEAR_ISSUE_ID, LINEAR_ISSUE_IDENTIFIER, SYMPHONY_TRACKER_KIND, LINEAR_API_KEY` (was missing both credentials). |
| P1 — no changeset | ✅ Fixed. `.changeset/fresh-linear-mcp-auth.md`, `@gh-symphony/cli` patch. |
| P2 — `docs/configuration.md` overclaims delivery | ✅ Fixed by the P1 landing, as predicted. One residual nit inline. |
| P3 — trim the credential before using it as a raw header | ✅ Fixed in both `resolveLinearAuthorizationHeader` and `createLinearGraphQLMcpServerEntry`, and it fails closed on blank. |

Reading `process.env.LINEAR_API_KEY` in `buildWorkerEnvironment` is the right resolution source: it matches how the daemon already authenticates its own polling (`orchestrator-adapter.ts:583`), so the worker and the poller now present the same identity. It also keeps the tracker-specific code in the Integration layer per `CLAUDE.md`.

### Smoke test

No Linear preview environment or live credentials, so I re-ran the black-box harness from round 1 — built `mcp-server.js` spawned with **only the declared env** (the codex child-spawn contract that Defect A was about), against a local HTTPS GraphQL stub on an allowlisted `*.linear.app` host — and extended it with the worker-boundary hop that this round fixed:

| Case | Input | `Authorization` the server received | Result |
| --- | --- | --- | --- |
| A | `linearApiKey: "  lin_api_TESTKEY \n"` | `lin_api_TESTKEY` — trimmed, no `Bearer` | ✅ query returned data |
| B | `linearAuthorization: " Bearer oauth_tok "` + api key | `Bearer oauth_tok` — verbatim, key ignored | ✅ priority correct |
| C | both whitespace-only | no request sent | ✅ fails closed with the actionable message |
| D | exported `LINEAR_API_KEY=" lin_api_EXPORTED \n"` → `buildWorkerEnvironment` → entry → server | `lin_api_EXPORTED` | ✅ **full chain, the round-1 P1 gap** |

Case D is the one that failed before: it walks the exported shell value through the real tracker adapter into the MCP child, which is exactly the `gh-symphony repo start` path #634 reported broken.

I also traced the remaining hops by reading rather than executing: `service.ts:2948` merges the adapter output into `buildProjectExecutionEnv`; `worker/src/index.ts:571-572` lifts it into the Codex config; `runtime-claude/src/adapter.ts:693-694` → `mcp-compose.ts:132-133` does the same for Claude. Both runtimes are covered.

Local checks on `288ec4f`: `pnpm build`, `pnpm typecheck`, and the four affected packages' tests all pass (`tool-linear-graphql` 19, `tracker-linear` 27, `runtime-codex` 56, `runtime-claude` 93). Scoped `prettier --check` over the 12 changed files is clean. CI `Test` and `Container Smoke` are green.

### #634 / #314 coverage

| Requirement | Status |
| --- | --- |
| Defect A — entry declares no auth env | ✅ Fixed |
| Defect B — API key sent as a Bearer token | ✅ Fixed |
| #314 AC — auth header attached without exposing the token to the model | ✅ Preserved. Claude's config is `0600` in `WORKSPACE_RUNTIME_DIR`, outside the checkout, and already carries `GITHUB_GRAPHQL_TOKEN` — no new exposure class. The non-Linear path still strips both vars (`runtime.ts:620-624`), and that leak case has a test. |
| #314 AC — worker can transition state, write workpads, post PR links | ✅ Now reachable in both the standalone and repo-embedded flows |

### Scope

No over-engineering and no unrequested decisions. Every change traces to a round-1 finding or to #634; the fourth commit is formatting-only. The trimming is scoped to normalization and fail-closed behavior rather than growing into credential validation, which is the right restraint.

Approving. The two nits are yours to take or leave — neither needs another round.

---

PR #597, opened 2026-08-20
fix(tracker-github): map GHES REST API base

moncher-dev:
## TL;DR

Map GHES GraphQL endpoints to the correct REST API base while preserving GitHub.com and `/api/v3/graphql` behavior.

## Issues — Closed #497

Fixes #497

## Change-Point Diagram

```text
GraphQL endpoint
      │
      ▼
resolveRestApiUrl ──► REST base URL ──► `/user`, issue-comment GET/POST/PATCH requests
```

## Start Here

- REST URL resolution and focused request-path tests are implemented in the GitHub tracker integration package.
- The key regression cases are GHES `/api/graphql`, GHES `/api/v3/graphql`, and github.com endpoints.
- The release metadata is recorded in `.changeset/ghes-rest-api-v3.md` for `@gh-symphony/cli`.

## Summary

- Map GHES `/api/graphql` to `/api/v3` for REST requests.
- Preserve existing GitHub.com and `/api/v3/graphql` mappings.
- Cover `/user` and issue-comment GET/POST/PATCH request paths for all supported endpoint forms.
- Preserve GHES `/api/graphql/` trailing-slash handling with an explicit regression case.

## Risks & Rollback

- Risk: changing endpoint normalization could affect non-GHES API requests.
- Rollback: revert the focused URL-resolution change and its tests.

## Changed Files

- `packages/tracker-github/src/adapter.ts` — GHES REST base normalization.
- `packages/tracker-github/src/tracker-github.test.ts` — endpoint and request-path regression coverage.
- `.changeset/ghes-rest-api-v3.md` — patch release metadata for `@gh-symphony/cli`.

## Evidence

- `pnpm --filter @gh-symphony/tracker-github test -- tracker-github.test.ts` — passed (94 tests).
- `pnpm --filter @gh-symphony/tracker-github lint` — passed.
- `pnpm --filter @gh-symphony/tracker-github test` — passed (114 tests).
- `pnpm lint` — passed.
- `pnpm test` — passed repository-wide in cycle 3.
- `pnpm typecheck` — passed.
- `pnpm build` — passed.
- `pnpm exec prettier --check .changeset/ghes-rest-api-v3.md packages/tracker-github/src/adapter.ts packages/tracker-github/src/tracker-github.test.ts` — passed.
- `pnpm changeset status` — passed; the new entry targets `@gh-symphony/cli` with a patch bump.
[...]

Issue #497, opened 2026-07-30
GHES GraphQL endpoint의 REST /api/v3 매핑


moncher-dev:
## 배경

PR #489 리뷰에서 GHES GraphQL endpoint가 `https://github.example/api/graphql`일 때 현재 `resolveRestApiUrl`이 REST base를 `/api`로 계산하는 문제가 확인됐다. GHES REST API는 일반적으로 `/api/v3`를 사용한다.

## 작업

- `/api/graphql` GraphQL endpoint를 `/api/v3` REST base로 매핑
- 기존 `/api/v3/graphql` 및 github.com endpoint 동작 보존
- `/user`와 issue comment GET/POST/PATCH 경로 TC 추가

## 완료 조건

- GHES endpoint별 REST URL 조립이 단위 테스트로 검증됨
- github.com 기존 동작에 회귀가 없음

관련: #476, PR #489

moncher-dev:
## Workpad — hojinzs/github-symphony#497 — Cycle 1

**Type:** fresh pickup
**Branch:** feat/497-ghes-rest-api-v3
**Draft PR:** https://github.com/hojinzs/github-symphony/pull/597
**Cycle opened:** 2026-08-20T10:16:50Z · **Trigger:** Issue #497 entered Ready for a fresh implementation cycle.

### Status Transitions



- 2026-08-20T10:17:10Z · `Ready` → `In progress` · Draft PR #597 created and implementation cycle opened (cycle 1 open)

### Plan

- [x] 1. Implement GHES REST URL mapping and add endpoint/path coverage tests.
- [x] 2. Wrap-up: re-verify the original issue · pass the Completion Bar · refresh the PR · transition to In review.

### Rework / PR Feedback

Not applicable; this is a fresh pickup.

### Completion Bar

- [x] In-scope requirements implemented
- [x] `pnpm lint`
- [x] `pnpm test`
- [x] `pnpm typecheck`
- [x] `pnpm build`
- [x] Docker E2E (tracker adapter integration behavior)
- [x] Tests for new functionality
- [x] Code conventions (CLAUDE.md)
- [x] Inline review comments answered (rework only; N/A for fresh cycle)

### Validation

- `pnpm --filter @gh-symphony/tracker-github test -- tracker-github.test.ts` — passed (94 tests).
- `pnpm --filter @gh-symphony/tracker-github lint` — passed.
- `pnpm --filter @gh-symphony/tracker-github test` — passed (113 tests).
[...]

PR #638, opened 2026-08-26
fix(cli): align subpath GHES REST mapping

### LINKED ISSUE #602: fix(cli): align subpath-hosted GHES REST URL mapping
## Background

PR #597 identified that `packages/cli/src/github/client.ts` has separate GHES GraphQL-to-REST normalization from the tracker adapter.

For a subpath-hosted endpoint such as `https://host/gh/api/graphql`, the tracker adapter resolves `https://host/gh/api/v3`, while the CLI helper currently resolves `https://host/gh/api`.

## Scope

- Align or consolidate CLI GHES REST endpoint normalization.
- Add regression coverage for subpath-hosted GHES endpoints.
- Preserve GitHub.com and existing `/api/v3/graphql` behavior.

## Context

- Follow-up from #497 and PR #597.
- This is intentionally outside #497's tracker-adapter scope.

moncher-dev:
## TL;DR

Align the CLI's GHES REST endpoint mapping with the tracker adapter for subpath-hosted GitHub Enterprise Server instances.

## Change-point diagram

`GHES GraphQL URL` → `deriveGitHubRestApiUrl` → `GHES REST base URL` → repository and token REST requests

## Start here

- `packages/cli/src/github/client.ts` contains endpoint normalization.
- `packages/cli/src/github/client.test.ts` covers public GitHub, canonical GHES, subpath-hosted GHES, and existing `/api/v3/graphql` URLs.

## Risks & rollback

The mapping changes only the REST base derived from custom GraphQL endpoints. Revert this change to restore the prior URL derivation.

## Changed files

- `packages/cli/src/github/client.ts`
- `packages/cli/src/github/client.test.ts`
- `.changeset/quiet-ghes-subpath.md`

## Issues — Closed #602

Fixes #602

## Evidence

- `pnpm --filter @gh-symphony/cli test -- client.test.ts` — passed (7 tests).
- `pnpm lint` — passed.
- `pnpm build` — passed.
- `pnpm test` — passed after building workspace package entrypoints.
- `pnpm typecheck` — passed.

- [ ] Verify CLI repository validation against a subpath-hosted GHES instance, if available.

### DIFF
--- a/packages/cli/src/github/client.ts
+++ b/packages/cli/src/github/client.ts
@@ -151,14 +151,13 @@ export function deriveGitHubRestApiUrl(graphqlApiUrl: string): string {
     if (url.hostname.toLowerCase() === "api.github.com") {
       return REST_API_URL;
     }
-    if (normalizedPath === "/api/graphql") {
-      url.pathname = "/api/v3";
-      url.search = "";
-      url.hash = "";
-      return url.toString().replace(/\/$/, "");
-    }
     if (normalizedPath.endsWith("/graphql")) {
-      url.pathname = normalizedPath.slice(0, -"/graphql".length) || "/";
+      const pathSegments = normalizedPath.split("/").filter(Boolean);
+      pathSegments.pop();
+      if (pathSegments.at(-1) === "api") {
+        pathSegments.push("v3");
+      }
+      url.pathname = `/${pathSegments.join("/")}`;
       url.search = "";
       url.hash = "";
       return url.toString().replace(/\/$/, "");
--- a/.changeset/quiet-ghes-subpath.md
+++ b/.changeset/quiet-ghes-subpath.md
@@ -1,0 +1,5 @@
+---
+"@gh-symphony/cli": patch
+---
+
+Map subpath-hosted GHES GraphQL endpoints to their `/api/v3` REST base URL (#602).
--- a/packages/cli/src/github/client.test.ts
+++ b/packages/cli/src/github/client.test.ts
@@ -24,6 +24,18 @@ describe("deriveGitHubRestApiUrl", () => {
       "https://github.example/api/v3"
     );
   });
+
+  it("preserves a GHES subpath when mapping /api/graphql to /api/v3", () => {
+    expect(
+      deriveGitHubRestApiUrl("https://github.example/gh/api/graphql")
+    ).toBe("https://github.example/gh/api/v3");
+  });
+
+  it("preserves an existing GHES /api/v3/graphql endpoint", () => {
+    expect(
+      deriveGitHubRestApiUrl("https://github.example/gh/api/v3/graphql")
+    ).toBe("https://github.example/gh/api/v3");
+  });
 });
 
 describe("discoverUserProjects", () => {


packages/cli/src/github/client.ts
```diff
@@ -151,14 +151,13 @@ export function deriveGitHubRestApiUrl(graphqlApiUrl: string): string {
     if (url.hostname.toLowerCase() === "api.github.com") {
       return REST_API_URL;
     }
-    if (normalizedPath === "/api/graphql") {
-      url.pathname = "/api/v3";
-      url.search = "";
-      url.hash = "";
-      return url.toString().replace(/\/$/, "");
-    }
     if (normalizedPath.endsWith("/graphql")) {
```
hojinzs, line 154: **P2 (non-blocking, pre-existing)** — the alignment this PR achieves stops at endpoints that end in `/graphql`. Everything else still silently falls through to `return REST_API_URL` (public GitHub).

Verified locally against this branch:

| input | this function | tracker `resolveRestApiUrl` |
| --- | --- | --- |
| `https://github.example/api/v3` | `https://api.github.com` | `https://github.example/api/v3/user` |
| `https://github.example/gh/api` | `https://api.github.com` | `https://github.example/gh/api/user` |

`tracker.endpoint` has no `/graphql` suffix validation anywhere (`packages/core/src/workflow/parser.ts` only enforces non-empty, and only for `kind: linear`), so a GHES operator who configures the REST base — or `https://host/gh/api`, an easy mistake given GHES documents `/api/v3` and `/api/graphql` side by side — ends up with the CLI sending their enterprise token to `api.github.com`, surfacing as a puzzling 404/401 against public GitHub rather than a config error.

Strictly outside #602's stated scope and it predates this diff, so I'm not blocking on it. But it is the same divergence class the issue is closing, so it is worth either handling here (for a non-`api.github.com` host, preserve the host instead of falling back — or raise a config error) or filing as the follow-up to #602.

---

packages/cli/src/github/client.ts
```diff
@@ -151,14 +151,13 @@ export function deriveGitHubRestApiUrl(graphqlApiUrl: string): string {
...
-      url.pathname = normalizedPath.slice(0, -"/graphql".length) || "/";
+      const pathSegments = normalizedPath.split("/").filter(Boolean);
+      pathSegments.pop();
+      if (pathSegments.at(-1) === "api") {
```
hojinzs, line 157: nit — case-sensitive segment compare: `https://github.example/gh/API/graphql` maps to `https://github.example/gh/API`, with no `v3` appended. The tracker adapter compares the same way, so alignment is preserved either way; only worth changing if both sides get normalized together.

---

packages/cli/src/github/client.ts
```diff
@@ -151,14 +151,13 @@ export function deriveGitHubRestApiUrl(graphqlApiUrl: string): string {
...
+        pathSegments.push("v3");
+      }
+      url.pathname = `/${pathSegments.join("/")}`;
```
hojinzs, line 160: **P3** — these five lines are now algorithmically identical to `resolveRestApiUrl` in `packages/tracker-github/src/adapter.ts:2687-2696` (split → pop `graphql` → append `v3` when the parent segment is `api`). #602 allowed "align **or** consolidate" and aligning is the smaller, lower-risk change, so this is the right call for this PR.

The concern is drift: this same mapping has now been corrected three times (#497 → #597 → #602), each time in one copy at a time. A shared helper would make round four impossible. `packages/tool-github-graphql` is the natural home — both `tracker-github` and `cli` already resolve it, and it keeps GitHub-specific semantics out of `core` per the CLAUDE.md layering rule.

If consolidation stays out of scope, a one-line comment on each side pointing at the other is enough to keep the next editor honest.

---

hojinzs (approved):
## Approve — requirements met, findings are non-blocking

### Requirement coverage vs #602

| Acceptance item | Status |
| --- | --- |
| Align or consolidate CLI GHES REST normalization | ✅ aligned — the algorithm now matches `resolveRestApiUrl` (`packages/tracker-github/src/adapter.ts:2687`) exactly |
| Regression coverage for subpath-hosted GHES | ✅ two new cases (`/gh/api/graphql`, `/gh/api/v3/graphql`) |
| Preserve GitHub.com and existing `/api/v3/graphql` behavior | ✅ verified below |
| Changeset (behavior-changing, per CLAUDE.md) | ✅ `.changeset/quiet-ghes-subpath.md`, patch / `@gh-symphony/cli` |

### Smoke test

No preview environment applies (CLI-internal URL normalization), so this ran locally against `bdd80af` with a matrix wider than the test file:

| input | output |
| --- | --- |
| `https://api.github.com/graphql` | `https://api.github.com` |
| `https://API.GitHub.com/graphql` | `https://api.github.com` |
| `https://github.example/api/graphql` | `https://github.example/api/v3` |
| `https://github.example/api/v3/graphql` | `https://github.example/api/v3` |
| `https://github.example/gh/api/graphql` | `https://github.example/gh/api/v3` ← the fix |
| `https://github.example/gh/api/v3/graphql` | `https://github.example/gh/api/v3` |
| `https://github.example/gh/api/graphql/` | `https://github.example/gh/api/v3` |
| `https://github.example:8443/gh/api/graphql?x=1#f` | `https://github.example:8443/gh/api/v3` |
| `https://github.example/graphql` | `https://github.example` (unchanged from before) |
| malformed / empty | `https://api.github.com` (unchanged) |

Port, credentials, query and fragment handling all behave. The only behavioral delta versus `main` is the intended one: `/<subpath>/api/graphql` now resolves to `/<subpath>/api/v3` instead of `/<subpath>/api`. Non-`api`-parented paths (`/foo/graphql` → `/foo`) are untouched, so the rewrite is genuinely behavior-preserving outside the bug.

- `npx vitest run packages/cli/src/github/client.test.ts` — 7 passed
- `tsc --noEmit` on `@gh-symphony/cli` — clean
- CI on `bdd80af`: **Test** ✅, **Container Smoke** ✅

### Scope / over-engineering

Clean. 23 additions across 3 files, one exported function touched, no new abstraction, no config surface added. Choosing "align" over "consolidate" is a defensible reading of the issue's own wording — see the P3 note.

### Prior review rounds

None with substance — the only prior review is the Codex bot's boilerplate header with no suggestions attached, so there is nothing outstanding to re-check.

### Findings

- **P2** (`client.ts:154`) — endpoints not ending in `/graphql` still fall back to `api.github.com`, which diverges from the tracker adapter and can send a GHES token to public GitHub. Pre-existing and outside #602's scope; handle here or file the follow-up.
- **P3** (`client.ts:160`) — the mapping is now duplicated verbatim in two packages; third drift on this same logic. Suggest a shared helper in `tool-github-graphql`, or cross-reference comments.
- **nit** (`client.ts:157`) — case-sensitive `api` segment compare, matching the tracker.

None of these block merge. Approving.

---

PR #622, opened 2026-08-23
fix(workflow): expose planning execution phase

moncher-dev:
## TL;DR

- Make `planning_states` behaviorally useful by exposing the normalized run phase as `{{ execution_phase }}` to workflow prompts.
- Use the same trimmed, case-insensitive state matching for worker metadata, orchestrator dispatch prompts, and CLI preview.
- Keep dispatch eligibility separate from phase classification; `WORKFLOW.md` chooses the phase-specific agent posture.
- Rebased onto current `main`, resolved the documentation/E2E conflicts, and incorporated the final runner review notes.

## Change-point diagram

```text
tracker issue state + workflow lifecycle config
                     │
                     ▼
       normalized phase classification
                     │
          ┌──────────┴──────────┐
          ▼                     ▼
 run metadata/status     {{ execution_phase }}
                                │
                                ▼
                   WORKFLOW.md prompt policy
```

## Start here

- `packages/core/src/workflow/lifecycle.ts` — shared normalized phase classification.
- `packages/core/src/workflow/render.ts` — prompt variable contract.
- `packages/orchestrator/src/service.ts` — dispatch-time phase injection.
- `packages/cli/src/commands/workflow.ts` — preview-time phase injection.
- `e2e/run-e2e.sh` and `e2e/stub-worker.ts` — black-box prompt and lifecycle assertions.

## Changes

- Add shared `resolveWorkflowExecutionPhase()` using normalized workflow-state matching and route worker classification through it.
- Add `execution_phase` to prompt variables and populate it for orchestrator dispatch, issue preview, validation, and retry preview paths.
- Document that `planning_states` classifies independently of dispatch eligibility and does not itself impose a plan-only gate.
- Add unit/CLI regression tests and a first-class Docker E2E `prompt-phase` runner with reliable polling, observable failures, final-state assertions, and bounded diagnostic output.
- Preserve current-main blocker and bounded-finalization documentation/fixtures while resolving the rebase conflicts.
- Add a patch changeset for `@gh-symphony/cli`.

## Evidence

- `pnpm lint`
- `pnpm test` — all workspace package suites passed.
- `pnpm typecheck`
- `pnpm build`
- `git diff --check`
- `bash -n e2e/run-e2e.sh`
- `./e2e/run-e2e.sh prompt-phase` — run `repository-test-owner-test-repo-1-mt6mtaum` observed `running`, accepted `phase=planning`, confirmed `Ready → Done`, and verified persisted `succeeded/succeeded`.

## Risks & rollback

[...]

PR #641, opened 2026-08-26
test(orchestrator): cover execution phase prompt dispatch

### LINKED ISSUE #632: test(orchestrator): cover execution phase prompt injection during dispatch
## Summary

Add an automated orchestrator unit regression test proving that dispatch injects the normalized `execution_phase` value into `SYMPHONY_RENDERED_PROMPT`.

## Context

PR #622 for #614 adds normalized planning-state classification and exposes `{{ execution_phase }}` to rendered worker prompts. Core, worker, CLI, and manual Docker E2E coverage exist, but the orchestrator dispatch wiring is not directly asserted by CI.

## Acceptance criteria

- Configure a workflow whose `planning_states` entry differs from the issue state by whitespace/case.
- Exercise the orchestrator dispatch path with `runOnce()`.
- Assert the spawned worker environment's `SYMPHONY_RENDERED_PROMPT` contains the normalized planning phase.
- Keep the test focused on dispatch wiring; production behavior should not change.

## Related

- Follow-up from #614 and PR #622 review.

moncher-dev:
## TL;DR

Adds a focused dispatch regression test proving the normalized execution phase is rendered into the worker prompt.

## Change-point diagram

`planning_states: "  rEaDy  "` → issue state `READY` → `runOnce()` → spawned worker `SYMPHONY_RENDERED_PROMPT` → `execution_phase=planning`

## Start here

- `packages/orchestrator/src/service.test.ts` — new dispatch-level `runOnce()` coverage.

## Risks & rollback

- Test-only change; production dispatch behavior is unchanged. Revert commit `64fa40f` to roll back.

## Changed files

- No manual validation required; automated unit and Docker black-box coverage completed.

## Evidence

- `pnpm lint`
- `pnpm test` (after `pnpm build` populated workspace package entries)
- `pnpm typecheck`
- `pnpm build`
- `./e2e/run-e2e.sh prompt-phase 60` exercised the prompt-phase Docker worker; its `phase=planning` guard passed. The runner's final container-log grep is stale because worker stderr is stored in `worker.log`.

### DIFF
--- a/packages/orchestrator/src/service.test.ts
+++ b/packages/orchestrator/src/service.test.ts
@@ -6011,6 +6011,49 @@ Prefer focused changes.
     expect(workerEnv?.SYMPHONY_ISSUE_IDENTIFIER).toBe("acme/platform#1");
   });
 
+  it("injects the normalized planning phase into the dispatched worker prompt", async () => {
+    process.env.GITHUB_GRAPHQL_TOKEN = "test-token";
+    const tempRoot = await mkdtemp(
+      join(tmpdir(), "orchestrator-execution-phase-prompt-")
+    );
+    const repository = await createRepositoryFixture(
+      tempRoot,
+      "acme",
+      "platform",
+      {
+        rawWorkflow: createReadyStateWorkflow(
+          "execution_phase={{ execution_phase }}\n",
+          ["  rEaDy  "]
+        ),
+      }
+    );
+    const store = new OrchestratorFsStore(tempRoot);
+    const projectConfig = createProjectConfig(tempRoot, repository);
+    await store.saveProjectConfig(projectConfig);
+
+    const spawnImpl = vi.fn().mockReturnValue({
+      pid: 4307,
+      unref: vi.fn(),
+    });
+    const service = new OrchestratorService(store, projectConfig, {
+      fetchImpl: vi
+        .fn()
+        .mockResolvedValue(createTrackerResponseWithState(repository, "READY")),
+      spawnImpl: spawnImpl as never,
+      now: () => new Date("2026-03-08T00:00:00.000Z"),
+    });
+
+    const result = await service.runOnce();
+    const workerEnv = spawnImpl.mock.calls[0]?.[2]?.env as
+      | Record<string, string>
+      | undefined;
+
+    expect(result.summary.dispatched).toBe(1);
+    expect(workerEnv?.SYMPHONY_RENDERED_PROMPT).toContain(
+      "execution_phase=planning"
+    );
+  });
+
   it("isolates an untrusted issue body from workflow instructions", async () => {
     process.env.GITHUB_GRAPHQL_TOKEN = "test-token";
     const tempRoot = await mkdtemp(
@@ -13932,7 +13975,15 @@ Prefer focused changes.
   await writeFile(join(repositoryRoot, "WORKFLOW.md"), content, "utf8");
 }
 
-function createReadyStateWorkflow(prompt = "{{ issue.title }}\n"): string {
+function createReadyStateWorkflow(
+  prompt = "{{ issue.title }}\n",
+  planningStates: string[] = []
+): string {
+  const planningStatesYaml = planningStates.length
+    ? `  planning_states:\n${planningStates
+        .map((state) => `    - ${JSON.stringify(state)}`)
+        .join("\n")}\n`
+    : "";
   return `---
 tracker:
   kind: github-project
@@ -13944,7 +13995,7 @@ tracker:
     - Done
   blocker_check_states:
     - Ready
-hooks:
+${planningStatesYaml}hooks:
   commands: []
 polling:
   interval_ms: 30000


packages/orchestrator/src/service.test.ts
```diff
@@ -6011,6 +6011,49 @@ Prefer focused changes.
...
+    });
+
+    const result = await service.runOnce();
+    const workerEnv = spawnImpl.mock.calls[0]?.[2]?.env as
+      | Record<string, string>
+      | undefined;
+
+    expect(result.summary.dispatched).toBe(1);
+    expect(workerEnv?.SYMPHONY_RENDERED_PROMPT).toContain(
+      "execution_phase=planning"
```
hojinzs, line 6053: nit: `toContain("execution_phase=planning")` is a prefix-loose match — it would also pass on a hypothetical `execution_phase=planning-something`. Since the template is `execution_phase={{ execution_phase }}\n`, asserting the trailing newline pins the rendered value exactly:

```suggestion
      "execution_phase=planning\n"
```

Non-blocking; the mutation behavior is already correct today (I verified that replacing `matchesWorkflowState(...)` with a strict `includes(...)` in `resolveWorkflowExecutionPhase` makes this assertion fail, so the test is genuinely load-bearing).

---

packages/orchestrator/src/service.test.ts
```diff
@@ -6011,6 +6011,49 @@ Prefer focused changes.
     expect(workerEnv?.SYMPHONY_ISSUE_IDENTIFIER).toBe("acme/platform#1");
   });
 
+  it("injects the normalized planning phase into the dispatched worker prompt", async () => {
+    process.env.GITHUB_GRAPHQL_TOKEN = "test-token";
+    const tempRoot = await mkdtemp(
+      join(tmpdir(), "orchestrator-execution-phase-prompt-")
```
hojinzs, line 6017: nit: this `tempRoot` is never removed. The immediately preceding test leaks its temp dir too, so this matches local style, but most fixture-based tests in this file end with `await rm(tempRoot, { recursive: true, force: true })` — each of these creates a real git checkout under `os.tmpdir()`. Adding the cleanup line at the end of the test keeps the suite from growing another leaked fixture.

---

packages/orchestrator/src/service.test.ts
```diff
@@ -13944,7 +13995,7 @@ tracker:
     - Done
   blocker_check_states:
     - Ready
-hooks:
+${planningStatesYaml}hooks:
```
hojinzs, line 13998: Verified this helper change is correct and minimal:

- The 2-space indent puts `planning_states` under `tracker:`, which matches the parser (`packages/core/src/workflow/parser.ts:78` reads `planning_states` off the `tracker` block). ✅
- `JSON.stringify(state)` emits a YAML double-quoted scalar, so leading/trailing whitespace survives the round-trip. I confirmed against the real parser that `"  rEaDy  "` comes back as `"  rEaDy  "` — so the test exercises both the whitespace *and* case halves of the normalization, as the acceptance criteria asked. ✅
- Default `planningStates = []` keeps all five existing callers byte-identical. ✅

nit only: `${planningStatesYaml}hooks:` is a slightly subtle interpolation — the reader has to notice that `planningStatesYaml` carries its own trailing newline for the `hooks:` key to land at column 0. A short comment on the `planningStatesYaml` const would save the next reader that trace.

---

hojinzs (approved):
## Approve

Test-only change, scoped tightly to #632, and it does what it claims. Three nits inline, none blocking.

### Acceptance criteria (#632) — all met

| Criterion | Status |
|---|---|
| `planning_states` entry differs from issue state by whitespace/case | ✅ `"  rEaDy  "` vs `READY` — verified against the real parser that YAML preserves both |
| Exercise dispatch via `runOnce()` | ✅ |
| Assert `SYMPHONY_RENDERED_PROMPT` contains the normalized planning phase | ✅ |
| Keep focused on dispatch wiring; no production change | ✅ diff is one test file, +53/−2 |

The wiring now covered is `packages/orchestrator/src/service.ts:2808-2814` (`buildPromptVariables` → `resolveWorkflowExecutionPhase`), which was previously untested at dispatch level — `dispatch.test.ts` only feeds `planningStates` into `explainIssueDispatch`, so there is no duplicated coverage.

### The test is non-vacuous (mutation-verified)

A regression test that passes when the feature is broken is worse than no test, so I checked it directly:

| Mutation in `packages/core/src/workflow/lifecycle.ts` | Result |
|---|---|
| baseline | ✅ passes |
| `matchesWorkflowState(issueState, planningStates)` → `planningStates.includes(issueState)` | ❌ **fails** — `expected '## Engine-Enforced Run Identity…' to contain 'execution_phase=planning'` |
| `normalizeWorkflowState` → identity | ❌ fails (at `dispatched === 1`) |

The targeted mutation fails on the exact assertion the test exists to protect. Good.

### Verification run locally (HEAD `64fa40f`)

- `pnpm build` ✅
- `pnpm lint` ✅
- `pnpm typecheck` ✅
- `pnpm test` ✅ — orchestrator 326/326, full monorepo green
- `pnpm format` — 80 pre-existing files flagged repo-wide; **`service.test.ts` is not among them**, so this PR is clean and the drift is out of scope
- CI on this head: Test ✅, Container Smoke ✅

No changeset needed — test-only, no behavior change, consistent with `CLAUDE.md`.

### No over-engineering

The `createReadyStateWorkflow` signature change is the minimum needed: one optional param defaulting to `[]`, all five existing callers unchanged. No speculative helpers, no scope creep.

### Smoke test note

Docker is unavailable in my review environment, so I could not re-run `./e2e/run-e2e.sh prompt-phase`. I did confirm the caveat in the PR description is real rather than an excuse: `e2e/run-e2e.sh:259` and `:263` grep `docker logs symphony-e2e` for `api-progress readback…"state":"Done"` and `scenario=prompt-phase`, but `e2e/stub-worker.ts` writes both via `console.error` (lines 256, 282) → worker **stderr** → `worker.log`, which the runner itself dumps separately at `run-e2e.sh:179`. So those two guards are stale for **both** the `api-progress` and `prompt-phase` scenarios and will report a false failure after the real assertions have already passed.

That defect predates this PR (it came in with #622, commit `cbdcdcc`) and fixing it here would widen a test-only PR, so I am not asking for it. **Suggest filing a follow-up issue** to point those greps at `worker.log`, otherwise the next person to run these scenarios will chase a phantom failure.

---

packages/orchestrator/src/service.test.ts
moncher-dev: Acknowledged. The current assertion already verifies the normalized phase is present; this non-blocking exactness tightening is intentionally left out to keep the approved test-only change narrowly scoped.

packages/orchestrator/src/service.test.ts
moncher-dev: Acknowledged. This matches the surrounding test style, and I am keeping the approved test-only change focused on dispatch prompt wiring rather than making unrelated fixture-cleanup adjustments.

packages/orchestrator/src/service.test.ts
moncher-dev: Acknowledged. The trailing newline is deliberate so `hooks:` remains a top-level YAML key. I am leaving the non-blocking explanatory-comment nit out of this narrowly scoped, already-approved test change.

Issue #675, opened 2026-08-27
docs: label existing divergences (C1–C13) as spec divergences and fix documentation errors (D1–D8)


moncher-dev:
Epic: #651
리포트 항목: **C1–C13, D1–D8** — `docs/reports/2026-08-28-upstream-spec-drift-research.md` (PR #650)

## 문제

AGENTS.md와 새 스펙(implementation-defined는 MUST document)은 divergence를 명시적으로 표기하라고 한다. 아래 동작은 문서에 "동작"으로만 있거나 아예 없다. 별도로 문서가 코드와 반대인 항목이 8건이다.

## 라벨링할 divergence (리포트 §4.1)

| ID | 내용 | 위치 |
| --- | --- | --- |
| C1 | 파일시스템 영속 스케줄러 상태(claims/retryEntry/run 레코드가 재시작 생존; 스펙 §14.3은 미복원) | ADR 또는 `docs/architecture.md` §3 |
| C2 | codex를 shell 없이 토큰화·`codex` 실행파일만 허용 (스펙 `bash -lc`) | ADR / `docs/configuration.md` |
| C3 | hook은 스크립트 경로만, shell 문법 거부, `SYMPHONY_ALLOW_WORKFLOW_HOOKS=1` 게이트 | `docs/configuration.md` + README |
| C4 | 신뢰 posture(`never`+`danger-full-access`, Claude `bypassPermissions`, user-input 즉시 실패) | README "Security posture" (§15.1) |
| C5 | control-plane API 형태(bearer, `:4680`, `--http/--web`, camelCase 스냅샷, `/refresh` 응답, 오류 envelope, worker POST 라우트, 포트 auto-increment) | `packages/control-plane/README.md` "Spec mapping" |
| C6 | `stall_timeout_ms <= 0`일 때 30분 하드코딩 fallback (`service.ts:160`) | `docs/configuration.md` |
| C7 | `retry_queued`가 슬롯 점유; 30초 exit grace | `docs/architecture.md` §3 |
| C9 | GitHub 합성 `Archived` 상태, blocker 상태 재작성 | 어댑터 프로필(S22)로 이관 |
| C10 | 트래커 설정이 `config.json`에 고정돼 live 반영 안 됨 | `docs/configuration.md` reload 절 |
| C11 | 스냅샷 오류 어휘 `stopped/degraded/idle/running` | control-plane README |
[...]

PR #681, opened 2026-08-27
docs(adr): define host-side tool isolation

### LINKED ISSUE #671: design: agent-tool isolation model — host-side execution via the token broker vs a documented MCP-subprocess divergence (ADR)
Epic: #651
리포트 항목: **A13 결정** — `docs/reports/2026-08-28-upstream-spec-drift-research.md` (PR #650)

## 결정 (2026-08-28 확정)

**(a) host-side 실행을 2단계로 도입한다.** 이 이슈의 범위는 ADR 작성으로 좁힌다.

- 1단계 = #672: 어댑터 `secretEnvironmentNames()` 선언 + Codex/Claude child env에서 raw 트래커 토큰 제거 + Claude `mcp.json`에 토큰 미기록. (a)/(b) 어느 쪽이든 필요한 공통 작업이라 먼저 간다.
- 2단계 = #673: Codex는 app-server dynamic tool(`item/tool/call`)을 워커가 host-side로 처리, Claude는 MCP 서버를 워커/오케스트레이터가 띄우고 HTTP transport로 노출(`mcp.json`에는 URL + 세션 토큰만). 툴 호출 인자로 `issue.id`/`nativeRef` 컨텍스트 전달.

이유: 사업화 시 고객 저장소에서 에이전트를 돌리므로 "에이전트가 토큰을 볼 수 없다"가 판매 조건이 된다. (b) MCP-subprocess 유지는 지금 편하지만 스펙 MUST 위반이 남고 Linear는 broker가 없어 raw 토큰이 필요해 결국 다시 해야 한다.

## 배경

새 스펙 §10.5/§15.3(MUST): provider-native 툴은 Symphony **host 프로세스**에서 어댑터 자격증명으로 실행하고, 코딩 에이전트 child는 raw 트래커 토큰을 env/디스크에서 읽어서는 안 된다. 현재는 툴이 에이전트가 spawn하는 MCP subprocess이고 토큰이 child env(`packages/runtime-codex/src/runtime.ts:598-619, 795-809`)와 `mcp.json`(`packages/runtime-claude/src/mcp-compose.ts:58-62, 119-126`)으로 전달된다.

재사용할 부품: GitHub 토큰 broker(`docs/configuration.md:168-198`, `GITHUB_TOKEN_BROKER_SECRET`), runtime env allowlist(`runtime.ts:110-122`, `service.ts:102-113`), 워커의 dead dynamic-tool 경로(`worker/src/index.ts:1255-1304, 2176-2219` — 실제 프로토콜로 교체 대상).

## 산출물

- `docs/adr/2026-XX-XX_agent-tool-isolation.md`:
  - 결정과 근거, 기각한 (b)의 트레이드오프
  - 런타임별 transport: Codex dynamic tool 흐름(advertise → `item/tool/call` → 워커 실행 → 응답), Claude host-side MCP over HTTP(포트/세션 토큰/수명)
  - 시크릿 흐름 다이어그램: 오케스트레이터 → 워커(broker/자격증명) → 툴 실행, child에는 도달하지 않음
  - Linear 자격증명 공급 방식(broker 확장 또는 워커 프로세스 보유)
  - README "Security posture" 섹션 초안(§15.1) — #675와 조율
- #672/#673 본문의 구현 범위 확정 코멘트

## 완료 기준

- ADR 병합, #672/#673에 반영 코멘트

moncher-dev:
## Issues — Closed #671

Fixes #671

## TL;DR

- Accepts host-side provider-native tool execution as the target architecture.
- Defines phase 1 as full child credential-boundary removal, including Git-auth and host-home isolation.
- Defines phase 2 as bounded, issue-aware Codex and Claude host-side transports.

## Change-point diagram

```text
current: child environment / mcp.json / host home → tracker or broker credential
                                                       │
phase 1: strip + isolate home + host Git transport ────┴─→ disable child-owned MCP tools
                                                       │
phase 2: orchestrator → worker host → bounded adapter tool → result only → child
```

## Start here

1. `docs/adr/2026-08-28_agent-tool-isolation.md` — decision, full child credential boundary, phase transports, and divergence.
2. `README.md` — current trusted-environment posture and target host-owned Git/auth boundary.
3. `docs/designs/2026-08-11-standalone-project-model-design.md` — corrected MCP/workflow-transition relationship.
4. #672 and #673 — implementation-scope confirmations.

## Risks & rollback

- This is documentation only; runtime credential paths remain until #672 and #673 land.
- Phase 1 temporarily removes child-authenticated Git and provider-native MCP tools; the worker performs bounded post-run Git transport.
- Reverting this commit removes the clarified ADR text without changing runtime behavior.

## Changed files

- `docs/adr/2026-08-28_agent-tool-isolation.md`
- `README.md`
- `docs/designs/2026-08-11-standalone-project-model-design.md`

## Evidence

- `pnpm lint` — pass
- `pnpm test` — pass
- `pnpm typecheck` — pass
- `pnpm build` — pass
- `pnpm exec prettier --check README.md docs/adr/2026-08-28_agent-tool-isolation.md docs/designs/2026-08-11-standalone-project-model-design.md` — pass
- `git diff --check` — pass

- [ ] Confirm the phase-1 host Git transport, isolated-home boundary, and temporary child-command limitation are actionable for #672.
- [ ] Deploy, external URL smoke tests, and manual UX are not in this documentation-only scope.

## Human Validation

- [ ] Confirm the full child credential boundary is unambiguous for tenant operators.
- [ ] Confirm the bounded, issue-aware host-side transport contract is actionable for #673.


README.md
```diff
@@ -959,6 +959,23 @@ pnpm typecheck
...
 
+## Security posture
+
+GitHub Symphony treats the coding-agent runtime and its workspace as a separate
+trust boundary. Tracker credentials are held by the Symphony host or a
+host-side credential broker; agents receive tool schemas and results, not raw
+tracker tokens. Provider-native tools are scoped to the selected adapter and
+the current normalized issue. Deployments should still use least-privilege
+credentials, dedicated workspace permissions, loopback-only local services,
+and the runtime approval/sandbox policy appropriate for their environment.
```
hojinzs: **P1 — this paragraph states the *target* posture in the present tense, and it is not true of the shipped code.**

"Tracker credentials are held by the Symphony host or a host-side credential broker; agents receive tool schemas and results, not raw tracker tokens" reads as a statement of current fact. Today it is false on both runtimes:

- `packages/runtime-codex/src/runtime.ts:598-619` puts `LINEAR_AUTHORIZATION` / `LINEAR_API_KEY` into the coding-agent child environment, and `:795-809` adds `GITHUB_GRAPHQL_TOKEN` **and** `GITHUB_TOKEN_BROKER_SECRET`.
- `packages/runtime-claude/src/mcp-compose.ts:119-126` writes those same values into the generated `mcp.json`.

Note the broker case is not a mitigation here: the child receives `GITHUB_TOKEN_BROKER_SECRET` + `GITHUB_TOKEN_BROKER_URL`, so it can mint tracker tokens itself. "Held by ... a host-side credential broker" is the one phrasing most likely to be read as already-safe, and it is the least safe of the two paths as implemented.

The following paragraph does disclose the divergence, but a README security section gets quoted a paragraph at a time, and this is the paragraph a prospective tenant will quote. Please tense-split it so the guarantee and the status are not separable — e.g. lead with what is true now ("Today, tracker credentials are passed to the coding-agent process; Symphony is not yet suitable for untrusted agents"), then describe the host-side boundary as the target the phased work delivers.

Related, from the same section: the ADR calls this text "required by §15.1", but §15.1's two SHOULDs are (a) state clearly whether the implementation targets trusted or restrictive environments, and (b) state clearly whether it relies on auto-approval, operator approval, or sandboxing. §10.5 additionally has a hard "Each implementation MUST document its chosen approval, sandbox, and operator-confirmation posture." This section does neither — "the runtime approval/sandbox policy appropriate for their environment" delegates the question back to the deployer rather than stating Symphony's own posture. Worth adding one sentence naming the actual default posture (auto-approve, trusted environment) so the §10.5 MUST is genuinely discharged.

---

docs/adr/2026-08-28_agent-tool-isolation.md
```diff
@@ -0,0 +1,195 @@
...
+The worker executes the selected adapter's tool implementation with its
+configured adapter credential and a normalized, internal tool context:
+
+```ts
+{
+  issue: {
+    (id, identifier, nativeRef);
+  }
+}
+```
```
hojinzs, line 88: **P2 — this `ts` block is not the type you meant; Prettier rewrote it.**

```ts
{
  issue: {
    (id, identifier, nativeRef);
  }
}
```

Written as a `ts` *statement*, `{ issue: { id, identifier, nativeRef } }` parses as a block containing a labelled statement `issue:` whose body is a block containing the sequence expression `id, identifier, nativeRef` — which is exactly what Prettier printed back. That is why `pnpm format` passed on it: the block is valid TypeScript, it just doesn't mean anything close to the context shape.

The scope-confirmation comment on #673 states the intended shape correctly as `{ issue: { id, identifier, nativeRef } }`, so the ADR — the artifact that actually merges — is currently the least accurate copy of the contract two implementation issues point at. Make it a declaration so Prettier keeps it:

```ts
type TrackerToolContext = {
  issue: {
    id: string;
    identifier: string;
    nativeRef: unknown;
  };
};
```

Typing `nativeRef` here would also let the ADR say in the type what the prose says below it ("`nativeRef` remains opaque to the scheduler").

---

docs/adr/2026-08-28_agent-tool-isolation.md
```diff
@@ -0,0 +1,195 @@
...
+the Codex and Claude runtime paths must not add them back. Generated Claude
+`mcp.json` must contain no literal tracker credential.
+
+This phase is deliberately a security-boundary prerequisite, not completion of
+host-side execution. Where a legacy MCP subprocess still needs a credential to
+function, the host supplies it over a non-environment, non-workspace channel
+such as inherited stdin or a file descriptor. That short-lived compatibility
+path remains a documented divergence until phase 2, because the agent still
+initiates the MCP subprocess rather than receiving host-executed tool results.
+
```
hojinzs, line 60: **P2 — the phase-1 compatibility channel names two mechanisms that are unavailable for the case that needs them, and #672 is already scoped to build it.**

"the host supplies it over a non-environment, non-workspace channel such as inherited stdin or a file descriptor" — for the Claude path this is the stdio MCP server described in `mcp-compose.ts`, and there:

- **stdin is taken.** The MCP stdio transport *is* stdin/stdout of that process; the host cannot put a credential on it.
- **fd inheritance doesn't reach it.** The worker does not spawn the MCP server — Claude spawns it from `mcp.json`. A host fd would have to survive worker → Claude child → MCP grandchild, and Claude Code does not forward arbitrary descriptors to MCP servers.

Codex is the same shape: the child owns the subprocess, the host doesn't hold either end.

This matters more than a wording nit because the #672 scope comment repeats the mechanism verbatim ("for example stdin or an inherited file descriptor"), so whoever picks up phase 1 will discover it is unbuildable and will be blocked on re-deciding the boundary — with an Accepted ADR saying it was already decided.

Two options that do work and stay non-env / non-workspace, either of which would make the ADR actionable:

1. a host-owned credential file **outside the workspace root**, mode `0600`, path passed via a non-secret env var (`mcp-compose.ts` already has `ensureSecureConfigParent` / `chmodExistingSecretFile` / `0o600` plumbing to build on) — weaker than phase 2 but honestly weaker, and it removes the value from `mcp.json` and the child env as §15.3 requires; or
2. state that where no such channel exists, phase 1 **disables** provider-native tracker tools for that runtime until #673, and name which tracker-write paths regress in the interim.

Please also fold the choice back into the #672 comment so the two don't diverge.

---

docs/adr/2026-08-28_agent-tool-isolation.md
```diff
@@ -0,0 +1,195 @@
...
+adopted as-is.
+
+## Decision
+
+Adopt host-side provider-native tool execution in two implementation phases.
+The host is the worker/orchestrator side of the Symphony runtime boundary; the
+coding-agent runtime is the untrusted child side. Adapter credentials remain
+available only to the host process or a host-owned broker.
+
+### Phase 1 — remove tracker secrets from the child boundary (#672)
```
hojinzs, line 37: **P2 — phase 1 says "tracker secrets" / "tracker credential" throughout, which leaves `GITHUB_TOKEN_BROKER_SECRET` ambiguously out of scope.**

The credential-flow diagram below gets this right ("no raw GitHub, Linear, or broker secret"), and the #672 scope comment gets it right for `mcp.json` ("no literal tracker **or broker** credential"). But this normative paragraph — the one an implementer reads to decide what `secretEnvironmentNames()` returns — only says *tracker* secret, and the child-env bullet in the #672 comment likewise says "every declared **tracker** secret".

A literal reading strips `GITHUB_GRAPHQL_TOKEN` and keeps `GITHUB_TOKEN_BROKER_SECRET` + `GITHUB_TOKEN_BROKER_URL` (`runtime.ts:795-809`), which leaves the child able to mint tracker tokens on demand — phase 1 would report done while the boundary it exists to establish is still open. That is the exact failure mode worth spending one word on.

Suggest naming the broker credential explicitly here, e.g. "…declare `secretEnvironmentNames()`, covering both raw tracker credentials and broker credentials that can be exchanged for one."

---

docs/README.md
```diff
@@ -53,3 +53,5 @@ Architecture documentation scoped to a single package lives in that package's
 
 See the [adr/](adr/) directory for the ADR list. When a newer decision replaces
 an older one, they are linked via the `Supersedes` header.
+
+- [2026-08-28 Agent Tool Isolation](adr/2026-08-28_agent-tool-isolation.md) — Accepted; host-side provider-native tool execution, runtime transports, and the temporary MCP-subprocess divergence.
```
hojinzs: **P3 — this turns a deliberate "no index" into a one-of-twelve index.**

The two lines directly above say "See the [adr/](adr/) directory for the ADR list" — i.e. this section intentionally does *not* enumerate. Appending a single entry after that sentence leaves `docs/adr/` holding 12 ADRs with exactly one listed, and the next ADR author has no signal about whether they're expected to add a row. Note the `designs/` and `reports/` sections above use full tables, so a partial bare-bullet list is a third convention.

Also worth noting this entry wasn't in the issue's 산출물 list — the ADR, the README draft, and the #672/#673 comments were. Either drop it (the `adr/` directory listing already covers discovery), or populate the section as a table like its siblings.

---

docs/adr/2026-08-28_agent-tool-isolation.md
```diff
@@ -0,0 +1,195 @@
...
+> and the current normalized issue. Deployments should still use least-privilege
+> credentials, dedicated workspace permissions, loopback-only local services,
+> and the runtime approval/sandbox policy appropriate for their environment.
+>
+> The host-side tool transport is being introduced in phases. The current
+> MCP-subprocess implementation is a documented divergence and is not suitable
+> for untrusted agents until phase 1 removes raw tracker values from
+> coding-agent environments and workspace configuration. Operators should track
+> the remaining divergence in this ADR until the host-side transport is
+> deployed.
```
hojinzs: **P3 — the "draft for #675" and the shipped README section are two copies, and they have already drifted inside this one PR.**

This blockquote and `README.md:962-977` differ in three places: "**The** host-side tool transport" vs "Host-side tool transport"; the blockquote's closing "Operators should track the remaining divergence in this ADR…" vs the README's "See [ADR 2026-08-28](…) for the target posture, remaining divergence, and removal plan."

Drift at birth is a good predictor of drift later, and #675 is being handed a "draft" that is no longer the text in the tree. There's also a scope wrinkle: the issue asked for a *draft* coordinated with #675, but the PR ships the section live into README while this heading still says "#675 owns final wording and placement" — so it's unclear which of the two is authoritative right now.

Simplest fix: keep the section in README as shipped, and replace this blockquote with a pointer ("§15.1 posture text now lives at `README.md` → Security posture; #675 owns final wording and placement"). One copy, and #675 still gets the handoff.

nit, same section: line 141 calls the current state "a §10.5/§15.3 MUST violation". §15.3's "Execute provider-native tracker tools in the Symphony host process" is unconditional, so that one holds. §10.5's host-side MUSTs sit inside the "Optional provider-native agent tool extension" block gated on "An adapter **MAY** expose provider-native tools to the app-server session" — since the current implementation doesn't advertise via app-server at all, §10.5 isn't strictly violated yet. Attributing it to §15.3 alone is both stronger and more accurate for a doc two implementation issues cite as authority.

---

hojinzs (changes requested):

Documentation-only ADR closing #671. The decision itself is right and the write-up is unusually careful about not overclaiming — the phase split, the rejected alternative, and the explicit divergence framing are all solid. I verified every factual claim in it against the tree rather than taking the PR body's word for it, and the claims hold. One README paragraph and one unbuildable phase-1 mechanism need fixing before merge.

### Requirement coverage (#671 산출물)

| Deliverable | Status |
| --- | --- |
| 결정 + 근거, 기각한 (b)의 트레이드오프 | ✅ |
| Codex transport (advertise → `item/tool/call` → 워커 실행 → 응답) | ✅ |
| Claude host-side MCP over HTTP (포트 / 세션 토큰 / 수명) | ✅ loopback, ephemeral port, per-run capability token, expiry + revoke + teardown on crash-recovery all covered |
| 시크릿 흐름 다이어그램 | ✅ |
| Linear 자격증명 공급 방식 | ✅ broker 확장 vs in-worker, with a stated default |
| README "Security posture" 초안 (§15.1) | ⚠️ present, but see P1 |
| #672 / #673 구현 범위 확정 코멘트 | ✅ both posted (#672 comment, #673 comment) |

완료 기준 is otherwise met once the ADR merges. No over-engineering: the ADR stays inside ADR scope and defers implementation to #672/#673. The one thing outside the requested scope is the `docs/README.md` index entry (P3 below).

### Verification performed

No API or UI in this PR, so "smoke test" here means checking the ADR against the artifacts it describes — a docs PR fails by being *wrong*, not by not running:

- **Spec citations** — §10.5, §11.5, §15.1, §15.3, §15.5, §17.5 all exist and say what the ADR attributes to them, with one nuance noted inline (§10.5's MUSTs are gated on an optional extension; §15.3's are not).
- **Code citations** — the child-env leak (`runtime-codex/src/runtime.ts:598-619, 795-809`), the `mcp.json` leak (`runtime-claude/src/mcp-compose.ts:119-126`), and the dead dynamic-tool path (`worker/src/index.ts:1261, 1284, 1296, 1310, 2176`) are all real and at the cited lines. The ADR's description of the current state is accurate.
- **Formatting / links** — `prettier --check` passes on all three files; relative links resolve; ADR header matches the existing convention (`Date` / `Status` / `Related Spec`).
- **CI** — Test and Container Smoke both green on `36bd596`.

### Findings

| Pri | Finding |
| --- | --- |
| **P1** | `README.md:964-970` asserts the *target* posture in the present tense. Agents do receive raw tracker tokens today, and the broker path hands the child `GITHUB_TOKEN_BROKER_SECRET` — the phrase most likely to read as safe is the least safe as implemented. |
| **P2** | Phase 1's compatibility channel ("inherited stdin or a file descriptor") is unavailable for the stdio MCP subprocess that needs it — stdin is the JSON-RPC transport, and the host doesn't spawn that process. #672 is already scoped to build it. |
| **P2** | Phase 1 says "tracker secret" throughout, leaving `GITHUB_TOKEN_BROKER_SECRET` ambiguously out of `secretEnvironmentNames()` — a literal reading ships phase 1 with the boundary still open. |
| **P2** | The `ts` context-shape block was reformatted by Prettier into a labelled block + sequence expression; the ADR now states the contract less accurately than the #673 comment does. |
| **P3** | `docs/README.md` adds 1 of 12 ADRs beneath a line that says the section intentionally doesn't enumerate. |
| **P3** | The README posture text exists in two copies (ADR blockquote + README) that already differ in three places. Plus a §10.5-vs-§15.3 attribution nit. |

Only P1 and the first P2 are merge-blocking for me — P1 because this is the paragraph a tenant quotes, and the P2 because an Accepted ADR that specifies an unbuildable mechanism will stall #672 with the decision nominally already made. The rest can ride along or be waved off with a reason.

---

README.md
```diff
@@ -959,6 +959,23 @@ pnpm typecheck
 pnpm build
 ```
 
+## Security posture
+
+GitHub Symphony treats the coding-agent runtime and its workspace as a separate
+trust boundary. Tracker credentials are held by the Symphony host or a
+host-side credential broker; agents receive tool schemas and results, not raw
+tracker tokens. Provider-native tools are scoped to the selected adapter and
+the current normalized issue. Deployments should still use least-privilege
```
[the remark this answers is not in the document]
moncher-dev: Resolved in the README and ADR: issue scoping is now explicitly a phase-2,
host-enforced requirement. The current implementation is described as a trusted
environment and does not claim that arbitrary GraphQL forwarding is bounded.

docs/adr/2026-08-28_agent-tool-isolation.md
```diff
@@ -0,0 +1,195 @@
...
+The host is the worker/orchestrator side of the Symphony runtime boundary; the
+coding-agent runtime is the untrusted child side. Adapter credentials remain
+available only to the host process or a host-owned broker.
+
+### Phase 1 — remove tracker secrets from the child boundary (#672)
+
+Every tracker adapter will declare `secretEnvironmentNames()`. Local and
+remote launchers will remove those names from coding-agent child environments;
+the Codex and Claude runtime paths must not add them back. Generated Claude
+`mcp.json` must contain no literal tracker credential.
```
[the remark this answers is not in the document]
moncher-dev: Resolved in the ADR and #672 scope correction: phase 1 must preserve
environment-indirection source names as secret metadata and strip those names
along with fixed tracker and broker credentials.

docs/adr/2026-08-28_agent-tool-isolation.md
```diff
@@ -0,0 +1,195 @@
...
+
+This phase is deliberately a security-boundary prerequisite, not completion of
+host-side execution. Where a legacy MCP subprocess still needs a credential to
+function, the host supplies it over a non-environment, non-workspace channel
+such as inherited stdin or a file descriptor. That short-lived compatibility
```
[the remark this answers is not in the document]
moncher-dev: Resolved by rejecting the unsafe bridge. The ADR now says agent-owned legacy
MCP has no safe credential channel and phase 1 disables provider-native MCP
tool exposure until #673 provides the host-owned transport.

README.md
moncher-dev: Resolved: the README now leads with the current trusted-environment limitation,
explicitly names the current default approval/sandbox posture, and describes
host-side isolation as the target rather than a current guarantee.

docs/adr/2026-08-28_agent-tool-isolation.md
moncher-dev: Resolved: the ADR now declares `TrackerToolContext`, including opaque
`nativeRef: unknown`, so the documented contract is valid TypeScript.

docs/adr/2026-08-28_agent-tool-isolation.md
moncher-dev: Resolved: phase 1 now disables agent-owned provider-native MCP tool exposure
rather than specifying an unavailable stdin or inherited-descriptor bridge.
The corrected decision is also posted to #672.

docs/adr/2026-08-28_agent-tool-isolation.md
moncher-dev: Resolved: phase 1 now explicitly covers raw tracker credentials, broker
credentials that can exchange for tracker tokens, and environment-indirection
source variables; the #672 scope correction uses the same wording.

docs/README.md
moncher-dev: Resolved by removing the one-off ADR index entry; `docs/README.md` again points
to the ADR directory without maintaining a partial list.

docs/adr/2026-08-28_agent-tool-isolation.md
moncher-dev: Resolved: the ADR now links to the single README posture text instead of
duplicating it, and attributes the current host-execution divergence to the
unconditional §15.3 requirement.

moncher-dev:
Rework is complete and ready for re-review.

- Corrected the current-versus-target security posture and documented the actual approval/sandbox defaults.
- Replaced the unavailable child-owned credential bridge with phase-1 MCP tool disablement until #673.
- Added required secret-source, broker-secret, bounded issue-context, and ADR/README ownership clarifications.
- Replied to every inline review comment and re-ran the full validation suite.

docs/adr/2026-08-28_agent-tool-isolation.md
```diff
@@ -0,0 +1,192 @@
...
+
+### Phase 1 — remove tracker secrets from the child boundary (#672)
+
+Every tracker adapter will declare `secretEnvironmentNames()`. The declaration
+must cover raw tracker credentials, broker credentials that can exchange for a
+tracker token, and the source names of supported environment indirections (for
+example, `$MY_LINEAR_SECRET`). Configuration parsing must preserve those source
+names as secret metadata rather than only their resolved values. Local and
+remote launchers will remove every declared name from coding-agent child
+environments; the Codex and Claude runtime paths must not add them back.
```
hojinzs: **P1 — as written, phase 1 breaks `git push` on the Codex runtime, and the ADR never mentions the git-credential channel in either direction.**

"remove every declared name from coding-agent child environments" collides with `packages/runtime-codex/src/runtime.ts:791-813`, which configures the child's git:

```
GIT_CONFIG_KEY_0: "credential.helper",
GIT_CONFIG_VALUE_0: `!node .../git-credential-helper.js`,
```

That helper (`git-credential-helper.ts:19-47` → `resolveGitHubGraphQLToken`) reads its token from `GITHUB_GRAPHQL_TOKEN` / `GITHUB_TOKEN_BROKER_URL` / `GITHUB_TOKEN_BROKER_SECRET` — **in the child environment**, the exact names phase 1 declares and strips. Strip them and HTTPS git auth in the agent goes away. This isn't an edge path: the injected `push` skill is plain `git push origin <branch>` (`packages/cli/src/skills/templates/push.ts:20-21`), so phase 1 lands on the main workflow. And phase 2 restores *tracker tools*, not git credentials, so nothing later in the ADR brings it back.

The same helper is the other half of the problem: because it is invoked by the child's own git, an agent can run `git credential fill` and read the token out of it directly. So it is simultaneously a raw-credential channel the boundary doesn't close and a dependency the boundary would break.

This is the same failure shape as the round-1 stdin/fd finding: whoever picks up #672 hits it on day one, with an Accepted ADR saying the boundary was already decided. Please spend one paragraph in phase 1 deciding it — either:

1. the credential helper moves host-side too (the host owns the helper process or proxies the remote), so `git push` keeps working and the token never sits in child env; or
2. the git-credential env vars are an explicit, named carve-out from `secretEnvironmentNames()` for phase 1, with the residual exposure recorded as part of the documented divergence and closed by a follow-up.

Either is fine. What doesn't work is leaving it unstated, because both literal readings of the current text are wrong — strip and push breaks, don't strip and the boundary phase 1 exists to establish is still open.

---

docs/adr/2026-08-28_agent-tool-isolation.md
```diff
@@ -0,0 +1,192 @@
...
+- it does not provide the host's normalized issue context or a single place to
+  enforce adapter-specific scope, error, and rate-limit contracts.
+
+The phase-1 compatibility channel is therefore temporary and explicitly
+tracked as a divergence, not an alternative architecture.
+
+## Consequences
+
+- Customer repository agents can use provider-native tracker operations without
+  possessing the corresponding tracker token.
```
hojinzs: **P2 — this consequence is not true after phase 2 either, because the agent still inherits `HOME`.**

`HOME` is in the safe-env allowlist of all three launchers — `runtime-codex/src/runtime.ts:113`, `runtime-claude/src/adapter.ts:551`, `orchestrator/src/service.ts:107` — so the agent reads `~/.config/gh/hosts.yml` and gets a GitHub token no matter what happens to env vars and `mcp.json`. That isn't incidental: `runtime-claude/src/preflight.ts:238-245` *fails the run* unless `gh auth status` succeeds, and the `land` skill is built on `gh pr view` / `gh pr merge` (`packages/cli/src/skills/templates/land.ts:21,34,56`). The `gh` credential is a load-bearing part of the current design.

So "Customer repository agents can use provider-native tracker operations without possessing the corresponding tracker token" overstates what the two phases deliver. The credential-flow block above has the same gap — "The child environment and child-readable workspace files receive no raw GitHub, Linear, or broker secret" is true of the env and the workspace, and false of `$HOME`.

Two ways to make it accurate, both cheap:

- narrow the claim to the adapter credential ("without possessing the tracker **adapter** credential"), and add one line to the credential-flow section naming inherited `HOME` / `gh` CLI auth as a separate credential channel outside this ADR's boundary; or
- state it as in-scope for a follow-up if you intend to close it.

I'd take the first — this ADR is about the tool boundary, and `gh`-CLI auth is its own decision. But an unqualified "without possessing the corresponding tracker token" in the Consequences of a security ADR is the sentence that gets quoted, so it should be the one that survives contact with `$HOME`.

---

docs/adr/2026-08-28_agent-tool-isolation.md
```diff
@@ -0,0 +1,192 @@
...
+Generated Claude `mcp.json` must contain no literal tracker or broker
+credential.
+
+This phase is deliberately a security-boundary prerequisite, not completion of
+host-side execution. A legacy MCP subprocess started by the coding agent has no
+safe credential channel: its stdio is the JSON-RPC transport, and any inherited
+descriptor would be readable by the untrusted parent. Phase 1 therefore disables
+provider-native MCP tool exposure for a runtime once its raw credential is
+removed, rather than passing a credential through the agent. It preserves
+non-agent host-side tracker operations, but agent tool access returns only in
```
hojinzs, line 57: **P3 — option 2 was taken, but the "name which paths regress" half of it is still missing.**

The round-1 ask was: state that phase 1 disables the tools, *and* name which tracker-write paths regress in the interim. The disablement decision landed cleanly; "It preserves non-agent host-side tracker operations, but agent tool access returns only in phase 2" is the only thing standing in for the blast radius.

Good news, and the reason this is P3 rather than P2 — from what I can see the radius is small, and saying so makes the phase-1 decision much easier to accept:

- **Workflow state transitions do not regress.** They already go through the orchestrator, not the tools: the `gh-project` skill sends a `transition-request` authenticated with `SYMPHONY_ORCHESTRATOR_TOKEN` (`packages/cli/src/skills/templates/gh-project.ts:12,23,48-66`).
- **Comments, issues, PR flow do not regress.** The injected skills use `gh` CLI, not `github_graphql` (`gh-project.ts:87-113`, `land.ts`).
- I could not find any injected skill that names `github_graphql` or `linear_graphql` at all, even though `mcp-compose.ts:109-133` always composes them.

Worth one clause — "no injected skill currently invokes the provider-native tools; workflow transitions run through the orchestrator transition endpoint and are unaffected" — because right now the ADR's phrasing reads as a bigger regression than it is.

Related, and worth fixing while you're here: `docs/designs/2026-08-11-standalone-project-model-design.md:118` still says *"If built-in tools get shadowed, workflow transitions break, so they are protected via reserved names."* That contradicts the current `gh-project` orchestrator-transition design and would make a reader think phase 1 breaks transitions. The ADR is a reasonable place to record which of the two is current.

---

hojinzs (changes requested):
## Review — round 2/6 · Request changes (one finding)

All six round-1 findings are fixed, and I checked the tree rather than taking the reply threads' word for it. This round has one new blocker, found by tracing phase 1's stripping rule against the launchers instead of reading it as prose.

### Round-1 rework verification

| # | Finding | Verified in `4f57ccf` |
| --- | --- | --- |
| P1 | README asserted target posture in present tense | ✅ `README.md:962-977` now leads with "Today … intended for trusted, operator-controlled environments … not suitable for untrusted agents until…" |
| P1 | §10.5 approval/sandbox posture MUST not discharged | ✅ defaults now named — and they check out: `never` + `danger-full-access` (`worker/src/codex-policy.ts:46,51`), `bypassPermissions` (`runtime-claude/src/argv.ts:12,86`) |
| P2 | unbuildable stdin/fd credential bridge | ✅ replaced with tool disablement + the reason it has no safe channel |
| P2 | "tracker secret" left `GITHUB_TOKEN_BROKER_SECRET` ambiguous | ✅ now covers raw credentials, broker credentials, **and** indirection source names |
| P2 | `ts` block Prettier-mangled into a labelled statement | ✅ real `type TrackerToolContext` declaration, `nativeRef: unknown` |
| P3 | partial ADR index in `docs/README.md` | ✅ reverted; PR is 2 files now |
| P3 | posture text duplicated ADR↔README + §10.5/§15.3 attribution | ✅ ADR links to the single README copy; divergence attributed to the unconditional §15.3 |

`prettier --check` passes on both files, the README anchor `../../README.md#security-posture` resolves, and CI is green on `4f57ccf` (Test, Container Smoke).

### New findings

| Pri | Finding |
| --- | --- |
| **P1** | Phase 1's "remove every declared name from child environments" strips the exact vars the child's git credential helper reads (`runtime.ts:791-813`), so `git push` breaks — and that same helper is a raw-token channel the boundary doesn't close. Both readings of the current text are wrong. |
| **P2** | The Consequences claim ("without possessing the corresponding tracker token") isn't true after phase 2 either: `HOME` is allowlisted in all three launchers and `gh auth` is required by preflight, so the agent reads a GitHub token from `~/.config/gh/hosts.yml`. |
| **P3** | Option 2 was taken but its "name which paths regress" half is missing — and the radius looks *smaller* than the text implies, which is worth saying. Includes a stale contradicting line in an existing design doc. |

Only the **P1** is merge-blocking. It's the round-1 stdin/fd problem in a new place: an Accepted ADR that specifies something #672 cannot implement as written, so phase 1 stalls with the decision nominally already made. One paragraph choosing host-side helper vs. explicit carve-out settles it.

P2 and P3 are wording/scoping and can ride along or be waved off with a reason. **Land the P1 paragraph and I'll approve on the next round** — the decision in this ADR is right and I don't want to keep it in review over prose.

nit: the PR body still lists `docs/README.md` under "Changed files"; the revert correctly removed it.

---

docs/adr/2026-08-28_agent-tool-isolation.md
moncher-dev: Resolved in 759492c. Phase 1 now removes the child Git credential helper and moves authenticated fetch/push to a bounded worker-host operation; it explicitly names the temporary loss of child `git fetch`, `git pull`, `git push`, and host-login-dependent `gh` commands. This closes the raw-token path rather than carving it out.

docs/adr/2026-08-28_agent-tool-isolation.md
moncher-dev: Resolved in 759492c. The target boundary now requires an isolated child home/configuration directory, removes `gh auth` preflight and inherited credential-helper access from the child, and keeps authenticated Git transport on the worker host. The consequence and README wording now include both the adapter-token and host-login boundary.

docs/adr/2026-08-28_agent-tool-isolation.md
moncher-dev: Resolved in 759492c. The ADR now states that no injected workflow skill invokes the provider-native MCP tools and that workflow transitions use the orchestrator transition endpoint, so they are unaffected. The standalone-project design now makes the same correction.

moncher-dev:
Rework is complete and ready for re-review.

- Closed the child Git credential-helper and host-home/`gh auth` token paths in the phase-1 and phase-2 design.
- Chose worker-host authenticated Git transport and documented the narrow temporary child-command limitation.
- Documented that workflow transitions are unaffected, corrected the stale design statement, and refreshed both phase-issue scope confirmations.
- Re-ran the full validation suite and replied to each new inline finding.

docs/adr/2026-08-28_agent-tool-isolation.md
```diff
@@ -0,0 +1,217 @@
...
+Authenticated Git transport follows the same boundary. The current child-side
+Git credential helper receives either a raw GitHub token or broker credentials
+and returns a password to Git, so merely stripping its environment variables
+would both break `git push` and leave an agent-readable token path through an
+allowlisted host home. #672 therefore removes that helper from child launches
+and moves authenticated fetch/push to a worker-host operation for the assigned
+branch. During phase 1, the child may make commits but cannot perform
+authenticated `git fetch`, `git pull`, `git push`, or `gh` commands that depend
+on host login; the worker performs the bounded post-run transport instead. This
+is a deliberately narrow compatibility regression, not a secret-name carve-out.
```
hojinzs, line 70: **P2 (impact P1) — the round-2 blocker is genuinely closed, and you took the stronger option. But "the worker performs the bounded post-run transport instead" understates what phase 1 now costs, in three specific ways. Not blocking this merge; it needs to be settled in #672 before code lands.**

Removing the helper rather than carving it out is the right call, and the paragraph is honest that `gh` breaks. The gap is that the named replacement covers only Git push, while the isolated home takes out more than Git:

**1. `land` loses its only mechanism, and nothing host-side replaces it.** The injected `land` skill is built entirely on host-login `gh` — `gh pr view --json state,mergeCommit` (`land.ts:21`), `gh pr view --json reviews` (`:34`), `gh pr checks` (`:39`), `gh pr merge --squash` (`:56`). I looked for a host-side equivalent and there isn't one: `tracker-github/src/adapter.ts:240,3452` has `addComment`, but no PR-merge path anywhere in `packages/worker/src` or `packages/extension-github-workflow/src`. So phase 1 as written leaves the `awaiting-merge` lifecycle phase with no route to `completed` at all. That's a bigger decision than a "narrow compatibility regression" and it deserves to be made explicitly rather than discovered.

**2. `pull` is mid-run, so a *post-run* transport cannot stand in for it.** The `pull` skill is `git fetch origin` → `git merge origin/main` (`pull.ts:20,24`) — the agent runs it to take base changes *during* the attempt. `push` is genuinely post-run and the host transport fits it; `pull` is not, and the current wording implies one mechanism covers both.

**3. `gh-project`'s comment paths regress too.** `gh issue comment` (`gh-project.ts:87`), `gh api -X PATCH .../issues/comments/<id>` (`:95`), `gh issue create` (`:103`), `gh issue edit --add-label` (`:113`) all need host login. Round 2 established that *transitions* are safe via the orchestrator endpoint (`SYMPHONY_ORCHESTRATOR_TOKEN`, `gh-project.ts:43,73`) — that still holds and lines 71-73 state it correctly. But that clause answers the MCP-disablement question, and sitting directly under this paragraph it reads as covering the `gh` question too, which it doesn't. The workpad-comment path has a host-side counterpart (`addComment`) but it isn't wired for agent use.

Also worth flagging for sizing: the host-side transport is **net-new**. `orchestrator/src/git.ts` does clone and `pull --ff-only` only; there is no `push` anywhere in `packages/orchestrator/src` or `packages/worker/src` outside tests. "The worker performs the bounded post-run transport" reads like reuse of an existing path, and #672 should be scoped knowing it isn't.

One added sentence enumerating what regresses — and a decision on whether `land` is acceptable to lose in the interim or needs a host-side merge operation — would make this actionable. I'd rather that land in the #672 scope comment than hold the ADR for it.

---

docs/adr/2026-08-28_agent-tool-isolation.md
```diff
@@ -0,0 +1,217 @@
...
+on normal completion, cancellation, timeout, and worker crash recovery. The
+port is an addressability detail, not an authorization boundary; the session
+token and loopback bind are both required.
+
+Both phase-2 launchers retain the isolated child home/configuration directory
+from phase 1. Agent preflight must not require `gh auth`, and neither `HOME`,
+`GH_CONFIG_DIR`, nor an inherited credential-helper configuration may expose a
+host GitHub login. The worker performs authenticated Git transport and tool
+calls with its host credential; the child receives only repository state and
+bounded results.
```
hojinzs, line 138: **P2 — "Agent preflight must not require `gh auth`" is filed under phase 2, but phase 1 is what trips it. One-line move.**

The preflight gate is toggled off today by the presence of the token phase 1 strips:

```ts
const hasGitHubGraphqlToken =
  typeof launcherEnv.GITHUB_GRAPHQL_TOKEN === "string" && ...;
const preflight = await runClaudePreflight({
  env: launcherEnv,
  includeGhAuth: !hasGitHubGraphqlToken,
});
```
(`worker/src/index.ts:533-542`)

Strip `GITHUB_GRAPHQL_TOKEN` per phase 1 → `hasGitHubGraphqlToken` goes false → `includeGhAuth` flips **on**, in the phase whose whole purpose is making host login unreachable from the child.

To be clear about severity, since it changes the priority: this does **not** hard-fail runs. `checkGhAuthentication` (`preflight.ts:225-232`) calls `execFileSync("gh", ["auth","status"], { cwd, ... })` with no `env`, so it inherits the *worker host's* `process.env` and passes against host login regardless of what the child's isolated `HOME` looks like. So the failure mode is a false green rather than a startup failure: preflight reports "GitHub CLI authentication ✓" while the child it is gating for cannot use `gh` at all — and on a run that fails at startup (`exitWorkerStartupFailure`, `:546-550`) that check is exactly the wrong signal to trust.

Cheapest fix is to move the sentence: phase 1 removes the credential, so phase 1 owns retiring the check that asserts it. Phase 2 then just inherits it, the same way it inherits the isolated home in the sentence above.

---

README.md
```diff
@@ -959,6 +959,24 @@ pnpm typecheck
...
+
+By default, Codex uses `approval_policy: never` and the
+`danger-full-access` thread sandbox; Claude uses `bypassPermissions`.
+Operators can configure the Codex approval and sandbox settings, but must use
+least-privilege credentials, dedicated workspaces, and controls appropriate to
+their environment. The target transport keeps credentials in the Symphony host
+or a host-side broker, returns only bounded issue-aware tool results to agents,
+uses loopback-only local services with scoped session capabilities, and gives
+the child an isolated home/configuration directory rather than a host `gh auth`
+store. Authenticated Git transport is performed by the host.
```
hojinzs, line 996: **P3 — this trailing sentence slipped back into the present tense the round-1 fix was about.**

The paragraph is careful right up to here: everything hangs off "The target transport keeps… returns… uses… gives…". Then `Authenticated Git transport is performed by the host.` starts a new sentence and drops the target framing, so read on its own it asserts current fact. It isn't — there is no `push` in `packages/orchestrator/src` or `packages/worker/src` today (`orchestrator/src/git.ts` does clone and `pull --ff-only`); the child's credential helper is still the Git auth path (`runtime-codex/src/runtime.ts:791-813`), and #672 hasn't landed.

This is the same tense-split as round 1, just in the sentence appended after it. Folding it into the preceding clause fixes it — `…rather than a host gh auth store, and performs authenticated Git transport on the host.` The opening "Today…" paragraph already carries the current-state disclosure, so nothing else needs to change.

---

docs/adr/2026-08-28_agent-tool-isolation.md
```diff
@@ -0,0 +1,217 @@
...
+  `mcp.json` no longer contains a literal secret;
+- a child-started process makes credential inheritance and disk exposure harder
+  to audit and reliably prevent;
+- Linear has no equivalent broker today, so a raw credential would remain
+  necessary or require a one-off side channel; and
+- it does not provide the host's normalized issue context or a single place to
+  enforce adapter-specific scope, error, and rate-limit contracts.
+
+The phase-1 compatibility channel is therefore temporary and explicitly
+tracked as a divergence, not an alternative architecture.
```
hojinzs, line 187: nit — "The phase-1 compatibility channel" is left over from the pre-round-1 text.

There is no phase-1 channel any more: round 1 replaced the stdin/fd bridge with disablement (lines 51-59), and round 2 replaced the Git carve-out with host-side transport (lines 61-70). Phase 1 now *closes* channels rather than providing one, so the sentence names a thing the ADR no longer describes — mildly confusing in the Rejected-alternative section, where a reader could take it as the subprocess model surviving into phase 1.

"The phase-1 tool disablement is therefore temporary…" says what you mean and keeps the rest of the sentence intact.

---

hojinzs (approved):
## Review — round 3/6 · **Approve**

The round-2 P1 landed, and you took the stronger of the two options I offered: `759492c` removes the child Git credential helper and moves authenticated transport host-side, rather than carving the credential names out. That closes the raw-token path instead of documenting it, which is the better outcome. I said I'd approve on the next round if that paragraph landed — it did, so I'm approving.

I did find one more thing, and I want to be explicit about why it isn't blocking rather than quietly dropping it.

### Round-2 rework verification (checked against the tree, not the reply threads)

| # | Finding | Verified in `759492c` |
| --- | --- | --- |
| P1 | Phase 1 stripped the vars the child's git credential helper reads — `git push` breaks, and the helper is itself a raw-token channel | ✅ lines 61-70: helper removed from child launches, transport moves to the worker host, regression named. Not a carve-out. |
| P2 | `HOME` allowlisted in all three launchers → agent reads a token from `~/.config/gh/hosts.yml` regardless | ✅ isolated child home/config dir now required in phase 1 (46-47) and retained in phase 2 (133-138); Consequences narrowed to "…or a readable host GitHub login" (191-192) |
| P3 | Option 2's "name which paths regress" half missing; stale contradicting line in the standalone-project design | ✅ lines 71-73 state that no injected skill invokes `github_graphql`/`linear_graphql` and transitions run through the orchestrator endpoint — both confirmed. Design doc line 118 corrected, and the secrets bullet now points at this ADR instead of implying the broker satisfies §10.5. |

`prettier --check` passes on all three files; CI green on `759492c` (Test, Container Smoke); PR body's "Changed files" now matches the 3-file diff.

### Requirement coverage (#671 산출물)

All six deliverables present and verified in earlier rounds; nothing regressed. 완료 기준 (ADR merged + #672/#673 반영 코멘트) is met on merge. No over-engineering — the ADR stays in ADR scope, and the one out-of-scope item from round 1 (`docs/README.md`) was correctly reverted.

### Findings this round

| Pri | Finding |
| --- | --- |
| **P2** (impact P1) | Phase 1's `gh` regression is wider than "post-run transport" covers: `land` loses `gh pr view`/`checks`/`merge` with **no host-side merge path existing anywhere**, so `awaiting-merge` has no route to `completed`; `pull` is mid-run so a post-run transport can't stand in for it; `gh-project`'s comment/label paths regress too. Also: the host-side push is net-new — there is no `push` in `orchestrator/src` or `worker/src`. |
| **P2** | "Agent preflight must not require `gh auth`" is filed under phase 2, but phase 1's stripping is what flips `includeGhAuth` on (`worker/src/index.ts:533-542`). Not a hard failure — the check inherits the *host* env, so it passes — which makes it a false green rather than a broken run. |
| **P3** | `README.md:978` — trailing "Authenticated Git transport is performed by the host." slipped back into the present tense the round-1 fix was about. |
| nit | "The phase-1 compatibility channel" (line 186) is stale; phase 1 no longer has one. |

### Why I'm approving rather than going to round 4

The first finding has P1-sized *impact* — losing the `land` phase is not small — but it isn't the failure shape that blocked rounds 1 and 2. Those specified mechanisms that **cannot exist** (a stdin/fd bridge with no endpoint; a stripping rule whose two readings were both wrong), so #672 would have stalled needing the architecture re-decided with an Accepted ADR saying it was settled. This one is different: the decision is sound and buildable, the ADR already admits `gh` breaks, and what's missing is the enumeration of blast radius and a call on whether `land` is acceptable to lose in the interim. That's scope work #672 can carry, not an architectural contradiction.

Holding an otherwise-correct ADR for a third round over prose would cost more than it buys — and the decision in it is one I want recorded. **Please fold the `land`/`gh` question into the #672 scope comment before implementation starts**, since whoever picks it up will hit it on day one; the other three can ride along in a follow-up or be waved off with a reason.

Note on the smoke-test requirement: this PR ships no API or UI, so there is nothing to exercise in a preview or E2E environment. The equivalent verification is checking the ADR's claims against the artifacts it describes — a docs PR fails by being wrong — which I did against the tree at `759492c` for every claim in the table above, plus `prettier` and CI.

---

moncher-dev:
**검증 통과 → Land로 이동 (프로젝트 상태 관리)** — docs-only PR: Prettier 통과, ADR이 확정 결정((a) host-side 실행, 2단계)·관련 이슈(#672/#673/#675)·README posture 초안과 일치합니다.

⚠️ **병합 전 오너 확인 요청 (영향도 상)** — ADR의 Phase 1(#672) 범위가 제가 결정 시점에 설명한 것보다 넓습니다:
- Phase 1에서 provider-native MCP 툴 노출을 **비활성화**하고(#673 전까지 에이전트가 `github_graphql`/`linear_graphql` 사용 불가)
- child 측 인증 Git 전송·`gh` 명령을 제거하고 워커가 post-run push를 대행하며, child HOME을 격리합니다.
ADR은 "워크플로 전이는 orchestrator transition endpoint를 쓰므로 영향 없음"이라고 하지만, 에이전트가 PR을 `gh pr create`로 만드는 경로가 있다면 Phase 1 동안 깨집니다. 이 범위로 진행할지, 아니면 #672를 "시크릿 제거 + broker secret만 전달(툴 유지)"로 축소하고 #673에서 툴/Git을 함께 옮길지 결정 후 #672를 Ready로 올리겠습니다. 결정 전까지 #672는 Backlog에 둡니다.

PR #691, opened 2026-08-27
feat(core): strict front-matter validation

### LINKED ISSUE #667: feat(core): strict front-matter validation — integer-only fields, hooks.timeout_ms, agent.max_turns, non-empty codex.command, supported tracker.kind, typed workflow errors
Epic: #651
리포트 항목: **A6, B8(일부), B17** — `docs/reports/2026-08-28-upstream-spec-drift-research.md` (PR #650)

## 문제

새 스펙 §5.3: 숫자 필드는 integer only, `hooks.timeout_ms` 무효값과 비양수 `agent.max_turns`는 **검증 실패**. §6.3: `tracker.kind`가 지원되는지, `codex.command`가 비어 있지 않은지 preflight. §5.5: typed error(`workflow_parse_error`, `workflow_front_matter_not_a_map`).

## 현재 상태 (증거)

- 문자열 정수 허용: `packages/core/src/workflow/parser.ts:846-848`
- `timeout_ms: -5` 저장 → 실행기가 timeout 비활성화: `parser.ts:148-149`, `hooks.ts` `if (timeoutMs > 0)`
- `max_turns: 0` 허용 → 워커 "exhausted before start": `packages/worker/src/turn-limits.ts:20-26`
- `command: ""` 통과: `parser.ts:88` (`??`가 빈 문자열을 놓침)
- `kind: jira` 파싱 통과 `parser.ts:66`; 미지원 kind는 tick 안에서 throw: `packages/orchestrator/src/tracker-adapters.ts:15-21`, `service.ts:1181`
- typed error는 `missing_workflow_file`만: `loader.ts:83`, `git.ts:498-500`; non-map은 산문 오류 `parser.ts:321`

## 할 일

- 타입/범위 검증과 `WorkflowValidationError { code, path, message }` 도입; 파서/로더/`doctor`/`workflow validate`가 같은 코드 사용
- 지원 kind 목록을 core에 등록(어댑터 wiring에서 주입) — 파싱 단계에서 미지원 kind 거부
- `docs/configuration.md`에 검증 규칙 표

## 완료 기준

- §17.1 행: invalid YAML typed, non-map typed, `tracker.kind` supported, defaults apply, per-state map — 각각 테스트
- `repo doctor`/`workflow validate`가 새 오류 코드를 표시

moncher-dev:
## TL;DR

- Enforce strict `WORKFLOW.md` validation before configuration reaches runtime execution.
- Preserve typed workflow error codes and qualified field paths across parser, loader, `workflow validate`, and `doctor`.
- Keep CLI and daemon tracker-kind preflight aligned with the orchestrator adapter registry.

## Change-point diagram

```text
adapter registry → parser/loader options → typed workflow validation → CLI diagnostics
```

## Start here

- `packages/core/src/workflow/parser.ts` owns typed front-matter validation and field paths.
- `packages/orchestrator/src/tracker-adapters.ts` supplies registered tracker kinds to workflow loading.
- `packages/cli/src/commands/{workflow,doctor}.ts` apply the same supported-kind registry to preflight commands.
- `packages/core/src/workflow-loader.test.ts` covers malformed YAML, scalar roots, strict integer/range checks, defaults, and injected kinds.

## Risks & rollback

- Previously accepted quoted numeric values, non-positive concurrency limits, and invalid configuration now fail early.
- Priority maps and per-state concurrency overrides must use YAML integers; global and per-state concurrency values must be positive.
- Roll back by reverting this PR to restore legacy permissive parsing.

## Changed files

- `packages/core/src/workflow/{parser,loader}.ts`: strict typed validation and parser coverage.
- `packages/orchestrator/src/{git,tracker-adapters}.ts`: adapter-owned supported-kind injection.
- `packages/cli/src/commands/{workflow,doctor,repo.test}.ts`: registry-aligned CLI preflight and diagnostic regression coverage.
- `docs/configuration.md`, `README.md`, `packages/cli/README.md`: validation behavior guidance.

## Evidence

- `pnpm lint`
- `pnpm test`
- `pnpm typecheck`
- `pnpm build`
- Docker E2E: rebuilt `symphony-e2e`; `/healthz` returned `{"ok":true}`, refresh succeeded, and `/api/v1/state` reported `health: "idle"` with no `lastError`.

- [ ] Confirm invalid workflows expose stable codes and qualified paths in `workflow validate --json`.
- [ ] Confirm `doctor` and `workflow validate` accept registered tracker kinds and reject an unregistered kind before dispatch.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -34,21 +34,23 @@ same metadata for automation.
 ## WORKFLOW.md Front-matter Validation
 
 `gh-symphony workflow validate` and `gh-symphony repo doctor` use the same
-strict parser as workflow loading. Failures include a stable error code and a
-field path so automation can distinguish malformed YAML from a policy error.
+strict parser as workflow loading. Failures include a stable error code; the
+`workflow validate --json` error also exposes its field path separately.
 
-| Rule | Required value | Error code/path |
-| --- | --- | --- |
-| Front matter syntax | A valid YAML mapping | `workflow_parse_error` or `workflow_front_matter_not_a_map` at `front_matter` |
-| Numeric fields | YAML integers; quoted numeric strings and fractions are rejected | `workflow_validation_error` at the field path |
-| `hooks.timeout_ms` | A positive integer when provided | `workflow_validation_error` at `hooks.timeout_ms` |
-| `agent.max_turns` | A positive integer when provided | `workflow_validation_error` at `agent.max_turns` |
-| `codex.command` | A non-empty string when provided | `workflow_validation_error` at `codex.command` |
-| `tracker.kind` | One of `github-project`, `linear`, or `file` | `workflow_validation_error` at `tracker.kind` |
+| Rule                | Required value                                                                                                             | Error code/path                                                               |
+| ------------------- | -------------------------------------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------- |
+| Front matter syntax | A valid YAML mapping                                                                                                       | `workflow_parse_error` or `workflow_front_matter_not_a_map` at `front_matter` |
+| Numeric fields      | YAML integers; quoted numeric strings and fractions are rejected, including priority maps and per-state concurrency values | `workflow_validation_error` at the field path                                 |
+| `hooks.timeout_ms`  | A positive integer when provided                                                                                           | `workflow_validation_error` at `hooks.timeout_ms`                             |
+| `agent.max_turns`   | A positive integer when provided                                                                                           | `workflow_validation_error` at `agent.max_turns`                              |
+| `codex.command`     | A non-empty string when provided                                                                                           | `workflow_validation_error` at `codex.command`                                |
+| `tracker.kind`      | One of `github-project`, `linear`, or `file`                                                                               | `workflow_validation_error` at `tracker.kind`                                 |
 
 Omitted optional values retain their documented defaults. Per-state concurrency
 maps remain supported through `agent.max_concurrent_agents_by_state` and their
-values also must be YAML integers.
+values must be positive YAML integers. This repository intentionally rejects an
+invalid override rather than ignoring it, so invalid configuration cannot silently
+disable dispatch for a state.
 
 ## Workflow Lifecycle Policy
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -31,6 +31,25 @@ Human-readable `gh-symphony repo status`, `gh-symphony project status`, and
 their `--watch` dashboards show the applied revision; `--json` exposes the
 same metadata for automation.
 
+## WORKFLOW.md Front-matter Validation
+
+`gh-symphony workflow validate` and `gh-symphony repo doctor` use the same
+strict parser as workflow loading. Failures include a stable error code and a
+field path so automation can distinguish malformed YAML from a policy error.
+
+| Rule | Required value | Error code/path |
+| --- | --- | --- |
+| Front matter syntax | A valid YAML mapping | `workflow_parse_error` or `workflow_front_matter_not_a_map` at `front_matter` |
+| Numeric fields | YAML integers; quoted numeric strings and fractions are rejected | `workflow_validation_error` at the field path |
+| `hooks.timeout_ms` | A positive integer when provided | `workflow_validation_error` at `hooks.timeout_ms` |
+| `agent.max_turns` | A positive integer when provided | `workflow_validation_error` at `agent.max_turns` |
+| `codex.command` | A non-empty string when provided | `workflow_validation_error` at `codex.command` |
+| `tracker.kind` | One of `github-project`, `linear`, or `file` | `workflow_validation_error` at `tracker.kind` |
+
+Omitted optional values retain their documented defaults. Per-state concurrency
+maps remain supported through `agent.max_concurrent_agents_by_state` and their
+values also must be YAML integers.
+
 ## Workflow Lifecycle Policy
 
 `tracker.blocker_check_states` selects the workflow states where unresolved
--- a/packages/cli/src/commands/workflow.ts
+++ b/packages/cli/src/commands/workflow.ts
@@ -4,6 +4,7 @@ import {
   buildPromptVariables,
   type OrchestratorProjectConfig,
   parseWorkflowMarkdown,
+  WorkflowValidationError,
   renderPrompt,
   resolveWorkflowExecutionPhase,
   type TrackedIssue,
@@ -23,6 +24,7 @@ import {
   validateGitHubToken,
 } from "../github/gh-auth.js";
 import type { GlobalOptions } from "../index.js";
+import { writeCliError } from "../cli-error.js";
 import {
   buildPriorityConfigDiagnostics,
   type PriorityDiagnostic,
@@ -1035,6 +1037,14 @@ const handler = async (
         return;
     }
   } catch (error) {
+    if (error instanceof WorkflowValidationError) {
+      writeCliError({
+        code: error.code,
+        message: `${error.path}: ${error.message}`,
+        json: options.json,
+      });
+      return;
+    }
     const message = error instanceof Error ? error.message : String(error);
     process.stderr.write(`Workflow command failed: ${message}\n`);
     process.exitCode = 1;
--- a/packages/cli/src/commands/workflow.ts
+++ b/packages/cli/src/commands/workflow.ts
@@ -1040,6 +1040,7 @@ const handler = async (
     if (error instanceof WorkflowValidationError) {
       writeCliError({
         code: error.code,
+        path: error.path,
         message: `${error.path}: ${error.message}`,
         json: options.json,
       });
--- a/packages/cli/src/commands/workflow.ts
+++ b/packages/cli/src/commands/workflow.ts
@@ -9,7 +9,10 @@ import {
   resolveWorkflowExecutionPhase,
   type TrackedIssue,
 } from "@gh-symphony/core";
-import { resolveTrackerAdapter } from "@gh-symphony/orchestrator";
+import {
+  getSupportedTrackerKinds,
+  resolveTrackerAdapter,
+} from "@gh-symphony/orchestrator";
 import { fetchGithubProjectIssueByRepositoryAndNumber } from "@gh-symphony/tracker-github";
 import { type CliProjectConfig } from "../config.js";
 import {
@@ -805,7 +808,9 @@ function validateWorkflow(
   workflowPath: string,
   markdown: string
 ): WorkflowValidationReport {
-  const workflow = parseWorkflowMarkdown(markdown);
+  const workflow = parseWorkflowMarkdown(markdown, process.env, {
+    supportedTrackerKinds: getSupportedTrackerKinds(),
+  });
   const samplePhase = resolveWorkflowExecutionPhase({
     issueState: SAMPLE_ISSUE.state,
     planningStates: workflow.lifecycle.planningStates,
@@ -951,7 +956,9 @@ async function runPreview(
     );
   }
   const { workflowPath, markdown } = await loadWorkflowMarkdown(flags.file);
-  const workflow = parseWorkflowMarkdown(markdown);
+  const workflow = parseWorkflowMarkdown(markdown, process.env, {
+    supportedTrackerKinds: getSupportedTrackerKinds(),
+  });
   if (
     flags.issue &&
     workflow.tracker.kind !== "github-project" &&
--- a/packages/core/src/workflow/loader.ts
+++ b/packages/core/src/workflow/loader.ts
@@ -2,7 +2,10 @@ import { createHash } from "node:crypto";
 import { access, readFile, stat } from "node:fs/promises";
 import { constants } from "node:fs";
 import { DEFAULT_WORKFLOW_DEFINITION, type ParsedWorkflow } from "./config.js";
-import { parseWorkflowMarkdown } from "./parser.js";
+import {
+  parseWorkflowMarkdown,
+  WorkflowValidationError,
+} from "./parser.js";
 import type { WorkflowResolution } from "../contracts/status-surface.js";
 
 type WorkflowCacheEntry = {
@@ -79,14 +82,19 @@ export class WorkflowConfigStore {
           isValid: false,
           usedLastKnownGood: true,
           validationError:
-            error instanceof Error
-              ? error.message
-              : "Invalid workflow definition.",
+            formatWorkflowValidationError(error),
         });
       }
       throw error;
     }
   }
+}
+
+export function formatWorkflowValidationError(error: unknown): string {
+  if (error instanceof WorkflowValidationError) {
+    return `${error.code} (${error.path}): ${error.message}`;
+  }
+  return error instanceof Error ? error.message : "Invalid workflow definition.";
 }
 
 export function createDefaultWorkflowResolution(): WorkflowResolution {
--- a/packages/core/src/workflow/loader.ts
+++ b/packages/core/src/workflow/loader.ts
@@ -5,6 +5,7 @@ import { DEFAULT_WORKFLOW_DEFINITION, type ParsedWorkflow } from "./config.js";
 import {
   parseWorkflowMarkdown,
   WorkflowValidationError,
+  type ParseWorkflowOptions,
 } from "./parser.js";
 import type { WorkflowResolution } from "../contracts/status-surface.js";
 
@@ -18,6 +19,8 @@ type WorkflowCacheEntry = {
 
 export class WorkflowConfigStore {
   private readonly cache = new Map<string, WorkflowCacheEntry>();
+
+  constructor(private readonly parseOptions: ParseWorkflowOptions = {}) {}
 
   async load(
     workflowPath: string,
@@ -57,7 +60,7 @@ export class WorkflowConfigStore {
     }
 
     try {
-      const workflow = parseWorkflowMarkdown(markdown, env);
+      const workflow = parseWorkflowMarkdown(markdown, env, this.parseOptions);
       const revision = createWorkflowRevision(workflow);
       const loadedAt = new Date().toISOString();
       this.cache.set(workflowPath, {
@@ -81,8 +84,7 @@ export class WorkflowConfigStore {
           loadedAt: cached.loadedAt,
           isValid: false,
           usedLastKnownGood: true,
-          validationError:
-            formatWorkflowValidationError(error),
+          validationError: formatWorkflowValidationError(error),
         });
       }
       throw error;
@@ -94,7 +96,9 @@ export function formatWorkflowValidationError(error: unknown): string {
   if (error instanceof WorkflowValidationError) {
     return `${error.code} (${error.path}): ${error.message}`;
   }
-  return error instanceof Error ? error.message : "Invalid workflow definition.";
+  return error instanceof Error
+    ? error.message
+    : "Invalid workflow definition.";
 }
 
 export function createDefaultWorkflowResolution(): WorkflowResolution {
--- a/packages/core/src/workflow/parser.ts
+++ b/packages/core/src/workflow/parser.ts
@@ -51,6 +51,7 @@ export class WorkflowValidationError extends Error {
   }
 }
 
+/** Compatibility list for consumers that do not own a tracker-adapter registry. */
 export const DEFAULT_SUPPORTED_TRACKER_KINDS = [
   "github-project",
   "linear",
@@ -68,10 +69,11 @@ export function parseWorkflowMarkdown(
     if (error instanceof WorkflowValidationError) {
       throw error;
     }
-    const message = error instanceof Error ? error.message : "Invalid workflow definition.";
+    const message =
+      error instanceof Error ? error.message : "Invalid workflow definition.";
     throw new WorkflowValidationError(
       "workflow_validation_error",
-      inferWorkflowErrorPath(message),
+      "front_matter",
       message
     );
   }
@@ -137,24 +139,37 @@ function parseWorkflowMarkdownInternal(
 
   const maxConcurrentAgentsByState = readNumberMap(
     agent,
-    "max_concurrent_agents_by_state"
+    "max_concurrent_agents_by_state",
+    "agent.max_concurrent_agents_by_state",
+    { positive: true }
   );
 
   const runtime = hasRuntime ? parseRuntimeConfig(runtimeNode, env) : null;
   const codexConfig = {
-    command: readOptionalNonEmptyString(codex, "command", env) ?? DEFAULT_AGENT_COMMAND,
+    command:
+      readOptionalNonEmptyString(codex, "command", env, "codex.command") ??
+      DEFAULT_AGENT_COMMAND,
     approvalPolicy: readOptionalString(codex, "approval_policy", env),
     threadSandbox: readOptionalString(codex, "thread_sandbox", env),
     turnSandboxPolicy: readOptionalString(codex, "turn_sandbox_policy", env),
     turnTimeoutMs:
-      readOptionalIntegerLike(codex, "turn_timeout_ms") ??
-      DEFAULT_TURN_TIMEOUT_MS,
+      readOptionalIntegerLike(
+        codex,
+        "turn_timeout_ms",
+        "codex.turn_timeout_ms"
+      ) ?? DEFAULT_TURN_TIMEOUT_MS,
     readTimeoutMs:
-      readOptionalIntegerLike(codex, "read_timeout_ms") ??
-      DEFAULT_READ_TIMEOUT_MS,
+      readOptionalIntegerLike(
+        codex,
+        "read_timeout_ms",
+        "codex.read_timeout_ms"
+      ) ?? DEFAULT_READ_TIMEOUT_MS,
     stallTimeoutMs:
-      readOptionalIntegerLike(codex, "stall_timeout_ms") ??
-      DEFAULT_STALL_TIMEOUT_MS,
+      readOptionalIntegerLike(
+        codex,
+        "stall_timeout_ms",
+        "codex.stall_timeout_ms"
+      ) ?? DEFAULT_STALL_TIMEOUT_MS,
   };
   const agentCommand = resolveWorkflowRuntimeCommand({
     runtime,
@@ -190,8 +205,11 @@ function parseWorkflowMarkdownInternal(
     },
     polling: {
       intervalMs:
-        readOptionalIntegerLike(polling, "interval_ms") ??
-        DEFAULT_POLL_INTERVAL_MS,
+        readOptionalIntegerLike(
+          polling,
+          "interval_ms",
+          "polling.interval_ms"
+        ) ?? DEFAULT_POLL_INTERVAL_MS,
     },
     repository: readOptionalExtensionObject(frontMatter, "repository"),
     workspace: {
@@ -208,21 +226,33 @@ function parseWorkflowMarkdownInternal(
     },
     agent: {
       maxConcurrentAgents:
-        readOptionalIntegerLike(agent, "max_concurrent_agents") ??
-        DEFAULT_MAX_CONCURRENT_AGENTS,
+        readOptionalIntegerLike(
+          agent,
+          "max_concurrent_agents",
+          "agent.max_concurrent_agents"
+        ) ?? DEFAULT_MAX_CONCURRENT_AGENTS,
       maxRetryBackoffMs:
-        readOptionalIntegerLike(agent, "max_retry_backoff_ms") ??
-        DEFAULT_MAX_RETRY_BACKOFF_MS,
+        readOptionalIntegerLike(
+          agent,
+          "max_retry_backoff_ms",
+          "agent.max_retry_backoff_ms"
+        ) ?? DEFAULT_MAX_RETRY_BACKOFF_MS,
       maxConcurrentAgentsByState,
       maxFailureRetries:
-        readOptionalIntegerLike(agent, "max_failure_retries") ??
-        DEFAULT_MAX_FAILURE_RETRIES,
+        readOptionalIntegerLike(
+          agent,
+          "max_failure_retries",
+          "agent.max_failure_retries"
+        ) ?? DEFAULT_MAX_FAILURE_RETRIES,
       maxTurns:
         readOptionalPositiveInteger(agent, "max_turns", "agent.max_turns") ??
         DEFAULT_MAX_TURNS,
       retryBaseDelayMs:
-        readOptionalIntegerLike(agent, "retry_base_delay_ms") ??
-        DEFAULT_BASE_DELAY_MS,
+        readOptionalIntegerLike(
+          agent,
+          "retry_base_delay_ms",
+          "agent.retry_base_delay_ms"
+        ) ?? DEFAULT_BASE_DELAY_MS,
     },
     runtime,
     codex: codexConfig,
@@ -376,7 +406,15 @@ function parseFrontMatter(
   const lines = frontMatter.replace(/\r\n/g, "\n").split("\n");
   let value: WorkflowFrontMatterNode;
   try {
-    [value] = parseBlock(lines, 0, 0);
+    const root = lines.find(
+      (line) => line.trim() && !line.trim().startsWith("#")
+    );
+    value =
+      root &&
+      findMappingSeparator(root.trim()) < 0 &&
+      !root.trim().startsWith("- ")
+        ? parseScalar(root)
+        : parseBlock(lines, 0, 0)[0];
   } catch (error) {
     if (error instanceof WorkflowValidationError) {
       throw error;
@@ -397,11 +435,6 @@ function parseFrontMatter(
   }
 
   return value as Record<string, WorkflowFrontMatterNode>;
-}
-
-function inferWorkflowErrorPath(message: string): string {
-  const field = message.match(/field "([^"]+)"/);
-  return field?.[1] ?? "front_matter";
 }
 
 function parseBlock(
@@ -818,16 +851,17 @@ function readRequiredObject(
 function readOptionalString(
   input: Record<string, WorkflowFrontMatterNode>,
   key: string,
-  env: NodeJS.ProcessEnv
+  env: NodeJS.ProcessEnv,
+  path = key
 ): string | null {
   const value = input[key];
   if (value === undefined || value === null) {
     return null;
   }
   if (typeof value !== "string") {
-    throw new Error(`Workflow front matter field "${key}" must be a string.`);
+    throw new Error(`Workflow front matter field "${path}" must be a string.`);
   }
-  return resolveEnvironmentValue(value, env);
+  return resolveEnvironmentValue(value, env, path);
 }
 
 function readOptionalWorkflowString(
@@ -914,7 +948,8 @@ function readOptionalBoolean(
 
 function readOptionalIntegerLike(
   input: Record<string, WorkflowFrontMatterNode>,
-  key: string
+  key: string,
+  path = key
 ): number | null {
   const value = input[key];
   if (value === undefined || value === null) {
@@ -923,7 +958,11 @@ function readOptionalIntegerLike(
   if (typeof value === "number" && Number.isInteger(value)) {
     return value;
   }
-  throw new Error(`Workflow front matter field "${key}" must be an integer.`);
+  throw new WorkflowValidationError(
+    "workflow_validation_error",
+    path,
+    `Workflow front matter field "${path}" must be an integer.`
+  );
 }
 
 function readOptionalPositiveInteger(
@@ -931,7 +970,7 @@ function readOptionalPositiveInteger(
   key: string,
   path: string
 ): number | null {
-  const value = readOptionalIntegerLike(input, key);
+  const value = readOptionalIntegerLike(input, key, path);
   if (value !== null && value <= 0) {
     throw new WorkflowValidationError(
       "workflow_validation_error",
@@ -945,14 +984,15 @@ function readOptionalPositiveInteger(
 function readOptionalNonEmptyString(
   input: Record<string, WorkflowFrontMatterNode>,
   key: string,
-  env: NodeJS.ProcessEnv
+  env: NodeJS.ProcessEnv,
+  path: string
 ): string | null {
-  const value = readOptionalString(input, key, env);
+  const value = readOptionalString(input, key, env, path);
   if (value !== null && value.trim().length === 0) {
     throw new WorkflowValidationError(
       "workflow_validation_error",
-      `codex.${key}`,
-      `Workflow front matter field "codex.${key}" must be a non-empty string when provided.`
+      path,
+      `Workflow front matter field "${path}" must be a non-empty string when provided.`
     );
   }
   return value;
@@ -961,7 +1001,8 @@ function readOptionalNonEmptyString(
 function readNumberMap(
   input: Record<string, WorkflowFrontMatterNode>,
   key: string,
-  path = key
+  path = key,
+  options: { positive?: boolean } = {}
 ): Record<string, number> {
   const value = input[key];
   if (value === undefined || value === null) {
@@ -974,6 +1015,13 @@ function readNumberMap(
   const result: Record<string, number> = {};
   for (const [entryKey, entryValue] of Object.entries(value)) {
     if (typeof entryValue === "number" && Number.isInteger(entryValue)) {
+      if (options.positive && entryValue <= 0) {
+        throw new WorkflowValidationError(
+          "workflow_validation_error",
+          `${path}.${entryKey}`,
+          `Workflow front matter field "${path}.${entryKey}" must be a positive integer.`
+        );
+      }
       result[entryKey] = entryValue;
       continue;
     }
@@ -986,14 +1034,15 @@ function readNumberMap(
 
 function resolveEnvironmentValue(
   value: string,
-  env: NodeJS.ProcessEnv
+  env: NodeJS.ProcessEnv,
+  path: string
 ): string {
   const envTokenMatch = value.match(/^(?:env:)?([A-Z0-9_]+)$/);
   if (value.startsWith("env:") && envTokenMatch) {
     const resolved = env[envTokenMatch[1]];
     if (!resolved) {
       throw new Error(
-        `Workflow front matter requires environment variable ${envTokenMatch[1]}.`
+        `Workflow front matter field "${path}" requires environment variable ${envTokenMatch[1]}.`
       );
     }
     return resolved;
@@ -1004,7 +1053,7 @@ function resolveEnvironmentValue(
     const resolved = env[dollarEnvTokenMatch[1]];
     if (!resolved) {
       throw new Error(
-        `Workflow front matter requires environment variable ${dollarEnvTokenMatch[1]}.`
+        `Workflow front matter field "${path}" requires environment variable ${dollarEnvTokenMatch[1]}.`
       );
     }
     return resolved;
@@ -1014,7 +1063,7 @@ function resolveEnvironmentValue(
     const resolved = env[name];
     if (!resolved) {
       throw new Error(
-        `Workflow front matter requires environment variable ${name}.`
+        `Workflow front matter field "${path}" requires environment variable ${name}.`
       );
     }
     return resolved;
--- a/packages/core/src/workflow/parser.ts
+++ b/packages/core/src/workflow/parser.ts
@@ -31,12 +31,56 @@ type WorkflowFrontMatterNode =
 
 export type ParseWorkflowOptions = {
   compatibilityMode?: "strict" | "legacy";
+  supportedTrackerKinds?: readonly string[];
 };
+
+export type WorkflowValidationErrorCode =
+  | "workflow_parse_error"
+  | "workflow_front_matter_not_a_map"
+  | "workflow_validation_error";
+
+/** A stable, machine-readable error raised while reading WORKFLOW.md. */
+export class WorkflowValidationError extends Error {
+  constructor(
+    public readonly code: WorkflowValidationErrorCode,
+    public readonly path: string,
+    message: string
+  ) {
+    super(message);
+    this.name = "WorkflowValidationError";
+  }
+}
+
+export const DEFAULT_SUPPORTED_TRACKER_KINDS = [
+  "github-project",
+  "linear",
+  "file",
+] as const;
 
 export function parseWorkflowMarkdown(
   markdown: string,
   env: NodeJS.ProcessEnv = process.env,
   options: ParseWorkflowOptions = {}
+): ParsedWorkflow {
+  try {
+    return parseWorkflowMarkdownInternal(markdown, env, options);
+  } catch (error) {
+    if (error instanceof WorkflowValidationError) {
+      throw error;
+    }
+    const message = error instanceof Error ? error.message : "Invalid workflow definition.";
+    throw new WorkflowValidationError(
+      "workflow_validation_error",
+      inferWorkflowErrorPath(message),
+      message
+    );
+  }
+}
+
+function parseWorkflowMarkdownInternal(
+  markdown: string,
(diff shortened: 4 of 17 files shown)


packages/core/src/workflow/parser.ts
```diff
@@ -864,14 +973,10 @@ function readNumberMap(
 
   const result: Record<string, number> = {};
   for (const [entryKey, entryValue] of Object.entries(value)) {
-    if (typeof entryValue === "number") {
+    if (typeof entryValue === "number" && Number.isInteger(entryValue)) {
```
hojinzs, line 1047: **P1 — `max_concurrent_agents_by_state` still accepts non-positive integers, and a negative value silently disables dispatch for that state.**

This hunk tightened the *type* check (string integers now rejected) but not the *range* check. The spec requires positive values here, in both places it defines the field:

- `docs/symphony-spec.md:458` — ``- `max_concurrent_agents_by_state` (map `state_name -> positive integer`)``
- `docs/symphony-spec.md:626` — ``- `agent.max_concurrent_agents_by_state`: map of positive integers, default `{}` ``

Reproduced against this branch (built `packages/core/dist`):

```
input:  agent.max_concurrent_agents_by_state: { Ready: -3 }
result: OK — byState={"Ready":-3}
```

For contrast, the same file *does* now reject `Ready: '2'` and `Ready: abc` (both throw `workflow_validation_error` at `max_concurrent_agents_by_state.Ready`). So an obviously-wrong value passes while a merely mistyped one is fatal.

The runtime consequence is the exact failure class #667 exists to prevent. `packages/orchestrator/src/explain.ts:565-571` and the limit map built at `packages/orchestrator/src/service.ts:4747-4750` both gate on `activeInState >= stateLimit`, so with `stateLimit = -3` the comparison is `0 >= -3` → true on every tick, and that state is blocked from dispatch forever with the operator-facing message `State concurrency is full for "Ready" (0/-3)`. That is the same shape as `max_turns: 0` → "exhausted before start", which this PR does fix.

Suggest routing the map entries through the same positivity gate as the two scalar fields:

```ts
    if (typeof entryValue === "number" && Number.isInteger(entryValue)) {
      if (entryValue <= 0) {
        throw new WorkflowValidationError(
          "workflow_validation_error",
          `${path}.${entryKey}`,
          `Workflow front matter field "${path}.${entryKey}" must be a positive integer.`
        );
      }
      result[entryKey] = entryValue;
      continue;
    }
```

Note `readNumberMap` is also used for `tracker.priority.values` / `tracker.priority.labels` (lines 317, 328), where zero and negatives are legitimate — so this needs to be opt-in per call site rather than applied unconditionally inside the helper.

One thing worth deciding explicitly and recording in the PR description either way: `docs/symphony-spec.md:2081` (§17.1) says *"Per-state concurrency override map normalizes state names and ignores invalid values"* — i.e. skip the bad entry rather than fail the whole workflow. This PR currently makes an invalid entry fatal. Fatal-vs-ignore is a defensible divergence given #667's "integer only" direction, but per the repo's spec-discipline rule it should be a stated divergence, not an accident.

---

packages/core/src/workflow/parser.ts
```diff
@@ -315,15 +374,36 @@ function parseFrontMatter(
...
+      "workflow_front_matter_not_a_map",
+      "front_matter",
+      "Workflow front matter must be a YAML object."
+    );
   }
 
   return value as Record<string, WorkflowFrontMatterNode>;
 }
 
+function inferWorkflowErrorPath(message: string): string {
```
hojinzs: **P2 — scraping `path` back out of the message means the reported path is not the documented path, and is ambiguous for some fields.**

`inferWorkflowErrorPath` recovers the path with `/field "([^"]+)"/` over the already-formatted message. The messages produced by `readOptionalIntegerLike` (line 923) carry only the bare leaf key, so the path this yields is the leaf, not the dotted path `docs/configuration.md` promises.

Measured on this branch:

| input | reported `path` | documented path |
| --- | --- | --- |
| `hooks.timeout_ms: '5000'` | `timeout_ms` | `hooks.timeout_ms` |
| `hooks.timeout_ms: 1.5` | `timeout_ms` | `hooks.timeout_ms` |
| `agent.max_turns: '2'` | `max_turns` | `agent.max_turns` |
| `agent.max_turns: 0` | `agent.max_turns` ✅ | `agent.max_turns` |
| `codex.turn_timeout_ms: '900'` | `turn_timeout_ms` | — |
| `polling.interval_ms: '30000'` | `interval_ms` | — |

So the *same field* reports two different paths depending on whether it failed the type check or the range check — and the PR's own test table encodes that split (`"string integer"` asserts `max_turns`, `"non-positive turn limit"` asserts `agent.max_turns`), which locks the inconsistency in.

`turn_timeout_ms` is the one that actually bites: it is read at line 150 for `codex.*` and again at line 734 for `runtime.timeouts.*`, so the emitted path cannot tell an operator which of the two blocks to edit.

The fix is mechanical — `readOptionalPositiveInteger` already threads an explicit `path`, so give `readOptionalIntegerLike` (and `readOptionalString`/`readRequiredString`, which have the same bare-key messages) the same parameter and pass the dotted path from each call site. Then `inferWorkflowErrorPath` can go away rather than being the mechanism the contract depends on.

Also `Workflow front matter requires environment variable X.` (line 599 et al.) contains no `field "..."` at all, so it falls through to `front_matter`:

```
input:  codex.command: ${SOME_UNSET_VAR}
result: workflow_validation_error at path "front_matter"
```

That one loses the field entirely — an explicit path parameter fixes it for free.

---

packages/core/src/workflow/parser.ts
```diff
@@ -31,12 +31,56 @@ type WorkflowFrontMatterNode =
...
+    public readonly code: WorkflowValidationErrorCode,
+    public readonly path: string,
+    message: string
+  ) {
+    super(message);
+    this.name = "WorkflowValidationError";
+  }
+}
+
+export const DEFAULT_SUPPORTED_TRACKER_KINDS = [
```
hojinzs, line 55: **P2 — the supported-kind list is hardcoded in core and never injected, which is the half of #667 that isn't delivered.**

Issue #667 asks for: *"지원 kind 목록을 core에 등록(**어댑터 wiring에서 주입**) — 파싱 단계에서 미지원 kind 거부"*. The rejection at parse time is done and works; the injection is not. `options.supportedTrackerKinds` has zero call sites outside this file:

```
$ grep -rn "supportedTrackerKinds" packages/ --include=*.ts
packages/core/src/workflow/parser.ts:34
packages/core/src/workflow/parser.ts:115,116,117,121
```

Every real caller (`loader.ts:60`, `git.ts`, `workflow.ts`, `doctor.ts`) passes no options, so `DEFAULT_SUPPORTED_TRACKER_KINDS` is what actually runs everywhere, and core now hardcodes the identities of the three tracker adapters. That cuts against `CLAUDE.md`'s "Keep tracker-specific behavior out of core layers".

The list is *correct today* — I checked both halves of the real registry:

- `packages/orchestrator/src/tracker-adapters.ts:10-13` → `file`, `linear`
- `packages/tracker-github/src/orchestrator-adapter.ts:343-345` → `github-project`

…but nothing keeps them in sync, so adding a fourth adapter now means editing core too, and forgetting to leaves a valid config rejected at parse time with no test failure.

Either wire it up (have the orchestrator/CLI pass the kinds derived from their adapter registries into `parseWorkflowMarkdown`, and add a test that a kind injected at the seam is accepted), or drop the unused option and say in the PR description that the hardcoded list is a deliberate deferral of #667's injection requirement. An unused seam is the worst of the three, since it reads as done.

Minor, related: `docs/symphony-spec.md:1290` lists `unsupported_tracker_kind` as a recommended error category. That's stated for the *adapter* error contract (§11.4), not for front-matter parsing, so `workflow_validation_error` at `tracker.kind` is a reasonable choice here — just flagging it in case you'd rather align the code names now while the surface is new.

---

packages/core/src/workflow/loader.ts
```diff
@@ -79,16 +82,21 @@ export class WorkflowConfigStore {
...
-              : "Invalid workflow definition.",
+            formatWorkflowValidationError(error),
         });
       }
       throw error;
     }
   }
 }
 
+export function formatWorkflowValidationError(error: unknown): string {
```
hojinzs, line 95: **P2 — Prettier violations, all on lines this PR adds.**

`npx prettier --check` fails on three of the changed files, and every reported deviation is new code:

```
[warn] packages/core/src/workflow/parser.ts
[warn] packages/core/src/workflow/loader.ts
[warn] packages/core/src/workflow-loader.test.ts
```

In this file: the `import { parseWorkflowMarkdown, WorkflowValidationError } from "./parser.js"` split at lines 5-8 collapses to one line, `validationError: formatWorkflowValidationError(error)` at 84-85 collapses to one line, and the ternary at line 93 wants wrapping. In `parser.ts` it's lines 71 and 145; in `workflow-loader.test.ts` the whole `it.each` table (66-72) reformats to multi-line entries.

`pnpm format:write` fixes all of it. Not CI-blocking — `.github/workflows/ci.yml` runs build/lint/typecheck/vitest but no format step, which is why this got through — but Prettier is a stated convention in `CLAUDE.md` and the next person to touch these files will pick up the reformat in their diff.

For the record, the rest of the gate is clean on this branch: `pnpm lint`, `pnpm typecheck`, `pnpm build`, and `pnpm test` all pass locally, matching the green CI run.

---

docs/configuration.md
```diff
@@ -31,6 +31,25 @@ Human-readable `gh-symphony repo status`, `gh-symphony project status`, and
...
+## WORKFLOW.md Front-matter Validation
+
+`gh-symphony workflow validate` and `gh-symphony repo doctor` use the same
+strict parser as workflow loading. Failures include a stable error code and a
+field path so automation can distinguish malformed YAML from a policy error.
+
+| Rule | Required value | Error code/path |
+| --- | --- | --- |
+| Front matter syntax | A valid YAML mapping | `workflow_parse_error` or `workflow_front_matter_not_a_map` at `front_matter` |
+| Numeric fields | YAML integers; quoted numeric strings and fractions are rejected | `workflow_validation_error` at the field path |
```
hojinzs: **P3 — the table under-states the blast radius of the integer-only change.**

"Numeric fields" is accurate but reads as if it only covers the two fields called out below it. In practice this PR makes quoted numeric strings fatal across *every* numeric front-matter field, including two the table never mentions:

- `tracker.priority.values` and `tracker.priority.labels` (`parser.ts:317,328` via `readNumberMap`) — previously accepted `"3"`, now throw.
- `agent.max_concurrent_agents_by_state` — mentioned at line 49 in prose, but as "their values also must be YAML integers" rather than as a breaking change.

Anyone who wrote `priority: { values: { P1: '1' } }` gets a hard parse failure on upgrade with no mention in the docs or the changeset. Worth one line in **Risks & rollback** and a sentence here naming the priority maps explicitly.

Two smaller things while you're in this table:

1. The `hooks.timeout_ms` / `agent.max_turns` rows promise a dotted error path that only holds for the range failure, not the type failure — see the `inferWorkflowErrorPath` comment. Whichever way that's resolved, these rows should describe what actually ships.
2. "so automation can distinguish malformed YAML from a policy error" oversells the current CLI surface. `path` never appears as a discrete field, and the two commands encode it differently:

```
$ gh-symphony workflow validate --file WORKFLOW.md --json
{ "error": { "code": "workflow_validation_error",
             "message": "tracker.kind: Unsupported workflow tracker.kind \"jira\". ..." } }

$ gh-symphony doctor --json      # checks[].details.error
"workflow_validation_error (tracker.kind): Unsupported workflow tracker.kind \"jira\". ..."
```

Both surface the code correctly, so #667's acceptance criterion is met — but automation has to string-split a human message, differently per command, to recover `path`. Adding `path` alongside `code` in `writeCliError` and in the doctor `details` object would make the sentence true; otherwise soften the claim to "a stable error code".

---

packages/core/src/workflow/parser.ts
```diff
@@ -840,15 +920,44 @@ function readOptionalIntegerLike(
...
+    throw new WorkflowValidationError(
+      "workflow_validation_error",
+      path,
+      `Workflow front matter field "${path}" must be a positive integer.`
+    );
+  }
+  return value;
+}
+
+function readOptionalNonEmptyString(
```
hojinzs, line 1014: nit — `readOptionalNonEmptyString` hardcodes the `codex.` prefix into the path and message, so it can't be reused for the other command field (`runtime.command`, read at line 730 via `readOptionalString`) without emitting a wrong path.

Its sibling immediately above, `readOptionalPositiveInteger`, already takes `path` as a parameter. Matching that signature keeps the two helpers symmetric and removes the one place where a path is baked into a helper rather than supplied by the caller:

```ts
function readOptionalNonEmptyString(
  input: Record<string, WorkflowFrontMatterNode>,
  key: string,
  env: NodeJS.ProcessEnv,
  path: string
): string | null {
```

Behaviour is correct as written for the single current call site — verified `codex.command: '   '` → `workflow_validation_error` at `codex.command`, and `command: ""` no longer slips past the `??` as reported in #667.

---

hojinzs (changes requested):
## 리뷰 1/6 — 변경 요청 (P1 1건)

#667의 요구사항 대부분이 충실히 반영됐고, 스코프도 타이트합니다(오버엔지니어링 없음). 다만 스펙이 positive integer로 정의한 필드 하나가 검증에서 빠져 있고, 그 값이 오케스트레이터를 조용히 멈추게 합니다.

### 요구사항 대조 (#667)

| 요구사항 | 상태 |
| --- | --- |
| 숫자 필드 integer only (문자열/소수 거부) | ✅ |
| `hooks.timeout_ms` 양수 | ✅ |
| `agent.max_turns` 양수 | ✅ |
| `codex.command` 비어 있지 않음 | ✅ (`''`, `'   '` 모두 거부) |
| `tracker.kind` 지원 목록 검증 | ⚠️ 파싱 단계 거부는 동작, **어댑터 wiring 주입은 미구현** |
| `WorkflowValidationError { code, path, message }` | ⚠️ 구조는 도입, **`path` 값이 문서와 불일치** |
| 파서/로더/`doctor`/`workflow validate` 동일 코드 사용 | ✅ |
| `docs/configuration.md` 검증 규칙 표 | ✅ (아래 P3 보완 필요) |
| §17.1 테스트 5행 (invalid YAML / non-map / kind / defaults / per-state map) | ✅ 전부 존재 |
| changeset (`@gh-symphony/cli` minor) | ✅ |

### Smoke Test

Preview 환경이 없어 로컬에서 실행했습니다. `pnpm install → build → lint → typecheck → test` 전부 통과(CI Test/Container Smoke도 green). 빌드된 `packages/core/dist`에 17개 입력 케이스를 넣고, CLI 두 표면을 직접 실행했습니다.

`workflow validate --json` / `doctor --json` 모두 새 코드를 정상 노출 — #667 완료 기준 충족:

```
$ gh-symphony workflow validate --file WORKFLOW.md --json     # exit 1
{ "error": { "code": "workflow_validation_error",
             "message": "tracker.kind: Unsupported workflow tracker.kind \"jira\". ..." } }

$ gh-symphony doctor --json                                    # checks[].details.error
"workflow_validation_error (tracker.kind): Unsupported ... Supported values: github-project, linear, file."
```

정상 거부 확인: `max_turns: '2'`, `max_turns: 0/-1`, `timeout_ms: 0`, `timeout_ms: 1.5`, `command: '   '`, `kind: jira`, front matter 없음, non-map front matter.

그 중 한 케이스만 통과했습니다:

```
agent.max_concurrent_agents_by_state: { Ready: -3 }   → OK, byState={"Ready":-3}
```

### 지적사항

- **P1** `max_concurrent_agents_by_state`가 비양수를 허용 — 스펙(§5.3 L458, §6.4 L626)은 positive integer로 정의하고, 음수면 `explain.ts:565-571` / `service.ts:4747`의 `activeInState >= stateLimit`가 매 tick 참이 되어 해당 state가 영구히 dispatch 차단됩니다. `max_turns: 0` → "exhausted before start"와 같은 실패 유형입니다. 같은 hunk가 `Ready: '2'`는 거부하면서 `Ready: -3`은 통과시키는 상태입니다. (`tracker.priority.*`도 같은 헬퍼를 쓰므로 헬퍼 내부가 아니라 호출부별 opt-in으로.)
- **P2** `inferWorkflowErrorPath`가 메시지를 정규식으로 되긁어 `path`를 만들어, 타입 오류와 범위 오류가 같은 필드에 서로 다른 path를 냅니다(`timeout_ms` vs `hooks.timeout_ms`). `turn_timeout_ms`는 `codex.*`/`runtime.timeouts.*` 구분이 불가능하고, env 변수 미해결은 `front_matter`로 떨어집니다.
- **P2** `DEFAULT_SUPPORTED_TRACKER_KINDS`가 core에 하드코딩되고 `supportedTrackerKinds` 옵션은 호출부가 0개 — #667의 "어댑터 wiring에서 주입"이 미이행이고, core가 트래커 어댑터 3종을 알게 됩니다. 값 자체는 현재 레지스트리와 일치하나 동기화 장치가 없습니다.
- **P2** 신규 추가 라인에 Prettier 위반 3개 파일. CI에 format 스텝이 없어 통과했습니다. `pnpm format:write`로 해결.
- **P3** `docs/configuration.md` 표가 `tracker.priority.values`/`labels`의 breaking change를 누락. 또 "automation can distinguish"라고 했지만 `path`가 별도 필드로 나오지 않고 두 커맨드의 인코딩도 다릅니다.
- **nit** `readOptionalNonEmptyString`이 `codex.` 접두사를 하드코딩 — 바로 위 `readOptionalPositiveInteger`처럼 `path` 파라미터로.

**임의 결정사항 1건 (판단 필요, 수정 요구 아님):** 스펙 §17.1(L2081)은 per-state 맵의 invalid value를 *무시*하라고 하는데, 이 PR은 워크플로 전체를 실패시킵니다. #667의 "integer only" 방향상 방어 가능한 선택이지만, 레포의 spec discipline 규칙상 명시적으로 문서화된 divergence여야 합니다. P1을 고칠 때 fail/ignore 중 어느 쪽을 택하든 PR 설명에 한 줄 남겨주세요.

P1 + P2 3건 처리되면 approve하겠습니다.

---

packages/core/src/workflow/parser.ts
```diff
@@ -315,15 +374,36 @@ function parseFrontMatter(
...
-  const [value] = parseBlock(lines, 0, 0);
+  let value: WorkflowFrontMatterNode;
+  try {
+    [value] = parseBlock(lines, 0, 0);
+  } catch (error) {
+    if (error instanceof WorkflowValidationError) {
+      throw error;
+    }
+    throw new WorkflowValidationError(
+      "workflow_parse_error",
```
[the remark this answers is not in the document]
moncher-dev: Resolved in a81454d. Scalar YAML roots now parse successfully and are then rejected as `workflow_front_matter_not_a_map` at `front_matter`; coverage was added to `workflow-loader.test.ts`.

packages/core/src/workflow/parser.ts
```diff
@@ -315,15 +374,36 @@ function parseFrontMatter(
...
+      "Workflow front matter must be a YAML object."
+    );
   }
 
   return value as Record<string, WorkflowFrontMatterNode>;
 }
 
+function inferWorkflowErrorPath(message: string): string {
+  const field = message.match(/field "([^"]+)"/);
+  return field?.[1] ?? "front_matter";
```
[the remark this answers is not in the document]
moncher-dev: Resolved in a81454d. Integer readers now receive and preserve qualified field paths, including `agent.max_turns`, `hooks.timeout_ms`, and runtime/codex timeout fields.

packages/cli/src/commands/workflow.ts
```diff
@@ -1020,6 +1022,14 @@ const handler = async (
         return;
     }
   } catch (error) {
+    if (error instanceof WorkflowValidationError) {
+      writeCliError({
+        code: error.code,
+        message: `${error.path}: ${error.message}`,
+        json: options.json,
+      });
```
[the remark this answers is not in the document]
moncher-dev: Resolved in a81454d. `README.md` and `packages/cli/README.md` now document stable strict-workflow error codes and the JSON `error.path` emitted by `workflow validate`.

packages/core/src/workflow/parser.ts
moncher-dev: Resolved in a81454d. `agent.max_concurrent_agents_by_state` now requires positive integers and reports the qualified map-entry path. The configuration guide explicitly records the intentional strict-validation divergence from the upstream map-normalization behavior.

packages/core/src/workflow/parser.ts
moncher-dev: Resolved in a81454d. The parser no longer infers paths from formatted messages for integer validation; readers carry explicit paths, and environment-resolution failures now include their originating field path.

packages/core/src/workflow/parser.ts
moncher-dev: Resolved in a81454d. Orchestrator tracker-adapter wiring now exports its supported kinds and injects them into `WorkflowConfigStore`; tests cover the registry export and an injected custom kind.

packages/core/src/workflow/loader.ts
moncher-dev: Resolved in a81454d. Applied Prettier to the affected implementation and test files; `pnpm format` now passes for the changed scope.

docs/configuration.md
moncher-dev: Resolved in a81454d. The validation table now names priority maps and per-state concurrency values, documents the strict positive constraint, and accurately limits the discrete JSON-path claim to `workflow validate --json`.

packages/core/src/workflow/parser.ts
moncher-dev: Resolved in a81454d. `readOptionalNonEmptyString` now accepts its path from the caller, removing the hardcoded `codex.` prefix.

packages/core/src/workflow/parser.ts
```diff
@@ -881,14 +1034,15 @@ function readNumberMap(
...
+  env: NodeJS.ProcessEnv,
+  path: string
 ): string {
   const envTokenMatch = value.match(/^(?:env:)?([A-Z0-9_]+)$/);
   if (value.startsWith("env:") && envTokenMatch) {
     const resolved = env[envTokenMatch[1]];
     if (!resolved) {
       throw new Error(
-        `Workflow front matter requires environment variable ${envTokenMatch[1]}.`
+        `Workflow front matter field "${path}" requires environment variable ${envTokenMatch[1]}.`
```
hojinzs, line 1079: **P1 — this message change breaks an existing test; `pnpm test` is red on a81454d.**

Adding `field "${path}"` to the env-resolution messages broke an assertion that pins the old wording:

```
FAIL packages/cli/src/commands/repo.test.ts:646
  > repo init runtime migration > fails Linear repo init when LINEAR_API_KEY reference is unresolved
  AssertionError: expected 'Workflow front matter field "api_key"…' to contain 'Workflow front matter requires enviro…'
  - Workflow front matter requires environment variable LINEAR_API_KEY
  + Workflow front matter field "api_key" requires environment variable LINEAR_API_KEY.
```

Full run on this branch: `Test Files 1 failed | 39 passed (40)`, `Tests 1 failed | 567 passed (568)`. The `Test` check on this head is still in progress as I write this, but it will land red. The PR's **Evidence** section still claims `pnpm test` passes — that was true for f74aed9 and is stale for a81454d; please re-run and refresh it before the next push.

The fix is one line in `repo.test.ts:647` — the new message is strictly better, so update the assertion rather than reverting the message.

---

**P2, same lines — `error.path` still doesn't carry the field, which is the half of my earlier thread that isn't done.**

The reply says environment-resolution failures "now include their originating field path", and they do — *in the message string*. But these still throw a bare `Error`, so the outer wrapper in `parseWorkflowMarkdown` re-labels them at `front_matter`, and the discrete `path` field is what my original comment was about:

```
input:  codex.command: ${SOME_UNSET_VAR}
result: workflow_validation_error  path="front_matter"
        message='...field "codex.command" requires environment variable SOME_UNSET_VAR.'
```

So automation reading `error.path` still gets `front_matter` and has to string-split the human message to recover `codex.command` — the exact thing the discrete `path` field exists to avoid. `resolveEnvironmentValue` already takes `path`; throwing `new WorkflowValidationError("workflow_validation_error", path, …)` here (and at 1056/1066) closes it for free.

---

packages/core/src/workflow/parser.ts
```diff
@@ -146,24 +221,38 @@ export function parseWorkflowMarkdown(
...
+        DEFAULT_HOOK_TIMEOUT_MS,
     },
     agent: {
       maxConcurrentAgents:
-        readOptionalIntegerLike(agent, "max_concurrent_agents") ??
-        DEFAULT_MAX_CONCURRENT_AGENTS,
+        readOptionalIntegerLike(
+          agent,
+          "max_concurrent_agents",
+          "agent.max_concurrent_agents"
```
hojinzs, line 251: **P1 — `agent.max_concurrent_agents: 0` still passes validation and silently disables dispatch for the whole project.**

Last round fixed the *per-state* map after I reported that `Ready: -3` blocks one state forever. The scalar sibling — the global limit, two lines up from `max_turns` which this PR *does* gate — kept `readOptionalIntegerLike` and so still accepts non-positive values.

Verified against this branch's built `packages/core/dist`, and end-to-end through the CLI:

```
$ cat W4.md
---
tracker: { kind: github-project }
agent:  { max_concurrent_agents: 0 }
codex:  { command: codex }
---
$ gh-symphony workflow validate --file W4.md --json
  → valid, maxConcurrentAgents = 0
```

Runtime consequence, tracing the same path as the per-state finding:

- `packages/orchestrator/src/service.ts:4801` reads `workflow.agent.maxConcurrentAgents`, and `getProjectConcurrency` guards it with `Number.isFinite(limit) && limit >= 0 ? limit : DEFAULT_CONCURRENCY` — `>= 0`, so **`0` passes the guard** and becomes the live limit.
- `packages/orchestrator/src/explain.ts:555` then evaluates `activeRunCount >= maxConcurrentAgents` → `0 >= 0` → **block on every tick**, with the operator-facing `Project concurrency is full (0/0)`.

That is the identical shape to `max_turns: 0` → "exhausted before start", which #667 names as its motivating example — only project-wide instead of per-issue. `readOptionalPositiveInteger` already exists and is already used at lines 224 and 248; this call site wants it too.

A second, quieter inconsistency at the same seam, worth deciding while you're here: `max_concurrent_agents: -4` **is** caught by the `>= 0` guard and silently swallowed into `DEFAULT_CONCURRENCY`. So a negative global limit is ignored, while a negative per-state limit is now fatal after last round's fix — and `docs/configuration.md` (as amended by this PR) states the opposite intent: *"This repository intentionally rejects an invalid override rather than ignoring it."* Gating this field positive resolves both halves and makes the doc true.

Note `explain.ts` also renders the raw workflow value rather than the guarded one, so with `-4` the explain output reports `(0/-4)` while dispatch is actually running at `DEFAULT_CONCURRENCY` — the two disagree. Out of scope to fix here, but it goes away on its own once the value can't be negative.

Please also add a `max_concurrent_agents: 0` row to the `it.each` table in `workflow-loader.test.ts`; the table currently covers `max_turns` and `hooks.timeout_ms` but not this field, which is why the gap survived a round.

---

packages/core/src/workflow/parser.ts
```diff
@@ -864,14 +1014,17 @@ function readNumberMap(
...
+      }
       result[entryKey] = entryValue;
       continue;
     }
-    if (typeof entryValue === "string" && /^-?\d+$/.test(entryValue)) {
-      result[entryKey] = Number.parseInt(entryValue, 10);
-      continue;
-    }
     throw new Error(
       `Workflow front matter field "${path}.${entryKey}" must be an integer.`
```
hojinzs, line 1061: **P2 — the "same field, two different paths depending on failure mode" defect from last round is fixed for scalars but still live for map entries.**

Removing `inferWorkflowErrorPath` and threading explicit paths through the integer readers genuinely fixed the scalar cases — I re-measured all of them and they're correct now. But `readNumberMap` only got the *positive* branch converted to `WorkflowValidationError`; this type-failure fallthrough still throws a bare `Error`, so it's re-labelled `front_matter` by the outer wrapper.

Measured on this branch, same field, two failure modes:

| input | `error.path` |
| --- | --- |
| `max_concurrent_agents_by_state: { Ready: -1 }` | `agent.max_concurrent_agents_by_state.Ready` ✅ |
| `max_concurrent_agents_by_state: { Ready: '2' }` | `front_matter` ❌ |
| `max_concurrent_agents_by_state: { Ready: 1.5 }` | `front_matter` ❌ |
| `tracker.priority.values: { P1: '1' }` | `front_matter` ❌ |

Confirmed at the CLI surface too, not just the library — note the message knows the path the `path` field is missing:

```
$ gh-symphony workflow validate --file W3.md --json
{ "error": { "code": "workflow_validation_error",
             "path": "front_matter",
             "message": "front_matter: Workflow front matter field \"agent.max_concurrent_agents_by_state.Ready\" must be an integer." } }
```

This is the same inconsistency I flagged last round, just relocated from the scalars to the map — and the new test only asserts the negative-value case, so the type-failure path is unpinned again. Constructing the `WorkflowValidationError` here with `` `${path}.${entryKey}` `` (as the branch above already does) closes it, and it's worth one `it.each` row per failure mode so the split can't come back a third time.

Scalar-root classification, qualified scalar paths, the per-state positivity gate, the README/`packages/cli/README.md` updates, and Prettier all verified fixed — `npx prettier --check` is clean on every changed file now. Thanks for the thorough pass on those.

---

packages/core/src/workflow/parser.ts
```diff
@@ -64,6 +114,15 @@ export function parseWorkflowMarkdown(
     : readRequiredObject(frontMatter, "codex");
 
   const trackerKind = readRequiredString(tracker, "kind", env);
+  const supportedTrackerKinds =
+    options.supportedTrackerKinds ?? DEFAULT_SUPPORTED_TRACKER_KINDS;
```
hojinzs, line 118: **P3 — injection is wired for the orchestrator but not for the two CLI commands, so `workflow validate` and `doctor` can disagree with the daemon.**

The seam is real now — `git.ts:31` passes `getSupportedTrackerKinds()` into `WorkflowConfigStore`, which was the load-bearing half of #667's injection requirement. Good fix.

The remaining gap is that the CLI still calls `parseWorkflowMarkdown` with no options:

```
$ grep -rn "supportedTrackerKinds" packages/ --include=*.ts
… packages/orchestrator/src/git.ts:31          ← only injection call site
```

`packages/cli/src/commands/workflow.ts` and `commands/doctor.ts` both fall through to `DEFAULT_SUPPORTED_TRACKER_KINDS`. No live divergence today — the hardcoded list and the registry happen to match, which I re-verified — but it means a fourth adapter is accepted by the daemon and rejected by `workflow validate`, with no test catching the drift. That's a preflight tool contradicting the runtime it's meant to preflight.

`@gh-symphony/orchestrator` is already a devDependency of `@gh-symphony/cli` and `getSupportedTrackerKinds` is already re-exported from its index (`orchestrator/src/index.ts:32`), so both call sites can pass it with an import and no new dependency edge.

If you'd rather defer, that's defensible for a P3 — but then say so in the PR description, because the current state reads as fully wired.

---

hojinzs (changes requested):
## Round 2 — most of round 1 is genuinely fixed; one CI-red regression and one repeat of the same bug class block approval

I re-verified all five of my round-1 threads against the code rather than the replies, built `packages/core` and the CLI, and smoke-tested `workflow validate --json` end to end.

### Verified fixed ✅

| Round-1 finding | Status |
| --- | --- |
| P2 scalar YAML root misclassified | Fixed — `---\nhello\n---` → `workflow_front_matter_not_a_map` (also with a leading comment line) |
| P2 `inferWorkflowErrorPath` message-scraping | Fixed — helper deleted; `hooks.timeout_ms`, `agent.max_turns`, `codex.turn_timeout_ms`, `polling.interval_ms` all report qualified paths for **both** type and range failures |
| P1 README / `packages/cli/README.md` not updated | Fixed |
| P1 per-state map accepts non-positive | Fixed — `Ready: -1` → `workflow_validation_error` at `agent.max_concurrent_agents_by_state.Ready`, divergence from §17.1 recorded in `docs/configuration.md` |
| P2 supported-kind list never injected | Fixed for the runtime path (`git.ts:31`); see P3 below for the CLI half |
| P2 Prettier violations | Fixed — `npx prettier --check` clean on all changed files |
| P3 doc table understates blast radius / nit hardcoded `codex.` prefix | Fixed |

#667's headline acceptance criterion holds at the user-facing surface:

```
$ gh-symphony workflow validate --file W1.md --json
{ "error": { "code": "workflow_validation_error", "path": "tracker.kind",
             "message": "tracker.kind: Unsupported workflow tracker.kind \"jira\". …" } }
```

### Blocking

**P1 — `pnpm test` is red on a81454d.** The env-var message change broke `packages/cli/src/commands/repo.test.ts:646`, which pins the old wording. `Tests 1 failed | 567 passed (568)`. One-line assertion update; the new message is better. The PR's **Evidence** section still claims a green `pnpm test` — that was true for f74aed9 and is stale now.

**P1 — `agent.max_concurrent_agents: 0` still passes validation and blocks all dispatch project-wide.** Round 1 fixed the per-state map; the scalar sibling kept `readOptionalIntegerLike`. `getProjectConcurrency` guards with `>= 0`, so `0` survives as the live limit and `explain.ts:555` evaluates `0 >= 0` → blocked every tick (`Project concurrency is full (0/0)`). Same failure class as `max_turns: 0`, which is #667's own motivating example — just project-wide. `readOptionalPositiveInteger` already exists and is used two lines away.

**P2 — the "two paths for one field" defect is fixed for scalars but repeats for map entries.** `Ready: -1` → correct qualified path; `Ready: '2'`, `Ready: 1.5`, and `tracker.priority.values.P1: '1'` → `front_matter`, because `readNumberMap`'s type fallthrough still throws a bare `Error`. The message names the field the `path` doesn't.

**P3 — CLI doesn't inject supported kinds.** `workflow validate` / `doctor` still use the hardcoded default, so a preflight tool can reject what the daemon accepts. No live divergence today; fix or state the deferral.

### Scope check

No overengineering found — the diff stays inside #667's brief, and I don't see invented requirements. Two of the three blockers are the same missed-sibling pattern (a gate applied to one field but not the one next to it), so the durable fix is the test table: it covers `max_turns` and `hooks.timeout_ms` but not `max_concurrent_agents`, and covers the map's range failure but not its type failure. Filling those rows would have caught both.

Ping-pong count: 2 of 6. The remaining work is mechanical and well-scoped — happy to approve on a push that turns CI green and gates the scalar concurrency field; the P2 and P3 are small enough to land alongside or be explicitly deferred.

---

packages/core/src/workflow/parser.ts
moncher-dev: Resolved in b496d59. Environment-variable resolution now throws WorkflowValidationError with the supplied field path, and repo init coverage asserts tracker.api_key is preserved in the diagnostic.

packages/core/src/workflow/parser.ts
moncher-dev: Resolved in b496d59. agent.max_concurrent_agents now uses the positive-integer reader; zero is rejected at agent.max_concurrent_agents with parser coverage.

packages/core/src/workflow/parser.ts
moncher-dev: Resolved in b496d59. Numeric-map type and decimal failures now preserve the qualified entry path, with coverage for both cases.

packages/core/src/workflow/parser.ts
moncher-dev: Resolved in b496d59. workflow validate and doctor now inject getSupportedTrackerKinds() from the orchestrator registry, matching daemon preflight behavior.

packages/core/src/workflow/parser.ts
```diff
@@ -834,25 +978,61 @@ function readOptionalBoolean(
...
   if (value === undefined || value === null) {
     return null;
   }
-  if (typeof value === "number") {
+  if (typeof value === "number" && Number.isInteger(value)) {
     return value;
   }
-  if (typeof value === "string" && /^-?\d+$/.test(value)) {
-    return Number.parseInt(value, 10);
+  throw new WorkflowValidationError(
```
hojinzs, line 991: **P2 (non-blocking, follow-up) — `runtime.timeouts.*` never passes `path`, so the exact `turn_timeout_ms` ambiguity from round 1 is still live.**

The `path = key` default here is correct as a default, but three call sites still rely on it. `parseRuntimeConfig` (lines ~795-802) calls the reader with two arguments:

```ts
turnTimeoutMs:  readOptionalIntegerLike(timeouts, "turn_timeout_ms")  ?? DEFAULT_TURN_TIMEOUT_MS,
readTimeoutMs:  readOptionalIntegerLike(timeouts, "read_timeout_ms")  ?? DEFAULT_READ_TIMEOUT_MS,
stallTimeoutMs: readOptionalIntegerLike(timeouts, "stall_timeout_ms") ?? DEFAULT_STALL_TIMEOUT_MS,
```

The `codex.*` half of that pair *was* converted this round, so the two blocks now disagree. Measured against this branch's built `packages/core/dist`:

| input | `error.path` |
| --- | --- |
| `codex.turn_timeout_ms: '900'` | `codex.turn_timeout_ms` ✅ |
| `runtime.timeouts.turn_timeout_ms: '900'` | `turn_timeout_ms` ❌ |
| `runtime.timeouts.read_timeout_ms: 1.5` | `read_timeout_ms` ❌ |

This is the specific case my round-1 comment called out — *"`turn_timeout_ms` is the one that actually bites: it is read at line 150 for `codex.*` and again at line 734 for `runtime.timeouts.*`, so the emitted path cannot tell an operator which of the two blocks to edit."* An operator who sets both blocks and gets `path: "turn_timeout_ms"` still can't tell which one to fix.

Three added arguments (`"runtime.timeouts.turn_timeout_ms"` etc.) closes it. Not blocking — this is diagnostic quality on fields #667 doesn't name, and everything #667 does name is correct.

---

packages/core/src/workflow/parser.ts
```diff
@@ -738,16 +880,17 @@ function readRequiredObject(
...
+  env: NodeJS.ProcessEnv,
+  path = key
 ): string | null {
   const value = input[key];
   if (value === undefined || value === null) {
     return null;
   }
   if (typeof value !== "string") {
-    throw new Error(`Workflow front matter field "${key}" must be a string.`);
+    throw new Error(`Workflow front matter field "${path}" must be a string.`);
```
hojinzs, line 891: **P2 (non-blocking, follow-up) — string fields are the last place the "message knows the path that `path` is missing" pattern survives.**

Two distinct leftovers here, both the shape we've now fixed twice (scalars in round 2, map entries in round 3):

**1. The type-failure throw on this line is still a bare `Error`,** so `parseWorkflowMarkdown`'s outer wrapper relabels it `front_matter` — while the message names the field:

```
input:  codex.approval_policy: 5
result: code=workflow_validation_error  path="front_matter"
        message='Workflow front matter field "approval_policy" must be a string.'
```

Same for `readRequiredString` at line ~916 (`Workflow front matter field "field" is required.` → `front_matter`) and `readNumberMap`'s non-object guard at line ~1042 (`max_concurrent_agents_by_state: 5` → `front_matter`).

**2. The env-resolution fix landed only where the caller passes `path`.** `resolveEnvironmentValue` now throws `WorkflowValidationError` correctly — that part is genuinely fixed and `tracker.api_key` / `codex.command` both verified good. But 13 `readOptionalString` call sites still pass three arguments and fall through to `path = key`:

```
input:  hooks.after_create: ${UNSET}   → path "after_create"
input:  runtime.auth.env:   ${UNSET}   → path "env"
```

`path: "env"` is the one worth fixing — as a machine-readable field path it's close to meaningless, and `runtime.auth.env` is exactly the kind of field that fails in a fresh environment where the operator most needs the path.

The `tracker.*` and `codex.command` sites were converted this round; the `hooks.*`, `workspace.root`, `runtime.command`, `runtime.auth.env`, `codex.approval_policy`/`thread_sandbox`/`turn_sandbox_policy` and lifecycle sites weren't. Mechanical to finish, and pairs naturally with the `runtime.timeouts` comment — one pass over the remaining unqualified call sites closes both.

Recording as follow-up rather than blocking: no runtime consequence, and every field #667 enumerates already reports correctly at the CLI surface.

---

docs/configuration.md
```diff
@@ -31,6 +31,27 @@ Human-readable `gh-symphony repo status`, `gh-symphony project status`, and
...
+| ------------------- | -------------------------------------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------- |
+| Front matter syntax | A valid YAML mapping                                                                                                       | `workflow_parse_error` or `workflow_front_matter_not_a_map` at `front_matter` |
+| Numeric fields      | YAML integers; quoted numeric strings and fractions are rejected, including priority maps and per-state concurrency values | `workflow_validation_error` at the field path                                 |
+| `hooks.timeout_ms`  | A positive integer when provided                                                                                           | `workflow_validation_error` at `hooks.timeout_ms`                             |
+| `agent.max_turns`   | A positive integer when provided                                                                                           | `workflow_validation_error` at `agent.max_turns`                              |
```
hojinzs, line 45: **P3 (non-blocking) — `agent.max_concurrent_agents` gained a positivity gate this round but has no row here, and it's an unstated spec divergence.**

The table has rows for the two other positive-integer fields (`hooks.timeout_ms`, `agent.max_turns`) but not the one added in b496d59. Worth a third row for symmetry:

```
| `agent.max_concurrent_agents` | A positive integer when provided | `workflow_validation_error` at `agent.max_concurrent_agents` |
```

More importantly, it's a **divergence** rather than spec conformance, and the repo's spec-discipline rule wants those explicit. `docs/symphony-spec.md` distinguishes the three deliberately:

- `:448` — ``- `max_concurrent_agents` (integer)`` ← plain integer
- `:452` — ``- `max_turns` (positive integer)``
- `:458` — ``- `max_concurrent_agents_by_state` (map `state_name -> positive integer`)``

So `max_turns` and the per-state map are conformance; the scalar is us going stricter than upstream. I asked for it last round and I still think it's right — `0` reaching `getProjectConcurrency`'s `>= 0` guard and blocking dispatch project-wide is the same failure class #667 exists to prevent — but that argument belongs in the docs, not just in a review thread.

The per-state divergence is already recorded two paragraphs down ("This repository intentionally rejects an invalid override rather than ignoring it"). Extending that sentence to cover the global limit is probably the whole fix.

Note **Risks & rollback** already covers the upgrade impact ("non-positive concurrency limits … now fail early"), so this is purely the reference table and the divergence record.

---

hojinzs (approved):
## Round 3/6 — Approve ✅

Both round-2 blockers are genuinely fixed. I re-verified every finding against the built code rather than the replies, rebuilt `packages/core` and the CLI, and re-ran the full gate.

### Round-2 findings — all verified fixed

| Finding | Status |
| --- | --- |
| **P1** `pnpm test` red on a81454d (`repo.test.ts:646`) | ✅ Fixed — assertion updated to the better message; **568/568 pass** |
| **P1** `agent.max_concurrent_agents: 0` blocks all dispatch | ✅ Fixed — `readOptionalPositiveInteger`; `0` and `-4` both rejected at `agent.max_concurrent_agents` |
| **P2** map entries report `front_matter` on type failure | ✅ Fixed — `Ready: '2'`, `Ready: 1.5`, `tracker.priority.values.P1: '1'` all report qualified paths |
| **P3** CLI doesn't inject supported kinds | ✅ Fixed — `workflow.ts` ×2 + `doctor.ts` now pass `getSupportedTrackerKinds()`; four injection sites total |

The positivity gate is correctly **opt-in per call site**, as needed — `tracker.priority.values: { P1: -1 }` and `P1: 0` still parse, so the priority maps didn't get caught in the concurrency fix. That was the trap in this change and it was avoided.

### Verification

Full gate on b496d59, all clean:

```
pnpm build      ✅
pnpm test       ✅  cli 568/568, orchestrator 339, worker 169, core+rest all pass
npx prettier --check (all changed files)  ✅
CI: Test ✅  Container Smoke ✅
```

Smoke-tested through the real CLI binary, not just the library — #667's acceptance criteria hold at the user-facing surface with a discrete `path`:

```
$ gh-symphony workflow validate --file Wjira.md --json
{ "error": { "code": "workflow_validation_error", "path": "tracker.kind",
             "message": "... Unsupported workflow tracker.kind \"jira\". Supported values: github-project, file, linear." } }

$ gh-symphony workflow validate --file Wzero.md --json
{ "error": { "code": "workflow_validation_error", "path": "agent.max_concurrent_agents", … } }

$ gh-symphony workflow validate --file Wmap.md --json
{ "error": { "code": "workflow_validation_error", "path": "agent.max_concurrent_agents_by_state.Ready", … } }
```

Note the supported-kind list is now genuinely registry-derived — the order changed to `github-project, file, linear`, matching `getSupportedTrackerKinds()` rather than the old hardcoded constant. That's the observable proof the injection is load-bearing.

Also checked `polling.interval_ms: 0`, which still parses: that's **correct**, not a gap. Spec `:616` defines it as a plain integer, and `clampPollInterval` floors it at `MIN_POLL_INTERVAL_MS`, so there's no busy-loop. Leaving it ungated is the right call — no overengineering.

### Requirement coverage (#667)

All acceptance criteria met: integer-only numeric fields, `hooks.timeout_ms` / `agent.max_turns` positive, non-empty `codex.command`, supported `tracker.kind` rejected at parse with adapter-injected kinds, typed `WorkflowValidationError { code, path, message }` shared across parser/loader/`doctor`/`workflow validate`, §17.1's five test rows, and the validation table in `docs/configuration.md`. Changeset present (`@gh-symphony/cli` minor). The test rows I asked for last round were added — the `max_concurrent_agents: 0` case and both map-entry failure modes — which is what stops this class from recurring a fourth time.

### Scope

No overengineering and no invented requirements across all three rounds; the diff stayed inside #667's brief the whole way.

### Non-blocking follow-ups

Three inline comments, none blocking — all diagnostic-quality polish on fields #667 doesn't enumerate, with no runtime consequence:

- **P2** `runtime.timeouts.*` still emits bare paths, so the `codex.turn_timeout_ms` vs `runtime.timeouts.turn_timeout_ms` ambiguity from round 1 survives on the runtime half.
- **P2** 13 `readOptionalString` call sites still fall through to `path = key`, and string type-failures still throw bare `Error` → relabeled `front_matter`. `runtime.auth.env` → `path: "env"` is the one worth fixing.
- **P3** `agent.max_concurrent_agents` needs a table row, and its positivity gate is a real divergence from spec `:448` (plain integer) that should be recorded alongside the per-state one.

These are one mechanical pass over the remaining unqualified call sites. Happy to see them in a follow-up issue rather than a fourth round here — the contract #667 asked for is delivered and correct, and holding the PR for path polish on adjacent fields would be the ping-pong we're trying to avoid.

Nice work on the last two rounds — the fixes were real and the added test rows were the right ones.

---

moncher-dev:

head `b496d59`(approve, CI green)를 격리 worktree에 빌드하고 `gh-symphony workflow validate --file … --json`으로 검증했습니다.

| 입력 | 결과 |
| --- | --- |
| 정상 front matter | `rc=0`, `ok: true` |
| `hooks.timeout_ms: -5` | `rc=1`, `workflow_validation_error` @ `hooks.timeout_ms` "must be a positive integer" |
| `agent.max_turns: 0` | `rc=1`, @ `agent.max_turns` "must be a positive integer" |
| `codex.command: ""` | `rc=1`, @ `codex.command` "must be a non-empty string when provided" |
| `polling.interval_ms: "30000"` | `rc=1`, @ `polling.interval_ms` "must be an integer" |
| `tracker.kind: jira` | `rc=1`, @ `tracker.kind` "Unsupported … Supported values: github-project, file, linear" |
| 로컬 E2E happy dispatch(file 트래커 + stub, `repo start --http 4785`) | 워커 완료, 오류 없음 |

스펙 §5.3/§6.3/§17.1 항목(integer-only, timeout/max_turns 검증, non-empty command, supported kind, typed error) 충족. 병합되면 #669/#670 unblock.

Issue #657, opened 2026-08-27
fix(worker): treat codex subprocess exit as a turn completion condition, reject pending requests, cap line size, stop injecting the prompt twice


moncher-dev:
Epic: #651
리포트 항목: **B12, B23, E1** — `docs/reports/2026-08-28-upstream-spec-drift-research.md` (PR #650)

## 문제

§10.3은 subprocess exit를 turn 실패 조건으로 둔다. `runCodexClientProtocol` 안에는 exit 리스너가 없어 codex가 turn 중 죽으면 워커가 `turn_timeout_ms`(기본 1시간)까지 대기한다.

## 현재 상태 (증거)

- exit 리스너 부재: `packages/worker/src/index.ts:1020-1960` (`598-624`는 상태만 갱신); pending JSON-RPC 요청 미거부
- 라인 버퍼 무제한: `index.ts:1531-1547` (스펙 RECOMMENDED 10MB)
- 렌더된 프롬프트를 `thread/start`의 `developerInstructions`(`index.ts:1573`)와 turn-1 input(`:1689-1690`)에 **두 번** 주입

## 할 일

- `child.on("exit")` → 진행 중 turn을 `port_exit`로 실패 처리, pending 요청 reject, `waitForTurnWithTimeout` 해제
- 라인 최대 크기 10MB, 초과 시 프로토콜 오류로 실패
- 프롬프트 1회 주입으로 정리 (developerInstructions에는 identity header/continuation 정책만, 혹은 그 반대 — 결정 후 `thread-resume.ts`와 정합)
- 오류 카테고리 `port_exit`, `codex_not_found`(spawn ENOENT) 매핑 (§10.6 RECOMMENDED)

## 완료 기준

- turn 중 프로세스 종료 → 즉시 실패 + 재시도 테스트 (`worker-protocol.test.ts`)
- 10MB 초과 라인 테스트
- 프롬프트 단일 주입 단언

moncher-dev:
## Workpad — hojinzs/github-symphony#657 — Cycle 1

**Type:** fresh pickup
**Branch:** `feat/657-codex-protocol-exit`
**Draft PR:** https://github.com/hojinzs/github-symphony/pull/680
**Cycle opened:** 2026-08-27T16:26:41Z · **Trigger:** Ready issue triaged as an actionable worker protocol fix

### Status Transitions



[...]

Issue #653, opened 2026-08-27
fix(orchestrator): rewrite the retry-fire path per spec §16.6 (single-ID refresh, terminal cleanup, requeue on refresh failure and slot exhaustion)


moncher-dev:
Epic: #651
리포트 항목: **A7, B10** — `docs/reports/2026-08-28-upstream-spec-drift-research.md` (PR #650)

## 문제

retry 만료 처리가 구 스펙의 "후보 전체 조회 후 id 탐색" 알고리즘 그대로이고, 트래커 실패 시에는 스펙과 반대로 **blind restart** 한다.

## 현재 상태 (증거)

- 전체 `listIssues` 조회: `packages/orchestrator/src/service.ts:4064-4078`; `service.test.ts:4975`가 `fetchIssueStatesByIds` 미호출을 단언해 구 동작을 고정
- 트래커 실패 → `"restart"` 반환(blind redispatch): `service.ts:4059-4060`, `service.test.ts:4992`
- `no available orchestrator slots` requeue 부재; retry restart는 슬롯 검사 없이 진행: `service.ts:3266-3284`, `:1472`는 `break`만
- `retry_queued` 레코드가 `currentRunId`를 가지면 슬롯을 점유: `service.ts:1415-1419` (미문서)
- terminal 발견 시 cleanup: `4034-4061`은 `"release"`만 반환, cleanup은 후속 루프(`1868-1893`)에 의존 — GitHub는 서버측 필터 때문에 도달 못 함

## 할 일 (새 스펙 §8.4, §16.6)

1. `fetchIssueStatesByIds([issueId])`로 단일 refresh
2. 결과 없음 → claim 해제
3. terminal → 워크스페이스 cleanup + claim 해제
4. active & routable → 슬롯 있으면 dispatch, 없으면 `no available orchestrator slots`로 requeue(attempt+1)
5. active 아님/routable 아님 → dispatch 없이 claim 해제
6. refresh 실패 → `retry refresh failed`로 requeue(attempt+1), restart 금지
7. `retry_queued` 슬롯 점유 정책은 유지하되 `docs/architecture.md` §3에 명시하거나 스펙대로 running만 세도록 변경 (결정 후 기록)

## 완료 기준

- `service.test.ts:4975`, `:4992` 단언을 새 의미론으로 교체
- refresh 실패 → requeue, 슬롯 부족 → requeue, terminal → cleanup 각각 테스트
[...]

PR #693, opened 2026-08-27
fix(e2e): isolate Docker Compose projects

### LINKED ISSUE #692: dx(e2e): isolate Docker E2E runs per worktree — unique compose project name instead of the fixed symphony-e2e container
## 문제

`docker-compose.e2e.yml`이 `container_name: symphony-e2e`를 고정하고 `e2e/run-e2e.sh` / `e2e/run-standalone-project-e2e.sh`가 compose 프로젝트명을 지정하지 않는다(`docker compose -f docker-compose.e2e.yml`, 프로젝트명 = 디렉터리명). 오케스트레이터가 `max_concurrent_agents: 10`으로 여러 이슈를 동시에 돌리면, 각 에이전트 worktree에서 실행되는 Docker E2E가 **같은 컨테이너 이름을 두고 서로 제거/재생성**한다.

실제 영향 (2026-08-27):
- #653 Cycle 3 블로커: "shared `symphony-e2e` Docker Compose container was repeatedly removed or conflicted by concurrent environment activity"
- #657 Cycle 3 블로커: `run-e2e.sh happy 60`이 워커 2개를 dispatch하고 `expected_one_workspace_record:[]`로 실패 — 동시 실행 컨테이너 간섭으로 보이나 단독 실행으로 재확인 필요
- 두 이슈 모두 코드·단위테스트는 통과한 상태에서 E2E만 못 돌려 Backlog로 회귀 → Epic #651 항목 대부분이 Docker E2E를 요구하므로 파이프라인 전체 병목

## 할 일

- `container_name` 제거(또는 `${COMPOSE_PROJECT_NAME}` 기반으로 파생)
- `run-e2e.sh`, `run-standalone-project-e2e.sh`, `test/e2e/claude/run-docker-e2e.sh`에서 `COMPOSE_PROJECT_NAME`을 worktree 경로 해시(또는 `SYMPHONY_E2E_PROJECT` 환경변수)로 기본 설정해 실행마다 고유한 compose 프로젝트/네트워크/볼륨을 사용
- 이미지 태그도 프로젝트별로 분리하거나 공유 이미지 빌드에 lock 적용
- 스크립트 종료 시 `docker compose -p <name> down -v`로 정리 (실패 경로 포함)
- 동시 실행 두 개가 서로 간섭하지 않음을 보이는 검증 절차를 `AGENT_TEST.md`에 기록
- (선택) `docker ps` 기반 사전 점검으로 같은 프로젝트명이 이미 떠 있으면 명확한 오류

## 완료 기준

- 두 worktree에서 `run-e2e.sh happy`를 동시에 실행해 둘 다 통과
- #653, #657의 E2E 재검증이 단독 실행 없이 통과

관련: #653, #657 (이 이슈가 해결될 때까지 blocked)

moncher-dev:
## TL;DR

Docker E2E runs remain isolated per worktree, and the supporting shell helper is now both correct under an image override and enforced in pull-request CI.

## Change-point diagram

`worktree path` → `project hash` → `Compose project + derived image + derived port` → `isolated container/network/volume` → `owned cleanup`

## Start here

- `e2e/lib/compose-project.sh` derives and validates isolated Compose resources.
- `e2e/lib/compose-project.test.sh` verifies the derivation and override/availability behavior.
- `.github/workflows/ci.yml` runs the helper test for every PR.

## Changes

- Remove fixed container names and use project-specific Compose resources in all three Docker E2E runners.
- Keep cleanup owned by a claimed project, reclaim only derived image tags, and preserve explicit `SYMPHONY_E2E_IMAGE` overrides.
- Fix the helper override test to use a real exported environment value and enforce it in PR CI.
- Remove the ineffective `--rmi local` flag from the explicitly tagged E2E Compose teardown.
- Document and exercise the concurrent-worktree Docker regression procedure.

## Evidence

- `bash -n e2e/lib/compose-project.sh e2e/lib/compose-project.test.sh e2e/run-e2e.sh e2e/run-standalone-project-e2e.sh test/e2e/claude/run-docker-e2e.sh`
- `bash e2e/lib/compose-project.test.sh` and `pnpm e2e:compose-project-test`
- `pnpm lint`, `pnpm test`, `pnpm typecheck`, `pnpm build`
- Concurrent `run-e2e.sh happy 90` in this worktree and a detached worktree: both passed with projects `symphony-e2e-97cc0337b402` and `symphony-e2e-33986c82d701`.

## Risks & rollback

- Hash-derived host ports remain overrideable through `SYMPHONY_E2E_PORT`; runner API access uses Compose exec while manual Compose retains port `4680`.
- Roll back by reverting the helper, runner, documentation, and CI helper-test changes.

## Changed files

- `docker-compose.e2e.yml`
- `e2e/lib/compose-project.sh`, `e2e/lib/compose-project.test.sh`
- `e2e/run-e2e.sh`, `e2e/run-standalone-project-e2e.sh`, `test/e2e/claude/run-docker-e2e.sh`
- `.github/workflows/ci.yml`, `AGENT_TEST.md`, affected `e2e/scenarios/*.md`, `docs/configuration.md`, `package.json`

## Issues — Closed #692

Fixes #692

- [ ] Re-run the blocked #653 and #657 Docker E2E scenarios without serializing Docker runs.


e2e/run-e2e.sh
```diff
@@ -20,34 +23,37 @@ log()  { echo -e "${GREEN}[e2e]${NC} $*"; }
...
   echo "[]" > e2e/fixtures/issues.json 2>/dev/null || true
 }
 trap cleanup EXIT
 
 # ── Setup ─────────────────────────────────────────────────────
 
 log "Scenario: ${SCENARIO} (timeout: ${TIMEOUT}s)"
+log "Compose project: ${COMPOSE_PROJECT_NAME}"
+
+assert_e2e_project_is_available docker-compose.e2e.yml
```
hojinzs, line 49: **P1 — the availability guard causes the exact cross-cleanup it is meant to prevent.**

`trap cleanup EXIT` is installed at line 42, *before* this guard runs. When the guard returns 1, `set -e` exits and the EXIT trap fires against the **already-running** project:

1. `exec -T symphony-e2e sh -lc 'find /e2e/evidence -mindepth 1 ... -exec rm -rf'` — wipes the other run's evidence inside its live container
2. `down --volumes --remove-orphans --timeout 5` — destroys the other run's container, network and volumes

I reproduced this with a stubbed `docker` on PATH whose `compose … ps -q` returns a container id (simulating a busy project):

```
[e2e] Compose project 'symphony-e2e-a3e122fce8d9' is already running. Set SYMPHONY_E2E_PROJECT to a unique value or stop it first.
[e2e] Cleaning up...
DOCKER-CALL: compose --project-name symphony-e2e-a3e122fce8d9 -f docker-compose.e2e.yml ps -q
DOCKER-CALL: compose --project-name symphony-e2e-a3e122fce8d9 -f docker-compose.e2e.yml exec -T symphony-e2e sh -lc  … rm -rf …
DOCKER-CALL: compose --project-name symphony-e2e-a3e122fce8d9 -f docker-compose.e2e.yml down --volumes --remove-orphans --timeout 5
```

Since the project name is derived from the worktree path, this guard can only ever fire when a second run starts in the **same** worktree (or with an explicit `SYMPHONY_E2E_PROJECT` collision) — i.e. precisely the case where a live run is about to be torn down by the newcomer. That is the #692 failure mode, reintroduced through the safety check.

Fix: run the guard before installing the trap.

```bash
assert_e2e_project_is_available docker-compose.e2e.yml
trap cleanup EXIT
```

The same ordering bug is in `e2e/run-standalone-project-e2e.sh:16` and `test/e2e/claude/run-docker-e2e.sh:23` — please fix all three.

---

e2e/run-e2e.sh
```diff
@@ -256,11 +262,11 @@ PY
       throw new Error(`unexpected_run_outcome:${JSON.stringify({status: run.status, runPhase: run.runPhase})}`);
     }
   '
-  if ! docker logs symphony-e2e 2>&1 | grep -q 'api-progress readback.*"state":"Done"'; then
+  if ! "${COMPOSE[@]}" logs symphony-e2e | grep -q 'api-progress readback.*"state":"Done"'; then
```
hojinzs: **P1 — dropping `2>&1` breaks these assertions; all the greppable output is on stderr.**

The previous form was `docker logs symphony-e2e 2>&1 | grep -q …`. The redirect was load-bearing: `e2e/stub-worker.ts` emits *every* line through `console.error` (lines 52, 111, 192, 255, 276, 282, 287, 311, 320, 341, 348, 362), including the two strings asserted here:

```ts
// e2e/stub-worker.ts:255
console.error(
  `[stub-worker] api-progress readback status=${readResponse.status} payload=${JSON.stringify(readback)}`
);
// e2e/stub-worker.ts:282
console.error(`[stub-worker] scenario=${SCENARIO} runId=${RUN_ID}`);
```

`e2e/seed/entrypoint.sh` does not redirect, so these stay on the container's stderr stream, and `docker compose logs` routes container stderr to the CLI's **stderr** (it uses a separate writer from stdout), not into the pipe. So `| grep -q` sees nothing and both checks fail.

The PR's evidence only covers `happy`; `api-progress`, `api-progress-unknown` and `prompt-phase` are the scenarios this touches, and none of them appear in the evidence list. Please add `2>&1` on both lines (265 and 269) and run those three scenarios:

```bash
if ! "${COMPOSE[@]}" logs symphony-e2e 2>&1 | grep -q 'api-progress readback.*"state":"Done"'; then
```

Same applies to the three failure-path diagnostics at lines 67, 197 and 356 — `logs --tail 20` without `2>&1` prints the orchestrator/worker output to the terminal but drops it from any capture, and more importantly `--tail 20` is now applied per-stream. These are diagnostics rather than assertions, so lower stakes, but the fix is the same.

---

docker-compose.e2e.yml
```diff
@@ -1,11 +1,9 @@
 services:
   symphony-e2e:
-    container_name: symphony-e2e
+    image: ${SYMPHONY_E2E_IMAGE:-gh-symphony-e2e:e2e}
```
hojinzs, line 3: **P1 — removing `ports: 4680:4680` is an unrequested decision that breaks 15 documents (74 command occurrences) which the PR did not migrate.**

Issue #692's 할 일 list asks for `container_name` removal, a path-derived `COMPOSE_PROJECT_NAME`, per-project image tags, `down -v` cleanup, an AGENT_TEST.md procedure, and an optional pre-check. Un-publishing the host port is not on that list. It *is* a real collision source (two concurrent runs would both want host 4680), so I'm not arguing against solving it — but the chosen solution silently removes the only host-side entry point, and the manual runbooks were left pointing at it.

Still broken after this PR:

| File | Broken commands |
| --- | --- |
| `AGENT_TEST.md` §1, §3, §4, §6, §7 | `curl http://localhost:4680/healthz`, `.../api/v1/refresh`, `.../api/v1/state`, `docker logs symphony-e2e`, `docker exec symphony-e2e …`, `docker compose -f docker-compose.e2e.yml down` (wrong project) |
| `e2e/scenarios/*.md` — 14 files (01, 02, 03, 04, 05, 06-retry, 06-unbounded, 07, 08, 09, 10, 14, 15, 17) | same host-port `curl` and fixed-container `docker exec` / `docker logs` |

`CLAUDE.md` routes agents to AGENT_TEST.md as *the* Docker E2E procedure, and #692's whole motivation is unblocking agent worktrees — so leaving the runbook broken re-blocks the same pipeline through a different door. The diagram at AGENT_TEST.md:137-158 was updated, but §1–§7 below it were not.

Two ways out, either is fine:

- **Keep a host port, make it per-project.** `ports: - "${SYMPHONY_E2E_PORT:-4680}:4680"`, or `- "4680"` for an ephemeral host port that the runner resolves via `compose port symphony-e2e 4680`. Manual docs keep working with a one-line change.
- **Commit to exec-only.** Then migrate all 15 documents to `docker compose --project-name "$(bash -c '. e2e/lib/compose-project.sh; e2e_compose_project_name "$PWD"')" -f docker-compose.e2e.yml exec …`, or expose a tiny `e2e/lib/compose.sh` wrapper so the docs stay readable.

I'd lean toward the first — it preserves the documented workflow and keeps the diff proportional to the issue.

---

e2e/lib/compose-project.sh
```diff
@@ -0,0 +1,28 @@
...
+  local worktree_path="$1"
+  local path_hash
+  path_hash=$(printf '%s' "$worktree_path" | shasum -a 256 | cut -c1-12)
+  printf 'symphony-e2e-%s' "$path_hash"
+}
+
+configure_e2e_compose_project() {
+  local worktree_path="$1"
+  export COMPOSE_PROJECT_NAME="${SYMPHONY_E2E_PROJECT:-$(e2e_compose_project_name "$worktree_path")}"
+  export SYMPHONY_E2E_IMAGE="${SYMPHONY_E2E_IMAGE:-${COMPOSE_PROJECT_NAME}:e2e}"
```
hojinzs: **P2 — per-project image tags are never reclaimed, so ephemeral agent worktrees leak images without bound.**

Every distinct worktree path produces a distinct tag (`symphony-e2e-<hash>:e2e`), and cleanup only runs `down --volumes --remove-orphans` — which removes containers, networks and volumes, but not images. With `max_concurrent_agents: 10` and worktrees that are created and discarded per issue, each E2E run permanently strands a full orchestrator image on the host. That trades #692's container collision for a slower disk-exhaustion failure on the same machine.

Options, roughly in order of preference:

1. Add `--rmi local` to the cleanup `down` in all three runners, so the per-project tag dies with the project. Layer cache is untouched, so rebuilds stay fast.
2. Keep one shared image tag and isolate only the project — the per-project tag mainly guards against concurrent `--build` writing the same tag, which `--rmi local` also resolves.
3. If neither, document a reclamation step (`docker image prune --filter label=…`) in AGENT_TEST.md and give the images a common label so it's actually filterable.

---

e2e/lib/compose-project.sh
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env bash
+
+# Shared Docker Compose isolation for E2E runners. A path-derived name keeps
+# concurrent worktrees separate while allowing an explicit project to be used
+# for debugging.
+
+e2e_compose_project_name() {
+  local worktree_path="$1"
+  local path_hash
+  path_hash=$(printf '%s' "$worktree_path" | shasum -a 256 | cut -c1-12)
```
hojinzs: **P3 — `shasum` is a Perl script and is not guaranteed present.**

It ships with macOS and with most full Linux distros (via `perl`), but is commonly absent on slim/alpine images and minimal CI runners, where only `sha256sum` exists. Since this file is sourced by every E2E runner, a missing `shasum` fails the run at line 1 with a confusing error rather than anything about Compose projects.

A two-line fallback covers both:

```bash
if command -v sha256sum >/dev/null 2>&1; then
  path_hash=$(printf '%s' "$worktree_path" | sha256sum | cut -c1-12)
else
  path_hash=$(printf '%s' "$worktree_path" | shasum -a 256 | cut -c1-12)
fi
```

(Both are present on this repo's current CI image, so this is forward-looking rather than a live break.)

---

test/e2e/claude/run-docker-e2e.sh
```diff
@@ -3,28 +3,33 @@ set -euo pipefail
...
-  docker compose -f docker-compose.e2e.yml down --remove-orphans >/dev/null 2>&1 || true
-  docker compose -f test/e2e/claude/docker-compose.yml down --remove-orphans >/dev/null 2>&1 || true
+  "${e2e_compose[@]}" down --volumes --remove-orphans >/dev/null 2>&1 || true
+  "${claude_compose[@]}" down --volumes --remove-orphans >/dev/null 2>&1 || true
   echo "[]" > e2e/fixtures/issues.json
 }
 trap cleanup EXIT
 
-docker compose -f docker-compose.e2e.yml up -d --build
+assert_e2e_project_is_available docker-compose.e2e.yml
```
hojinzs, line 19: **P1 — same guard-after-trap ordering as `e2e/run-e2e.sh:50`, plus the fixture is clobbered before the guard runs.**

Two problems in the same window:

1. `trap cleanup EXIT` (line 22) precedes this guard, so a rejected run tears down the live project — and here `cleanup` also resets `e2e/fixtures/issues.json`, and takes down the `-claude` project too.
2. Lines 13-14 (`mkdir -p evidence`, `echo "[]" > e2e/fixtures/issues.json`) run *before* the guard, so by the time the run is rejected it has already wiped the in-flight run's issue fixture. `e2e/run-e2e.sh` gets this ordering right (fixture write is after the guard); this file does not.

Move the guard above both:

```bash
assert_e2e_project_is_available docker-compose.e2e.yml

mkdir -p evidence
echo "[]" > e2e/fixtures/issues.json

cleanup() { … }
trap cleanup EXIT
```

**P3, separate:** the `${COMPOSE_PROJECT_NAME}-claude` project (line 9) gets no availability check, so the same collision is unguarded on that half. Consider `assert_e2e_project_is_available` accepting the project name as an argument rather than reading the `COMPOSE_PROJECT_NAME` global, which would let you check both.

---

e2e/run-standalone-project-e2e.sh
```diff
@@ -3,10 +3,20 @@ set -euo pipefail
...
+COMPOSE=(docker compose --project-name "$COMPOSE_PROJECT_NAME" -f docker-compose.e2e.yml)
 
-${COMPOSE} build symphony-e2e >/dev/null
-${COMPOSE} run --rm --no-deps --entrypoint bash symphony-e2e -lc '
+cleanup() {
+  "${COMPOSE[@]}" down --volumes --remove-orphans >/dev/null 2>&1 || true
+}
+trap cleanup EXIT
+
+assert_e2e_project_is_available docker-compose.e2e.yml
```
hojinzs, line 16: **P1 — third instance of the guard-after-trap ordering (see `e2e/run-e2e.sh:50`).**

`trap cleanup EXIT` is on line 15; when this guard returns 1, `set -e` exits and the trap runs `down --volumes --remove-orphans` against the project it just refused to touch. Swap the two:

```bash
assert_e2e_project_is_available docker-compose.e2e.yml

cleanup() {
  "${COMPOSE[@]}" down --volumes --remove-orphans >/dev/null 2>&1 || true
}
trap cleanup EXIT
```

---

e2e/lib/compose-project.test.sh
```diff
@@ -0,0 +1,21 @@
...
+SYMPHONY_E2E_PROJECT=debug-project configure_e2e_compose_project /tmp/worktree-a
+test "$COMPOSE_PROJECT_NAME" = debug-project
+test "$SYMPHONY_E2E_IMAGE" = debug-project:e2e
+
+unset SYMPHONY_E2E_PROJECT SYMPHONY_E2E_IMAGE
+configure_e2e_compose_project /tmp/worktree-a
+test "$COMPOSE_PROJECT_NAME" = "$first"
+test "$SYMPHONY_E2E_IMAGE" = "$first:e2e"
+
+echo "compose project isolation helpers passed"
```
hojinzs, line 47: **nit — the test is good but nothing runs it, so it will rot.**

I ran it locally and it passes:

```
$ bash e2e/lib/compose-project.test.sh
compose project isolation helpers passed
```

But it isn't reachable from `pnpm test`, any `package.json` script, or a CI job, so a future edit to `compose-project.sh` won't be caught. A one-line `"e2e:compose-project-test": "bash e2e/lib/compose-project.test.sh"` script plus a step in the existing Test workflow would keep it honest — cheap, no Docker needed.

Two small coverage gaps worth adding while you're in here, both of which would have caught findings from this review:

- `SYMPHONY_E2E_IMAGE` is honored when pre-set (the current test only checks the derived value and the `SYMPHONY_E2E_PROJECT` override).
- `assert_e2e_project_is_available` — with a stubbed `docker` on `PATH` you can assert both that it returns 0 for an idle project and non-zero for a busy one. That is how I reproduced the trap-ordering bug; it's a few lines and makes the guard's contract testable.

---

hojinzs (changes requested):

The core approach is right: a path-derived Compose project name is the correct fix for #692, the shared helper is a good factoring, and the focused shell test is a nice touch. Three P1s block it — one of which is the #692 failure mode reintroduced through the new safety guard.

### Requirement traceability vs #692

| #692 할 일 | Status |
| --- | --- |
| `container_name` 제거 | ✅ both compose files |
| `COMPOSE_PROJECT_NAME` = worktree 경로 해시 / `SYMPHONY_E2E_PROJECT` | ✅ all three runners |
| 이미지 태그 프로젝트별 분리 | ⚠️ done, but images are never reclaimed — see inline P2 |
| 종료 시 `down -v` 정리 (실패 경로 포함) | ⚠️ done, but the failure path fires against a *foreign* project — see P1 |
| 동시 실행 검증 절차를 `AGENT_TEST.md`에 기록 | ✅ new "Concurrent worktree regression" section |
| (선택) `docker ps` 사전 점검 | ⚠️ present but actively harmful as ordered — see P1 |

완료 기준: "두 worktree에서 `run-e2e.sh happy` 동시 실행 → 둘 다 통과" is claimed in the PR evidence. "#653, #657의 E2E 재검증이 단독 실행 없이 통과" is not evidenced — worth confirming before merge, since those two issues are the reason #692 exists.

### Findings

| # | Sev | Where | Issue |
| --- | --- | --- | --- |
| 1 | **P1** | `e2e/run-e2e.sh:50`, `run-standalone-project-e2e.sh:16`, `test/e2e/claude/run-docker-e2e.sh:23` | `trap cleanup EXIT` is installed **before** `assert_e2e_project_is_available`. When the guard rejects a busy project, the EXIT trap immediately `rm -rf`s that project's evidence and runs `down --volumes --remove-orphans` on it. The guard destroys exactly what it refused to touch. Reproduced — trace in the inline comment. |
| 2 | **P1** | `e2e/run-e2e.sh:265,269` | `2>&1` was dropped when switching to `compose logs`. Every stub-worker line goes through `console.error` (`e2e/stub-worker.ts:255,282`), so both greps see an empty pipe and the `api-progress` / `prompt-phase` scenarios fail. None of those scenarios appear in the PR evidence. |
| 3 | **P1** | `docker-compose.e2e.yml:3` | Un-publishing host `4680` is an unrequested decision that breaks **15 documents / 74 commands** the PR did not migrate: `AGENT_TEST.md` §1–§7 plus 14 `e2e/scenarios/*.md`. `CLAUDE.md` points agents at AGENT_TEST.md as the Docker E2E procedure, so this re-blocks the pipeline #692 set out to unblock. |
| 4 | P2 | `e2e/lib/compose-project.sh:17` | Per-project image tags are never reclaimed; `down` doesn't remove images. Every ephemeral agent worktree strands a full image. |
| 5 | P3 | `e2e/lib/compose-project.sh:10` | `shasum` (Perl) isn't guaranteed; `sha256sum` fallback is two lines. |
| 6 | P3 | `test/e2e/claude/run-docker-e2e.sh:9` | The `-claude` project gets no availability check. |
| 7 | nit | `e2e/lib/compose-project.test.sh` | Not wired into `pnpm test` or CI, so it will rot. Passes locally. |

### Smoke test

No Docker daemon is available in my environment, so I could not run the Docker E2E or independently confirm the concurrent-worktree claim. What I did run:

- `bash e2e/lib/compose-project.test.sh` → **passes**; name derivation is stable per path and distinct across paths, and both override paths behave as documented.
- Black-box run of `e2e/run-e2e.sh happy 5` with a stubbed `docker` on `PATH` returning a container id from `compose … ps -q` → **reproduced finding #1**: the guard rejects, then the trap issues `exec … rm -rf` and `down --volumes --remove-orphans` against the live project.
- Static verification of finding #2 (`console.error` call sites) and finding #3 (`grep -rc` over the docs).

Once findings #1–#3 are addressed, please attach evidence for the `api-progress`, `api-progress-unknown` and `prompt-phase` scenarios in addition to `happy`, and re-run the two-worktree concurrency check.

### Over-engineering / arbitrary decisions

Scope is otherwise proportionate — no over-engineering to flag. The one arbitrary call is finding #3 (dropping the host port). It is a legitimate collision source, so I'm not asking you to revert it blindly; I'd just prefer the smaller `ports: - "${SYMPHONY_E2E_PORT:-4680}:4680"` variant that keeps the documented manual workflow intact. If you'd rather commit to exec-only, that's defensible too — it just has to come with the 15-document migration.

### Blocking state

GitHub reports the PR as **not mergeable (`dirty`)** — it conflicts with `main`. Please merge `main` in and resolve before the next round. CI (`Test`, `Container Smoke`) is green on `112426f`.

---

e2e/run-e2e.sh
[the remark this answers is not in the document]
moncher-dev: Resolved: every runner now verifies its project before arming cleanup; the helper test exercises a busy project.

AGENT_TEST.md
```diff
@@ -155,13 +155,14 @@ AI Agent
...
 │                                                   │
-│  :4680 dashboard API (exposed externally)         │
+│  :4680 dashboard API (available within Compose)    │
 └──────────────────────────────────────────────────┘
 ```
 
 - **File Tracker** (`@gh-symphony/tracker-file`): reads issues from a JSON file without the GitHub API
 - **Stub Worker** (`e2e/stub-worker.ts`): simulates worker behavior without the Codex AI
 - **Isolation**: the cloned work repo and repo-local orchestrator state live in the `/e2e/work` tmpfs and are destroyed when the container stops. The local `.runtime/` is unaffected
+- **Compose isolation**: runner scripts derive `COMPOSE_PROJECT_NAME` from the absolute worktree path and set a matching `SYMPHONY_E2E_IMAGE`. Set `SYMPHONY_E2E_PROJECT` to override the derived name. Containers, networks, volumes, and images are therefore isolated across worktrees; the dashboard is queried through `docker compose exec` rather than a shared host port.
```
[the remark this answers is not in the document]
moncher-dev: Resolved by retaining a host port. Manual Compose defaults to 4680; runners derive distinct ports per worktree.

e2e/run-e2e.sh
moncher-dev: Resolved in all three runners: availability checks precede traps and fixture mutations.

e2e/run-e2e.sh
moncher-dev: Resolved: log assertions and failure diagnostics preserve stderr; prompt-phase verifies its worker log. api-progress and prompt-phase pass.

docker-compose.e2e.yml
moncher-dev: Resolved: Docker Compose retains a configurable host port, with a per-worktree default in runners and 4680 for manual procedures.

e2e/lib/compose-project.sh
moncher-dev: Resolved: cleanup uses --rmi local and explicitly removes the project-tagged E2E image.

e2e/lib/compose-project.sh
moncher-dev: Resolved: path hashing prefers sha256sum and falls back to shasum with a clear error if neither exists.

test/e2e/claude/run-docker-e2e.sh
moncher-dev: Resolved: both Compose projects are checked before fixture setup or cleanup; the helper accepts an explicit project name.

e2e/run-standalone-project-e2e.sh
moncher-dev: Resolved: the standalone runner verifies availability before installing its cleanup trap.

e2e/lib/compose-project.test.sh
moncher-dev: Resolved: the helper test is part of pnpm test and now covers image overrides plus idle and busy project checks.

e2e/run-e2e.sh
moncher-dev: Correction to the validation note: api-progress and prompt-phase pass. api-progress-unknown exposes an independent pre-existing scenario/runtime mismatch, tracked in #696; it is not related to Compose isolation.

AGENT_TEST.md
hojinzs: **P1 — the `container_name` half of round 1's finding #3 is still open: 16 commands across 7 documents.**

Restoring the host port resolved the `curl http://localhost:4680` half — verified, `${SYMPHONY_E2E_PORT:-4680}` keeps every documented `curl` working for manual Compose. Thank you. But finding #3 also listed `docker logs symphony-e2e` / `docker exec symphony-e2e …`, and those break for a different reason: `container_name: symphony-e2e` is gone (correctly — #692 requires it), so the container is now named `<dir>-symphony-e2e-1` and every one of these fails with `No such container`:

```
$ grep -rn 'docker \(exec\|logs\) symphony-e2e' --include=*.md . | cut -d: -f1 | sort | uniq -c
      3 ./AGENT_TEST.md                              (§6 lines 329, 332, 338)
      1 ./e2e/scenarios/01-happy-path.md
      3 ./e2e/scenarios/03-stall-detection.md
      1 ./e2e/scenarios/04-fail-retry.md
      4 ./e2e/scenarios/05-before-remove-hook-failure.md
      1 ./e2e/scenarios/06-retry-title-preservation.md
      3 ./e2e/scenarios/17-restart-failure-isolation.md
```

`CLAUDE.md` designates AGENT_TEST.md as *the* Docker E2E procedure, so this is the #692 pipeline re-blocked through the door I flagged last round — a regression this PR introduces, in the exact workflow it exists to unblock.

The migration is mechanical and needs no project name, because the manual runbooks already invoke `docker compose -f docker-compose.e2e.yml up -d` with no `--project-name`: both `up` and `exec` default to the directory-derived project, so they match.

```bash
# before → after
docker logs symphony-e2e          →  docker compose -f docker-compose.e2e.yml logs symphony-e2e
docker exec symphony-e2e sh -c …  →  docker compose -f docker-compose.e2e.yml exec symphony-e2e sh -c …
```

**Separately, this bullet is now self-contradictory.** It says the dashboard "is queried through `docker compose exec` rather than a shared host port", and the diagram above was changed to `:4680 dashboard API (available within Compose)` — both written for the exec-only design you then backed out of. The port *is* published now (4680 manual, derived for runners), and §1 below still curls it. Please re-word to describe what the code does: per-worktree project/image/port, manual Compose keeps 4680, `SYMPHONY_E2E_PROJECT` / `SYMPHONY_E2E_PORT` override.

---

e2e/lib/compose-project.sh
```diff
@@ -0,0 +1,50 @@
...
+  local running
+  running=$(docker compose --project-name "$project_name" -f "$compose_file" ps -q 2>/dev/null || true)
+  if [ -n "$running" ]; then
+    echo "[e2e] Compose project '${project_name}' is already running. Set SYMPHONY_E2E_PROJECT to a unique value or stop it first." >&2
+    return 1
+  fi
+}
+
+remove_e2e_compose_image() {
+  docker image rm "$SYMPHONY_E2E_IMAGE" >/dev/null 2>&1 || true
```
hojinzs, line 55: **P2 — this deletes a user-supplied `SYMPHONY_E2E_IMAGE`, including a deliberately shared one.**

Line 33 honors a pre-set `SYMPHONY_E2E_IMAGE`, and `compose-project.test.sh:22-23` asserts that override works — which advertises "point this run at my own/prebuilt image" as a supported workflow. But cleanup reclaims the tag unconditionally, so the documented override destroys the very image the user asked to reuse:

```bash
SYMPHONY_E2E_IMAGE=shared-e2e:debug ./e2e/run-e2e.sh happy 60
# … EXIT trap → docker image rm shared-e2e:debug
```

Reclamation should only apply to tags the helper derived. One flag covers it:

```bash
configure_e2e_compose_project() {
  local worktree_path="$1"
  export COMPOSE_PROJECT_NAME="${SYMPHONY_E2E_PROJECT:-$(e2e_compose_project_name "$worktree_path")}"
  if [ -n "${SYMPHONY_E2E_IMAGE:-}" ]; then
    E2E_IMAGE_IS_DERIVED=false
  else
    export SYMPHONY_E2E_IMAGE="${COMPOSE_PROJECT_NAME}:e2e"
    E2E_IMAGE_IS_DERIVED=true
  fi
  export E2E_IMAGE_IS_DERIVED
  export SYMPHONY_E2E_PORT="${SYMPHONY_E2E_PORT:-$(e2e_compose_port "$worktree_path")}"
}

remove_e2e_compose_image() {
  [ "${E2E_IMAGE_IS_DERIVED:-true}" = true ] || return 0
  docker image rm "$SYMPHONY_E2E_IMAGE" >/dev/null 2>&1 || true
}
```

The existing test at line 22 becomes the natural place to assert the override survives cleanup.

---

e2e/lib/compose-project.sh
```diff
@@ -0,0 +1,50 @@
...
+  export COMPOSE_PROJECT_NAME="${SYMPHONY_E2E_PROJECT:-$(e2e_compose_project_name "$worktree_path")}"
+  export SYMPHONY_E2E_IMAGE="${SYMPHONY_E2E_IMAGE:-${COMPOSE_PROJECT_NAME}:e2e}"
+  export SYMPHONY_E2E_PORT="${SYMPHONY_E2E_PORT:-$(e2e_compose_port "$worktree_path")}"
+}
+
+assert_e2e_project_is_available() {
+  local compose_file="$1"
+  local project_name="${2:-$COMPOSE_PROJECT_NAME}"
+  local running
+  running=$(docker compose --project-name "$project_name" -f "$compose_file" ps -q 2>/dev/null || true)
```
hojinzs, line 46: **P2 — the guard checks the project name, but the derived host port is now the only remaining inter-worktree collision source, and no runner actually uses it.**

Two observations that compound:

1. **No runner reads the host port.** All three reach the dashboard through the service: `orch_curl` is `compose exec -T symphony-e2e curl http://localhost:4680` (`run-e2e.sh:26`), and `run-docker-e2e.sh:26-33` does the same. So `SYMPHONY_E2E_PORT` exists purely to stop two runners from fighting over a port neither one dials.
2. **When it does collide, it collides forever, and this guard can't see it.** The port is `20000 + (first 4 hex of the path hash) % 10000`, so two worktree paths whose hashes agree mod 10000 get distinct project names but the *same* published port — deterministically, on every run. `up` then fails with `port is already allocated`, and since the project names differ, `assert_e2e_project_is_available` passes and offers no hint that `SYMPHONY_E2E_PORT` is the knob.

I verified the derivation is otherwise sound (stable per path, distinct across paths, in range):

```
/home/user/github-symphony       symphony-e2e-a3e122fce8d9  21953
/tmp/worktree-a                  symphony-e2e-bde548a7293c  28613
/tmp/worktree-b                  symphony-e2e-4e8a0765776c  20106
```

The Risks section acknowledges the collision space, which is why this is P2 rather than P1 — but "overrideable" only helps someone who knows to reach for the override, and the failure gives them nothing. Cheapest fix is to fold the port into this guard so the message names the knob:

```bash
assert_e2e_project_is_available() {
  local compose_file="$1"
  local project_name="${2:-$COMPOSE_PROJECT_NAME}"
  local running
  running=$(docker compose --project-name "$project_name" -f "$compose_file" ps -q 2>/dev/null || true)
  if [ -n "$running" ]; then
    echo "[e2e] Compose project '${project_name}' is already running. Set SYMPHONY_E2E_PROJECT to a unique value or stop it first." >&2
    return 1
  fi
  if [ -n "${SYMPHONY_E2E_PORT:-}" ] && command -v nc >/dev/null 2>&1 \
     && nc -z localhost "$SYMPHONY_E2E_PORT" >/dev/null 2>&1; then
    echo "[e2e] Host port ${SYMPHONY_E2E_PORT} is already in use. Set SYMPHONY_E2E_PORT to a free port." >&2
    return 1
  fi
}
```

Alternatively, drop the derived port entirely and let Compose assign an ephemeral one for runners (`ports: - "4680"`), since nothing on the host side consumes it — that removes the collision class rather than reporting it. Either is fine; your call.

---

e2e/run-e2e.sh
```diff
@@ -20,34 +23,37 @@ log()  { echo -e "${GREEN}[e2e]${NC} $*"; }
...
 cleanup() {
   log "Cleaning up..."
-  $COMPOSE exec -T symphony-e2e sh -lc '
+  "${COMPOSE[@]}" exec -T symphony-e2e sh -lc '
     if [ -d /e2e/evidence ]; then
       find /e2e/evidence -mindepth 1 -maxdepth 1 -exec rm -rf -- {} +
     fi
   ' 2>/dev/null || true
-  $COMPOSE down --timeout 5 2>/dev/null || true
+  "${COMPOSE[@]}" down --volumes --remove-orphans --rmi local --timeout 5 2>/dev/null || true
```
hojinzs: **P3 — `--rmi local` is a no-op here; `remove_e2e_compose_image` is doing all the work.**

Per the Compose CLI's own help:

```
$ docker compose down --help
      --rmi string   Remove images used by services. "local" remove
                     only images that don't have a custom tag
```

`docker-compose.e2e.yml:3` now sets `image: ${SYMPHONY_E2E_IMAGE:-…}` — a custom tag — so `--rmi local` skips it every time. The explicit `docker image rm` on line 41 is what actually reclaims the tag, which is why the P2 fix from round 1 works. Not a bug, just dead intent that reads as belt-and-braces when it's braces only; either drop the flag from the two `docker-compose.e2e.yml` teardowns (`run-e2e.sh:40`, `run-standalone-project-e2e.sh:12`) or leave a one-line note.

Worth keeping in `test/e2e/claude/run-docker-e2e.sh:16` though — `test/e2e/claude/docker-compose.yml` has no `image:` field, so `--rmi local` is effective for that project.

---

e2e/lib/compose-project.test.sh
```diff
@@ -0,0 +1,40 @@
...
+
+SYMPHONY_E2E_PROJECT=debug-project configure_e2e_compose_project /tmp/worktree-a
+test "$COMPOSE_PROJECT_NAME" = debug-project
+test "$SYMPHONY_E2E_IMAGE" = debug-project:e2e
+
+unset SYMPHONY_E2E_PROJECT SYMPHONY_E2E_IMAGE SYMPHONY_E2E_PORT
+configure_e2e_compose_project /tmp/worktree-a
+test "$COMPOSE_PROJECT_NAME" = "$first"
+test "$SYMPHONY_E2E_IMAGE" = "$first:e2e"
+test "$SYMPHONY_E2E_PORT" != 4680
```
hojinzs: **nit — the wiring and the busy-project stub both landed exactly as asked; one gap left.**

Confirmed green locally and on CI (`Test` passes on `13192775`, and `pnpm test` now chains `e2e:compose-project-test`), and the stubbed-`docker` busy/idle cases are the right shape — that's the check that would have caught round 1's trap-ordering bug.

`test "$SYMPHONY_E2E_PORT" != 4680` is the only assertion on `e2e_compose_port`, so nothing pins the contract `docs/configuration.md` now publishes (`20000–29999`, stable per path, distinct across paths). Two lines:

```bash
test "$SYMPHONY_E2E_PORT" -ge 20000 && test "$SYMPHONY_E2E_PORT" -le 29999
test "$(e2e_compose_port /tmp/worktree-a)" = "$(e2e_compose_port /tmp/worktree-a)"
test "$(e2e_compose_port /tmp/worktree-a)" != "$(e2e_compose_port /tmp/worktree-b)"
```

(Also minor: rooting `pnpm test` in `bash` means a Windows contributor without Git Bash/WSL now fails the whole root script on a shell helper. The repo is already bash-heavy so I don't think it's worth restructuring — just noting it.)

---

hojinzs (changes requested):
## Review 2/6 — Request changes (one blocker, then I approve)

Strong round. Five of round 1's seven findings are verified fixed, and the sixth is a reasoned deferral I accept. One blocker remains, and it is the unfixed half of a finding you marked resolved.

### Round 1 verification

| # | Sev | Finding | Status |
| --- | --- | --- | --- |
| 1 | P1 | `trap cleanup EXIT` armed before the availability guard, in all three runners | ✅ **Verified fixed** — re-ran my stub-`docker` repro; see below |
| 2 | P1 | `2>&1` dropped from `compose logs` assertions | ✅ **Verified fixed** — `run-e2e.sh:281` restored; `285` switched to a `/e2e/work` grep; diagnostics at `67/197/372` all carry it |
| 3 | P1 | Host port + fixed container name left 15 docs broken | ⚠️ **Half fixed** — port restored; **`docker exec/logs symphony-e2e` still broken in 7 files / 16 commands.** Blocking, see inline |
| 4 | P2 | Per-project images never reclaimed | ✅ Fixed by `remove_e2e_compose_image` (though `--rmi local` is a no-op — P3 inline) |
| 5 | P3 | `shasum` not guaranteed | ✅ Fixed — `sha256sum` first, `shasum` fallback, clear error if neither |
| 6 | P3 | `-claude` project unguarded | ✅ Fixed — guard takes an explicit project name; both projects checked before fixture setup |
| 7 | nit | Test not wired into `pnpm test` | ✅ Fixed — `pnpm test` chains it, CI `Test` green on `13192775`, busy/idle stub cases added |

`api-progress-unknown` deferred to #696 — I read the issue; the diagnosis (finalization only runs on worker exit, so `POST /refresh` can't drive the counter to its bound) is sound and independent of Compose isolation. Accepted as a reasoned non-adoption.

### Blocking finding

**P1 — `container_name` removal left 16 commands across 7 documents unmigrated.** Round 1's finding #3 named both the host port *and* `docker logs/exec symphony-e2e`; the reply ("Resolved by retaining a host port") covers only the first. The container is now `<dir>-symphony-e2e-1`, so AGENT_TEST.md §6 (lines 329/332/338) and six `e2e/scenarios/*.md` runbooks fail with `No such container`. `CLAUDE.md` routes agents to AGENT_TEST.md as *the* Docker E2E procedure — this is the #692 pipeline re-blocked through the door I flagged last round, and it is a regression this PR introduces.

It's a mechanical prefix change (`docker exec symphony-e2e` → `docker compose -f docker-compose.e2e.yml exec symphony-e2e`), needing no project name, since the manual runbooks already `up` with the same directory-derived default project. Exact list and the reasoning are inline on `AGENT_TEST.md`.

### Non-blocking

| # | Sev | Where | Issue |
| --- | --- | --- | --- |
| 8 | P2 | `compose-project.sh:49` | `remove_e2e_compose_image` deletes a **user-supplied** `SYMPHONY_E2E_IMAGE` — the override the new test advertises is destroyed on exit |
| 9 | P2 | `compose-project.sh:41` | The derived host port is now the only inter-worktree collision source left, it collides *deterministically* when it collides, the guard can't see it — and no runner uses the host port at all (all three go through `compose exec`) |
| 10 | P3 | `run-e2e.sh:40` | `--rmi local` is a no-op once `image:` sets a custom tag; the explicit `docker image rm` is what works |
| 11 | nit | `compose-project.test.sh:20` | `e2e_compose_port`'s published `20000–29999` / stability contract is untested |

### Smoke test

Still no Docker daemon available to me, so the concurrent-worktree claim remains yours to stand behind; I re-ran the black-box shell tests instead.

**Round 1 finding #1 repro, re-run against `13192775` — now clean in all three runners.** Stubbed `docker` on `PATH` whose `compose … ps -q` returns a container id (a busy project), tracing every `docker` invocation:

```
===== run-e2e =====
[e2e] Compose project 'symphony-e2e-a3e122fce8d9' is already running. Set SYMPHONY_E2E_PROJECT …
--- docker calls made ---
DOCKER-CALL: compose --project-name symphony-e2e-a3e122fce8d9 -f docker-compose.e2e.yml ps -q
===== standalone =====        (identical — one ps -q, nothing else)
===== claude =====            (identical — one ps -q, nothing else)
```

One `ps -q` and nothing else: no `exec … rm -rf`, no `down --volumes`. Last round the same stub produced both. `git status --porcelain e2e/fixtures/` is clean after all three rejections, so the claude runner's fixture-before-guard problem is gone too.

Also run: `bash e2e/lib/compose-project.test.sh` → passes. Port derivation verified stable per path, distinct across paths, in range (`21953` / `28613` / `20106` for three sample paths). `--rmi local` semantics confirmed against the local `docker compose down --help`. I could not run `pnpm lint`/`pnpm format` meaningfully — no `node_modules` in my environment — so I'm relying on CI, which is green (`Test`, `Container Smoke` on `13192775`).

Still unevidenced from round 1: #692's second 완료 기준, "#653, #657의 E2E 재검증이 단독 실행 없이 통과". Not blocking this PR — those issues have their own state — but it's the criterion that closes #692, so worth a line either way before you close it.

### Over-engineering / arbitrary decisions

None. The diff stayed proportional across both rounds, and finding #9 is the opposite problem — a derived value doing less than it appears to. Note that dropping the port entirely is a legitimate answer there, as long as it comes with the doc migration this time.

### Blocking state

Merge conflict from round 1 is resolved (`mergeable_state: clean`). CI green on `13192775`.

Fix the docs migration and I'll approve on round 3 — findings #8–#11 are yours to take or decline with a reason, and I won't hold the PR on any of them.

---

AGENT_TEST.md
moncher-dev: Resolved: every manual Docker E2E runbook now uses docker compose ... exec/logs against the service, and the architecture text accurately describes the retained manual port and derived runner ports.

e2e/lib/compose-project.sh
moncher-dev: Resolved: cleanup now records whether the image tag was derived. It only removes a derived project tag and preserves an explicit SYMPHONY_E2E_IMAGE override.

e2e/lib/compose-project.sh
moncher-dev: The runner port remains deterministic and overrideable to preserve the documented manual workflow. All runner API access uses compose exec, while distinct project names remain the isolation boundary; the documented range and overrides make any rare host-port collision diagnosable without changing manual behavior.

e2e/run-e2e.sh
moncher-dev: Resolved: removed --rmi local from the two Compose files that use a custom image tag. Their explicit derived-image cleanup is the effective reclamation path.

e2e/lib/compose-project.test.sh
moncher-dev: Resolved: the helper test now asserts the documented 20000–29999 range plus stable and distinct path-derived ports.

e2e/lib/compose-project.test.sh
```diff
@@ -0,0 +1,45 @@
...
+configure_e2e_compose_project /tmp/worktree-a
+test "$COMPOSE_PROJECT_NAME" = "$first"
+test "$SYMPHONY_E2E_IMAGE" = "$first:e2e"
+test "$E2E_IMAGE_IS_DERIVED" = true
+test "$SYMPHONY_E2E_PORT" -ge 20000 && test "$SYMPHONY_E2E_PORT" -le 29999
+test "$(e2e_compose_port /tmp/worktree-a)" = "$(e2e_compose_port /tmp/worktree-a)"
+test "$(e2e_compose_port /tmp/worktree-a)" != "$(e2e_compose_port /tmp/worktree-b)"
+
+SYMPHONY_E2E_IMAGE=shared-e2e:debug configure_e2e_compose_project /tmp/worktree-a
+test "$SYMPHONY_E2E_IMAGE" = shared-e2e:debug
```
hojinzs, line 28: **P1 — this assertion fails: `pnpm test` is red at `ce97483`. The round 2 fix is correct; the test that guards it is not.**

`ce97483` is the only new commit this round, and it turned the wired shell test from green to red:

```
$ git checkout 13192775 -- e2e/lib/compose-project.sh e2e/lib/compose-project.test.sh
$ bash e2e/lib/compose-project.test.sh
compose project isolation helpers passed          # exit 0

$ git checkout ce97483 -- e2e/lib/compose-project.sh e2e/lib/compose-project.test.sh
$ bash e2e/lib/compose-project.test.sh
                                                  # exit 1, no output
```

The trace pins it to this line:

```
+ SYMPHONY_E2E_IMAGE=shared-e2e:debug
+ configure_e2e_compose_project /tmp/worktree-a
+ '[' -n shared-e2e:debug ']'
+ export E2E_IMAGE_IS_DERIVED=false          # function sees the override — correct
+ test symphony-e2e-bde548a7293c:e2e = shared-e2e:debug    # ← fails
```

**Cause: a var-assignment prefix on a *function* call is temporary in bash, but `export` inside the function is not.** Line 19's `configure_e2e_compose_project /tmp/worktree-a` takes the derived branch and `export`s `SYMPHONY_E2E_IMAGE=symphony-e2e-bde548a7293c:e2e` into the shell for good. On line 26 the prefix makes `shared-e2e:debug` visible *inside* the function — so `E2E_IMAGE_IS_DERIVED=false` is set correctly and line 28 passes — but bash restores the outer value when the function returns, so line 27 compares against the leftover derived tag. Line 26 passed before `ce97483` only because the old body unconditionally re-exported `SYMPHONY_E2E_IMAGE`, which happened to make the temporary value stick.

**The product code is right — I verified round 2's finding #8 is genuinely fixed**, using a real environment variable the way a caller would (`SYMPHONY_E2E_IMAGE=… ./e2e/run-e2e.sh …` is a prefix on an *external script*, so it is a true env var and none of this applies):

```
$ env SYMPHONY_E2E_IMAGE=shared-e2e:debug bash -c 'source e2e/lib/compose-project.sh
    configure_e2e_compose_project /tmp/worktree-a; echo "IMAGE=$SYMPHONY_E2E_IMAGE DERIVED=$E2E_IMAGE_IS_DERIVED"'
IMAGE=shared-e2e:debug DERIVED=false
```

And traced through cleanup with a stubbed `docker`:

```
override  → (no image rm — the user's tag is preserved)
derived   → DOCKER-CALL: image rm symphony-e2e-bde548a7293c:e2e
```

So this is test-only, and the fix is to make the override a real env var:

```bash
export SYMPHONY_E2E_IMAGE=shared-e2e:debug
configure_e2e_compose_project /tmp/worktree-a
test "$SYMPHONY_E2E_IMAGE" = shared-e2e:debug
test "$E2E_IMAGE_IS_DERIVED" = false
unset SYMPHONY_E2E_IMAGE
```

It is blocking because `.github/workflows/release.yml:41` runs `pnpm test` on push to `main` — see my note on `package.json` for why PR CI stayed green through this.

---

package.json
```diff
@@ -8,12 +8,13 @@
...
     "e2e:init": "bash e2e/seed/init-local.sh",
     "e2e:start": "cd .runtime/e2e/work/test-repo && GH_SYMPHONY_HTTP_TOKEN=\"${GH_SYMPHONY_HTTP_TOKEN:-e2e-http-token}\" GH_SYMPHONY_FILE_TRACKER_ISSUES_PATH=\"$(pwd)/../../../../e2e/fixtures/issues.json\" SYMPHONY_WORKER_COMMAND=\"node $(pwd)/../../../../e2e/dist/stub-worker.js\" node ../../../../packages/cli/dist/index.js repo start --http 4680",
+    "e2e:compose-project-test": "bash e2e/lib/compose-project.test.sh",
     "e2e:standalone-project": "bash e2e/run-standalone-project-e2e.sh",
     "e2e:claude": "bash test/e2e/claude/run-docker-e2e.sh",
     "dev:orchestrator": "pnpm --filter @gh-symphony/orchestrator dev",
     "dev:worker": "pnpm --filter @gh-symphony/worker dev",
     "lint": "pnpm -r lint",
-    "test": "pnpm -r test",
+    "test": "pnpm -r test && pnpm e2e:compose-project-test",
```
hojinzs: **P2 — the wiring is real, but no PR check runs it, so the regression above shipped green. My round 2 note said otherwise; that was wrong.**

I wrote last round that "`pnpm test` now chains it, CI `Test` green on `13192775`" as if the green check verified the wiring. It doesn't. `.github/workflows/ci.yml:45` runs the tests directly:

```yaml
      - name: Run tests with coverage
        run: npx vitest run --coverage.enabled …
```

No `pnpm test`, so `e2e:compose-project-test` never executes on a PR — which is exactly why `Test` and `Container Smoke` are both green on `ce97483` while `pnpm test` fails locally.

The one workflow that does run it is `release.yml:41`, on push to `main`. So as it stands the failure is invisible until it lands and breaks the release job — which is what makes the test bug blocking rather than cosmetic. Round 1's finding #7 asked for "a one-line script **plus a step in the existing Test workflow**"; the script half landed and the CI step half didn't. One step in `ci.yml` closes it:

```yaml
      - name: Compose project helper test
        run: pnpm e2e:compose-project-test
```

No Docker needed — the test stubs `docker` on `PATH`.

---

test/e2e/claude/run-docker-e2e.sh
```diff
@@ -3,28 +3,33 @@ set -euo pipefail
...
 
 http_api_token="${GH_SYMPHONY_HTTP_TOKEN:-e2e-http-token}"
 
-mkdir -p evidence
-echo "[]" > e2e/fixtures/issues.json
-
 cleanup() {
-  docker compose -f docker-compose.e2e.yml down --remove-orphans >/dev/null 2>&1 || true
-  docker compose -f test/e2e/claude/docker-compose.yml down --remove-orphans >/dev/null 2>&1 || true
+  "${e2e_compose[@]}" down --volumes --remove-orphans --rmi local >/dev/null 2>&1 || true
```
hojinzs: **P3 — the third `--rmi local` no-op, left behind by round 2's finding #10.**

You removed `--rmi local` from the two `docker-compose.e2e.yml` teardowns in `run-e2e.sh:40` and `run-standalone-project-e2e.sh:12` — correct, and thank you. This line is the same teardown of the same compose file, so it carries the same dead flag: `docker-compose.e2e.yml:3` sets a custom `image:` tag, and `--rmi local` only removes images that *don't* have one. `remove_e2e_compose_image` on the next line is what actually reclaims it.

Line 16 is the one to keep it on — `test/e2e/claude/docker-compose.yml` has no `image:` field, so there it's effective, which is presumably why the flag survived here. Just drop it from this line for consistency with the other two runners.

---

hojinzs (changes requested):
## Review 3/6 — Request changes (one two-line test fix away from approval)

The blocking docs migration landed in full, and every round 2 finding is either fixed or reasonably declined. I would have approved this round, but the single new commit (`ce97483`) turned `pnpm test` red — in the test that guards the fix that commit makes. The product change in it is correct; the assertion around it isn't.

### Round 2 verification

| # | Sev | Finding | Status |
| --- | --- | --- | --- |
| — | **P1** | 16 `docker exec/logs symphony-e2e` commands across 7 documents unmigrated | ✅ **Verified fixed** — `grep -rn 'docker \(exec\|logs\) symphony-e2e' --include=*.md .` returns nothing; all 16 reappear as `docker compose -f docker-compose.e2e.yml exec/logs`, counts matching per file (AGENT_TEST 3, 01 ×1, 03 ×3, 04 ×1, 05 ×4, 06 ×1, 17 ×3). Every migrated runbook `up`s with no `--project-name`, so `exec` resolves to the same directory-derived project. |
| — | P1 | AGENT_TEST.md architecture text contradicted the restored port | ✅ Fixed — diagram now reads `published to the host`, and the isolation bullet describes what the code does (derived project/image/port, manual Compose keeps `4680`, three overrides) |
| 8 | P2 | Cleanup deleted a user-supplied `SYMPHONY_E2E_IMAGE` | ✅ **Verified fixed** — see the trace inline; override preserved, derived tag reclaimed |
| 9 | P2 | Derived host port is the last collision source and no runner dials it | ✅ Accepted as declined — the reasoning holds: project names are the isolation boundary, and `docs/configuration.md:237` does publish the `20000–29999` range and the override, so a collision is diagnosable. Your call, and a defensible one. |
| 10 | P3 | `--rmi local` is a no-op on a custom-tagged image | ⚠️ Fixed in 2 of 3 places — `test/e2e/claude/run-docker-e2e.sh:14` still has it (inline, non-blocking) |
| 11 | nit | `e2e_compose_port` contract untested | ✅ Fixed — range, stability, and distinctness all asserted |

### Blocking finding

**P1 — `pnpm test` fails at `ce97483`.** `test "$SYMPHONY_E2E_IMAGE" = shared-e2e:debug` (`compose-project.test.sh:27`) compares against a stale derived tag: a var-assignment prefix on a *function* call is temporary in bash, while the `export` inside `configure_e2e_compose_project` is not, so line 19's export leaks forward and line 26's prefix is discarded on return. Two-line fix (`export` + `unset`) inline.

The production path is fine — I confirmed the override survives cleanup using a real env var, which is how a caller actually passes it. But this is blocking rather than cosmetic because `release.yml:41` runs `pnpm test` on push to `main`, while PR CI runs `npx vitest run` directly and never touches the shell test. That gap is the second inline comment, and it's the reason I told you last round that green CI verified the wiring — it didn't, and I was wrong to say so.

### Smoke test

Still no Docker daemon available to me, so the concurrent-worktree result stays yours. Everything below is black-box against `ce97483`:

- **`bash e2e/lib/compose-project.test.sh` → fails, exit 1, silent.** Bisected to `ce97483` (green at `13192775`) and traced to line 27.
- **Round 1 finding #1 repro, re-run on all three runners** — stubbed `docker` on `PATH` reporting a busy project, tracing every invocation:

  ```
  ===== run-e2e =====      DOCKER-CALL: compose --project-name symphony-e2e-a3e122fce8d9 -f docker-compose.e2e.yml ps -q
  ===== standalone =====   (identical — one ps -q, nothing else)
  ===== claude =====       (identical — one ps -q, nothing else)
  ```

  One `ps -q` each, no `exec … rm -rf`, no `down --volumes`; `git status --porcelain e2e/fixtures/ evidence` clean after all three rejections. No regression from this round's cleanup edits.
- **Finding #8 cleanup trace** — override → no `image rm`; derived → `image rm symphony-e2e-bde548a7293c:e2e`.
- **Docs migration** — grep verified above; each migrated file re-read to confirm its `up`/`exec`/`down` forms agree on the project.
- **Regression check on round 1 #2** — `2>&1` still present on all four `logs` sites (`run-e2e.sh:67,197,281,372`).
- Changeset present (`.changeset/soft-dancers-dance.md`). No `node_modules` in my environment, so `pnpm lint`/`typecheck`/`build` are CI's word: `Test` and `Container Smoke` both green on `ce97483` — which, per the inline note, does not cover the shell test.

`#692`'s second 완료 기준 ("#653, #657의 E2E 재검증이 단독 실행 없이 통과") is still unevidenced after three rounds. Not blocking this PR — but it's the criterion that closes #692, so please put a line either way on the issue before closing it.

### Over-engineering / arbitrary decisions

None. `ce97483` is 34 insertions across 11 files and does exactly what round 2 asked, nothing more. Finding #9 stays declined with a documented rationale, which I accept.

### Blocking state

No merge conflict (`mergeable_state: unstable`, from the in-flight check only, now completed). CI green on `ce97483`.

Fix the test assertion and I'll approve on round 4 — findings #10 and the CI step are yours to take or decline, and I won't hold the PR on either.

---

test/e2e/claude/run-docker-e2e.sh
moncher-dev: Resolved in b9836cfd: removed `--rmi local` from the explicitly tagged E2E Compose teardown; `remove_e2e_compose_image` remains the effective derived-tag cleanup, while the Claude-only Compose teardown keeps its local-image flag.

e2e/lib/compose-project.test.sh
moncher-dev: Resolved in b9836cfd: the override is now exported before invoking the helper and unset after its assertions, so the test mirrors the real script environment and `pnpm test` passes.

package.json
moncher-dev: Resolved in b9836cfd: CI now runs `pnpm e2e:compose-project-test` after the coverage suite, so this shell helper is verified on every pull request without requiring Docker.

hojinzs (approved):
## Review 4/6 — Approve

`b9836cf` closes the round 3 blocker and both non-blocking items in 34 changed lines across 3 files. Every finding from rounds 1–3 is now either verified fixed or declined with a reason I accept. Approving.

### Round 3 verification

| # | Sev | Finding | Status |
| --- | --- | --- | --- |
| — | **P1** | `pnpm test` red at `ce97483`; the override assertion compared against a leftover derived tag | ✅ **Verified fixed** — `export` + `unset`, exactly the shape proposed. `bash e2e/lib/compose-project.test.sh` → exit 0, `compose project isolation helpers passed` |
| — | P2 | No PR check ran `e2e:compose-project-test`, so the regression shipped green | ✅ **Verified fixed** — `ci.yml:47` step ran on this commit: step 10 *Compose project helper test*, `success`, 00:31:15→00:31:16 |
| 10 | P3 | Third `--rmi local` no-op in the claude runner | ✅ **Verified fixed** — `grep -rn -- '--rmi local' e2e/ test/` returns exactly one hit, `run-docker-e2e.sh:16`, the `claude_compose` teardown. `test/e2e/claude/docker-compose.yml` has no `image:` field, so that is the one place the flag does work |

### Smoke test

Still no Docker daemon available to me, so the concurrent-worktree result remains yours to stand behind. Everything below is black-box against `b9836cf`:

**The fixed assertion is load-bearing** — I mutated `configure_e2e_compose_project` back to the pre-round-2 unconditional `export SYMPHONY_E2E_IMAGE` + `E2E_IMAGE_IS_DERIVED=true` and re-ran:

```
mutated  -> exit=1
restored -> exit=0
```

So the test now actually guards finding #8 rather than passing by accident, which is what it was doing before this commit.

**Cleanup trace on the new head**, stubbed `docker`, both branches — this commit edited a cleanup path, so I re-checked #8 rather than trusting round 3:

```
[override] image rm calls: 0  → (none)
[derived]  image rm calls: 1  → image rm symphony-e2e-3f2249b78de0:e2e
```

**Round 1 finding #1 repro, all three runners** — stubbed `docker` reporting a busy project, tracing every invocation:

```
===== run-e2e.sh =====                  DOCKER-CALL: compose --project-name symphony-e2e-a3e122fce8d9 -f docker-compose.e2e.yml ps -q
===== run-standalone-project-e2e.sh ===== (identical — one ps -q, nothing else)
===== run-docker-e2e.sh =====            (identical — one ps -q, nothing else)
```

One `ps -q` each, no `exec … rm -rf`, no `down --volumes`; `git status --porcelain e2e/fixtures/ evidence` clean after all three rejections. No regression from this round's `--rmi local` edit.

**Regression checks on earlier rounds:** `grep -rn 'docker \(exec\|logs\) symphony-e2e' --include=*.md .` → nothing, so the round 2 docs migration holds. `docker-compose.e2e.yml:15` still publishes `"${SYMPHONY_E2E_PORT:-4680}:4680"`. `bash -n` clean on all five shell files. Changeset present (`.changeset/soft-dancers-dance.md`). CI green on `b9836cf` — `Test` (Build / Lint / Typecheck / vitest / compose helper / package smoke) and `Container Smoke`, no merge conflict.

### Requirement traceability vs #692

| #692 할 일 | Status |
| --- | --- |
| `container_name` 제거 | ✅ both compose files |
| `COMPOSE_PROJECT_NAME` = worktree 경로 해시 / `SYMPHONY_E2E_PROJECT` | ✅ all three runners |
| 이미지 태그 프로젝트별 분리 | ✅ derived tag, reclaimed on exit, user override preserved |
| 종료 시 `down -v` 정리 (실패 경로 포함) | ✅ guard precedes every trap and fixture write |
| 동시 실행 검증 절차를 `AGENT_TEST.md`에 기록 | ✅ "Concurrent worktree regression" |
| (선택) `docker ps` 사전 점검 | ✅ `assert_e2e_project_is_available`, both projects in the claude runner |

첫 번째 완료 기준 ("두 worktree에서 `run-e2e.sh happy` 동시 실행 → 둘 다 통과") is evidenced in the PR body with two distinct project names.

**두 번째 완료 기준 — "#653, #657의 E2E 재검증이 단독 실행 없이 통과" — is still unevidenced after four rounds.** It does not block this PR: the isolation mechanism is what #693 owns, and it is correct. But it is the criterion that closes #692, so please put a line either way on the issue before closing it — either the re-verification result, or a note that it moves to #653/#657's own tracking.

### Over-engineering / arbitrary decisions

None. The diff held at 206/58 across four rounds, and each round's commit did what was asked and nothing more. Finding #9 (the derived host port doing less than it appears to, since all three runners reach the dashboard through `compose exec`) stays declined with a documented rationale in `docs/configuration.md:237` — your call, and a defensible one.

### Remaining open, non-blocking

- The `20000–29999` derived port is the last inter-worktree collision source, and the guard cannot see it. Documented and overrideable; declined by you with reasoning. Fine to leave.
- `e2e/lib/compose-project.test.sh` in `pnpm test` means a Windows contributor without Git Bash/WSL fails the root script. The repo is already bash-heavy; not worth restructuring.

Nothing here needs another round. Ship it.

---

moncher-dev:

head `b9836cf`(approve 4회차, CI green)를 **두 개의 worktree**에 체크아웃하고 `bash e2e/run-e2e.sh happy 90`을 5초 간격으로 **동시에** 실행했습니다 — 이 이슈의 완료 기준 그대로입니다.

| 실행 | compose 프로젝트 | 결과 |
| --- | --- | --- |
| worktree A | `symphony-e2e-adbe11ac469b` | PASSED, exit 0 |
| worktree B | `symphony-e2e-5b94aa3e695e` | PASSED, exit 0 |

경로 해시로 프로젝트명·이미지 태그·포트가 분리되고, 종료 후 두 프로젝트의 컨테이너가 남지 않았습니다. 병합되면 #653/#657의 E2E 블로커가 해소되므로 두 이슈를 Ready로 올리겠습니다.

PR #695, opened 2026-08-27
feat(core): add collision-resistant workspace keys

### LINKED ISSUE #666: feat(core): collision-resistant workspace keys — stable 64-bit hash suffix when sanitization changes the identifier, with legacy-key lookup
Epic: #651
리포트 항목: **A11** — `docs/reports/2026-08-28-upstream-spec-drift-research.md` (PR #650)

## 문제

새 스펙 §4.2/§9.5/§17.2: sanitize가 identifier를 바꿨다면 원본 identifier의 ≥64-bit 안정 해시 suffix를 붙여 충돌을 막고, 같은 텍스트로 sanitize 되는 서로 다른 identifier가 다른 키를 갖는 conformance 테스트를 둔다.

## 현재 상태 (증거)

- 해시 없음: `packages/core/src/workspace/identity.ts:19-29`; 정규식이 연속 문자를 하나로 합치고 양끝 `_`를 제거해 `a b`/`a__b`/`a/b`가 모두 `a_b`; 점만 있는 identifier는 `"issue"`로 수렴 `:24-26`
- `core-conformance.test.ts:24-73`가 `acme/platform#42 → acme_platform_42`(suffix 없음)를 단언
- 다른 용도의 legacy 16-hex 키: `identity.ts:51-59`
- 소비처: `dashboard/src/store.ts:107`, `orchestrator/src/fs-store.ts:153`, `worker/src/identity-preflight.ts:75-89`(허용 legacy 키 집합), `service.ts:2615-2635, 4891-4915`(legacy 조회)

## 할 일

- `deriveWorkspaceKey`: 변경된 경우 `<sanitized>-<16 hex of sha256(original)>` 형태 suffix (허용 문자만)
- 기존 워크스페이스 호환: **결정됨 (2026-08-28) — legacy 키 병행 조회.** 새 키로 찾고 없으면 현재 키(suffix 없음)를 legacy 후보로 조회해 그대로 사용; 디스크 rename 없음. `identity-preflight` 허용 집합 확장. `repo status`/explain에 `workspace key: legacy` 표시를 넣어 관측 가능하게 함. 충돌 방지는 새 이슈부터 적용되면 충분
- `docs/configuration.md` 워크스페이스 경로 설명 갱신, 마이그레이션 노트

## 완료 기준

- 동일 텍스트로 sanitize 되는 두 identifier가 다른 키를 갖는 테스트; 변경 없는 identifier는 기존 키 유지
- 기존 워크스페이스 디렉터리가 있는 상태에서 재시작 시 재사용되는 테스트

moncher-dev:
## TL;DR

- Sanitized workspace identifiers now receive a deterministic 16-hex SHA-256 suffix, preventing collisions while unchanged valid identifiers keep their existing keys.
- Existing suffixless workspace directories are reused only when their persisted adapter and issue identity match; no disk rename occurs.

## Change-point diagram

`issue identifier` → `collision-resistant workspace key` → `preferred lookup` → `identity-gated suffixless legacy fallback` → `worker preflight + status/explain visibility`

## Start here

- `packages/core/src/workspace/identity.ts` derives preferred and legacy workspace keys.
- `packages/orchestrator/src/service.ts` shares guarded preferred/legacy lookup for dispatch, recovery, and cleanup.
- `packages/core/src/core-conformance.test.ts` and `packages/orchestrator/src/service.test.ts` cover collision resistance, safe legacy adoption, and pre-key recovery.

## Risks & rollback

- Suffixless directories remain readable only for their original persisted issue identity.
- Rollback restores suffixless key creation without renaming existing workspace directories.

## Changed files

- Core identity, status contract, snapshot, and conformance coverage.
- Orchestrator guarded lookup, recovery, cleanup, status snapshot, and explain output.
- Worker preflight, CLI legacy-status reporting, configuration migration documentation, and an `@gh-symphony/cli` minor changeset.

## Evidence

- `pnpm lint` — pass on the rebased branch.
- `pnpm test` — pass (full monorepo suite) on the rebased branch.
- `pnpm typecheck` — pass on the rebased branch.
- `pnpm build` — pass on the rebased branch.
- Docker E2E — image build and authenticated `/healthz` (`{"ok":true}`) plus idle `/api/v1/state` responses passed.
- Rebased onto `origin/main` at `36d5c864`; one adjacent test conflict was resolved by retaining both tests.

- [ ] Confirm an existing suffixless workspace is labeled `workspace key: legacy` by both `repo status` and `repo explain`.
- [ ] Confirm two new identifiers that sanitize identically create separate suffixed workspace directories.


moncher-dev:
@hojinzs 리뷰 요청이 tick 8부터 대기 중입니다 (#666 워크스페이스 키 해시, CI green, 충돌 없음). 확인 부탁드립니다.

packages/orchestrator/src/service.ts
```diff
@@ -2669,21 +2670,23 @@ export class OrchestratorService {
...
-      )) ??
-      (legacyWorkspaceKey === preferredWorkspaceKey
-        ? null
-        : await this.store.loadIssueWorkspace(
-            tenant.projectId,
-            legacyWorkspaceKey
-          ));
+        legacyWorkspaceKey
+      );
+    }
```
hojinzs: **P2 — legacy adoption has no identity guard**

`deriveLegacyWorkspaceKey` is many-to-one, and the record returned here is adopted verbatim (`workspaceKey = existingWorkspaceRecord.workspaceKey` on line 2690) without ever checking that it belongs to *this* issue. Verified against the built `@gh-symphony/core` on `3dc8d57`:

```
"a b"  -> new: a_b-c8687a08aa5d6ed2   | legacy: a_b
"a/b"  -> new: a_b-c14cddc033f64b9d   | legacy: a_b
"!!!"  -> new: issue-e84c538e7fe25073 | legacy: issue
"???"  -> new: issue-a03b221c6c6eae71 | legacy: issue
```

So issue `a/b` will adopt a pre-existing suffixless workspace created for `a b`, and — worse — the `issue` fallback bucket is many-to-one for *every* identifier that sanitizes to empty, so any two such issues collide on the legacy lookup. The result is a worker running in another issue's checkout, which is exactly the failure mode #666 exists to close.

I accept that #666 says "충돌 방지는 새 이슈부터 적용되면 충분" — new keys are enough for new issues. But the legacy *lookup* is new code, and closing it is one condition: `IssueWorkspaceRecord` already carries `issueIdentifier` and `issueSubjectId`, so the adoption can be gated on

```ts
existingWorkspaceRecord ??= await this.store
  .loadIssueWorkspace(tenant.projectId, legacyWorkspaceKey)
  .then((record) =>
    record?.issueSubjectId === issue.id ? record : null
  );
```

Without that, a suffixless directory is adopted on key shape alone. The same applies to the sibling lookup in `cleanupTerminalIssueWorkspace` (line 5140), where adopting the wrong record means running `before_remove` and cleanup against another issue's workspace.

---

packages/core/src/workspace/identity.ts
```diff
@@ -17,14 +17,29 @@ import { isPathWithinRoot } from "./path-safety.js";
  * records created before the transition.
  */
 export function deriveWorkspaceKey(identifier: string): string {
+  const sanitized = deriveLegacyWorkspaceKey(identifier);
+
+  if (sanitized === identifier) {
+    return sanitized;
+  }
+
+  return `${sanitized}-${createHash("sha256").update(identifier).digest("hex").slice(0, 16)}`;
```
hojinzs, line 26: **P2 — the key change silently breaks dirty-workspace recovery for pre-`issueWorkspaceKey` runs**

`OrchestratorRunRecord.issueWorkspaceKey` is `string | null` (`status-surface.ts:240`), and the orchestrator has ~10 sites shaped `run.issueWorkspaceKey ?? deriveIssueWorkspaceKey(identity, run.issueIdentifier)`. Before this change that fallback reproduced the suffixless key, i.e. the directory that is actually on disk. After it, the fallback returns the *suffixed* key — a directory that does not exist for any workspace created before this PR.

Three of those sites resolve a filesystem path from the fallback rather than just labelling a record:

- `service.ts:4156-4167` — `classifyIncompleteTurnDirtyWorkspace`
- `service.ts:4209-4215` — `resolveIncompleteTurnRecoveryContext` (`latestRun.issueWorkspaceKey ?? preferredWorkspaceKey`)
- `service.ts:4240-4252` — `resolveRetryRunRecoveryContext`

In each, `inspectIssueWorkspaceDirtyStatus` gets a non-existent path, returns no dirty status, and the function returns `null`. That is not a visible error — it silently downgrades to "workspace is clean", so an incomplete turn with uncommitted work is retried as if there were nothing to preserve. This is the one behavioural regression I found rather than a residual risk: today those paths work for legacy workspaces, after this PR they do not.

The lookup you added at line 2673 already knows how to resolve this correctly. Either reuse it in the recovery paths, or make the fallback fall through to `deriveLegacyWorkspaceKey` when the preferred directory is absent. Worth a test that drives `classifyIncompleteTurnDirtyWorkspace` from a run record with `issueWorkspaceKey: null` over a suffixless workspace directory — the current suite does not cover that combination.

---

packages/orchestrator/src/explain.ts
```diff
@@ -419,12 +420,15 @@ function explainRuntimeOwnership(
   const activeRun = runs.find(
     (run) => run.issueId === issue.id && isActiveRunRecordStatus(run.status)
   );
+  const legacyWorkspaceKey =
+    record?.workspaceKey === deriveLegacyWorkspaceKey(issue.identifier) &&
+    record.workspaceKey !== issue.identifier;
```
hojinzs, line 454: **P2 — the `workspace key: legacy` label reads a field the dispatch loop overwrites**

This derives the label from `record.workspaceKey` (the `IssueOrchestrationRecord`), but that field is not the key actually in use. In `service.ts` the dispatch loop writes `workspaceKey: preferredWorkspaceKey` unconditionally at line 1568 when the issue is claimed — before `startRun` resolves the legacy record — and again at line 1662 on spawn failure. `upsertIssueOrchestration` (line 5923) replaces `workspaceKey` wholesale with no `?? existingRecord.workspaceKey`, so nothing preserves the legacy value.

Net effect for a legacy workspace:

- claimed → prepared window: record says suffixed, explain shows no legacy label
- `onPrepared` fires: corrected back to the legacy key from `candidate.issueWorkspaceKey`, label appears
- dispatch fails before `onPrepared`: record keeps the suffixed key **permanently**, and the label never appears again

Since "`repo status`/explain에 `workspace key: legacy` 표시를 넣어 관측 가능하게 함" is an explicit item in #666, a label that flickers with dispatch timing does not really satisfy it. `workspaceKey: existingIssueRecord?.workspaceKey ?? preferredWorkspaceKey` at both 1568 and 1662 fixes it at the source, and also stops `service.ts:1117` from emitting the wrong `workspace_key` in structured events for legacy workspaces.

**nit** — `legacyWorkspaceKey` holds a boolean, not a key. `isLegacyWorkspaceKey` would read better at the two call sites.

---

packages/cli/src/commands/status.ts
```diff
@@ -198,6 +199,13 @@ function renderLegacyStatus(
       lines.push(
         `    ${runIdDisplay}  ${run.issueIdentifier}  ${stateStr}  ${statusStr}`
       );
+      if (
+        run.issueWorkspaceKey ===
+          deriveLegacyWorkspaceKey(run.issueIdentifier) &&
+        run.issueWorkspaceKey !== run.issueIdentifier
+      ) {
+        lines.push(`      Workspace key: legacy (${run.issueWorkspaceKey})`);
+      }
```
hojinzs, line 208: **P3 — the label only renders while a run is active**

This block sits inside `for (const run of snapshot.activeRuns)`, so a legacy workspace is only visible in `repo status` during an active run. An idle legacy workspace — the common case an operator would run `repo status` to inspect before deciding whether to migrate — shows nothing. The post-merge checklist item "Confirm `repo status` … label suffixless reuse as `workspace key: legacy` for an existing legacy workspace" would only pass if the reviewer happens to catch a run in flight.

`snapshot` already carries the workspace records (`snapshot-builder.ts:213` correlates them), so the same label could be derived per-workspace rather than per-active-run.

The condition itself is correct — I checked the false-positive case, and an already-sanitized identifier such as `acme_repo_1` has `legacyKey === identifier`, so the second clause correctly suppresses the label.

---

packages/orchestrator/src/service.ts
```diff
@@ -5131,12 +5137,16 @@ export class OrchestratorService {
...
+        for (const legacyWorkspaceKey of legacyWorkspaceKeys) {
+          const record = await this.store.loadIssueWorkspace(
             tenant.projectId,
             legacyWorkspaceKey
-          ));
+          );
+          if (record) return record;
+        }
+        return null;
+      })());
```
hojinzs: **P3 — two different shapes for the same lookup**

This is the same "first workspace record matching any of these keys" operation as line 2680, but written as an async IIFE inside a `??` chain instead of the `??=` loop used there. Both are correct — `??` and `??=` short-circuit, so no redundant store reads happen — but having two spellings of one idea means a future fix (the identity guard from my other comment, for instance) has to be applied twice and can drift.

A small private helper would collapse both:

```ts
private async loadIssueWorkspaceByKeys(
  projectId: string,
  keys: readonly string[]
): Promise<IssueWorkspaceRecord | null> {
  for (const key of keys) {
    const record = await this.store.loadIssueWorkspace(projectId, key);
    if (record) return record;
  }
  return null;
}
```

---

packages/core/src/contracts/status-surface.ts
```diff
@@ -363,6 +363,7 @@ export type ProjectStatusSnapshot = {
   activeRuns: Array<{
     runId: string;
     issueIdentifier: string;
+    issueWorkspaceKey?: string | null;
```
hojinzs: **nit** — `string | null` rather than `?: string | null`.

`OrchestratorRunRecord.issueWorkspaceKey` at line 240 in this same file is required-nullable, and `snapshot-builder.ts:92` always populates the field (`run.issueWorkspaceKey ?? null`). Optional here adds a third state that nothing produces, and every other field in this `activeRuns` shape is required. Making it required keeps consumers of the public snapshot contract from having to distinguish "absent" from "null".

---

packages/orchestrator/src/service.test.ts
```diff
@@ -8517,12 +8517,10 @@ Prefer focused changes.
...
-    const events = (
-      await readFile(
-        join(store.runDir("run-1", "tenant-1"), "events.ndjson"),
-        "utf8"
-      )
-    )
+    const events = (await readFile(
+      join(store.runDir("run-1", "tenant-1"), "events.ndjson"),
+      "utf8"
+    ))
```
hojinzs: **nit** — unrelated formatting churn, and it flip-flops.

This hunk reformats `(await readFile(...))` one way here, while the hunk at line ~8840 reformats the *identical* expression the opposite way. Neither is related to workspace keys, and one of the two will be reverted by the next `pnpm format:write`, which suggests a local Prettier version differing from CI's. Dropping both hunks keeps the diff to the change under review.

Same class of noise in `docs/configuration.md`: widening one cell in the env-var table reflowed all ~25 rows. Harmless, but it makes the doc change hard to read against #666's "워크스페이스 경로 설명 갱신, 마이그레이션 노트" — the actual new content is just the `Workspace-key migration` section.

---

hojinzs (changes requested):
## Review — round 1/6

Reviewed `3dc8d57` against #666. The core design is right and I'd expect this to land after one round: sanitize-then-suffix is the correct shape, the legacy lookup avoids a disk rename as decided on 2026-08-28, and both stated 완료 기준 are genuinely covered by tests rather than asserted in the description.

### 요구사항 반영 — 전부 반영됨

| #666 할 일 / 완료 기준 | 상태 |
| --- | --- |
| `deriveWorkspaceKey` 에 `-<16 hex sha256>` suffix (변경된 경우만) | ✅ `identity.ts:19-27`, 16 hex = 64-bit, spec §4.2 의 ≥64-bit 충족 |
| legacy 키 병행 조회, 디스크 rename 없음 | ✅ `service.ts:2673-2691`, `5119-5149` |
| `identity-preflight` 허용 집합 확장 | ✅ `identity-preflight.ts:80` |
| `repo status`/explain 에 `workspace key: legacy` | ⚠️ 반영되었으나 조건부로만 노출 — 인라인 P2/P3 참고 |
| `docs/configuration.md` 마이그레이션 노트 | ✅ `Workspace-key migration` 섹션 |
| 같은 텍스트로 sanitize 되는 두 identifier 가 다른 키 | ✅ `core-conformance.test.ts` — `a b` vs `a/b` |
| 변경 없는 identifier 는 기존 키 유지 | ✅ `My.Issue-1` → `My.Issue-1` |
| 기존 디렉터리 재사용 재시작 테스트 | ✅ `service.test.ts:13179` — `acme_platform_1` suffixless 재사용, `run.issueWorkspaceKey` 단언 추가됨 |

오버엔지니어링이나 요구사항 밖 임의 결정은 없습니다. `deriveLegacyIssueWorkspaceKey` 를 배열에 넣어 일반화한 것도 기존 2-key 조회를 그대로 옮긴 것이라 범위 확장이 아닙니다.

### 검증 (로컬, `3dc8d57`)

- `pnpm build` — pass
- `pnpm typecheck` — pass (전 패키지)
- `core-conformance` / `status` / `identity-preflight` — 27+44 tests pass
- 레거시 재사용 테스트 단독 실행 — pass
- CI `Test`, `Container Smoke` — green

빌드된 `@gh-symphony/core` 로 스모크한 키 파생 결과:

```
"acme/platform#42" -> new: acme_platform_42-cbb20472b0ece3db | legacy: acme_platform_42
"a b"              -> new: a_b-c8687a08aa5d6ed2              | legacy: a_b
"a/b"              -> new: a_b-c14cddc033f64b9d              | legacy: a_b
"!!!"              -> new: issue-e84c538e7fe25073            | legacy: issue
"???"              -> new: issue-a03b221c6c6eae71            | legacy: issue
"a__b"             -> new: a__b                              | legacy: a__b
```

새 키의 충돌 저항성은 확인했습니다. 아래 P2 는 전부 **legacy 경로** 쪽입니다.

### 지적사항

- **P2** `identity.ts:26` — `run.issueWorkspaceKey ?? deriveIssueWorkspaceKey(...)` fallback 이 이제 suffixed 키를 돌려주면서, `issueWorkspaceKey: null` 인 구 run record + suffixless 워크스페이스 조합에서 dirty-workspace 복구 3개 경로가 존재하지 않는 디렉터리를 보게 됩니다. 에러가 아니라 "clean" 으로 조용히 강등되는 형태라, 미커밋 작업이 있는 incomplete turn 이 그냥 재시도됩니다. 이번 PR 에서 유일하게 **동작 회귀**로 판단한 항목입니다.
- **P2** `service.ts:2684` — legacy 조회가 키 모양만 보고 레코드를 채택합니다. `deriveLegacyWorkspaceKey` 는 many-to-one 이고 (`issue` 버킷은 특히), `IssueWorkspaceRecord` 에 `issueSubjectId` 가 이미 있으니 한 줄로 막을 수 있습니다.
- **P2** `explain.ts:423` — 라벨이 읽는 `record.workspaceKey` 를 dispatch 루프가 `service.ts:1568`/`1662` 에서 preferred 키로 덮어씁니다. dispatch 가 `onPrepared` 전에 실패하면 라벨이 영구히 사라집니다.
- **P3** `status.ts:202` — 라벨이 active run 동안에만 노출됩니다.
- **P3** `service.ts:5140` — 같은 조회가 두 가지 형태(IIFE / `??=` 루프)로 존재합니다.
- **nit** `status-surface.ts:366`, `service.test.ts:8520` (prettier flip-flop), `docs/configuration.md` 테이블 리플로우.

### 판단

P2 세 건 중 첫 번째는 실제 회귀이고 나머지 둘은 #666 이 명시한 관측성 요구를 불완전하게 만드는 것이라, 이번 라운드는 request changes 로 둡니다. 세 건 모두 국소 수정이고 구조 변경은 필요 없습니다 — 다음 푸시에서 정리되면 approve 하겠습니다. P3/nit 은 근거 있는 미반영 의사를 스레드에 남겨주시면 그대로 수용합니다.

---

packages/orchestrator/src/service.ts
```diff
@@ -2669,21 +2670,23 @@ export class OrchestratorService {
...
-        preferredWorkspaceKey
-      )) ??
-      (legacyWorkspaceKey === preferredWorkspaceKey
-        ? null
-        : await this.store.loadIssueWorkspace(
-            tenant.projectId,
-            legacyWorkspaceKey
-          ));
+        legacyWorkspaceKey
+      );
```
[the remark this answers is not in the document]
moncher-dev: Resolved in 2b6444e. Legacy candidates are now accepted only when their persisted adapter and issue subject ID match the current issue; the same guard is used by terminal cleanup.

packages/orchestrator/src/service.ts
```diff
@@ -5131,12 +5137,16 @@ export class OrchestratorService {
...
-        ? null
-        : await this.store.loadIssueWorkspace(
+      (await (async () => {
+        for (const legacyWorkspaceKey of legacyWorkspaceKeys) {
+          const record = await this.store.loadIssueWorkspace(
             tenant.projectId,
             legacyWorkspaceKey
-          ));
+          );
+          if (record) return record;
```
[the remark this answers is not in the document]
moncher-dev: Resolved in 2b6444e. Legacy candidates are now accepted only when their persisted adapter and issue subject ID match the current issue; the same guard is used by terminal cleanup.

packages/orchestrator/src/service.ts
moncher-dev: Resolved in 2b6444e. Legacy candidates are now accepted only when their persisted adapter and issue subject ID match the current issue; the same guard is used by terminal cleanup.

packages/core/src/workspace/identity.ts
moncher-dev: Resolved in 2b6444e. Pre-key runs now resolve a matching persisted preferred/legacy workspace before inspecting dirty state, preserving dirty-workspace recovery for suffixless directories.

packages/orchestrator/src/explain.ts
moncher-dev: Resolved in 2b6444e. Dispatch preserves the selected legacy key through claimed and spawn-failure states, while `repo status` now exposes idle legacy workspaces from the snapshot. The lookup implementation is shared, and the active-run contract is required-nullable.

packages/cli/src/commands/status.ts
moncher-dev: Resolved in 2b6444e. Dispatch preserves the selected legacy key through claimed and spawn-failure states, while `repo status` now exposes idle legacy workspaces from the snapshot. The lookup implementation is shared, and the active-run contract is required-nullable.

packages/orchestrator/src/service.ts
moncher-dev: Resolved in 2b6444e. Dispatch preserves the selected legacy key through claimed and spawn-failure states, while `repo status` now exposes idle legacy workspaces from the snapshot. The lookup implementation is shared, and the active-run contract is required-nullable.

packages/core/src/contracts/status-surface.ts
moncher-dev: Resolved in 2b6444e. Dispatch preserves the selected legacy key through claimed and spawn-failure states, while `repo status` now exposes idle legacy workspaces from the snapshot. The lookup implementation is shared, and the active-run contract is required-nullable.

packages/orchestrator/src/service.test.ts
moncher-dev: Partially resolved in 2b6444e: the unrelated test-expression formatting is restored to its original forms. The configuration-table reflow remains in the earlier commit and will be narrowed before re-handoff.

packages/core/src/observability/snapshot-builder.ts
```diff
@@ -110,6 +111,11 @@ export function buildProjectSnapshot(
         cumulativeTokenUsageByIssue.get(run.issueId)
       ),
     })),
+    issueWorkspaces: (issueWorkspaces ?? []).map((workspace) => ({
```
hojinzs, line 114: **P3 — the new snapshot array is unfiltered on both status and project**

Two things this mapping does not do, that the only other consumer of the same input does at `findRecoveryWorkspaceRecord` (line 212):

1. **No `projectId` filter.** `fs-store.ts:377-393` builds `loadIssueWorkspaces(projectId)` from `projectDir(projectId)` *plus* `legacyProjectDir()`, and records read out of the legacy dir are not guaranteed to carry this project's `projectId` — which is exactly why line 212 filters before correlating. Here they land in the project's own `status.json`, so `repo status` can print another project's suffixless workspace under this tenant's `Legacy Workspaces:` heading.
2. **No status filter.** `cleanupTerminalIssueWorkspace` (`service.ts:5498`) ends by re-saving the record as `status: "removed"` rather than deleting it, so this array accumulates one entry per issue the project has ever processed, and the whole thing is re-serialized into `status.json` on every poll (`service.ts:1996`). `status.ts:214` then filters `status === "active"` at render, so every `removed` entry is pure payload.

Both close with one filter, and it matches the shape already used further down this file:

```ts
issueWorkspaces: (issueWorkspaces ?? [])
  .filter(
    (workspace) =>
      workspace.projectId === project.projectId &&
      workspace.status === "active"
  )
  .map((workspace) => ({ ... })),
```

Non-blocking — the render-side filter keeps the operator-visible output correct today for the single-project case.

---

packages/core/src/contracts/status-surface.ts
```diff
@@ -400,6 +401,11 @@ export type ProjectStatusSnapshot = {
       cumulativeTotalTokens?: number;
     };
   }>;
+  issueWorkspaces?: Array<{
```
hojinzs, line 404: **nit** — same shape as the `issueWorkspaceKey` nit you just resolved above, in the other direction.

`snapshot-builder.ts:114` always populates this (`(issueWorkspaces ?? []).map(...)` — never `undefined`), so `?:` adds a third state nothing produces, and `status.ts:214` has to write `snapshot.issueWorkspaces ?? []` to consume it. `issueWorkspaces: Array<...>` would be consistent with how you made `activeRuns[].issueWorkspaceKey` required-nullable four lines up.

Fine to leave if you'd rather not widen the public contract in this PR — just flagging the inconsistency.

---

packages/orchestrator/src/service.ts
```diff
@@ -4236,6 +4237,57 @@ export class OrchestratorService {
...
+        record.adapter === adapter &&
+        record.issueSubjectId === issueSubjectId
+      ) {
+        return record;
+      }
+    }
+    return null;
+  }
+
+  private async resolveRunWorkspaceKey(
```
hojinzs, line 4285: **Verified — the regression is closed.** I drove this method directly against a fixture (suffixless `workspace.json` on disk, run record with `issueWorkspaceKey: null`) on `ea612bc`:

```
preferred: acme_platform_42-cbb20472b0ece3db | legacy: acme_platform_42
resolved for pre-key run:      acme_platform_42                    ← legacy dir found
resolved for foreign subject:  acme_platform_42-cbb20472b0ece3db   ← identity guard rejects
```

So both my round-1 P2s are genuinely fixed, not just asserted: dirty-workspace recovery resolves the suffixless directory again, and `loadWorkspaceForIssue` refuses a record whose persisted `adapter`/`issueSubjectId` doesn't match. I also confirmed `adapter` and `issueSubjectId` have been required on `IssueWorkspaceRecord` for the whole life of the type, so the new guard can't reject a genuinely-owned legacy record for want of the fields.

**P3 — the fix has no test.** Round 1 asked for a test driving the pre-key path over a suffixless directory; the suite still has none. What was added (`service.test.ts:14145`) covers the *guard* by calling the private `loadWorkspaceForIssue` through a cast, which is the weaker half — it proves rejection, not adoption, and it never touches `classifyIncompleteTurnDirtyWorkspace` / `resolveIncompleteTurnRecoveryContext` / `resolveRetryRunRecoveryContext`, the three call sites the regression actually lived in. The fixture above is ~40 lines and would pin all three.

**nit** — eight `run.issueWorkspaceKey ?? deriveIssueWorkspaceKey(...)` sites remain (`3262`, `3610`, `3690`, `4682`, `4741`, `4834`, `5007`), and for a pre-key run they now write the *suffixed* key into `IssueOrchestrationRecord.workspaceKey` — the field `explain.ts:454` reads for the legacy label. Unlike round 1 this is no longer a path resolution, so nothing breaks: the label just drops until the next dispatch claim restores it from `selectedWorkspaceKey` at line 1561. `resolveRunWorkspaceKey` is already async and in scope at most of them if you want them consistent.

---

packages/orchestrator/src/explain.ts
```diff
@@ -448,12 +449,15 @@ function explainRuntimeOwnership(
...
```
hojinzs, line 454: **P3 — `repo explain` is the one #666 surface with no test on it.**

`grep -rn "Workspace key: legacy"` over the tree returns four production sites and exactly two assertions, both in `status.test.ts`. The `repo status` half of "`repo status`/explain에 `workspace key: legacy` 표시를 넣어 관측 가능하게 함" is well covered now — I rendered it and both the active-run and idle paths print correctly:

```
  Legacy Workspaces:
    acme/repo#1  Workspace key: legacy (acme_repo_1)
```

The explain half is only ever exercised transitively. Given that the label here is prefixed onto two different `message` strings and reads `record?.workspaceKey` — a field the dispatch loop rewrites, which is what made this a P2 last round — one `explain.test.ts` case asserting the prefix on a legacy record and its absence on a suffixed one would be worth having before the post-merge checklist item is signed off by hand.

**nit** (carried from round 1, no reply) — `legacyWorkspaceKey` holds a boolean. `isLegacyWorkspaceKey` at both call sites.

---

hojinzs (approved):

`ea612bc` 검토. 라운드 1의 P2 3건 전부 해소되었고, 문서상 주장이 아니라 코드를 직접 돌려서 확인했습니다. 남은 것은 P3/nit 뿐이라 여기서 approve 합니다 — 추가 라운드로 끌 만한 항목이 없습니다.

### 라운드 1 지적사항 처리

| 라운드 1 | 상태 |
| --- | --- |
| **P2** `identity.ts:26` — pre-key run 의 dirty-workspace 복구 회귀 | ✅ `resolveRunWorkspaceKey` 로 3개 경로 전부 교체, 픽스처로 재현 확인 |
| **P2** `service.ts:2684` — legacy 채택에 identity guard 없음 | ✅ `loadWorkspaceForIssue` 가 `adapter` + `issueSubjectId` 로 게이팅 |
| **P2** `explain.ts:423` — dispatch 루프가 라벨 필드를 덮어씀 | ✅ `selectedWorkspaceKey` 로 1561/1655 둘 다 보존 |
| **P3** `status.ts:202` — idle 워크스페이스 미노출 | ✅ 스냅샷 `issueWorkspaces` + `Legacy Workspaces:` 섹션 |
| **P3** `service.ts:5140` — 같은 조회의 두 가지 형태 | ✅ `loadWorkspaceForIssue` 로 단일화 |
| **nit** `status-surface.ts` — required-nullable | ✅ `issueWorkspaceKey: string \| null` |
| **nit** prettier flip-flop / 문서 테이블 리플로우 | ✅ 둘 다 원복, 문서 변경은 1행 + 신규 섹션으로 축소 |

미반영 nit 하나: `explain.ts:452` `legacyWorkspaceKey` boolean 네이밍 (인라인 참고).

### 검증 (로컬, `ea612bc`)

- `pnpm lint` / `pnpm typecheck` / `pnpm build` — pass
- `pnpm test` — 전 패키지 pass (orchestrator 358, cli 581, worker 176 …)
- CI `Test`, `Container Smoke` — green

**Smoke — legacy 채택 경로 직접 구동.** 픽스처(디스크상 suffixless `workspace.json`, `issueWorkspaceKey: null` 인 run 레코드)로 `resolveRunWorkspaceKey` 실행:

```
preferred: acme_platform_42-cbb20472b0ece3db | legacy: acme_platform_42
resolved for pre-key run:      acme_platform_42                    ← 레거시 디렉터리 채택
resolved for foreign subject:  acme_platform_42-cbb20472b0ece3db   ← identity guard 로 거부
```

라운드 1 의 회귀(존재하지 않는 경로를 보고 조용히 "clean" 으로 강등)와 충돌 채택이 같은 픽스처에서 동시에 닫히는 것을 확인했습니다. `IssueWorkspaceRecord` 의 `adapter`/`issueSubjectId` 는 타입 도입 이래 계속 required 였으므로, 새 guard 가 정당한 레거시 레코드를 필드 부재로 거부할 위험도 없습니다.

**Smoke — `repo status` 렌더.** idle 경로 실제 출력:

```
  Legacy Workspaces:
    acme/repo#1  Workspace key: legacy (acme_repo_1)
```

라벨 조건의 false-positive(이미 sanitize 된 `acme_repo_1` 형태 identifier)도 두 번째 절이 정상 억제합니다.

### 요구사항 반영

#666 의 할 일·완료 기준은 라운드 1에서 전부 확인했고 이번 푸시에서 회귀 없습니다. 오버엔지니어링이나 요구사항 밖 임의 결정 없음 — 신규 코드는 `loadWorkspaceForIssue` / `resolveRunWorkspaceKey` 두 헬퍼와 스냅샷 필드 하나가 전부이고, 셋 다 라운드 1 지적사항에 직접 대응합니다.

### 남은 항목 (전부 non-blocking)

- **P3** `snapshot-builder.ts:114` — 신규 `issueWorkspaces` 배열이 `projectId`·`status` 로 필터링되지 않음. 같은 파일 212행이 하는 필터를 그대로 적용하면 닫힙니다.
- **P3** 테스트 공백 두 건 — pre-key 복구 경로(위 픽스처로 재현 가능), `repo explain` 라벨(#666 명시 항목인데 어서션이 없음).
- **nit** `status-surface.ts:404` optional/required 불일치, `explain.ts:452` boolean 네이밍, 잔여 8개 orchestration upsert 의 라벨 일관성.
- **nit** `status.ts:219` 가 prettier printWidth 초과입니다. CI 는 `pnpm format` 을 돌리지 않고 main 도 현재 prettier-clean 이 아니라 실패하지 않으니, 다음에 이 파일을 만질 때 같이 정리하면 됩니다.

머지 후 체크리스트 2건은 그대로 유효합니다. P3 는 후속 이슈로 빼거나 근거 있는 미반영 의사를 스레드에 남겨주셔도 좋습니다 — 이 PR 을 막지 않습니다.

---

hojinzs:

head `ea612bc`(approve, CI green)를 격리 worktree에 빌드하고 file 트래커 + stub(repo 모드)으로 확인했습니다.

| 케이스 | 결과 |
| --- | --- |
| 새 dispatch: identifier `test-owner/test-repo#1`(sanitize 변경됨) → 워크스페이스 키 `test-owner_test-repo_1-484e36eec6eba867`(16-hex suffix), issues.json 동일 | ✅ |
| pre-upgrade 상태 에뮬레이션(레코드·디렉터리 모두 suffix 없음, identity 일치) → 기존 디렉터리 **재사용**, 새 디렉터리 생성 없음, disk rename 없음 | ✅ |
| `repo status` → `Workspace key: legacy (test-owner_test-repo_1)` 표시 | ✅ |
| identity 레코드가 없는 suffixless 디렉터리는 채택하지 않고 새 키로 생성(안전한 fallback, PR Risks 절과 일치) | ✅ (관찰) |

병합되면 A11(§4.2/§9.5 충돌 방지) 해소.

Issue #663, opened 2026-08-27
refactor(core,trackers,orchestrator): move blocker eligibility into adapter-derived dispatchable and retire blocker_check_states from core


moncher-dev:
Epic: #651
리포트 항목: **A2** — `docs/reports/2026-08-28-upstream-spec-drift-research.md` (PR #650)

## 문제

새 스펙 §8.2/§11.2: 오케스트레이터는 provider blocker 의미론으로 분기하면 안 된다(MUST NOT). 구 스펙의 "Todo + non-terminal blocker → skip" 규칙은 어댑터의 `dispatchable`로 흡수됐다. 현재 blocker 판정은 coordination 계층에 있다.

## 현재 상태 (증거)

- `packages/orchestrator/src/explain.ts:149-163, 595-635`; 호출 `service.ts:2390-2401, 3957, 4052`
- `blockerCheckStates`: `packages/core/src/workflow/lifecycle.ts:5, 13`; `docs/configuration.md:11-17`은 **구** 규칙 대비 divergence로 문서화
- GitHub는 blocker 상태를 `lifecycle.terminalStates[0]`/`null`로 재작성: `adapter.ts:2663-2681` (provider가 lifecycle 도메인 상태명을 발명)
- 테스트: `dispatch.test.ts:628-843`, `service.test.ts:8134`

## 할 일

- GitHub/Linear 어댑터가 blocker 관계를 평가해 `dispatchable=false` + 이유 반환; `blocked_by`는 best-effort 메타데이터로만 유지
- `blocker_check_states`: **결정됨 (2026-08-28) — 이동.** `tracker.provider.blocker_check_states`(GitHub/Linear 어댑터 소유)로 옮긴다. "Todo에서만 blocker 검사"는 실제 운영 요구(README:735)였고, 어댑터가 `dispatchable`을 파생할 때 참조하면 스펙 그대로다. #669 전에 착수하면 flat 키 alias로 읽고 #669에서 provider로 승격
- `explain.ts` blocker 로직 제거, `dispatchReason`으로 대체
- `docs/configuration.md:11-17` 갱신

## 완료 기준

- 오케스트레이터 코드에 blocker 관련 분기 없음 (grep)
- 기존 blocker dispatch 테스트를 어댑터 단위 테스트 + `dispatchable` 스케줄러 테스트로 이전

Blocked by: #661

moncher-dev:
## Workpad — hojinzs/github-symphony#663 — Cycle 1

[...]

Issue #689, opened 2026-08-27
docs+e2e: document dispatchable semantics and add a Docker E2E scenario for dispatchable=false


moncher-dev:
Parent: #661 (Epic #651)
리포트 항목: **A1, A18** — `docs/reports/2026-08-28-upstream-spec-drift-research.md`

#661 분할 마지막 조각.

## 범위

- `docs/architecture.md` §3/§5: 스케줄러는 `dispatchable`만 적용하고 provider 규칙은 어댑터가 파생한다는 경계 명시
- `README.md`의 `--assigned-only` 설명을 "어댑터가 `dispatchable` 파생 시 사용하는 입력"으로 갱신
- `AGENT_TEST.md` 시나리오 표 + `e2e/fixtures/`에 `dispatchable=false` 이슈가 dispatch 되지 않고 explain에 이유가 남는 Docker E2E 추가
- 어댑터 프로필(#674)에 들어갈 `dispatchable` 파생 규칙 초안을 각 패키지 README에 기록

## 완료 기준

- Docker E2E 통과, 문서 병합

Blocked by: #687, #688

moncher-dev:
## Workpad — hojinzs/github-symphony#689 — Cycle 1

**Type:** fresh pickup
**Branch:** feat/689-dispatchable-docs-e2e
**Draft PR:** https://github.com/hojinzs/github-symphony/pull/713
**Cycle opened:** 2026-08-28T04:29:12Z · **Trigger:** Ready issue is actionable; no linked delivery PR or review feedback.

### Status Transitions



- 2026-08-28T04:41:30Z · FAILED `In progress` → `In review` · expected-state mismatch; canonical issue remained `Ready`
- 2026-08-28T04:42:08Z · `Ready` → `In progress` · completed delivery requires the missed cycle-opening status readback (cycle 1 open)
- 2026-08-28T04:42:31Z · `In progress` → `In review` · Docker dispatchable=false regression and documentation are complete (cycle 1 close)

### Plan

- [x] 1. Document the adapter-owned `dispatchable` contract and add the Docker non-dispatchable black-box runner path
- [x] 2. Wrap-up: re-verify the original issue · pass the Completion Bar · changeset (if needed) · PR ready · transition to In review

### Rework / PR Feedback

Not applicable.

### Completion Bar

- [x] In-scope requirements implemented
- [x] `pnpm lint`
- [x] `pnpm test`
- [x] `pnpm typecheck`
- [x] `pnpm build`
[...]

PR #697, opened 2026-08-27
feat(tracker-github): derive dispatchable eligibility

### LINKED ISSUE #687: feat(tracker-github): derive dispatchable from assignment, board membership, and PR head repository; return unassigned issues
Parent: #661 (Epic #651)

#661 분할 두 번째 조각 — GitHub Project V2 어댑터.

## 현재 상태 (증거)

- 미할당 아이템 drop: `packages/tracker-github/src/adapter.ts:614-617`; assignees는 필터에만 사용 `:1904-1915`
- pickup label / assignedOnly를 어댑터가 자체 적용: `tracker-github/src/orchestrator-adapter.ts:19-30`

## 범위

- scope 내 active 이슈를 **모두 반환**하고, `--assigned-only`/할당 규칙, board membership, fork PR head repo 등 provider 규칙을 `dispatchable=false` + `dispatchReason`으로 표현
- `assigneeId` 채움 (첫 assignee login 또는 node ID — 프로필 문서에 명시)
- blocker 규칙은 여기서 옮기지 않음 (#663)
- `docs/architecture.md` §5 GitHub 줄 갱신

## 완료 기준

- 미할당 이슈가 `dispatchable=false`로 반환되고 dispatch 되지 않으며 explain에 이유가 보이는 테스트
- 기존 `--assigned-only` 동작이 동일하게 유지되는 테스트

Blocked by: #686

moncher-dev:
## TL;DR

- GitHub Project polling retains every active in-scope record and exposes provider-derived eligibility to `repo explain`.
- Rework restores invariant repository matching, prevents retained records from resolving foreign workflows, and preserves structured eligibility diagnostics.

## Change-point diagram

Project item → assignment/repository/PR-head/pickup-label policy → `dispatchable` + `dispatchReason` → explain-visible record → dispatch and mutation/workflow gates

## Start here

- `packages/tracker-github/src/adapter.ts`: assignment, repository scope, PR-head eligibility, and operator event derivation.
- `packages/core/src/workflow/pickup-labels.ts`: shared pickup-label reason resolution.
- `packages/orchestrator/src/service.ts`: prevents retained ineligible records from mutation and workflow-resolution paths.
- `packages/core/src/workflow/pickup-labels.test.ts` and tracker/orchestrator tests: regression coverage.

## Risks & rollback

- Eligibility is provider-specific; blocker policy remains deferred to #663.
- Fork PR Project items are now non-dispatchable rather than degrading project status; `repo explain` supplies the reason.
- Rollback: revert this PR to restore the prior GitHub Project filtering behavior.

## Changed files

- GitHub tracker adapter and orchestrator adapter
- Core pickup-label policy and regression test
- Orchestrator workflow-resolution gate
- Architecture/design documentation and CLI changeset

## Evidence

- `pnpm lint` — pass
- `pnpm test` — pass
- `pnpm typecheck` — pass
- `pnpm build` — pass
- Docker E2E: a `Ready` non-dispatchable fixture resulted in `health: idle`, `activeRuns: 0`, and no run event directory.

- [ ] Confirm `repo explain` reports the GitHub reason for an unassigned Project item under `--assigned-only`.
- [ ] Confirm project logs provide the configured eligibility context and reason breakdown.

### DIFF
--- a/packages/orchestrator/src/service.test.ts
+++ b/packages/orchestrator/src/service.test.ts
@@ -1827,9 +1827,23 @@ describe("OrchestratorService", () => {
       requestFails: false,
       targetIdentifier: "acme/platform#99",
     },
+    {
+      name: "non-dispatchable terminal candidate",
+      metadata: { sourceState: "CLOSED" },
+      terminalFact: "issue_closed",
+      requestFails: false,
+      targetIdentifier: null,
+      dispatchable: false,
+    },
   ])(
     "reconciles a $name to Done without dispatching",
-    async ({ metadata, terminalFact, requestFails, targetIdentifier }) => {
+    async ({
+      metadata,
+      terminalFact,
+      requestFails,
+      targetIdentifier,
+      dispatchable = true,
+    }) => {
       const tempRoot = await mkdtemp(
         join(tmpdir(), "orchestrator-terminal-candidate-")
       );
@@ -1852,7 +1866,7 @@ describe("OrchestratorService", () => {
         branchName: null,
         url: "https://github.com/acme/platform/issues/1",
         labels: [],
-        dispatchable: true,
+        dispatchable,
         assigneeId: null,
         blockedBy: [],
         createdAt: "2026-03-08T00:00:00.000Z",
@@ -1912,7 +1926,7 @@ describe("OrchestratorService", () => {
       expect(result.summary.dispatched).toBe(0);
       expect(result.summary.suppressed).toBe(0);
       expect(spawnImpl).not.toHaveBeenCalled();
-      if (targetIdentifier) {
+      if (targetIdentifier || !dispatchable) {
         expect(requestState).not.toHaveBeenCalled();
         expect(info).not.toHaveBeenCalledWith(
           expect.stringContaining(
@@ -1952,6 +1966,78 @@ describe("OrchestratorService", () => {
       }
     }
   );
+
+  it("does not publish active PR advisories for non-dispatchable items", async () => {
+    const tempRoot = await mkdtemp(
+      join(tmpdir(), "orchestrator-nondispatchable-advisory-")
+    );
+    const repository = await createRepositoryFixture(
+      tempRoot,
+      "acme",
+      "platform"
+    );
+    const store = new OrchestratorFsStore(tempRoot);
+    const projectConfig = createProjectConfig(tempRoot, repository);
+    await store.saveProjectConfig(projectConfig);
+    const issue = {
+      id: "issue-1",
+      identifier: "acme/platform#1",
+      number: 1,
+      title: "Out-of-scope issue",
+      description: null,
+      priority: null,
+      state: "Todo",
+      branchName: null,
+      url: "https://github.com/acme/platform/issues/1",
+      labels: [],
+      dispatchable: false,
+      dispatchReason: "repository does not match the configured scope",
+      assigneeId: null,
+      blockedBy: [],
+      createdAt: "2026-03-08T00:00:00.000Z",
+      updatedAt: "2026-03-08T00:00:00.000Z",
+      repository,
+      tracker: {
+        adapter: "github-project" as const,
+        bindingId: "project-123",
+        itemId: "item-1",
+      },
+      metadata: {
+        sourceState: "OPEN",
+        linkedPullRequests: [
+          {
+            id: "pr-2",
+            number: 2,
+            identifier: "acme/platform#2",
+            url: "https://github.com/acme/platform/pull/2",
+            state: "OPEN",
+            merged: false,
+          },
+        ],
+      },
+    };
+    const issues = Object.assign([issue], {
+      rateLimits: null,
+      skippedItems: [],
+    }) as TrackedIssueList;
+    const upsertIssueComment = vi.fn();
+    vi.spyOn(trackerAdapters, "resolveTrackerAdapter").mockReturnValue({
+      listIssues: vi.fn().mockResolvedValue(issues),
+      listIssuesByStates: vi.fn().mockResolvedValue([]),
+      fetchIssueStatesByIds: vi.fn().mockResolvedValue([]),
+      buildWorkerEnvironment: vi.fn(),
+      reviveIssue: vi.fn(),
+      upsertIssueComment,
+    });
+    const service = new OrchestratorService(store, projectConfig, {
+      spawnImpl: vi.fn() as never,
+      now: () => new Date("2026-03-08T00:00:00.000Z"),
+    });
+
+    await service.runOnce();
+
+    expect(upsertIssueComment).not.toHaveBeenCalled();
+  });
 
   it("passes worktree-cache settings into issue populate", async () => {
     process.env.GITHUB_GRAPHQL_TOKEN = "test-token";
@@ -7051,10 +7137,8 @@ Prefer focused changes.
 
     const result = await service.runOnce();
 
-    expect(result.health).toBe("degraded");
-    expect(result.lastError).toMatch(
-      /fork pull requests are unsupported for automatic checkout\/push \(contributor\/platform -> acme\/platform\)/
-    );
+    expect(result.health).toBe("idle");
+    expect(result.lastError).toBeNull();
     expect(spawnImpl).not.toHaveBeenCalled();
   });
 
@@ -7090,10 +7174,8 @@ Prefer focused changes.
 
     const result = await service.runOnce();
 
-    expect(result.health).toBe("degraded");
-    expect(result.lastError).toMatch(
-      /fork pull requests are unsupported for automatic checkout\/push \(unknown fork -> acme\/platform\)/
-    );
+    expect(result.health).toBe("idle");
+    expect(result.lastError).toBeNull();
     expect(spawnImpl).not.toHaveBeenCalled();
   });
 
--- a/packages/orchestrator/src/service.ts
+++ b/packages/orchestrator/src/service.ts
@@ -1306,11 +1306,12 @@ export class OrchestratorService {
         );
       }
       const canonicalIssues = resolveCanonicalSubjectIssues(issues);
-      const terminalCandidateIssues = issueIdentifier
+      const terminalCandidateIssues = (issueIdentifier
         ? canonicalIssues.filter((issue) =>
             matchesTargetIssueIdentifier(issue, issueIdentifier)
           )
-        : canonicalIssues;
+        : canonicalIssues
+      ).filter((issue) => issue.dispatchable);
       const terminalCandidateReconciliation =
         await this.reconcileTerminalCandidates(
           tenant,
@@ -1465,11 +1466,12 @@ export class OrchestratorService {
             matchesTargetIssueIdentifier(issue, issueIdentifier)
           )
         : trackedActionableIssues;
-      const targetedIssues = issueIdentifier
+      const targetedIssues = (issueIdentifier
         ? canonicalIssues.filter((issue: TrackedIssue) =>
             matchesTargetIssueIdentifier(issue, issueIdentifier)
           )
-        : canonicalIssues;
+        : canonicalIssues
+      ).filter((issue) => issue.dispatchable);
       const advisoryRateLimits =
         await this.publishLinkedPullRequestActiveAdvisories(
           tenant,
--- a/packages/orchestrator/src/service.ts
+++ b/packages/orchestrator/src/service.ts
@@ -362,6 +362,8 @@ function resolvePullRequestBranchCheckoutTarget(
     const source = headRepository
       ? `${headRepository.owner}/${headRepository.name}`
       : "unknown fork";
+    // PR-content items are excluded earlier by tracker dispatchability; this
+    // remains for issues linked to a fork PR.
     throw new NonRetryableDispatchError(
       `Cannot checkout pull request branch for ${pullRequest.identifier}: fork pull requests are unsupported for automatic checkout/push (${source} -> ${issue.repository.owner}/${issue.repository.name}).`
     );
@@ -1441,8 +1443,11 @@ export class OrchestratorService {
         lifecyclesByIssueIdentifier,
       } = await this.resolveActionableCandidates(
         tenant,
+        // Retained provider-ineligible records are kept for explain/status but
+        // must not load or cache workflow policy for another repository.
         canonicalIssues.filter(
           (issue) =>
+            issue.dispatchable &&
             !terminalCandidateReconciliation.suppressedIdentifiers.has(
               issue.identifier
             )
--- a/packages/tracker-github/src/adapter.ts
+++ b/packages/tracker-github/src/adapter.ts
@@ -581,7 +581,7 @@ export async function fetchProjectIssues(
   const currentUserLogin = config.assignedOnly
     ? await fetchCurrentUserLogin(config, fetchImpl)
     : null;
-  let nonDispatchableCount = 0;
+  const nonDispatchableByReason: Record<string, number> = {};
   let latestRateLimits: GitHubRateLimitPayload | null = null;
   const skippedItems: NonNullable<TrackedIssueList["skippedItems"]> = [];
 
@@ -635,7 +635,9 @@ export async function fetchProjectIssues(
         repositoryFilter: config.repositoryFilter,
       });
       if (!issue.dispatchable) {
-        nonDispatchableCount += 1;
+        const reason = getDispatchabilityReasonCategory(issue.dispatchReason);
+        nonDispatchableByReason[reason] =
+          (nonDispatchableByReason[reason] ?? 0) + 1;
       }
       return [issue];
     });
@@ -656,8 +658,10 @@ export async function fetchProjectIssues(
 
   emitDispatchabilityDerivationEvent({
     projectId: config.projectId,
+    currentUserLogin,
+    repository: config.repositoryFilter ?? null,
     includedCount: issues.length,
-    nonDispatchableCount,
+    nonDispatchableByReason,
   });
 
   latestRateLimits = finalizeGraphQLRateLimitCycle(
@@ -1935,7 +1939,7 @@ function getAssignedOnlyDispatchReason(
   }
 
   if (item.content.__typename !== "Issue") {
-    return `Pull request is not assigned to ${currentUserLogin}.`;
+    return "Pull request items are not dispatched under assigned-only mode.";
   }
 
   const isAssigned = (item.content.assignees?.nodes ?? []).some(
@@ -1953,12 +1957,9 @@ function getRepositoryDispatchReason(
   }
 
   const isInScope =
-    issue.repository.owner.localeCompare(repositoryFilter.owner, undefined, {
-      sensitivity: "accent",
-    }) === 0 &&
-    issue.repository.name.localeCompare(repositoryFilter.name, undefined, {
-      sensitivity: "accent",
-    }) === 0;
+    issue.repository.owner.toLowerCase() ===
+      repositoryFilter.owner.toLowerCase() &&
+    issue.repository.name.toLowerCase() === repositoryFilter.name.toLowerCase();
   return isInScope
     ? null
     : `Repository ${issue.repository.owner}/${issue.repository.name} is outside configured repository scope ${repositoryFilter.owner}/${repositoryFilter.name}.`;
@@ -1982,17 +1983,41 @@ function getPullRequestHeadDispatchReason(
     : `Fork pull requests are unsupported for automatic checkout/push (${headRepository ? `${headRepository.owner}/${headRepository.name}` : "unknown fork"} -> ${issue.repository.owner}/${issue.repository.name}).`;
 }
 
+function getDispatchabilityReasonCategory(reason: string | null | undefined): string {
+  if (reason?.startsWith("Issue is not assigned") || reason?.startsWith("Pull request items")) {
+    return "assignment";
+  }
+  if (reason?.startsWith("Repository ")) {
+    return "repository";
+  }
+  return "forkPullRequest";
+}
+
 function emitDispatchabilityDerivationEvent(input: {
   projectId: string;
+  currentUserLogin: string | null;
+  repository: { owner: string; name: string } | null;
   includedCount: number;
-  nonDispatchableCount: number;
+  nonDispatchableByReason: Record<string, number>;
 }): void {
+  if (!input.currentUserLogin && !input.repository) {
+    return;
+  }
+
   console.info(
     JSON.stringify({
       event: "tracker-dispatchability-derived",
       projectId: input.projectId,
+      ...(input.currentUserLogin
+        ? { currentUserLogin: input.currentUserLogin }
+        : {}),
+      ...(input.repository ? { repository: input.repository } : {}),
       includedCount: input.includedCount,
-      nonDispatchableCount: input.nonDispatchableCount,
+      nonDispatchableCount: Object.values(input.nonDispatchableByReason).reduce(
+        (count, value) => count + value,
+        0
+      ),
+      nonDispatchableByReason: input.nonDispatchableByReason,
     })
   );
 }
--- a/packages/tracker-github/src/adapter.ts
+++ b/packages/tracker-github/src/adapter.ts
@@ -581,8 +581,7 @@ export async function fetchProjectIssues(
   const currentUserLogin = config.assignedOnly
     ? await fetchCurrentUserLogin(config, fetchImpl)
     : null;
-  let excludedCount = 0;
-  let repositoryExcludedCount = 0;
+  let nonDispatchableCount = 0;
   let latestRateLimits: GitHubRateLimitPayload | null = null;
   const skippedItems: NonNullable<TrackedIssueList["skippedItems"]> = [];
 
@@ -604,22 +603,6 @@ export async function fetchProjectIssues(
       }
 
       if (!isRepositoryFilterableProjectItem(item)) {
-        if (config.repositoryFilter) {
-          repositoryExcludedCount += 1;
-        }
-        return [];
-      }
-
-      if (currentUserLogin && !isIssueAssignedToLogin(item, currentUserLogin)) {
-        excludedCount += 1;
-        return [];
-      }
-
-      if (
-        config.repositoryFilter &&
-        !isIssueInRepository(item, config.repositoryFilter)
-      ) {
-        repositoryExcludedCount += 1;
         return [];
       }
 
@@ -647,7 +630,14 @@ export async function fetchProjectIssues(
         return [];
       }
 
-      return [normalized];
+      const issue = applyDispatchabilityRules(normalized, item, {
+        currentUserLogin,
+        repositoryFilter: config.repositoryFilter,
+      });
+      if (!issue.dispatchable) {
+        nonDispatchableCount += 1;
+      }
+      return [issue];
     });
 
     issues.push(...pageIssues);
@@ -664,23 +654,11 @@ export async function fetchProjectIssues(
     });
   }
 
-  if (currentUserLogin) {
-    emitAssignedOnlyFilterEvent({
-      projectId: config.projectId,
-      currentUserLogin,
-      includedCount: issues.length,
-      excludedCount,
-    });
-  }
-
-  if (config.repositoryFilter) {
-    emitRepositoryFilterEvent({
-      projectId: config.projectId,
-      repository: `${config.repositoryFilter.owner}/${config.repositoryFilter.name}`,
-      includedCount: issues.length,
-      excludedCount: repositoryExcludedCount,
-    });
-  }
+  emitDispatchabilityDerivationEvent({
+    projectId: config.projectId,
+    includedCount: issues.length,
+    nonDispatchableCount,
+  });
 
   latestRateLimits = finalizeGraphQLRateLimitCycle(
     latestRateLimits,
@@ -1915,19 +1893,6 @@ async function fetchCurrentUserLogin(
   return payload.login;
 }
 
-function isIssueAssignedToLogin(
-  item: GraphQLProjectItem,
-  login: string
-): boolean {
-  if (item.content?.__typename !== "Issue") {
-    return false;
-  }
-
-  return (item.content.assignees?.nodes ?? []).some(
-    (assignee) => assignee?.login === login
-  );
-}
-
 function isRepositoryFilterableProjectItem(
   item: GraphQLProjectItem
 ): item is GraphQLProjectItem & {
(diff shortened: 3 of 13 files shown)


packages/tracker-github/src/adapter.ts
```diff
@@ -654,7 +637,14 @@ export async function fetchProjectIssues(
...
 
-      return [normalized];
+      const issue = applyDispatchabilityRules(normalized, item, {
+        currentUserLogin,
+        repositoryFilter: config.repositoryFilter,
+      });
+      if (!issue.dispatchable) {
+        nonDispatchableCount += 1;
+      }
+      return [issue];
```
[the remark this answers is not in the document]
moncher-dev: Resolved in c2272df. Retained non-dispatchable candidates remain available for explain, but are now filtered out before terminal reconciliation and linked-PR advisory mutations. Added regression coverage for both paths.

packages/tracker-github/src/adapter.ts
```diff
@@ -1931,51 +1896,95 @@ function isRepositoryFilterableProjectItem(
...
+    return null;
+  }
+
+  const isInScope =
+    issue.repository.owner.localeCompare(repositoryFilter.owner, undefined, {
+      sensitivity: "accent",
+    }) === 0 &&
+    issue.repository.name.localeCompare(repositoryFilter.name, undefined, {
+      sensitivity: "accent",
+    }) === 0;
```
hojinzs: **P1 — locale-dependent repository scope matching can silently stall dispatch**

The removed `isIssueInRepository` compared with `toLowerCase()`, which is locale-invariant. This replaces it with `localeCompare(..., undefined, { sensitivity: "accent" })`, which resolves against the **host's default locale**. On a daemon host with a Turkish locale the dotted/dotless `I` mapping differs, so a repository that is genuinely in scope compares unequal:

```js
const opts = { sensitivity: "accent" };
"Iris-Platform".localeCompare("iris-platform", "en-US", opts); // 0  → in scope
"Iris-Platform".localeCompare("iris-platform", "tr-TR", opts); // -1 → OUT of scope
"Iris-Platform".toLowerCase() === "iris-platform".toLowerCase(); // true (old behavior)
```

(Reproduced on Node 24 in this checkout.)

Failure scenario: operator runs the orchestrator on a host with `LANG=tr_TR.UTF-8`, scope `acme/iris-platform`, Project item `acme/Iris-Platform#12`. Before this PR the item was included and dispatched. After it, `getRepositoryDispatchReason` returns non-null, the item is marked `dispatchable: false`, and it is never dispatched — silently, since the only surface is `repo explain`. Previously this class of item at least disappeared from the list; now it looks present-but-stuck.

This is also inconsistent with `getPullRequestHeadDispatchReason` fifteen lines below, which still uses `toLowerCase()` for the same owner/name comparison.

GitHub repository names are ASCII-restricted, so locale-aware collation buys nothing here. Suggest restoring the invariant comparison:

```ts
const isInScope =
  issue.repository.owner.toLowerCase() === repositoryFilter.owner.toLowerCase() &&
  issue.repository.name.toLowerCase() === repositoryFilter.name.toLowerCase();
```

The existing `matches repository filters case-insensitively` test passes either way because it runs under `en-US`, so a regression test would need to pin the locale (or simply assert the invariant helper is used).

---

packages/orchestrator/src/service.ts
```diff
@@ -1397,11 +1398,12 @@ export class OrchestratorService {
...
           )
         : trackedActionableIssues;
-      const targetedIssues = issueIdentifier
+      const targetedIssues = (issueIdentifier
         ? canonicalIssues.filter((issue: TrackedIssue) =>
             matchesTargetIssueIdentifier(issue, issueIdentifier)
           )
-        : canonicalIssues;
+        : canonicalIssues
+      ).filter((issue) => issue.dispatchable);
```
hojinzs, line 1479: **P2 — the retained-item side effect is gated for tracker mutations but not for workflow-policy resolution**

`c2272df` correctly gates `reconcileTerminalCandidates` (L1278) and `publishLinkedPullRequestActiveAdvisories` (here). But `resolveActionableCandidates` at L1372 is still fed the unfiltered `canonicalIssues`, and inside its loop `resolveIssueLifecycle` runs **before** the eligibility check:

```
resolveActionableCandidates (2304)
  → resolveIssueLifecycle (2305)         ← runs for non-dispatchable items
    → loadProjectWorkflow(tenant, issue.repository)
  → isIssueCandidateEligible (2311)      ← only here does !dispatchable short-circuit
```

For an out-of-scope repository, `resolveWorkflowRepositoryDirectory` (4912) has no `repository.path` and a non-`file:` `cloneUrl`, so it falls back to `process.cwd()`. The daemon therefore reads **its own** `WORKFLOW.md`, treats it as the foreign repository's policy, and `resolveWorkflowResolution` (4835-4848) persists it as last-known-good under `cache/<foreignOwner>/<foreignName>/` and caches it in `this.lastKnownGoodWorkflows`.

Failure scenario: a shared Project spanning 20 repositories, watched by a daemon scoped to one of them. Every poll cycle now resolves and persists a workflow cache for all 19 foreign repositories, seeded from the local repo's `WORKFLOW.md`. Nothing dispatches (dispatch is gated), but the daemon writes policy cache directories for repositories outside its scope on a poll cadence, and `lastKnownGoodWorkflows` grows with the Project's repository count rather than the daemon's scope — the same "out-of-scope daemon does work it has no business doing" concern the P1 raised, one layer up.

Smallest fix consistent with the two gates already added — filter before the call at L1372:

```ts
canonicalIssues.filter(
  (issue) =>
    issue.dispatchable &&
    !terminalCandidateReconciliation.suppressedIdentifiers.has(issue.identifier)
)
```

If `lifecyclesByIssueIdentifier` must stay populated for the explain path, the alternative is to hoist the `!issue.dispatchable` check above line 2305 inside `resolveActionableCandidates`.

---

packages/tracker-github/src/adapter.ts
```diff
@@ -1931,51 +1896,95 @@ function isRepositoryFilterableProjectItem(
...
   console.info(
     JSON.stringify({
-      event: "tracker-repository-filtered",
+      event: "tracker-dispatchability-derived",
       projectId: input.projectId,
-      repository: input.repository,
       includedCount: input.includedCount,
-      excludedCount: input.excludedCount,
+      nonDispatchableCount: input.nonDispatchableCount,
     })
```
hojinzs, line 2021: **P2 — Observability-layer regression not captured anywhere**

Two structured events are removed for the GitHub tracker and replaced by one with strictly less information:

| removed | payload |
|---|---|
| `tracker-assigned-only-filtered` | `projectId`, `currentUserLogin`, `includedCount`, `excludedCount` |
| `tracker-repository-filtered` | `projectId`, `repository`, `includedCount`, `excludedCount` |

`tracker-dispatchability-derived` keeps only `projectId`, `includedCount`, `nonDispatchableCount` — so `currentUserLogin`, the scoped `repository`, and the per-cause breakdown are all gone. An operator who sees `nonDispatchableCount: 14` can no longer tell whether that is an assignment mismatch, a scope mismatch, a pickup label, or a fork PR without running `repo explain` per issue, which is exactly the triage the counters existed for.

Three follow-ons:

1. **Cross-tracker divergence.** `packages/tracker-linear/src/orchestrator-adapter.ts:786` still emits `tracker-assigned-only-filtered`. After this PR the two trackers emit different event names for the same operator-visible concern.
2. **Stale doc.** `docs/designs/2026-07-06-github-project-repo-dispatch-filter-design.md:97` still specifies `tracker-repository-filtered` and its payload. This PR updates `docs/architecture.md` but not that design note.
3. **Changeset.** `.changeset/green-items-dispatch.md` describes the eligibility feature but not the event rename. Anything parsing these log lines breaks on upgrade with no release-note signal.

Cheapest resolution that keeps the feature: carry the cause breakdown on the new event, e.g. `nonDispatchableByReason: { assignment: n, repository: n, pickupLabel: n, forkPullRequest: n }` plus `currentUserLogin` / `repository` when configured — and add a line to the changeset.

**nit (same block):** this is now emitted unconditionally. Previously the emit was guarded by `if (currentUserLogin)` / `if (config.repositoryFilter)`, so a project with no eligibility rules configured logged nothing. It now writes one line per poll forever with `nonDispatchableCount: 0`. Consider keeping the emit conditional on at least one rule being active.

---

packages/tracker-github/src/adapter.ts
```diff
@@ -1931,51 +1896,95 @@ function isRepositoryFilterableProjectItem(
...
+  },
+  currentUserLogin: string | null
+): string | null {
+  if (!currentUserLogin) {
+    return null;
+  }
+
+  if (item.content.__typename !== "Issue") {
+    return `Pull request is not assigned to ${currentUserLogin}.`;
+  }
```
hojinzs, line 1943: **P3 — this reason asserts a fact the code never checks**

The dispatch decision is unchanged from `isIssueAssignedToLogin` (which also returned `false` for non-`Issue` content), so no behavior regression here. But the *reason string* is new, and it is now operator-facing through `repo explain`.

The branch is taken for every PR item purely because `__typename !== "Issue"` — PR assignees are never read. A pull request that genuinely **is** assigned to the machine user will be explained as "Pull request is not assigned to machine-user.", which is false and sends the operator looking for a missing assignment that isn't missing.

Suggest stating the actual rule:

```ts
return `Pull request items are not dispatched under assigned-only mode.`;
```

---

packages/tracker-github/src/orchestrator-adapter.ts
```diff
@@ -165,6 +156,50 @@ export async function findGithubProjectIssue(
...
+  result.rateLimits = issues.rateLimits;
+  result.skippedItems = issues.skippedItems;
+  return result;
+}
+
+function readPickupLabels(value: unknown): string[] {
+  return Array.isArray(value)
+    ? value.filter((label): label is string => typeof label === "string")
+    : [];
+}
```
hojinzs: **P3 — third copy of pickup-label semantics; drift risk across trackers**

I checked this against `packages/core/src/workflow/pickup-labels.ts` line by line — the settings lookup, the `!object || Array.isArray` guard, the `readLabelList`/`readPickupLabels` string filter, the empty-include-and-exclude short circuit, and the exclude-then-include precedence are all identical. So the semantics are correct today; this is purely about where they live.

The repo now carries three implementations of the same rule: core (used by `tracker-file`), a private copy in `tracker-linear/src/orchestrator-adapter.ts:747`, and this one. Per CLAUDE.md's "keep workflow-policy behavior separate from orchestration-core behavior", pickup labels are Configuration/Policy, not GitHub Integration — the label semantics themselves should not be forking into the tracker packages.

The concrete drift risk: the next change to core's matching (case-insensitive labels, glob support, an `include` precedence tweak) silently applies to `tracker-file` and not to GitHub. Nothing in the test suite would catch it, because each package tests its own copy.

Since the filter→annotate shape is what forced the copy, the reusable piece is the *reason*, not the filter. Suggest exporting from core alongside the existing function:

```ts
export function resolvePickupLabelDispatchReason<T extends TrackedIssue>(
  issue: T,
  project: Pick<OrchestratorProjectConfig, "tracker">
): string | null
```

then this adapter becomes a `map` over that, and `filterIssuesByPickupLabels` can be expressed in terms of it so the two can't diverge. That also gives `tracker-linear` a landing spot when it migrates.

**nit:** L182 is over the print width and only survives because it's inside a file Prettier already formats — worth reflowing:
```ts
const included =
  include.length === 0 || include.some((label) => labels.has(label));
```

---

packages/orchestrator/src/service.test.ts
```diff
@@ -6849,10 +6935,8 @@ Prefer focused changes.
 
     const result = await service.runOnce();
 
-    expect(result.health).toBe("degraded");
-    expect(result.lastError).toMatch(
-      /fork pull requests are unsupported for automatic checkout\/push \(contributor\/platform -> acme\/platform\)/
-    );
+    expect(result.health).toBe("idle");
```
hojinzs, line 7140: **P3 — flipping these two assertions removes the only passive operator signal for fork PRs**

Both fork-PR cases previously surfaced as `health: "degraded"` with a `lastError` naming the fork. They now assert `"idle"` / `null`, so a fork PR sitting in the Project is indistinguishable from an empty queue on `repo status` — the reason exists only if someone runs `repo explain` on that specific issue.

Two notes rather than a request to revert, since the silencing follows from what #687 asked for:

1. Worth an explicit line in the changeset. "Expose … with explainable reasons" reads as additive; "fork pull requests no longer mark the project degraded" is the part an operator upgrading would want to know.
2. The equivalent guard at `service.ts:350` (`NonRetryableDispatchError`, same message text) is now only reachable for **issues with a linked fork PR** — `getPullRequestHeadDispatchReason` keys off `issue.metadata.pullRequest`, i.e. PR-content items only. So fork handling is now split across two layers with two different operator outcomes (silent non-dispatchable vs. degraded). That split is defensible but not obvious; a comment at one of the two sites pointing at the other would save the next reader the trace I just did.

---

hojinzs (changes requested):
## Review — round 2 of 6 · `c2272df`

Verified against issue #687. **Requirements are met and the mechanism works end-to-end**; the blocking item is one regression introduced incidentally, not a gap in the feature.

### Requirements vs. implementation (#687)

| Requirement | Status |
|---|---|
| Return **all** active in-scope items, express provider rules as `dispatchable=false` + `dispatchReason` | ✅ assignment, repository scope, pickup label, fork PR head |
| Fill `assigneeId` (first assignee login), document in adapter profile | ✅ `adapter.ts:487`, documented in `packages/tracker-github/README.md` |
| Blocker rules **not** moved here (deferred to #663) | ✅ untouched |
| Update `docs/architecture.md` §5 GitHub line | ✅ |
| Test: unassigned issue returned non-dispatchable, not dispatched, reason visible in explain | ⚠️ split across two suites, wiring itself untested — see below |
| Test: existing `--assigned-only` behavior preserved | ✅ for dispatch; ⚠️ not for the emitted events (P2) |

No over-engineering found. Every derived rule traces to a line in the issue, and blocker semantics were correctly left alone.

### Acceptance criterion verified manually

The PR body defers *"Confirm `repo explain` reports the GitHub reason for an unassigned Project item under `--assigned-only`"* to post-merge human validation — but that is issue #687's stated 완료 기준, not a post-merge item. I wired the real GitHub adapter to `explainIssueDispatch` and ran it:

```
SMOKE summary: Not dispatchable: Issue is not assigned to machine-user.
SMOKE check:   {"id":"tracker_dispatchability","status":"block",
                "message":"not dispatchable: Issue is not assigned to machine-user.",
                "details":{"dispatchReason":"Issue is not assigned to machine-user."}}
```

**It works** — the GitHub-derived reason reaches the explain surface intact. Please land that as a test in `packages/orchestrator` rather than a checkbox; today `tracker-github.test.ts` proves derivation and `dispatch.test.ts:301-322` proves explain rendering from a *synthetic* reason, but nothing proves the two are connected.

### Previous round

Codex's P1 (retained ineligible items reaching mutation passes) is **genuinely fixed** in `c2272df` — `terminalCandidateIssues` and `targetedIssues` are both gated, with regression coverage for each, and the inline reply is accurate. I did find that the same retention still reaches workflow-policy resolution one layer up (P2 below), which that fix didn't cover.

### Findings

| | Finding | File |
|---|---|---|
| **P1** | Locale-dependent repository scope matching can silently stall dispatch | `tracker-github/src/adapter.ts` |
| **P2** | Retained items still drive workflow resolution + last-known-good cache writes for out-of-scope repos | `orchestrator/src/service.ts` |
| **P2** | Two structured events removed, replaced by a lossier one; `tracker-linear` diverges; design doc + changeset stale | `tracker-github/src/adapter.ts` |
| **P3** | `"Pull request is not assigned to X"` asserts a fact never checked | `tracker-github/src/adapter.ts` |
| **P3** | Third copy of pickup-label semantics in the repo | `tracker-github/src/orchestrator-adapter.ts` |
| **P3** | Fork PRs lose the `degraded` signal; fork handling now split across two layers | `orchestrator/src/service.test.ts` |
| nit | Derivation event now emitted every poll even with no rules configured | `tracker-github/src/adapter.ts` |

The P1 is a two-line revert to `toLowerCase()` — `localeCompare` with the host's default locale replaces a locale-invariant comparison, and under `LANG=tr_TR` an in-scope repository is judged out of scope and never dispatched. Reproduced on Node 24; details inline.

### Validation performed

`pnpm install` · `pnpm build` · `pnpm test` (all packages) · `pnpm lint` · `pnpm typecheck` — all pass on `c2272df`. CI Test + Container Smoke both green. Smoke test as above.

### Verdict

Requesting changes on the **P1 only**. The two P2s are worth fixing in the same round since both are small and both are consequences of this PR's retention model; the P3s are yours to take or decline with a reason. Once the locale comparison is restored this is an approve from my side — the design is sound and the requirements are fully covered.

---

packages/tracker-github/src/adapter.ts
moncher-dev: Resolved in f24205d by restoring locale-invariant `toLowerCase()` repository comparisons; existing case-insensitive coverage protects the intended matching contract.

packages/orchestrator/src/service.ts
moncher-dev: Resolved in f24205d: non-dispatchable retained records are filtered before lifecycle resolution, so foreign workflow policy is neither loaded nor cached.

packages/tracker-github/src/adapter.ts
moncher-dev: Resolved in f24205d: the derived event now retains configured assignment/scope context plus a non-dispatchable reason breakdown, and is suppressed when neither rule is active. The design note and changeset are updated.

packages/tracker-github/src/adapter.ts
moncher-dev: Resolved in f24205d: the explain reason now states the actual rule: PR items are not dispatched under assigned-only mode.

packages/tracker-github/src/orchestrator-adapter.ts
moncher-dev: Resolved in f24205d by exporting the pickup-label dispatch reason from core and using it for both filtering and retained GitHub-item annotation; a core regression test covers exclude-first semantics.

packages/orchestrator/src/service.test.ts
moncher-dev: Addressed in f24205d: the changeset explicitly documents non-dispatchable fork PR Project items, and the linked-fork dispatch guard now documents its distinct issue-linked path.

packages/tracker-github/src/adapter.ts
```diff
@@ -1931,51 +1900,116 @@ function isRepositoryFilterableProjectItem(
...
+    headRepository != null &&
+    headRepository.owner.toLowerCase() === issue.repository.owner.toLowerCase() &&
+    headRepository.name.toLowerCase() === issue.repository.name.toLowerCase();
+  return sameRepository
+    ? null
+    : `Fork pull requests are unsupported for automatic checkout/push (${headRepository ? `${headRepository.owner}/${headRepository.name}` : "unknown fork"} -> ${issue.repository.owner}/${issue.repository.name}).`;
 }
 
-function emitRepositoryFilterEvent(input: {
+function getDispatchabilityReasonCategory(reason: string | null | undefined): string {
```
hojinzs, line 1986: **P3 — the reason breakdown is derived by prefix-matching operator-facing English copy, with a silent catch-all**

This gives me exactly the triage data I asked for, and I verified it works (smoke output in the review body). The concern is only how the category is recovered.

`applyDispatchabilityRules` (L1903) already *knows* which of the three rules fired — it calls them in order and takes the first non-null. That information is then thrown away and reconstructed here by matching the prefix of a human-readable string. Two consequences:

1. **The log schema is now coupled to message copy.** Reword `"Repository X is outside configured repository scope Y."` to `"Out of scope: repository X..."` and the counter silently reclassifies every scope exclusion as `forkPullRequest`. Nothing fails — not a test, not a type — the dashboard just starts lying.
2. **The fallback is a catch-all, not a default.** Any reason that isn't recognised (a fourth rule added later, or `dispatchReason` left null while `dispatchable` is false) lands in `forkPullRequest`. That's the one bucket an operator is least likely to question, because fork PRs are genuinely expected to be rare.

Neither is reachable today — I checked all three current reasons classify correctly — so this is a latent-fragility note, not a bug report. The cheap version is to have the rule functions return the category alongside the reason:

```ts
const RULES = [
  { category: "assignment", resolve: () => getAssignedOnlyDispatchReason(item, options.currentUserLogin) },
  { category: "repository", resolve: () => getRepositoryDispatchReason(issue, options.repositoryFilter) },
  { category: "forkPullRequest", resolve: () => getPullRequestHeadDispatchReason(issue) },
] as const;
```

then `applyDispatchabilityRules` returns `{ issue, category }` and this function disappears. Your call — take it or decline with a reason; it isn't blocking.

**nit (L1978, L1979):** both lines exceed the print width. `adapter.ts` is already Prettier-dirty on `main` and CI doesn't run `pnpm format`, so this changes nothing today — just noting it so the file doesn't drift further.

---

packages/tracker-github/src/adapter.ts
```diff
@@ -1931,51 +1900,116 @@ function isRepositoryFilterableProjectItem(
...
+function emitDispatchabilityDerivationEvent(input: {
   projectId: string;
-  repository: string;
+  currentUserLogin: string | null;
+  repository: { owner: string; name: string } | null;
   includedCount: number;
-  excludedCount: number;
+  nonDispatchableByReason: Record<string, number>;
 }): void {
+  if (!input.currentUserLogin && !input.repository) {
```
hojinzs, line 2003: **nit — the guard suppresses the event for the one rule that is never configured**

This is the conditional emit I asked for, and for the common case it's right: `repositoryFilter` defaults to the project's own repository, so in practice one of the two is almost always set and the event fires. I confirmed that — an unconfigured project still logs, because `repository` is populated from `project.repository`.

The gap is the documented opt-out. With `tracker.settings.repository: "*"` (covered by *"allows tracker repository '\*' to opt out of repository dispatch scoping"*) and no `assignedOnly`, both inputs are null and this returns early — but `getPullRequestHeadDispatchReason` (L1958) is **not** gated on any config, so fork-PR eligibility still runs and still marks items non-dispatchable.

Reproduced on this head with a fork-PR Project item and `settings: { projectId, repository: "*" }`:

```
issue:  { dispatchable: false,
          reason: "Fork pull requests are unsupported for automatic checkout/push
                   (contributor/platform-fork -> acme/platform)." }
dispatchability events emitted: 0
```

So on a `"*"` project, fork PRs become silently non-dispatchable with no passive signal at all — which is the same "reason exists only if someone runs `repo explain` on that exact issue" gap I flagged for fork PRs last round, just narrowed to one config. Since the fork rule is always active, the guard is more accurately "did any rule *find* anything":

```ts
if (!input.currentUserLogin && !input.repository && input.nonDispatchableCount === 0) {
  return;
}
```

That keeps the quiet-when-nothing-configured property you were after — an unconfigured project with no fork PRs still logs nothing per poll — while never dropping a derivation that actually happened. Non-blocking.

---

hojinzs (approved):
## Review — round 3 of 6 · `f24205d` · ✅ Approve

All seven findings from round 2 (one Codex P1, my P1 + 2×P2 + 3×P3) are **verified fixed in the code**, not just claimed. Requirements for #687 are fully covered. The two comments left inline are P3/nit and explicitly non-blocking — take them or decline with a reason; neither should hold the merge.

### Round-2 findings — verification

| | Finding | Verified |
|---|---|---|
| **P1** | Locale-dependent repository scope matching | ✅ `toLowerCase()` restored at `adapter.ts:1952-1955`; now consistent with `getPullRequestHeadDispatchReason` fifteen lines below, which is what made the divergence suspicious in the first place |
| **P2** | Retained items drove workflow resolution + foreign last-known-good cache writes | ✅ `issue.dispatchable &&` added at `service.ts:1379` with a comment naming the reason. I traced the consequence: `resolveActionableCandidates` now receives a pool with non-dispatchable items removed, which also narrows the blocker-resolution pool passed to `unresolvedBlockers` — that path treats a *missing* blocker as **unresolved**, so the change can only make an issue more blocked, never less. It also restores exactly the pool `main` had, since `main` dropped these items at the adapter. Safe in both directions |
| **P2** | Two structured events replaced by a lossier one; linear diverges; doc + changeset stale | ✅ `currentUserLogin`, `repository`, and `nonDispatchableByReason` all restored on the event; design note `:97` rewritten; changeset now names the event rename **and** the fork-PR status change. `tracker-linear` keeps its private copy but core now exports the shared reason, which is the landing spot it needed |
| **P3** | `"Pull request is not assigned to X"` asserted an unchecked fact | ✅ now `"Pull request items are not dispatched under assigned-only mode."` — states the rule the code actually applies |
| **P3** | Third copy of pickup-label semantics | ✅ genuinely deduplicated, not just moved: `filterIssuesByPickupLabels` is now *expressed in terms of* `resolvePickupLabelDispatchReason`, so filter and annotation cannot drift. `pickup-labels.test.ts` asserts both against the same fixture. I checked the one behavioral edge — the function no longer returns the input array by identity when unconfigured, which would have dropped `rateLimits`/`skippedItems` off a `TrackedIssueList`; the only core caller is `tracker-file`, which passes a plain array, and the GitHub path copies both properties explicitly at `orchestrator-adapter.ts:170-171`. No leak |
| **P3** | Fork PRs lose the `degraded` signal | ✅ changeset documents it; `service.ts:349` comment now points at the split |
| **P1** (Codex) | Retained ineligible items reaching mutation passes | ✅ still fixed; the three gates (`terminalCandidateIssues`, `targetedIssues`, `resolveActionableCandidates`) are now consistent |

### Requirements vs. #687

| Requirement | Status |
|---|---|
| Return **all** active in-scope items; provider rules as `dispatchable=false` + `dispatchReason` | ✅ assignment, repository scope, pickup label, fork PR head |
| `assigneeId` filled, documented in adapter profile | ✅ |
| Blocker rules **not** moved here (deferred to #663) | ✅ untouched |
| `docs/architecture.md` §5 GitHub line | ✅ |
| Test: unassigned issue non-dispatchable, not dispatched, reason visible in explain | ⚠️ covered in three pieces, wiring still untested — see below |
| Test: existing `--assigned-only` behavior preserved | ✅ |

Still no over-engineering. Every rule traces to a line in the issue; blocker semantics were left alone across all three rounds.

### Smoke test — acceptance criterion, re-run on this head

Wired the real GitHub adapter (`resolveTrackerAdapter` → `listIssues`) to `explainIssueDispatch` with all four eligibility rules active:

```
acme/platform#1  dispatchable: true   tracker_dispatchability: pass
acme/platform#2  block  "not dispatchable: Issue is not assigned to machine-user."
acme/web#3       block  "not dispatchable: Repository acme/web is outside configured
                         repository scope acme/platform."
acme/platform#4  block  "not dispatchable: Issue is missing a required pickup label (symphony)."

{"event":"tracker-dispatchability-derived","projectId":"project-123",
 "currentUserLogin":"machine-user","repository":{"owner":"acme","name":"platform"},
 "includedCount":4,"nonDispatchableCount":2,
 "nonDispatchableByReason":{"assignment":1,"repository":1}}
```

All four GitHub-derived reasons reach the explain surface intact, and the event carries the triage context it lost last round. **Both open post-merge checkboxes in the PR body are satisfied by this run** — worth ticking them rather than carrying them past merge.

One observation from the output rather than a request: `nonDispatchableByReason` counts 2 while `listIssues` returns 3 non-dispatchable items. Pickup labels are applied a layer later in `orchestrator-adapter.ts`, after the event is emitted in `fetchProjectIssues`, so the breakdown under-reports what the adapter actually hands back. Defensible given where the event lives — just don't let anyone read `nonDispatchableCount` as "how many items the tracker suppressed".

### The one thing I asked for that didn't land

Round 2 asked for the explain wiring to land as a test in `packages/orchestrator` instead of a post-merge checkbox, since it's #687's stated 완료 기준. It isn't in `f24205d` and the thread wasn't answered on that point. The three legs each have coverage — derivation in `tracker-github.test.ts`, non-dispatch in `service.test.ts`, explain rendering from a synthetic reason in `dispatch.test.ts:301-322` — so the criterion is met *in substance*, and my smoke proves the seam holds today. I'm **not** blocking on it: I said last round that restoring the locale comparison made this an approve, and re-litigating a test placement after every other finding was addressed is exactly the ping-pong this process is meant to avoid. Please pick it up as a follow-up so the seam has a guard.

### CI — red, and not this PR's

The **Test** check fails on `f24205d`, but not on anything in this diff:

```
Test Files  121 passed (121)
      Tests  1733 passed (1733)
...
Run pnpm e2e:compose-project-test
```

Every test passes; the job then dies on a workflow step. Root cause is on `main`, not here: **#693 (`17f158e`) added the `Compose project helper test` step to `.github/workflows/ci.yml` but never added the `e2e:compose-project-test` script to the root `package.json`** — it added `e2e/lib/compose-project.test.sh` and wired CI to a script name that doesn't exist. `main`'s own script list confirms the gap. This PR touches neither file, and its branch point (`de8c001`) predates that commit; the step reaches this run because `pull_request` workflows execute the merged `ci.yml`.

So: **not this PR's failure, and not this PR's to fix.** It will red every PR against `main` until someone adds the script (or points the step at `bash e2e/lib/compose-project.test.sh`). Flagging it as its own issue — merging #697 shouldn't wait on it, but it does mean this PR can't show green through no fault of its own.

### Validation performed

`pnpm install` · `pnpm build` · `pnpm lint` · `pnpm typecheck` — all pass on `f24205d`. `npx vitest run` — **121 files, 1715 tests, all passing** locally (CI counts 1733 with coverage instrumentation). Smoke tests as above, both scenarios. `pnpm format` fails on 83 files including `adapter.ts`, but it fails identically on `de8c001` and CI doesn't run it — pre-existing, not introduced here.

### Verdict

**Approve.** Three rounds in, every substantive finding is fixed, the design is sound, the requirements are fully covered, and what remains is two non-blocking notes and a test-placement follow-up. Merge once the base-branch CI script is sorted — nothing in this PR needs to change first.

---

moncher-dev:

head `f24205d`(approve)를 격리 worktree에 빌드하고 **실제 GitHub**로 검증: 임시 Project V2(Status Todo)에 미할당 이슈 #2와 `moncher-dev` 할당 이슈 #3을 넣고 PR 빌드의 `fetchProjectIssues({ assignedOnly, lifecycle: { stateFieldName: "Status", … } })`를 직접 호출.

| 호출 | 결과 |
| --- | --- |
| `assignedOnly: true` | 2건 **모두 반환**; #2 `dispatchable: false, dispatchReason: "Issue is not assigned to moncher-dev."`, #3 `dispatchable: true` |
| `assignedOnly: false` | 2건 모두 `dispatchable: true` |
| 이벤트 | `tracker-dispatchability-derived { includedCount: 2, nonDispatchableCount: 1, nonDispatchableByReason: { assignment: 1 } }` |

스펙 §11.1(후보 폴링에 `dispatchable=false` 포함)·§11.3(명시적 플래그) 충족. CI red는 main의 `e2e:compose-project-test` 결함(#702)뿐이며 vitest 121/121 통과 — #702 병합 후 재실행하면 green. 임시 Project는 삭제, 임시 저장소는 archive(토큰에 delete_repo 없음).

moncher-dev:

head `f24205d`(hojinzs approve)를 독립 clone에 빌드하고 **실제 GitHub Project V2**로 검증했습니다: 임시 저장소 `moncher-dev/symphony-smoke-687` + Project #4(Status: Todo/In Progress/Done), 이슈 #1(moncher-dev 할당)·#2(미할당) 둘 다 Status=Todo.

| 검증 | 결과 |
| --- | --- |
| `repo start --once --assigned-only`(stub worker): `tracker-dispatchability-derived` 이벤트 `includedCount:2, nonDispatchableCount:1, nonDispatchableByReason:{assignment:1}, currentUserLogin:moncher-dev` — 미할당 이슈가 **제거되지 않고 반환**되며 `dispatchable=false`로 표시 | ✅ |
| run 레코드 1개(#1만), #2 미dispatch | ✅ |
| assigned-only 아님: `repo explain` #1/#2 모두 `Tracker marks this issue as dispatchable` | ✅ |

비차단 관찰: `repo explain`은 `--assigned-only` 런타임 플래그를 알 수 없어 미할당 이슈에도 "dispatchable"로 답합니다(WORKFLOW.md `tracker.assignedOnly`도 반영 안 됨). 운영 중 "왜 dispatch 안 되나"를 explain으로 볼 때 혼동 소지가 있으니 #689(문서/E2E) 또는 별도 후속에서 explain에 assigned-only 컨텍스트를 전달하길 권합니다.

CI `Test` 잡 FAILURE는 vitest 121/121 통과 후 `pnpm e2e:compose-project-test` 미정의(main 인프라, #702)로 실패한 것이므로 #702 병합 후 rerun하면 green입니다.

PR #699, opened 2026-08-28
feat(trackers): derive dispatch eligibility locally

### LINKED ISSUE #688: feat(tracker-linear,tracker-file): derive dispatchable instead of server-side assignee filtering; return unassigned issues
Parent: #661 (Epic #651)

#661 분할 세 번째 조각 — Linear 및 file 어댑터.

## 현재 상태 (증거)

- Linear: `assignee.isMe`를 GraphQL 필터로 밀어넣어 미할당 이슈가 아예 조회되지 않음: `packages/tracker-linear/src/orchestrator-adapter.ts:375`; pickup label 자체 적용 `:189-197`
- file: pickup label 자체 적용 `packages/tracker-file/src/file-tracker-adapter.ts:120-122`; fixture에 `dispatchable` 필드 없음

## 범위

- Linear: assignee 필터를 쿼리에서 제거하고 응답의 `assignee.id`로 `assigneeId`/`dispatchable` 파생; `--assigned-only`는 파생 입력으로 유지
- file: fixture의 `dispatchable`(기본 true)·`assigneeId` 필드 지원 — E2E에서 `dispatchable=false` 시나리오를 만들 수 있게
- Linear 요청 수 변화(미할당 포함 조회) 검토 및 프로필 문서 메모

## 완료 기준

- Linear 미할당 이슈가 `dispatchable=false`로 반환되는 테스트; file fixture 테스트

Blocked by: #686

moncher-dev:
## TL;DR

- Return every matching Linear issue, including unassigned items, and derive dispatch eligibility locally.
- Fail closed when `--assigned-only` cannot resolve the authenticated Linear viewer.
- Preserve file-fixture `dispatchable`/`assigneeId` modeling and candidate pickup-label filtering.

## Change-point diagram

```text
Linear issue query + viewer id / file fixture
  → normalize assigneeId + dispatchable
  → candidate pickup-label filtering
  → orchestration dispatches only dispatchable issues
```

## Start here

- `packages/tracker-linear/src/orchestrator-adapter.ts` removes server-side `assignee.isMe`, derives viewer-scoped eligibility, and emits derivation-specific observability.
- `packages/tracker-linear/src/tracker-linear.test.ts` covers assigned, unassigned, other-user, and unavailable-viewer responses.
- `packages/tracker-file/src/file-tracker-adapter.test.ts` covers fixture-supplied `dispatchable` and `assigneeId`; implementation support is already on `main`.

## Risks & rollback

- A partial Linear response without a viewer id now fails loudly under `--assigned-only`, rather than silently dispatching or suppressing candidates.
- `viewer { id }` remains part of each paginated request but adds no round trip; the first non-empty id is retained for the fetch.
- Revert the #688 commits to restore the previous server-side filtering behavior.

## Changed files

- Linear adapter and regression tests; file fixture regression tests; request-profile documentation; Docker E2E fixture/test guide; and a minor CLI changeset.

## Evidence

- `pnpm --filter @gh-symphony/tracker-linear test -- tracker-linear.test.ts` — 29 passed.
- `pnpm lint`, `pnpm test`, `pnpm typecheck`, and `pnpm build` — passed.
- Docker E2E with `non-dispatchable.json` — `/api/v1/state` was healthy idle with zero active runs and zero retries after refresh.

- [ ] Confirm an unassigned Linear issue is visible in candidate diagnostics but is not dispatched when `--assigned-only` is enabled.
- [ ] Confirm a missing Linear viewer response reports a clear assigned-only configuration error.
- [ ] Confirm configured pickup labels remain enforced for Linear and file candidate polling.


packages/tracker-linear/src/orchestrator-adapter.ts
```diff
@@ -424,7 +413,7 @@ export function normalizeLinearIssue(
     labels: (issue.labels?.nodes ?? [])
       .map((label) => label.name)
       .filter((label): label is string => typeof label === "string"),
-    dispatchable: true,
+    dispatchable: !assignedOnly || Boolean(issue.assignee?.id),
```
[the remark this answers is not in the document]
moncher-dev: Resolved in 1c423ad0. The Linear query now reads `viewer.id` in the same request and compares it with each returned `assignee.id`; only the authenticated viewer's issues are dispatchable under `--assigned-only`. Unassigned and other-user issues remain visible with `dispatchable: false`, covered by the regression test.

packages/tracker-linear/src/orchestrator-adapter.ts
```diff
@@ -194,9 +188,7 @@ export const linearTrackerAdapter: OrchestratorTrackerAdapter = {
     return listLinearIssues(
       project,
       project.tracker.settings?.activeStates,
-      dependencies,
-      undefined,
-      { applyPickupLabels: true }
+      dependencies
     );
```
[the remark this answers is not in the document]
moncher-dev: Resolved in 1c423ad0. Restored `filterIssuesByPickupLabels` on the candidate-listing paths for both Linear and file adapters. State and recovery lookups remain deliberately unfiltered. Added/restored regression expectations for both adapters.

packages/tracker-linear/src/orchestrator-adapter.ts
```diff
@@ -424,7 +443,7 @@ export function normalizeLinearIssue(
     labels: (issue.labels?.nodes ?? [])
       .map((label) => label.name)
       .filter((label): label is string => typeof label === "string"),
-    dispatchable: true,
+    dispatchable: !assignedOnly || issue.assignee?.id === viewerId,
```
hojinzs: **P1 — `null === null` makes an unassigned issue dispatchable when the viewer id is unresolved**

`viewerId` defaults to `null`, and `LinearIssueNode` declares `assignee?: { id?: string | null } | null` (L64) — so an `assignee` object carrying a null `id` is a modeled state. When both are null this comparison is `null === null` → `true`, and the issue is dispatched under `--assigned-only` with `assigneeId: null`. That is a fail-open on the exact guard the flag exists for.

Reproduced against this head (`08390af`), driving `listIssues` with `assignedOnly: true`:

| response | expected | actual |
| --- | --- | --- |
| `assignee: {id:"user-1"}`, no `viewer` | dispatchable | `dispatchable: false` |
| `assignee: {id:null}`, no `viewer` | **not** dispatchable | **`dispatchable: true`** |
| `assignee: {id:null}`, `viewer: {id:"user-1"}` | not dispatchable | `dispatchable: false` ✅ |

Row 2 is the fail-open. Row 1 is the mirror-image problem and worth fixing in the same change: when `viewer` is absent, *every* issue becomes non-dispatchable, so orchestration silently stops picking up all work — and the event emitted at L730 reports `includedCount: 0`, which reads as "no assigned work" rather than "we could not identify the viewer". The previous server-side `assignee.isMe` filter could not fail either way, so both are new failure modes introduced here.

Both rows need `viewer` to be missing from the payload. That is not hypothetical: a GraphQL partial error can resolve `issues` while nulling `viewer`, and would plausibly null `assignee.id` in the same response.

Suggested fix — resolve the viewer strictly and compare only non-empty strings:

```ts
// in listLinearIssues, before normalizing
if (config.assignedOnly && !result.viewerId) {
  throw new Error(
    "Linear assignedOnly is enabled but the authenticated viewer id could not be resolved; refusing to derive dispatch eligibility."
  );
}
```

```ts
// normalizeLinearIssue
const assigneeId = issue.assignee?.id ?? null;
// ...
dispatchable:
  !assignedOnly || (assigneeId !== null && assigneeId === viewerId),
assigneeId,
```

Failing closed with a loud error is better than either silent outcome: an operator can act on a thrown error, but not on an empty candidate list. Please add regression tests for both rows above.

---

packages/tracker-linear/src/orchestrator-adapter.ts
```diff
@@ -400,7 +417,9 @@ export function normalizeLinearIssue(
   project: OrchestratorProjectConfig,
   projectSlug: string,
   issue: LinearIssueNode,
-  rateLimits: Record<string, unknown> | null = null
+  assignedOnly = false,
```
hojinzs: **P3 — positional-parameter break on an exported function, with a silent failure mode**

`normalizeLinearIssue` is re-exported from `src/index.ts`, so this is package public API. `assignedOnly` was inserted at position 4, which is where `rateLimits` used to sit:

```ts
// before: (project, projectSlug, issue, rateLimits)
// after:  (project, projectSlug, issue, assignedOnly, rateLimits, viewerId)
```

Any caller still passing `rateLimits` fourth now passes an object as `assignedOnly`. That is truthy, so every issue evaluates `issue.assignee?.id === viewerId` against `viewerId = null` and comes back non-dispatchable — no type error, no runtime error, just silently zero dispatch. In-repo callers are all updated, so this is not currently broken; the concern is the shape of the mistake it invites.

Six positional params with three optional trailing ones is also past the point where an options object reads better:

```ts
export function normalizeLinearIssue(
  project: OrchestratorProjectConfig,
  projectSlug: string,
  issue: LinearIssueNode,
  options: {
    assignedOnly?: boolean;
    rateLimits?: Record<string, unknown> | null;
    viewerId?: string | null;
  } = {}
): TrackedIssue
```

This makes the existing 3-arg test call sites keep working unchanged, and turns the misordering above into a compile error.

---

packages/tracker-linear/src/orchestrator-adapter.ts
```diff
@@ -736,77 +721,17 @@ function readStringArray(value: unknown): string[] | undefined {
...
 }): void {
   console.info(
     JSON.stringify({
-      event: "tracker-pickup-label-filtered",
+      event: "tracker-assigned-only-filtered",
       tracker: "linear",
       projectSlug: input.projectSlug,
-      include: input.include,
-      exclude: input.exclude,
+      assigneeFilter: "assignee.id",
```
hojinzs: **P3 — observability contract drift (Layer 6)**

Two breaking changes to an emitted event in one commit:

1. `assigneeFilter` changed `"isMe"` → `"assignee.id"`. Any consumer keying on the old value stops matching.
2. `tracker-assigned-only-filtered` is now a misnomer. Nothing is filtered any more — all issues are returned and `includedCount`/`excludedCount` describe a *derived flag*, not a filtered set. A dashboard reading `excludedCount` as "issues withheld from the candidate list" now reports the wrong thing, since those issues *are* in the list.

`excludedCount` also flipped from a literal `null` to a real number, which is the one strictly-better part of the change.

Consider renaming to something like `tracker-dispatchable-derived` with a `dispatchableCount`/`nonDispatchableCount` pair, so the event name matches what the code now does. At minimum, note the field-value change in the changeset — it is user-visible for anyone consuming structured events.

---

packages/tracker-linear/src/orchestrator-adapter.ts
```diff
@@ -8,6 +8,7 @@ import {
   type TrackedIssue,
   type TrackedIssueList,
 } from "@gh-symphony/core";
+import { filterIssuesByPickupLabels } from "@gh-symphony/core";
```
hojinzs: nit: second import from `@gh-symphony/core` — fold `filterIssuesByPickupLabels` into the existing import block at L1-11.

---

packages/tracker-linear/src/orchestrator-adapter.ts
```diff
@@ -354,6 +371,7 @@ async function fetchPaginatedLinearIssues(
       after,
     });
     latestRateLimits = response.rateLimits ?? latestRateLimits;
+    viewerId = response.data.viewer?.id ?? viewerId;
```
hojinzs: nit: `viewer { id }` is selected on all three queries and so is re-resolved on every page of a paginated fetch, though only one value is ever used. The PR description is right that no extra round trip is added, but on a 10-page poll this is 10 redundant `viewer` resolutions.

`viewerId ??= response.data.viewer?.id ?? null` would at least make "first page wins" explicit. Not worth a second request to avoid; just noting the selection is per-page rather than per-fetch.

---

hojinzs (changes requested):
## Review — round 2

Scope is tight and matches #688. One P1 blocks: the new local eligibility derivation can fail open.

### Requirements coverage vs #688

| --- | --- |
| Linear: `assignee.isMe` removed from the GraphQL filter | ✅ |
| Linear: `assigneeId`/`dispatchable` derived from response `assignee.id` | ⚠️ derived, but see P1 |
| `--assigned-only` retained as a derivation input | ✅ |
| Test: unassigned Linear issue returns `dispatchable=false` | ✅ `tracker-linear.test.ts:283` |
| file: fixture `dispatchable` (default true) / `assigneeId` support | ✅ (already on `main`; see note) |
| file fixture test | ✅ |
| Linear request-count note in profile docs | ✅ `packages/tracker-linear/README.md` |
| Changeset | ✅ |

No over-engineering or unrequested scope found. The `rateLimits` `defineProperty` re-attach in `listIssues` is a little awkward but is doing necessary work — the non-enumerable property does not survive `Array.prototype.filter`.

### Smoke test

Ran against this head (`08390af`) locally, since there is no preview environment for this change.

Full chain, fixture → adapter → scheduler gate, using the new `e2e/fixtures/non-dispatchable.json`:

```
issues:       [{"id":"test-owner/test-repo#1","state":"Ready","dispatchable":false}]
gate verdict: {"eligible":false,"reason":"not_dispatchable"}
```

The flag this PR derives is genuinely consumed — `isIssueCandidateEligibleWithReason` (`packages/orchestrator/src/explain.ts:161`) rejects it before dispatch. `pickupLabels` filtering is applied on the candidate path of all three adapters (github/linear/file) and correctly left off state and recovery lookups.

Also confirmed the PR's evidence claims: `pnpm lint`, `pnpm typecheck`, `pnpm test` (1730 tests), and `pnpm build` all pass locally.

Driving `listIssues` directly with `assignedOnly: true` is where it breaks — details in the inline comment on L446.

### CI

The red **Test** check is **not this PR's**. All 1730 unit tests pass; the job then fails on:

```
```

`.github/workflows/ci.yml:48` invokes `pnpm e2e:compose-project-test`, but no such script exists in `package.json` — on this branch **or on `main`**. This PR touches neither file. It is pre-existing base-branch breakage and needs a separate fix; the empty `chore(ci): refresh PR checks` commit will not clear it.

### Prior review threads

Both Codex P1s were answered in `1c423ad0`:

- **pickup-label filtering** — fully resolved, verified above.
- **authenticated-user assignee check** — resolved in the happy path, but the replacement introduces the two new edge-case failures in my L446 comment. The original concern (dispatching work assigned to someone else) is fixed; the null-handling around it is not.

### Findings

| Sev | Location | Issue |
| --- | --- | --- |
| **P1** | `orchestrator-adapter.ts:446` | `null === null` marks an unassigned issue dispatchable when the viewer id is unresolved; missing `viewer` also silently halts all dispatch |
| P3 | `orchestrator-adapter.ts:420` | `normalizeLinearIssue` positional-param break with a silent failure mode |
| P3 | `orchestrator-adapter.ts:734` | `tracker-assigned-only-filtered` event name and `assigneeFilter` value drift |
| nit | `orchestrator-adapter.ts:11` | duplicate `@gh-symphony/core` import |
| nit | `orchestrator-adapter.ts:374` | `viewer { id }` re-resolved per page |

### Note on the description

"Start here" says `packages/tracker-file/src/file-tracker-adapter.ts` supports fixture `dispatchable` and `assigneeId` — but this PR does not change that file. That support already landed on `main`, and only `file-tracker-adapter.test.ts` changed here. Not a code defect; worth correcting so the diff and the description agree.

---

Fix the P1 and this is good to go. The P3s and nits are yours to take or decline with a reason.

---

packages/tracker-linear/src/orchestrator-adapter.ts
moncher-dev: Resolved in 5dbf387. When `assignedOnly` is active, the adapter now requires a non-empty authenticated viewer id before normalizing candidates; it throws a clear error if unavailable. It also normalizes missing/empty assignee ids to `null` and requires a non-null assignee match, preventing the `null === null` fail-open. Added regression coverage for a missing viewer response.

packages/tracker-linear/src/orchestrator-adapter.ts
moncher-dev: Resolved in 5dbf387. `normalizeLinearIssue` now accepts a trailing options object (`assignedOnly`, `viewerId`, and `rateLimits`), preserving all existing three-argument calls and preventing positional optional-argument mistakes.

packages/tracker-linear/src/orchestrator-adapter.ts
moncher-dev: Resolved in 5dbf387. The event now reports local derivation explicitly as `tracker-dispatchable-derived`, with `assignmentScope: "viewer"`, `dispatchableCount`, and `nonDispatchableCount`; it no longer claims candidates were filtered.

packages/tracker-linear/src/orchestrator-adapter.ts
moncher-dev: Resolved in 5dbf387 by folding `filterIssuesByPickupLabels` into the existing `@gh-symphony/core` import block.

packages/tracker-linear/src/orchestrator-adapter.ts
moncher-dev: Resolved in 5dbf387. Pagination now preserves the first non-empty viewer id with `??=`; the selected field remains part of the existing request, so this adds no request round trip.

packages/tracker-linear/src/orchestrator-adapter.ts
```diff
@@ -288,36 +303,46 @@ async function listLinearIssues(
...
   if (config.assignedOnly) {
-    emitAssignedOnlyFilterEvent({
+    emitDispatchableDerivedEvent({
       projectSlug: config.projectSlug,
-      includedCount: filteredIssues.length,
+      dispatchableCount: fetchedIssues.filter((issue) => issue.dispatchable)
+        .length,
+      nonDispatchableCount: fetchedIssues.filter(
+        (issue) => !issue.dispatchable
+      ).length,
```
hojinzs, line 341: **P3 — the counts are computed over the pre-pickup-label set, so `dispatchableCount` overstates actual candidates**

This emits from inside `listLinearIssues`, over `fetchedIssues`. But `listIssues` (L203-215) applies `filterIssuesByPickupLabels` *after* this function returns, so the numbers here describe the raw fetch, not the candidate list the scheduler sees.

The old code did not have this gap: `filterIssuesByPickupLabels` used to run inside `listLinearIssues`, and `includedCount` was taken from `filteredIssues`. Moving the filter out to the adapter method (correct, and what the Codex P1 asked for) left the emit behind.

Reproduced on this head, with `pickupLabels: { include: ["agent"], exclude: ["no-agent"] }` and two viewer-assigned issues, one of each label:

```
event:    {"event":"tracker-dispatchable-derived",...,"dispatchableCount":2,"nonDispatchableCount":0}
returned: ["ENG-1"]
```

Two reported as dispatchable, one actually dispatchable. Any operator reading this event to answer "how much work is the poller picking up" gets the wrong number whenever pickup labels are configured — and pickup labels are exactly the feature that makes two projects on one repo disjoint, so that is the case where the count matters most.

Emitting from `listIssues` after the filter would fix it, at the cost of losing the event on the state/revive paths — which arguably should not be reporting candidate counts anyway.

Minor, same lines: `fetchedIssues.filter(...)` runs twice over the same array; a single reduce or a `partition` would do.

---

.changeset/linear-dispatchable-candidates.md
```diff
@@ -0,0 +1,5 @@
+---
+"@gh-symphony/cli": minor
+---
+
+Return unassigned Linear candidates with locally derived dispatch eligibility and allow file fixtures to model non-dispatchable issues. (#688)
```
hojinzs, line 5: **P3 — the structured-event contract change is still not mentioned here**

Carrying this over from round 2, where the ask was "at minimum, note the field-value change in the changeset." The rename itself was done properly (thanks — `tracker-dispatchable-derived` reads far better than the old name), but the changeset still only describes the candidate-listing behavior. Two user-visible event changes ship silently:

1. `tracker-assigned-only-filtered` is gone for Linear, replaced by `tracker-dispatchable-derived` with a different field set (`assignmentScope` / `dispatchableCount` / `nonDispatchableCount` instead of `assigneeFilter` / `includedCount` / `excludedCount`).
2. `tracker-pickup-label-filtered` is gone for Linear entirely, with no replacement — the shared `filterIssuesByPickupLabels` helper does not emit. The tests encode this as `expect(infoSpy).not.toHaveBeenCalled()` (`tracker-linear.test.ts:554`, `:595`, `:643`).

Consumers of either event break with no error and no log line — they just stop matching. That is worth a sentence, e.g.:

```
Linear structured events: `tracker-assigned-only-filtered` is replaced by
`tracker-dispatchable-derived` (new field set), and `tracker-pickup-label-filtered`
is no longer emitted.
```

Non-blocking, but this is Layer 6 and the changeset is the only place a consumer would find out.

Worth noting for the #661 follow-up rather than this PR: `tracker-github/src/adapter.ts:1965` still emits `tracker-assigned-only-filtered` with the old shape, so the two adapters now disagree on this event. That is the GitHub slice's problem, not yours.

---

packages/tracker-linear/src/orchestrator-adapter.ts
```diff
@@ -288,36 +303,46 @@ async function listLinearIssues(
...
 
   const fetchedIssues = result.nodes.map((node) =>
-    normalizeLinearIssue(project, config.projectSlug, node, result.rateLimits)
+    normalizeLinearIssue(
+      project,
+      config.projectSlug,
+      node,
+      {
+        assignedOnly: config.assignedOnly,
+        rateLimits: result.rateLimits,
```
hojinzs, line 322: nit: this file does not pass `pnpm format` on this head. Two hunks, both introduced by this PR:

```diff
-    normalizeLinearIssue(
-      project,
-      config.projectSlug,
-      node,
-      {
-        assignedOnly: config.assignedOnly,
-        rateLimits: result.rateLimits,
-        viewerId: result.viewerId,
-      }
-    )
+    normalizeLinearIssue(project, config.projectSlug, node, {
+      assignedOnly: config.assignedOnly,
+      rateLimits: result.rateLimits,
+      viewerId: result.viewerId,
+    })
```

and at L339:

```diff
-      nonDispatchableCount: fetchedIssues.filter(
-        (issue) => !issue.dispatchable
-      ).length,
+      nonDispatchableCount: fetchedIssues.filter((issue) => !issue.dispatchable)
+        .length,
```

`pnpm format:write` fixes both. (The repo is not prettier-clean overall — 81 files fail — so CI does not catch this; but these two are yours, and CLAUDE.md lists `pnpm format` in the pre-ship checks.)

---

packages/tracker-file/src/file-tracker-adapter.test.ts
```diff
@@ -223,9 +246,9 @@ describe("fileTrackerAdapter", () => {
         pickupLabels: { include: ["alpha"] },
       };
 
-      await expect(fileTrackerAdapter.listIssues(project)).resolves.toEqual([
-        expect.objectContaining({ id: "alpha" }),
-      ]);
+      await expect(
+        fileTrackerAdapter.listIssues(project)
+      ).resolves.toHaveLength(1);
```
hojinzs, line 251: nit: this weakens the assertion rather than adapting it.

```diff
-      await expect(fileTrackerAdapter.listIssues(project)).resolves.toEqual([
-        expect.objectContaining({ id: "alpha" }),
-      ]);
+      await expect(
+        fileTrackerAdapter.listIssues(project)
+      ).resolves.toHaveLength(1);
```

The test is about `pickupLabels: { include: ["alpha"] }`, so *which* issue survives is the thing under test — and `toHaveLength(1)` no longer checks it. The test would still pass if the filter kept the wrong issue.

I assume `toEqual` started failing because the normalized issue grew fields. `toMatchObject` keeps the identity check without pinning the full shape:

```ts
await expect(fileTrackerAdapter.listIssues(project)).resolves.toMatchObject([
  { id: "alpha" },
]);
```

which is the same form used in the new test at L104-124.

---

hojinzs (approved):
## Review — round 3 · Approve

The round-2 P1 is fixed, and I verified the fix rather than taking the reply at its word. Everything left is P3/nit and yours to take or decline.

### Round-2 P1: fixed — re-ran the exact table

Driving `listIssues` with `assignedOnly: true` against this head (`5dbf387`):

| response | round 2 (`08390af`) | this head |
| --- | --- | --- |
| `assignee:{id:"user-1"}`, no `viewer` | `dispatchable: false` (silent halt) | **throws** ✅ |
| `assignee:{id:null}`, no `viewer` | **`dispatchable: true`** (fail-open) | **throws** ✅ |
| `assignee:{id:null}`, `viewer:{id:"user-1"}` | `false` ✅ | `false` ✅ |
| `assignee:{id:"user-1"}`, `viewer:{id:"user-1"}` | `true` ✅ | `true` ✅ |

Both failure modes now fail closed with the same actionable error. I also probed four cases beyond what was asked, all correct:

- `viewer: {id: ""}` → throws (`normalizeLinearUserId` rejects the empty string, so it can't match an empty `assignee.id`)
- `viewer: null` explicit → throws
- `assignedOnly` off + no `viewer` → no throw, everything dispatchable (the guard is correctly scoped to the flag)
- `viewer` null on page 1, present on page 2 → picked up (`??=` first-non-empty-wins works across pages)

End-to-end through the scheduler gate, unchanged from round 2:

```
ENG-1 (viewer)      → {"eligible":true,"reason":null}
ENG-2 (unassigned)  → {"eligible":false,"reason":"not_dispatchable"}
ENG-3 (other user)  → {"eligible":false,"reason":"not_dispatchable"}
```

One thing I checked because I was the one who proposed throwing: the throw is contained. The poll loop catches it at `service.ts:1991` into `lastError`/`trackerError`, so it surfaces as a project error in the status snapshot instead of crashing the orchestrator. Startup cleanup (`:2107`) and `classifyCurrentTrackerProgress` (`:4152`) also catch it. Fail-closed and loud, as intended.

### Other round-2 items

| Item | Status |
| --- | --- |
| P3 `normalizeLinearIssue` positional break | ✅ options object; 3-arg call sites unchanged |
| P3 observability contract drift | ✅ renamed to `tracker-dispatchable-derived` with `assignmentScope`/`dispatchableCount`/`nonDispatchableCount` — but see the changeset comment |
| nit duplicate core import | ✅ folded |
| nit per-page `viewer` re-resolution | ✅ `??=`, verified across pages |
| Description said `file-tracker-adapter.ts` was changed | ✅ corrected to name the test file |

### Requirements vs #688

All acceptance criteria met: `assignee.isMe` gone from the filter, `assigneeId`/`dispatchable` derived from the response, `--assigned-only` retained as a derivation input, unassigned-issue regression test, file fixture test, request-count note in the README, changeset present. No over-engineering or unrequested scope — the diff is 7 files and every one traces to the issue.

### Verification

`pnpm lint`, `pnpm typecheck`, `pnpm test` (1742 tests, 122 files), `pnpm build` — all pass locally on this head, confirming the PR's evidence claims.

### Behavior change worth a line in "Risks"

Removing the filter from `buildLinearIssueFilter` necessarily removes it from all three queries, so `fetchIssueStatesByIds` now derives `dispatchable` too — confirmed, it returns `dispatchable: false` for an other-user issue. That feeds `classifyCurrentTrackerProgress` (`service.ts:4146`), so under `--assigned-only` an issue reassigned mid-run changes from `unknown` → deferral loop to `non-actionable` → run completed.

This is consistent with the core contract ("Whether this issue may enter the orchestration dispatch lifecycle") and I think it's the better behavior. Not a defect — just not mentioned anywhere, and it changes finalization, so it belongs in the PR's Risks section.

### CI

The red **Test** check is **still not this PR's**, and unchanged from round 2. All 1742 unit tests pass; the job then fails on:

```
```

`.github/workflows/ci.yml:48` invokes a script that does not exist in `package.json`, on this branch or on `main`. This PR touches neither file. It needs a separate fix and should not hold this PR.

### Findings — all non-blocking

| Sev | Location | Issue |
| --- | --- | --- |
| P3 | `orchestrator-adapter.ts:335` | event counts computed pre-pickup-label-filter; `dispatchableCount` overstates candidates |
| P3 | `.changeset/linear-dispatchable-candidates.md` | structured-event rename + loss of `tracker-pickup-label-filtered` not noted |
| nit | `orchestrator-adapter.ts:316` | two hunks fail `pnpm format` |
| nit | `file-tracker-adapter.test.ts:249` | `toEqual` → `toHaveLength(1)` drops the identity check |

Approving. The blocking issue is resolved and verified; none of the above needs another round — fold them in if you agree, or skip them.

---

moncher-dev:
**스모크테스트(file 어댑터) 통과 → Land로 이동 (프로젝트 상태 관리)**

head `5dbf387`(approve)를 격리 worktree에 빌드하고 로컬 E2E(file 트래커 + stub, `repo start --http 4788`)에 fixture 2건을 주입: #1 `dispatchable:false, dispatchReason:"Issue is not assigned…", assigneeId:null`, #2 `assigneeId:"moncher-dev"`.

| 검증 | 결과 |
| --- | --- |
| #1 run/`run-dispatched` 0건, `repo explain` → `Not dispatchable: Issue is not assigned to the worker account.` | ✅ |
| #2 정상 dispatch, explain `✓ Tracker marks this issue as dispatchable.` | ✅ |
| 데몬 오류 없음 | ✅ |

⚠️ **Linear 어댑터 부분은 실 스모크를 하지 않았습니다** — 로컬에 운영(HJ-company) Linear 키만 있어 테스트 팀/키 확보 여부를 오너에게 문의 중입니다. Linear 변경(서버측 assignee 필터 제거 → `assignee.id`로 `dispatchable` 파생)은 이 PR의 `tracker-linear.test.ts` 단위테스트로만 검증된 상태입니다. 병합 판단 시 참고해 주세요. CI red는 #702 미병합 때문(vitest 122/122 통과).

moncher-dev:
**사후 Linear 실 스모크 통과 (병합 후 main `32f17d1` 기준)**

테스트 워크스페이스 `moncher-symphony-test`(팀 MON, 봇 계정 키)에 프로젝트 `symphony-smoke`를 만들고 봇 할당 이슈 MON-5와 미할당 MON-6을 넣은 뒤 `linearTrackerAdapter.listIssues/fetchIssueStatesByIds`를 직접 호출했습니다.

| 호출 | 결과 |
| --- | --- |
| `listIssues({assignedOnly:true})` | 2건 모두 반환; MON-6 `dispatchable:false`, MON-5 `dispatchable:true` (`assigneeId` = viewer id); `tracker-dispatchable-derived { dispatchableCount:1, nonDispatchableCount:1 }` |
| `listIssues({assignedOnly:false})` | 2건 모두 `dispatchable:true` |
| `fetchIssueStatesByIds(…, {assignedOnly:true})` | 전체 스냅샷 + 동일한 dispatchable 판정 |

관찰: Linear 경로는 `dispatchReason`이 `null`입니다(GitHub는 "Issue is not assigned to …"). `repo explain`이 이유를 못 보여주므로 #689(문서/E2E) 또는 어댑터 프로필 작업에서 Linear에도 이유 문자열을 채우는 것을 권합니다. 앞선 Land 코멘트의 caveat는 해소되었습니다.

PR #706, opened 2026-08-28
feat(cli): preflight repository workflow startup

### LINKED ISSUE #670: feat(cli,orchestrator): repo start workflow preflight, unsupported tracker.kind must not abort reconciliation, honor repo init --workflow-file
Epic: #651
리포트 항목: **B8(나머지), B19** — `docs/reports/2026-08-28-upstream-spec-drift-research.md` (PR #650)

## 문제

§6.3/§8.1: 시작 시 설정을 검증하고, 검증 실패는 dispatch만 막고 reconciliation은 계속돼야 한다. §6.1/§17.7: 명시적 워크플로 경로를 존중한다.

## 현재 상태 (증거)

- `repo start`는 auth preflight만: `packages/cli/src/commands/start.ts:176-262, 905-1000`; `service.run()`은 cleanup만: `service.ts:949-982` → 잘못된 WORKFLOW.md로도 데몬이 뜨고 stderr만 반복
- 미지원 kind가 reconciliation **이전**에 throw: `service.ts:1181`, `tracker-adapters.ts:15-21` → 매 tick 전체 실패
- `repo init --workflow-file`은 한 번 파싱(`repo-runtime.ts:91-95`)되지만 데몬은 항상 `<repoDir>/WORKFLOW.md` 로드: `git.ts:482-488`, `service.ts:2504-2510`; standalone은 `workflowSource.path` 존중

## 할 일

- `repo start`/`project start`에서 S15의 검증기로 preflight; 실패 시 nonzero exit + 명확한 오류(§17.7)
- 어댑터 해석 실패를 tick의 dispatch 단계로 이동해 reconciliation 보존
- `config.json`에 `workflowSource.path` 기록 후 데몬이 사용 (repo 모드)

## 완료 기준

- 잘못된 kind로 `repo start` 시 시작 실패 테스트; 런타임 중 kind 변경 시 reconciliation이 계속되는 테스트
- `--workflow-file` 경로가 데몬 로드에 반영되는 테스트

Blocked by: #667

moncher-dev:
## TL;DR

- Preflight invalid configured workflows before daemon startup.
- Persist and honor `repo init --workflow-file` paths across daemon and diagnostics.
- Preserve reconciliation when runtime tracker adapter resolution fails.

## Change-point diagram

```text
repo init --workflow-file → workflowSource.path → preflight / daemon / diagnostics
invalid tracker.kind → reconciliation continues → dispatch skips
```

## Start here

- `packages/cli/src/commands/start.ts`

## Risks & rollback

- Legacy projects without a configured workflow path retain daemon defaults.
- Reverting this PR restores prior startup behavior.

## Changed files

- CLI workflow preflight and configured-source consumers.
- Orchestrator dispatch/reconciliation and workflow-loading coverage.
- User documentation and E2E scenario table.

## Validation

- `pnpm lint`, `pnpm test`, `pnpm typecheck`, and `pnpm build` passed.
- `./e2e/run-e2e.sh happy 60` passed.
- Rebased head `ec8248a7`: GitHub Test and Container Smoke passed.

- [ ] Exercise `repo init --workflow-file <custom-path>` followed by `repo start` in a real repository.
- [ ] Confirm invalid configured workflow errors before daemon launch.


packages/cli/src/commands/start.ts
```diff
@@ -198,6 +200,36 @@ async function preflightRepoStartAuth(
...
+    (repositoryDirectory ? join(repositoryDirectory, "WORKFLOW.md") : null);
+  if (!workflowPath) {
+    // Legacy managed-project configs did not record a workflow source. Repo
+    // init now persists one, while existing installations keep their prior
+    // runtime behavior until they are reinitialized.
+    return true;
+  }
+
+  try {
+    parseWorkflowMarkdown(await readFile(workflowPath, "utf8"), process.env, {
```
hojinzs: **P1 — preflight uses `process.env` only, but the daemon resolves the workflow against the project `.env` + `process.env`. This blocks startup for a documented, currently-working configuration.**

`service.ts:2620` loads the workflow with `resolveProjectEnvironment(tenant)`, which is `{ ...readProjectEnv(tenant), ...process.env }` (`service.ts:4517-4524`) — the project `.env` documented at `docs/configuration.md:72` ("The folder owns `WORKFLOW.md`, optional `.mcp.json`, `.env` …") and `:193` (layer 1 of the env precedence table). The preflight here passes bare `process.env`, so any front-matter value sourced from the project `.env` (`env:VAR` / `$VAR` → `parser.ts:1067-1094` throws `WorkflowValidationError`) now aborts startup before the daemon ever gets a chance to resolve it.

Reproduced against this branch with a `file`-tracker repo, `codex.command: $SYMPHONY_CODEX_CMD`, and `SYMPHONY_CODEX_CMD` present only in the project `.env` (mode 0600):

```
$ gh-symphony repo start --once
Workflow preflight failed for .../policy/custom-workflow.md: Workflow front matter field
"codex.command" requires environment variable SYMPHONY_CODEX_CMD.
exit=1
```

With the `preflightWorkflowStart` call removed and nothing else changed, the identical configuration starts fine and the tick resolves the value from the project `.env`:

```
▲ Starting orchestrator for project: repository
· Running one orchestration tick
● Repository acme/platform connected (idle)
exit=0
```

This is the realistic daemon shape: `repo init` runs once from an interactive shell with the secret exported, then `repo start --daemon` runs from systemd/a container where the project `.env` is the intended source. `tracker.api_key: env:LINEAR_API_KEY` hits this the same way.

Suggested fix: resolve the same environment the daemon will use before parsing — export the project-env read (`readProjectEnv`) or reuse `loadWorkflowFile`'s resolution so preflight and runtime agree by construction rather than by parallel implementation.

---

packages/cli/src/commands/start.ts
```diff
@@ -198,6 +200,36 @@ async function preflightRepoStartAuth(
...
+    projectConfig.repositoryDir ?? projectConfig.repository.path;
+  const workflowPath =
+    configuredPath ??
+    (repositoryDirectory ? join(repositoryDirectory, "WORKFLOW.md") : null);
+  if (!workflowPath) {
+    // Legacy managed-project configs did not record a workflow source. Repo
+    // init now persists one, while existing installations keep their prior
+    // runtime behavior until they are reinitialized.
+    return true;
+  }
```
hojinzs: **P2 — the legacy fallback path validates a file the daemon may not load, and turns "missing" into a hard failure the daemon treats as a soft default.**

Two divergences, both only on the `configuredPath == null` branch (i.e. exactly the legacy configs the comment says should "keep their prior runtime behavior"):

1. **Different file.** Preflight resolves `repositoryDir ?? repository.path`; the daemon resolves `resolveWorkflowRepositoryDirectory()` = `repository.path` → local `cloneUrl` path → `process.cwd()` (`service.ts:5003-5018`). When `repositoryDir` and `repository.path` differ, preflight validates one `WORKFLOW.md` and the daemon loads another, so a pass here says nothing about the run.

2. **Missing file.** `loadWorkflowFile` returns `createDefaultWorkflowResolution()` on ENOENT (`git.ts:501-504`) — the daemon starts on defaults. Preflight instead `readFile`s and rejects. Startup now fails for a repo that previously ran on the default workflow, and the message is a raw fs error:

```
Workflow preflight failed for .../policy/custom-workflow.md: ENOENT: no such file or
directory, open '.../policy/custom-workflow.md'
```

Issue #670 asks for "nonzero exit + 명확한 오류(§17.7)", and every neighbouring preflight in this file pairs the failure with a remediation line. Suggest: mirror `loadWorkflowFile`'s missing-file semantics (or, if hard-failing on a *configured-but-absent* path is intended, say so explicitly — `Configured workflow not found at <path>. Run 'gh-symphony repo init --workflow-file <path>' …`), and reuse the daemon's directory resolution for the fallback.

---

packages/orchestrator/src/service.ts
```diff
@@ -2617,14 +2618,13 @@ export class OrchestratorService {
...
     const environment = this.resolveProjectEnvironment(tenant);
-    const resolution =
-      tenant.workflowSource?.type === "external"
-        ? await loadWorkflowFile(tenant.workflowSource.path, environment)
-        : await loadRepositoryWorkflow(
-            this.resolveWorkflowRepositoryDirectory(repository),
-            repository,
-            environment
-          );
+    const resolution = tenant.workflowSource?.path
```
hojinzs, line 2536: **P3 — this call site was widened to `workflowSource?.path`, but `resolveWorkflowWarnings` right below still keys on `type === "external"`.**

`resolveWorkflowWarnings` (`service.ts:2631-2636`) early-returns for anything that isn't `external`. After this PR, `repo init --workflow-file policy/custom.md` produces a `repo`-typed path that can sit next to a committed root `WORKFLOW.md` — precisely the "which policy is actually in effect?" confusion the shadow warning exists to surface — and the operator now gets no warning for it.

Widening that guard to `if (!tenant.workflowSource?.path) return []` would keep the two paths consistent. Low priority; flagging so they don't drift further.

Also, on the same hunk: no orchestrator test covers this new branch. Every `workflowSource` fixture in `service.test.ts` is `type: "external"` (13350, 13548, 13614, 13670, 13710, 13742, 13800), and `repo.test.ts` asserts only that `repo init` *persists* the path. Issue #670's 완료 기준 asks for "`--workflow-file` 경로가 데몬 로드에 반영되는 테스트" — a `type: "repo"` + `path` fixture asserting `loadProjectWorkflow` reads that file instead of `<repoDir>/WORKFLOW.md` would close it.

---

packages/cli/src/commands/start.test.ts
```diff
@@ -275,6 +276,33 @@ describe("shutdownForegroundOrchestrator", () => {
 });
 
 describe("start command foreground locking", () => {
+  it("fails before constructing the daemon when the configured workflow is invalid", async () => {
```
hojinzs: **P2 — this test does not cover the acceptance criterion it is standing in for.**

Issue #670's 완료 기준 reads "잘못된 **kind**로 `repo start` 시 시작 실패 테스트". This fixture writes `"not a workflow"`, so it exercises the front-matter parse failure, not tracker-kind rejection. And because `getSupportedTrackerKinds` is stubbed at line 62 of this file, the one thing that *would* exercise kind rejection — the `supportedTrackerKinds` option threaded into `parseWorkflowMarkdown` — is never actually asserted against the real supported set.

A second case with valid front matter and `tracker.kind: retired-kind` would cover the stated criterion and pin the `getSupportedTrackerKinds()` wiring at the same time.

---

hojinzs (changes requested):
## Review — round 1/6

The three things #670 asked for are all present, and I verified each of them working against `f3557e5` with a real `file`-tracker repo, not just by reading the diff. One blocker: the new preflight refuses to start a configuration that the daemon handles correctly today.

### Requirement coverage vs #670

| 할 일 | |
|---|---|
| `repo start`/`project start` preflight, nonzero exit + 명확한 오류 (§17.7) | ⚠️ works, but see P1 and the ENOENT message |
| 어댑터 해석 실패를 dispatch 단계로 이동해 reconciliation 보존 | ✅ `service.ts:1281`, covered by a new test |
| `config.json`에 `workflowSource.path` 기록 후 데몬이 사용 | ✅ verified end-to-end |

| 완료 기준 | |
|---|---|
| 잘못된 kind로 `repo start` 시 시작 실패 테스트 | ⚠️ the new test uses an unparseable workflow, not an unsupported kind |
| 런타임 중 kind 변경 시 reconciliation 계속 테스트 | ✅ |
| `--workflow-file` 경로가 데몬 로드에 반영되는 테스트 | ⚠️ persistence is tested; daemon loading is not |

### Smoke test (built CLI, `file` tracker, fresh repo)

| Case | Result |
|---|---|
| `repo init --workflow-file policy/custom-workflow.md` | ✅ persists `{"type":"repo","path":"/abs/…"}` |
| `repo start --once`, valid custom workflow | ✅ `● Repository acme/platform connected (idle)` |
| `repo start --once`, invalid workflow | ✅ `Workflow preflight failed …: WORKFLOW.md must use YAML front matter.`, exit 1 |
| `repo start --once`, workflow file deleted | ⚠️ blocks on a raw `ENOENT: no such file or directory, open '…'` |
| `repo start --once`, value from project `.env` | ❌ **blocked** — see P1 |
| `doctor` after `--workflow-file` | ❌ two false required-check failures — see the open Codex thread |

`pnpm build` and `vitest run start.test.ts repo.test.ts service.test.ts` (247 tests) pass locally.

### Findings

- **P1** — `preflightWorkflowStart` parses with bare `process.env`, while the daemon uses `{ ...projectEnv, ...process.env }`. Any front-matter value sourced from the project `.env` (`tracker.api_key: env:LINEAR_API_KEY`, `codex.command: $VAR`) now aborts startup. Confirmed by removing only the preflight call and watching the identical config start cleanly. Inline on `start.ts:220`.
- **P2** — legacy fallback validates `repositoryDir ?? repository.path` while the daemon resolves `repository.path → cloneUrl → cwd`; and a missing file is a soft default for the daemon but a hard failure here, with an unactionable message. Inline on `start.ts:209-217`.
- **P2** — `doctor` and `repo explain` do not consult the persisted `repo`-typed path. Reproduced; details in my reply on the existing Codex thread.
- **P2** — test gaps against the 완료 기준 above. Inline on `start.test.ts:279` and `service.ts:2621`.
- **P3** — `resolveWorkflowWarnings` still keys on `type === "external"`, so a `repo`-typed path no longer produces a shadowing warning. Inline on `service.ts:2621`.
- **nit** — `service.ts:1394-1400` and `4145-4149` and the `docs/configuration.md` table are pure Prettier reflow, unrelated to #670. Harmless, just diff noise.

No over-engineering or unrequested scope found otherwise; `normalizeWorkflowSource` keeping `path` optional is the right call for back-compat, and moving `resolveTrackerAdapter` into the dispatch `try` is a minimal, correct fix.

### CI — not this PR's fault

The red `Test` check is **not** caused by this change. All 1743 unit tests pass; the job then fails on:

```
```

`.github/workflows/ci.yml:48` invokes `pnpm e2e:compose-project-test`, which does not exist in the root `package.json` (nearest is `e2e:standalone-project`). Neither file is touched here, so this is red on `main` too and needs a separate fix — please don't chase it in this PR.

---

packages/cli/src/repo-runtime.ts
```diff
@@ -166,6 +166,7 @@ export async function initRepoRuntime(flags: RepoInitFlags): Promise<{
       priority: workflow.tracker.priority,
       settings: trackerSettings,
     },
+    workflowSource: { type: "repo", path: workflowPath },
```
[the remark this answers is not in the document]
hojinzs: Confirming this finding — it reproduces against `f3557e5`. Treating it as **P2** to address before merge.

Repro (fresh git repo, `file` tracker, no root `WORKFLOW.md`):

```
$ gh-symphony repo init --workflow-file policy/custom-workflow.md
Workflow: .../policy/custom-workflow.md

$ cat .runtime/orchestrator/project.json   # workflowSource
{"type":"repo","path":".../policy/custom-workflow.md"}    # persisted correctly

$ gh-symphony repo start --once
● Repository acme/platform connected (idle)                # startup honors it

$ gh-symphony doctor
FAIL Repository WORKFLOW.md
  WORKFLOW.md was not found in the repository root.
  Fix: Run 'gh-symphony workflow init' in this repository or add a valid WORKFLOW.md at the repo root.
FAIL Runtime command detection
  Runtime command detection could not run because WORKFLOW.md is missing or invalid.
```

So `doctor` reports two required-check failures for a configuration this PR makes fully supported, and the false `workflow_file` failure cascades into `runtime_command`. `doctor.ts:2059-2062` narrowing to `type === "external"` is the cause, as noted.

`repo-explain.ts:304-315` has the same shape: `resolveExplainWorkflowCandidates` takes `explicitWorkflowPath` only from the `--workflow` flag (`repo-explain.ts:173`) and otherwise falls back to `repository.path/WORKFLOW.md`, so `repo explain` can evaluate a different policy than the daemon is running.

Both consumers should read `workflowSource.path` regardless of source type — the same widening `service.ts:2621` already does.

---

packages/cli/src/commands/start.ts
moncher-dev: Resolved in a55f6a3: startup preflight now resolves the managed project .env before inherited process variables, matching daemon precedence. A regression test covers an env-only codex.command.

packages/cli/src/commands/start.ts
moncher-dev: Resolved in a55f6a3: legacy configs no longer take a divergent preflight path, preserving daemon fallback semantics; an explicitly configured missing path now fails with a remediation command.

packages/orchestrator/src/service.ts
moncher-dev: Resolved in a55f6a3: repo-typed workflow sources now produce the same shadow warning, and the service regression test verifies the daemon loads the configured repo path.

packages/cli/src/commands/start.test.ts
moncher-dev: Resolved in a55f6a3: the start-command regression uses valid front matter with tracker.kind: retired-kind and verifies startup stops before daemon construction.

packages/cli/src/repo-runtime.ts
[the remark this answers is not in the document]
moncher-dev: Resolved in a55f6a3: doctor and repo explain now use workflowSource.path for repo-typed sources, with direct diagnostics and explanation-path regression coverage.

packages/cli/src/commands/doctor.ts
```diff
@@ -2058,7 +2058,7 @@ export async function runDoctorDiagnostics(
 
   const externalWorkflowPath =
     resolvedProjectConfig?.kind === "resolved" &&
-    resolvedProjectConfig.projectConfig.workflowSource?.type === "external"
+    resolvedProjectConfig.projectConfig.workflowSource?.path
```
hojinzs, line 2063: **P2 — the path widening landed, but `doctor` still parses with bare `process.env`, so it keeps reporting two false required-check failures for the configuration `repo start` now accepts.**

This is the same divergence as the round-1 P1, just in a consumer that didn't get the fix. `checkWorkflow` parses at `doctor.ts:700` with `process.env`, while `start.ts:217-220` (correctly) uses `{ ...readEnvFile(projectDir/.env), ...process.env }`.

The symptom from my round-1 repro is unchanged from the operator's seat — only the message moved from "not found" to "could not be parsed". Verified against `a55f6a3` with the built CLI, `file` tracker, `codex.command: $SYMPHONY_CODEX_CMD`, value present **only** in the project `.env` (0600):

```
$ gh-symphony repo start --once
● Repository acme/platform connected (idle)          # starts fine — P1 fix works

$ gh-symphony doctor
FAIL Project WORKFLOW.md
  WORKFLOW.md could not be parsed.
  Fix: Fix the WORKFLOW.md front matter or re-run 'gh-symphony workflow init' to regenerate it.
FAIL Runtime command detection
  Runtime command detection could not run because WORKFLOW.md is missing or invalid.
```

With the identical file and `SYMPHONY_CODEX_CMD` merely exported, both flip to PASS — so the front matter is fine and the remediation text sends the operator to fix a file that isn't broken. The `workflow_file` failure still cascades into `runtime_command` via `blockedBy`.

`resolvedProjectConfig` is already in scope here with `projectDir`/`projectId`, so threading the same environment into `checkWorkflow` is the small fix. The existing `doctor.test.ts` change only flips the fixture `external` → `repo`, which is why this slipped through — a case with an `env:`/`$VAR` front-matter value sourced from the project `.env` would pin it.

---

packages/cli/src/commands/repo-explain.ts
```diff
@@ -303,13 +305,17 @@ async function loadExplainWorkflow(input: {
...
+  configuredWorkflowPath?: string;
   repository: RepositoryRef;
   runs: readonly OrchestratorRunRecord[];
 }): string[] {
   const paths: string[] = [];
   if (input.explicitWorkflowPath) {
     paths.push(resolve(input.explicitWorkflowPath));
   }
+  if (input.configuredWorkflowPath) {
+    paths.push(resolve(input.configuredWorkflowPath));
```
hojinzs, line 299: **P2 — same `process.env` divergence as the `doctor` comment, and here it fails *silently* into the wrong policy rather than reporting an error.**

Adding the configured path as a candidate is right, but `loadExplainWorkflow` (`repo-explain.ts:288`) calls `new WorkflowConfigStore().load(workflowPath)` with no `env` argument, and that parameter defaults to `process.env` (`core/src/workflow/loader.ts:27`). When the configured workflow resolves a value from the project `.env`, the load throws, the `catch` at `:295` pushes it onto `failures`, and the loop falls through to the **next** candidate — `repository.path/WORKFLOW.md`. Nothing is printed unless *every* candidate fails, so `repo explain` reports a policy the daemon is not running, with no indication anything was substituted.

Reproduced against `a55f6a3` by driving the real `WorkflowConfigStore` over the same candidate list this function builds, configured workflow using `codex.command: $SYMPHONY_CODEX_CMD` (project `.env` only) and a root `WORKFLOW.md` carrying a deliberately distinct `active_states`:

```
FAILED: …/policy/custom-workflow.md -> Workflow front matter field "codex.command" requires environment variable SYMPHONY_CODEX_C…
USED:   …/WORKFLOW.md
activeStates: ["RootOnlyState","In Progress"]
```

That is exactly the "evaluates a different policy than the daemon is running" hazard from the round-1 thread — the candidate ordering closes it for the missing-file case but not for the env case. Passing the resolved project environment into `.load(workflowPath, env)` fixes it; the ordering change here is then load-bearing rather than best-effort.

(I could not exercise this through the `repo explain` CLI end-to-end — it requires GitHub auth, which this environment doesn't have — so the repro above drives the same loader and candidate order directly rather than the command.)

---

packages/cli/src/commands/start.ts
```diff
@@ -198,6 +200,46 @@ async function preflightRepoStartAuth(
...
+    // Legacy managed-project configs rely on the daemon's repository lookup,
+    // including its default workflow fallback. Avoid a divergent preflight.
+    return true;
+  }
+
+  const projectDirectory =
+    projectConfig.projectDir ??
+    join(runtimeRoot, "projects", projectConfig.projectId);
+  const environment = {
+    ...readEnvFile(join(projectDirectory, ".env")),
```
hojinzs: **nit — this is the correct precedence (verified working), but it re-implements the daemon's project-dir resolution instead of sharing it.**

The daemon uses `tenant.projectDir ?? this.store.projectDir(tenant.projectId)` (`service.ts:4494`), and `OrchestratorFsStore.projectDir` applies `encodeProjectId` = `encodeURIComponent` (`fs-store.ts:48, 521`). Line 216 here uses the raw `projectConfig.projectId` — note `start.ts:659` in this same file does apply `encodeURIComponent` for the same purpose.

Not reachable today, so genuinely a nit: repo mode uses the literal `"repository"`, and `standaloneProjectId` is a slug + hex hash (`standalone-project.ts:12-14`), both encode-stable — and standalone always persists `projectDir` anyway. Flagging only because parallel implementations of the daemon's resolution are what produced the P1 in the first place; exporting the resolution (or the whole `resolveProjectEnvironment`) would make preflight and runtime agree by construction. That same helper is what the `doctor` and `repo explain` comments need, so there may be one shared fix for all three.

Also minor: `readEnvFile` on line 218 sits outside the `try`, so an unreadable-but-present `.env` (EACCES) propagates out of `preflightWorkflowStart` as an unhandled throw, where the daemon warns and continues with `{}` (`service.ts:4509-4517`). Missing files are already safe — `readEnvFile` returns `{}`.

---

hojinzs (changes requested):
## Review — round 2/6

All five round-1 threads are genuinely fixed, and I verified each one against `a55f6a3` with the built CLI rather than by reading the diff. One finding remains: the P1 env-precedence fix was applied to `start.ts` but not to the two consumers this PR newly pointed at the configured path.

### Round-1 findings — re-verified

| Finding | Claim | Verified |
|---|---|---|
| **P1** preflight ignored project `.env` | fixed in `a55f6a3` | ✅ env-only `codex.command` now starts cleanly, exit 0 |
| **P2** legacy fallback divergence | fixed | ✅ `configuredPath == null` returns early — no divergent path |
| **P2** missing file → raw ENOENT | fixed | ✅ `Configured workflow not found at …. Restore the file or run 'gh-symphony repo init --workflow-file <path>'.`, exit 1 |
| **P2** `doctor` / `repo explain` ignore persisted path | fixed | ⚠️ path widening landed; **env resolution did not** — see below |
| **P3** `resolveWorkflowWarnings` keyed on `external` | fixed | ✅ shadow warning now fires for `repo`-typed paths |
| **P2** test used unparseable file, not bad kind | fixed | ✅ real CLI rejects `tracker.kind: retired-kind` |

### Smoke test (built CLI, `file` tracker, fresh repo, no root `WORKFLOW.md`)

| Case | Result |
|---|---|
| `repo init --workflow-file policy/custom-workflow.md` | ✅ persists `{"type":"repo","path":"/abs/…"}` |
| `repo start --once`, value from project `.env` only | ✅ **the round-1 P1 blocker is gone** — starts, exit 0 |
| `repo start --once`, configured file deleted | ✅ actionable remediation, exit 1 |
| `repo start --once`, `tracker.kind: retired-kind` | ✅ `Unsupported workflow tracker.kind "retired-kind". Supported values: github-project, file, linear.`, exit 1 |
| daemon loads configured path, not root | ✅ proved with a deliberately **invalid** root `WORKFLOW.md` — still starts |
| root `WORKFLOW.md` alongside configured path | ✅ shadow warning present in `status.json` |
| `doctor` with the value in project `.env` only | ❌ two false required-check failures |
| `repo explain` with the value in project `.env` only | ❌ silently evaluates the root `WORKFLOW.md` |

`pnpm build` passes; `vitest run start.test.ts doctor.test.ts repo-explain.test.ts repo.test.ts service.test.ts` — 312 passed.

### Findings

- **P2** — `doctor.ts:700` and `repo-explain.ts:288` both parse with bare `process.env` while `start.ts:217-220` now correctly uses `{ ...readEnvFile(projectDir/.env), ...process.env }`. One root cause, two call sites. `doctor` reports FAIL on a config `repo start` accepts, with remediation text pointing at front matter that is fine; `repo explain` is the worse of the two — its candidate loop swallows the parse failure and falls through to the root `WORKFLOW.md`, reporting a **different policy than the daemon is running, with no error shown**. Inline on `doctor.ts:2061` and `repo-explain.ts:317`, each with a repro.
- **nit** — the preflight re-implements the daemon's project-dir resolution (missing `encodeURIComponent`) instead of sharing it; unreachable today. Plus `readEnvFile` sits outside the `try`. Inline on `start.ts:218`.

I'll note #670's 할 일 doesn't itself name `doctor`/`repo explain` — they entered scope because round 1 raised them and this PR changed them. So if you'd rather land the env alignment for all three consumers as a tracked follow-up rather than growing this PR, say so on the thread and I'll approve on that basis; what I don't want to wave through unremarked is `repo explain` silently answering with the wrong policy. Either resolution closes this round.

No over-engineering or unrequested scope in `a55f6a3`. The `normalizeWorkflowSource` absolute-path validation is proportionate, and the `resolveWorkflowWarnings` label switch is the minimal way to widen the guard.

### CI — still not this PR's fault

Unchanged from round 1, and I re-confirmed it on this head: **1748 unit tests pass**, then the job dies on

```
```

`.github/workflows/ci.yml:48` calls a script absent from the root `package.json` — I verified it is missing on `main` too, and this PR touches neither file. Needs a separate fix; please don't chase it here.

---

packages/cli/src/commands/doctor.ts
moncher-dev: Resolved in c3fe4a13: `doctor` now uses the shared managed-project environment resolver, so configured workflows receive project `.env` values with process-environment override precedence. The existing registered repo-source diagnostic test now supplies `codex.command` only through the project `.env`.

packages/cli/src/commands/repo-explain.ts
moncher-dev: Resolved in c3fe4a13: `repo explain` passes the same shared managed-project environment into `WorkflowConfigStore.load`, preventing a configured workflow from failing on an `.env` value and silently falling back to a different root policy. A regression test covers the persisted repo path with an env-only command.

packages/cli/src/commands/start.ts
moncher-dev: Resolved in c3fe4a13: start preflight now also uses the shared managed-project resolver. It follows daemon-compatible encoded project-directory resolution and treats an unreadable managed `.env` as absent, rather than throwing before workflow validation.

packages/cli/src/commands/doctor.ts
```diff
@@ -2058,13 +2060,19 @@ export async function runDoctorDiagnostics(
...
```
hojinzs, line 2063: **P3 — the widening is correct, but the downstream message is still standalone-project-only, so a missing `repo`-typed path now gets remediation for the wrong command.**

This line now feeds `repo`-typed paths into `checkWorkflow`'s `externalWorkflowPath`, which is right. But `checkWorkflow:695-696` branches on that same value to emit standalone wording:

```ts
remediation: externalWorkflowPath
  ? `Restore ${workflowPath}; the folder's workflow is derived on every 'gh-symphony project start'.`
```

Verified against `c3fe4a13` with the built CLI (`file` tracker, `repo init --workflow-file policy/custom-workflow.md`, then the file deleted):

```
$ gh-symphony doctor
FAIL Project WORKFLOW.md
  WORKFLOW.md was not found at the registered project path .../policy/custom-workflow.md.
  Fix: Restore .../policy/custom-workflow.md; the folder's workflow is derived on
       every 'gh-symphony project start'.
```

There is no folder-derived workflow here and `project start` is not the command this operator runs — `start.ts:216-219` gets exactly this case right (`Restore the file or run 'gh-symphony repo init --workflow-file <path>'.`), so the two consumers now disagree about the same failure. Branching the remediation on `workflowSource.type` (or threading the type in) would align them; the `externalWorkflowPath` parameter name is also a misnomer now that it carries both types.

Non-blocking — the check itself is correct and this is only the failure text.

---

packages/cli/src/commands/repo-explain.ts
```diff
@@ -283,7 +288,10 @@ async function loadExplainWorkflow(input: {
   const failures: string[] = [];
   for (const workflowPath of workflowPaths) {
     try {
-      const resolution = await new WorkflowConfigStore().load(workflowPath);
+      const resolution = await new WorkflowConfigStore().load(
+        workflowPath,
+        input.environment
+      );
```
hojinzs, line 270: **P3 — env resolution confirmed fixed; the silent-substitution shape behind it survives for non-env parse failures.**

The `input.environment` threading here does what round 2 asked for, and I verified it end-to-end rather than by reading the diff. With the configured workflow using `codex.command: $SYMPHONY_CODEX_CMD` and the value present **only** in `projects/repository/.env`, and a deliberately **invalid** root `WORKFLOW.md`:

```
$ gh-symphony repo explain acme/platform#1
Issue dispatch explanation: acme/platform#1     # loads the configured policy, root ignored
```

Remove the `.env` and it now reports both candidates instead of silently substituting:

```
Error: Unable to load WORKFLOW.md for repo explain. Checked:
  .../policy/custom-workflow.md: Workflow front matter field "codex.command" requires
    environment variable SYMPHONY_CODEX_CMD.;
  .../WORKFLOW.md: WORKFLOW.md must use YAML front matter.
```

The residual: the `catch` at `:301` still pushes onto `failures` and continues, so when the configured workflow fails for any *other* reason and a root `WORKFLOW.md` happens to parse, the fall-through is still silent. Confirmed on this head — configured file replaced with `garbage not a workflow`, root valid with a distinct `active_states: [RootOnlyState]`:

```
$ gh-symphony repo explain acme/platform#1
Issue dispatch explanation: acme/platform#1     # succeeded, on the root policy, no notice
```

Flagging for the record, not as a blocker: this isn't a regression (before this PR `repo explain` always read the root file), and the env case that made it likely in practice is now closed. If you want it fully shut, stopping at the configured candidate — or printing the swallowed failure — would do it. Fine as a follow-up.

---

hojinzs (approved):

Both round-2 P2s and the nit are genuinely fixed, and I verified each against `c3fe4a13` with the built CLI plus a mutation test — not by reading the diff. Nothing blocking remains in the code. **One mechanical item before merge: this PR currently conflicts with `main`** (see below).

### Round-2 findings — re-verified

| Finding | Claim | Verified |
|---|---|---|
| **P2** `doctor` parsed with bare `process.env` | fixed in `c3fe4a13` | ✅ `PASS Project WORKFLOW.md` + `PASS Runtime command detection` on an env-only config |
| **P2** `repo explain` silently fell through to root policy | fixed | ✅ loads the configured policy; reports both candidates instead of substituting |
| **nit** re-implemented project-dir resolution, `readEnvFile` outside `try` | fixed | ✅ `managed-project-environment.ts` uses `encodeURIComponent` and swallows an unreadable `.env`, matching `service.ts:4498-4520` |

The shared resolver is the right shape: one 28-line module, three call sites, and it now agrees with the daemon by construction rather than by parallel implementation — which is what produced the round-1 P1 in the first place.

### Smoke test (built CLI, `file` tracker, fresh repo, `repo init --workflow-file`)

Value for `codex.command` present **only** in `projects/repository/.env` (0600), never exported:

| Case | Result |
|---|---|
| `repo init --workflow-file policy/custom-workflow.md` | ✅ persists `{"type":"repo","path":"/abs/…"}` |
| `repo start --once` | ✅ `● Repository acme/platform connected (idle)`, exit 0 |
| `doctor` | ✅ both required checks PASS |
| `doctor` with `.env` removed *(negative control)* | ✅ flips to FAIL — the pass is genuinely caused by the fix, not a fallback |
| `repo explain`, root `WORKFLOW.md` deliberately invalid | ✅ succeeds on the configured policy |
| `repo explain` with `.env` removed | ✅ errors listing both candidates |
| `repo start --once`, configured file deleted | ✅ `Configured workflow not found at …. Restore the file or run 'gh-symphony repo init --workflow-file <path>'.`, exit 1 |
| `repo start --once`, `tracker.kind: retired-kind` | ✅ `Unsupported workflow tracker.kind "retired-kind". Supported values: github-project, file, linear.`, exit 1 |

**Mutation test on the new coverage.** Round 2 noted the previous `doctor` test only flipped a fixture `external` → `repo`, which is why the env bug slipped through. The new fixtures use a `$VAR` front-matter value sourced from the project `.env`, so I checked they actually bite: reverting just the two env threadings fails exactly 2 of 63 tests, one per file. The coverage is load-bearing.

`pnpm lint`, `pnpm typecheck`, `pnpm build` pass; `pnpm test` passes across all packages (574 in `cli`, 253 in `core`, all others green).

### Requirement coverage vs #670 — complete

| 할 일 | |
|---|---|
| `repo start`/`project start` preflight, nonzero exit + 명확한 오류 (§17.7) | ✅ |
| 어댑터 해석 실패를 dispatch 단계로 이동해 reconciliation 보존 | ✅ `service.ts:1281` |
| `config.json`에 `workflowSource.path` 기록 후 데몬이 사용 | ✅ |

| 완료 기준 | |
|---|---|
| 잘못된 kind로 `repo start` 시 시작 실패 테스트 | ✅ real CLI rejects `retired-kind` |
| 런타임 중 kind 변경 시 reconciliation 계속 테스트 | ✅ |
| `--workflow-file` 경로가 데몬 로드에 반영되는 테스트 | ✅ persistence **and** daemon loading both covered |

No over-engineering or unrequested scope in `c3fe4a13`. `doctor`/`repo explain` entered scope only because round 1 raised them and this PR already changed them; the fix stayed to the smallest thing that closes it.

### Remaining findings — both P3, neither blocking

- **P3** — `doctor`'s missing-file remediation still uses standalone wording (`the folder's workflow is derived on every 'gh-symphony project start'`) for a `repo`-typed path, disagreeing with `start.ts`, which gets the same case right. Inline on `doctor.ts:2063` with the repro.
- **P3** — `repo explain` still falls through to the root `WORKFLOW.md` without notice when the configured workflow fails to parse for a *non-env* reason. Not a regression, and the env case that made it likely is now closed. Inline on `repo-explain.ts:294`.

Approving over both — they are message/edge polish, and holding a fourth round for them would be exactly the ping-pong this process is meant to avoid. Take them here or as a follow-up, your call.

### Before merge — merge conflict

`mergeable_state` is `dirty`. Confirmed locally against `origin/main`:

```
CONFLICT (content): Merge conflict in docs/configuration.md
CONFLICT (content): Merge conflict in packages/cli/src/commands/start.test.ts
```

Please merge `main` in and resolve; the approval stands through that.

### CI — the rounds 1–2 blocker is gone

The red `Test` check I twice reported as not this PR's fault is **resolved**. `e2e:compose-project-test` now exists in the root `package.json` (landed on `main` via #702), and this branch picked it up in the `8b72ef4` merge. I ran it on this head: `compose project isolation helpers passed`, exit 0. No check runs have reported on `c3fe4a13` yet — worth a glance once they do, but there is no known-red check left to chase.

---

packages/cli/src/commands/start.ts
```diff
@@ -204,6 +207,42 @@ async function preflightRepoStartAuth(
...
+      runtimeRoot
+    );
+    parseWorkflowMarkdown(await readFile(configuredPath, "utf8"), environment, {
+      supportedTrackerKinds: getSupportedTrackerKinds(),
+    });
+    return true;
+  } catch (error) {
+    if ((error as NodeJS.ErrnoException | undefined)?.code === "ENOENT") {
+      process.stderr.write(
+        `Configured workflow not found at ${configuredPath}. Restore the file or run 'gh-symphony repo init --workflow-file <path>'.\n`
```
hojinzs, line 233: **nit — the remediation is repo-mode-only, but this preflight also runs for `external` (standalone) sources.**

`preflightWorkflowStart` is called from the one shared handler at `start.ts:1002`, so it gates `gh-symphony project start` as well as `repo start`. This string is unconditional, so a standalone project whose `WORKFLOW.md` is gone is told to run `gh-symphony repo init --workflow-file <path>` — not a command that applies to it. The handler already computes the right branch 7 lines above for the auth preflight's `retryCommand` (`start.ts:995`), so threading `options.invocation` (or `workflowSource.type`) in here would be a one-line fix.

This is the mirror image of the still-open `doctor.ts` P3 from round 3, which gives *standalone* wording (`the folder's workflow is derived on every 'gh-symphony project start'`) for a `repo`-typed path — I re-confirmed that one still reproduces on `ec8248a`:

```
$ gh-symphony doctor          # after repo init --workflow-file policy/custom-workflow.md, file deleted
FAIL Project WORKFLOW.md
  WORKFLOW.md was not found at the registered project path .../policy/custom-workflow.md.
  Fix: Restore .../policy/custom-workflow.md; the folder's workflow is derived on
       every 'gh-symphony project start'.
```

So the two consumers are each emitting the *other* mode's remediation. One small helper that maps `workflowSource.type` → remediation text would close both at once.

Evidence that this branch is reachable for `external` sources: this PR had to change the existing `"starts the active external project daemon from its project directory"` test (`start.test.ts:955-1008`) to write a real `WORKFLOW.md` — it previously pointed at a nonexistent `/tmp/standalone-project/WORKFLOW.md` and passed. Being straight about the limit of my check: I could not drive it through a real `gh-symphony project start`, because the standalone folder-derivation step fails earlier in my no-auth environment, so I can't rule out that the derivation makes this unreachable in practice. Message-only either way — not blocking, and fine as a follow-up alongside the two open P3s.

---

hojinzs (approved):
## Review — round 4/6 · Approve

Round 3 approved `c3fe4a13` subject to a merge conflict. The head has since been rebased to `ec8248a`, so this round re-verifies that the rebase preserved every round 1–3 fix — checked against the built CLI on `ec8248a`, not by reading the diff. It did. Nothing blocking; approving.

### Rebase integrity — every earlier fix re-verified on `ec8248a`

| Round | Finding | Still fixed on `ec8248a` |
|---|---|---|
| 1 · **P1** | preflight ignored project `.env` | ✅ env-only `codex.command` starts cleanly, exit 0 |
| 1 · **P2** | legacy fallback divergence | ✅ `configuredPath == null` still returns early |
| 1 · **P2** | missing file → raw ENOENT | ✅ actionable remediation, exit 1 |
| 1 · **P3** | `resolveWorkflowWarnings` keyed on `external` | ✅ shadow warning fires for `repo`-typed paths |
| 1 · **P2** | test used unparseable file, not bad kind | ✅ real CLI rejects `tracker.kind: retired-kind` |
| 2 · **P2** | `doctor` parsed with bare `process.env` | ✅ `PASS Project WORKFLOW.md` on an env-only config |
| 2 · **P2** | `repo explain` silent fall-through on env failure | ✅ `input.environment` threading intact at `repo-explain.ts:267` |
| 2 · **nit** | re-implemented project-dir resolution | ✅ `managed-project-environment.ts` intact, `encodeURIComponent` + `try` |

The round-3 nit about unrelated Prettier reflow (`service.ts:1394-1400`, `4145-4149`, the `docs/configuration.md` table) is also gone — the rebase dropped that diff noise. The change is now 17 files, all on-topic.

### Smoke test (built CLI on `ec8248a`, `file` tracker, fresh git repo)

`SYMPHONY_CODEX_CMD` present **only** in `.runtime/orchestrator/projects/repository/.env` (0600), never exported:

| Case | Result |
|---|---|
| `repo init --workflow-file policy/custom-workflow.md` | ✅ persists `{"type":"repo","path":"/abs/…"}` |
| `repo start --once`, value from managed `.env` only | ✅ `● Repository acme/platform connected (idle)`, exit 0 |
| `repo start --once`, `tracker.kind: retired-kind` | ✅ `Unsupported workflow tracker.kind "retired-kind". Supported values: github-project, file, linear.`, exit 1 |
| `repo start --once`, configured file deleted | ✅ `Configured workflow not found at …. Restore the file or run 'gh-symphony repo init --workflow-file <path>'.`, exit 1 |
| daemon loads configured path, not root | ✅ proved with a deliberately **invalid** root `WORKFLOW.md` — still starts, exit 0 |
| root `WORKFLOW.md` alongside configured path | ✅ `status.json` carries `Configured workflow source … shadows repository WORKFLOW.md at …` |
| `doctor` | ✅ `PASS Project WORKFLOW.md` |
| `doctor` with the managed `.env` removed *(negative control)* | ✅ flips to `FAIL … could not be parsed` — the pass is caused by the fix, not a fallback |

### Requirement coverage vs #670 — complete

| 할 일 | |
|---|---|
| `repo start`/`project start` preflight, nonzero exit + 명확한 오류 (§17.7) | ✅ one shared handler, `start.ts:1002`, ahead of both `acquireProjectLock` and the `--daemon` spawn |
| 어댑터 해석 실패를 dispatch 단계로 이동해 reconciliation 보존 | ✅ `resolveTrackerAdapter` moved inside the tick `try` (`service.ts:1167`) |
| `config.json`에 `workflowSource.path` 기록 후 데몬이 사용 | ✅ verified end-to-end above |

| 완료 기준 | |
|---|---|
| 잘못된 kind로 `repo start` 시 시작 실패 테스트 | ✅ |
| 런타임 중 kind 변경 시 reconciliation 계속 테스트 | ✅ `service.test.ts:340` asserts `reconcileRun` ran and `summary.recovered === 1` while the adapter throws |
| `--workflow-file` 경로가 데몬 로드에 반영되는 테스트 | ✅ persistence **and** daemon loading both covered |

No over-engineering or unrequested scope in the rebased head. `doctor`/`repo explain` entered scope only because round 1 raised them and this PR already changed them, and the fix stayed at the smallest thing that closes it.

### Mergeability and CI

- **Merge conflict: resolved.** `mergeable_state` is `clean`, and `origin/main` (`e33afbc`) is an ancestor of `ec8248a`.
- **CI: green** on this head — `Test` and `Container Smoke` both `success`.
- Locally: `pnpm lint`, `pnpm typecheck`, `pnpm build` pass. `pnpm test` had one failure, `start.test.ts > increments the port when the requested HTTP port is already in use` (`expected '32941' to be '32940'`) — that test binds an ephemeral port and asserts the next one is free, and my sandbox had it taken during the parallel run. It passes 3/3 in isolation on this branch, and CI is green on the same commit, so it is a port race in my environment, not this PR.

### Remaining findings — all non-blocking

- **P3** *(open since round 3, no reply)* — `doctor`'s missing-file remediation still uses standalone wording for a `repo`-typed path. Re-confirmed reproducing on `ec8248a`.
- **P3** *(open since round 3, no reply)* — `repo explain` still falls through to the root `WORKFLOW.md` without notice when the configured workflow fails to parse for a non-env reason. Not a regression; the env case that made it likely is closed.
- **nit** *(new)* — the mirror of the first P3: the preflight's missing-file remediation is repo-mode-only but this preflight also gates `project start`. Inline on `start.ts:233`; one helper would close both.

All three are message/edge polish on paths whose *behavior* is correct. Both P3s were approved over in round 3 with "take them here or as a follow-up, your call" — the threads are still open and unanswered, so before merging please either fix them or reply on each thread; leaving them silently open is the only process item outstanding. Not holding the approval for it, and I'd rather not spend a fifth round on wording.

---

moncher-dev:

head `ec8248a`(approve, CI green)를 격리 worktree에 빌드하고 격리 HOME에서 file 트래커 + stub으로 확인했습니다.

| 검증 | 결과 |
| --- | --- |
| `repo init --workflow-file custom/WORKFLOW.md` → `project.json.workflowSource = {type:"repo", path:<abs custom>}` | ✅ |
| 설정된 워크플로에 `tracker.kind: jira` → `repo start`가 데몬 생성 전 종료: `Workflow preflight failed for …: Unsupported workflow tracker.kind "jira"…` | ✅ |
| 정상 워크플로로 시작 → `repo status`에 revision + "custom path shadows repository WORKFLOW.md" 경고, dispatch 정상 | ✅ |
| 실행 중 커스텀 워크플로를 `kind: jira`로 변경 → 데몬 생존, tick 계속(`lastTickAt` 전진), `workflow_validation_error (tracker.kind)` 로그, health running(last-known-good) | ✅ |

병합되면 W3의 #669 분할 이슈들과 함께 진행 가능.

PR #713, opened 2026-08-28
docs(e2e): cover non-dispatchable issues

### LINKED ISSUE #689: docs+e2e: document dispatchable semantics and add a Docker E2E scenario for dispatchable=false

moncher-dev:
## TL;DR

Documents the adapter-owned dispatchability boundary and proves in Docker that a `dispatchable: false` candidate is reconciled without worker dispatch while retaining its explain reason.

## Change-point diagram

```text
adapter derives dispatchable + reason
          ↓
scheduler applies normalized gate only
          ↓
Docker E2E observes two post-injection ticks → no worker → explain reason retained
```

## Start here

- `./e2e/run-e2e.sh non-dispatchable 30`
- `docs/architecture.md` §§3 and 5

## Changes

- Documents scheduler/adapter ownership and the explicit Linear pickup-label filtering divergence from the upstream scheduler-owned label boundary.
- Adds a `dispatchable: false` file-tracker fixture and a Docker black-box scenario that verifies suppression plus the explain reason.
- Hardens the absence-based proof with two post-injection reconciliation ticks and resilient polling.

## Evidence

- `pnpm lint` — passed.
- `pnpm build` — passed.
- `pnpm test` — passed.
- `pnpm typecheck` — passed.
- `./e2e/run-e2e.sh non-dispatchable 30` — passed: zero workers, no `run-dispatched` event, and retained adapter reason in `repo explain`.

## Risks & rollback

- The E2E waits for two ticks for proof strength; reverting `e15b0305` and `42ab9577` restores the previous documentation and runner behavior.

## Changed files

- `README.md`, `docs/architecture.md`, `AGENT_TEST.md`
- `e2e/fixtures/non-dispatchable.json`, `e2e/run-e2e.sh`, `e2e/scenarios/18-dispatchable-eligibility.md`
- `packages/tracker-file/README.md`, `packages/tracker-linear/README.md`

- [ ] Review Docker E2E output in CI.
- [ ] Confirm the documented Linear divergence is acceptable.


moncher-dev:
**보류 (프로젝트 상태 관리, WIP 상한 8)** — 이 이슈는 WIP 상한 적용으로 Backlog에 있고, PR은 dispatch가 먼저 일어나 생성됐습니다. 리뷰 부하와 rebase 연쇄를 줄이기 위해 draft로 전환합니다. 보드에서 이슈가 다시 Ready/In review로 올라가면 draft를 해제하고 리뷰를 요청하겠습니다. 그때까지 push하지 마세요.

e2e/run-e2e.sh
```diff
@@ -120,6 +124,38 @@ fi
 
 log "Issues injected; refresh trigger accepted (202). Falling back to polling until dispatch is observed"
 
+if [ "$SCENARIO" = "non-dispatchable" ]; then
+  STATUS_JSON=$(orch_curl -s http://localhost:4680/api/v1/state)
```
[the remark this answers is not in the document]
moncher-dev: Resolved in f726fc64: the scenario now records the pre-refresh `lastTickAt` and waits for a new reconciliation tick before asserting zero active runs, explain output, and no dispatch event.

docs/architecture.md
```diff
@@ -46,6 +46,7 @@ touches a layer, check that its slice (and the linked documents) still holds.
 ### 3. Coordination — the orchestrator
 
 - Dispatch loop, concurrency, retry, reconciliation: `packages/orchestrator/src/service.ts`
+- The scheduler applies only the normalized `TrackedIssue.dispatchable` gate before loading workflow policy or starting a worker. It does not encode provider assignment, repository-scope, label, or fork rules; each tracker adapter derives those rules and supplies an explainable `dispatchReason` with non-dispatchable candidates.
```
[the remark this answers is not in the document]
moncher-dev: Resolved in f726fc64: the architecture text now states that project workflow and polling policy load before candidates are fetched; `dispatchable` gates issue-specific workflow loading and worker startup.

docs/architecture.md
```diff
@@ -66,8 +67,8 @@ touches a layer, check that its slice (and the linked documents) still holds.
 ### 5. Integration — tracker adapters (tracker-specific code lives only here)
 
 - GitHub Project V2: `packages/tracker-github` (including the adapter-owned linked-PR canonical-subject extension; opaque `nativeRef` data never crosses into orchestration). Source issue state and linked-PR metadata remain distinct from Project workflow status; it returns all active scoped items and derives GitHub assignment, repository-scope, pickup-label, and fork-PR eligibility as `dispatchable` with an explainable reason.
-- Linear: `packages/tracker-linear`
-- File-based (E2E only): `packages/tracker-file`
+- Linear: `packages/tracker-linear`; it derives provider-native assignment and pickup-label eligibility as the same normalized `dispatchable` contract.
```
hojinzs: **P2 — Linear does not express pickup-label eligibility as `dispatchable`**

I verified this independently of the existing Codex thread, and it holds. The sentence is correct for assignment but wrong for pickup labels, and the two use opposite mechanisms:

- Linear: `packages/tracker-linear/src/orchestrator-adapter.ts:204` calls `filterIssuesByPickupLabels`, which **removes** the records (`packages/core/src/workflow/pickup-labels.ts:13-16`). They never reach the scheduler and never reach `repo explain`.
- GitHub: `packages/tracker-github/src/orchestrator-adapter.ts:314-319` calls `resolvePickupLabelDispatchReason` and **retains** the record as `{ dispatchable: false, dispatchReason }`.

I ran `packages/tracker-linear/src/tracker-linear.test.ts` to confirm: the include-label case at L515-557 asserts `issues.map(i => i.identifier)` equals exactly `["ENG-1", "ENG-2"]` — the two label-ineligible issues are gone, not marked. So "the same normalized `dispatchable` contract" is only half true for Linear.

This matters more than a typical wording nit because the observability contract differs: a GitHub issue blocked by pickup labels is explainable via `repo explain`, a Linear one is invisible.

Suggested wording:

```
- Linear: `packages/tracker-linear`; it derives provider-native assignment eligibility as the
  normalized `dispatchable` contract. Pickup labels are applied as a candidate-listing filter
  rather than a retained `dispatchable: false` record, so label-ineligible Linear issues are not
  observable on explain surfaces — unlike the GitHub adapter, which retains them with a reason.
```

Related: `docs/symphony-spec.md:1196-1197` assigns `required_labels` application to the orchestrator, so adapter-side pickup-label handling is a repository divergence. Per `CLAUDE.md` ("Divergences from the spec must be explicit and documented"), this is a good place to name it rather than smooth it over — this PR is specifically about getting the adapter/scheduler boundary described accurately.

The same implication is carried by `packages/tracker-linear/README.md:15` ("Pickup-label eligibility is also derived here before candidate dispatch.") directly after the sentence about the scheduler consuming the normalized result; worth qualifying there too.

---

e2e/run-e2e.sh
```diff
@@ -120,6 +126,49 @@ fi
...
 
+if [ "$SCENARIO" = "non-dispatchable" ]; then
+  for i in $(seq 1 "$TIMEOUT"); do
+    STATUS_JSON=$(orch_curl -s http://localhost:4680/api/v1/state)
+    LAST_TICK=$(echo "$STATUS_JSON" | python3 -c "import sys,json; print(json.load(sys.stdin).get('lastTickAt', ''))")
+    if [ -n "$LAST_TICK" ] && [ "$LAST_TICK" != "$INITIAL_LAST_TICK" ]; then
+      break
+    fi
+    if [ "$i" = "$TIMEOUT" ]; then
+      fail "Refresh reconciliation did not complete before timeout"
```
hojinzs: **P2 — the tick-wait can still break on a tick that never saw the fixture**

This confirms the open Codex thread, with one correction to the mechanism that also makes the fix easier to reason about.

`lastTickAt` is the tick's **start** timestamp, not its completion time: `reconcileProject` captures `const now = this.now()` at `packages/orchestrator/src/service.ts:1096` and publishes that same value at `service.ts:1958`, when the snapshot is saved at the end of the tick. `requestReconcile()` only sets a flag (`service.ts:1029-1031`).

So the failing interleave is:

1. A periodic tick `T` starts at `t0` and reads the still-empty `issues.json`.
2. Line 106 copies the fixture; line 112 captures `INITIAL_LAST_TICK` — `T` has not published yet, so this is the *previous* tick's start time.
3. The refresh POST returns 202 and queues a later tick.
4. `T` completes and publishes `lastTickAt = t0`, which differs from `INITIAL_LAST_TICK`, so this loop breaks on a tick that read the empty fixture.

Every assertion downstream then passes vacuously: `activeRuns == 0` is trivially true, `repo explain` re-reads the fixture from disk independently of the orchestrator, and no `run-dispatched` event exists yet.

That is worse here than in a typical E2E because there is no positive signal to fall back on — a `dispatchable: false` candidate is silently filtered at `service.ts:1339` with no `suppressed` counter and no event. The scenario is entirely absence-based, so "the wait observed the right tick" is its only proof that the scheduler gate ran at all. If this breaks early, the scenario reports `PASSED` for a gate it never exercised.

Because `lastTickAt` is a start time, the cheapest sound fix is to require **two** distinct values: the second tick necessarily started after the first one finished, hence after the fixture copy.

```bash
  TICKS_SEEN=0
  PREV_TICK="$INITIAL_LAST_TICK"
  for i in $(seq 1 "$TIMEOUT"); do
    STATUS_JSON=$(orch_curl -s http://localhost:4680/api/v1/state 2>/dev/null || echo '{}')
    LAST_TICK=$(echo "$STATUS_JSON" | python3 -c "import sys,json; print(json.load(sys.stdin).get('lastTickAt') or '')" 2>/dev/null || echo '')
    if [ -n "$LAST_TICK" ] && [ "$LAST_TICK" != "$PREV_TICK" ]; then
      PREV_TICK="$LAST_TICK"
      TICKS_SEEN=$((TICKS_SEEN + 1))
      [ "$TICKS_SEEN" -ge 2 ] && break
    fi
    ...
```

Codex's alternative — comparing `lastTickAt` against the refresh request time — is also sound *precisely because* `lastTickAt` is a start timestamp, so please don't dismiss it on the assumption that it is a completion time. It just needs the reference timestamp taken from the container clock (`compose exec … date -u +%FT%TZ`) rather than the host's.

---

e2e/run-e2e.sh
```diff
@@ -103,7 +103,13 @@ log "Initial state: idle"
...
 # ── Inject issues ─────────────────────────────────────────────
 
-cp e2e/fixtures/happy-path.json e2e/fixtures/issues.json
+if [ "$SCENARIO" = "non-dispatchable" ]; then
+  cp e2e/fixtures/non-dispatchable.json e2e/fixtures/issues.json
+else
+  cp e2e/fixtures/happy-path.json e2e/fixtures/issues.json
+fi
+
+INITIAL_LAST_TICK=$(orch_curl -s http://localhost:4680/api/v1/state | python3 -c "import sys,json; print(json.load(sys.stdin).get('lastTickAt', ''))")
```
hojinzs: **P3 — missing the failure fallbacks every other poll site in this script uses**

Under `set -euo pipefail` (line 2), this assignment aborts the whole run on any transient hiccup: if `curl` exits non-zero the pipeline fails via `pipefail`, and if the body is empty or truncated `json.load` raises and `python3` exits 1. Either way the script dies with a bare non-zero status and no `fail` diagnostic, so the operator sees a crash rather than "reconciliation did not complete".

Every other polling site guards against exactly this — e.g. line 185: `orch_curl -s … 2>/dev/null || echo '{}'`, and each extractor with `2>/dev/null || echo "?"`. The startup wait at line 60 uses `-sf … || true`.

Same applies to lines 131-132 and 143 in the new block. Suggested for this line:

```bash
INITIAL_LAST_TICK=$(orch_curl -s http://localhost:4680/api/v1/state 2>/dev/null | python3 -c "import sys,json; print(json.load(sys.stdin).get('lastTickAt') or '')" 2>/dev/null || echo '')
```

Note `.get('lastTickAt', '')` also returns `None` (printed as the literal `None`) when the key is present but null, which then compares unequal to a later real timestamp — `or ''` avoids that.

---

AGENT_TEST.md
```diff
@@ -193,6 +193,7 @@ Control worker behavior with the `STUB_SCENARIO` environment variable:
...
+| Adapter-owned dispatch eligibility                                           | `packages/orchestrator/src/dispatch.test.ts` verifies the scheduler suppresses `dispatchable: false` issues, while each tracker adapter's suite verifies its own provider-specific derivation.                                                                                                                                                                                                                                                                                                                                                                                                                  | `./e2e/run-e2e.sh non-dispatchable 15` injects a file-tracker issue with `dispatchable: false`, confirms no worker or `run-dispatched` event exists, and verifies `repo explain` retains the adapter reason.                                                                                                                                      |
```
hojinzs: **nit — documented timeout is not the one that was verified**

This row and `e2e/scenarios/18-dispatchable-eligibility.md:31` both document `./e2e/run-e2e.sh non-dispatchable 15`, but the PR description's evidence line is `./e2e/run-e2e.sh non-dispatchable 30`. The second argument doubles as the tick-wait iteration count (`for i in $(seq 1 "$TIMEOUT")`), so `15` is a materially different — and unverified — budget. It gets tighter still if the wait moves to two ticks per the comment above.

Suggest aligning both to `30`, which also matches the runner's documented default.

The rest of this row checks out: `packages/orchestrator/src/dispatch.test.ts:677` ("does not dispatch issues the tracker marks non-dispatchable") exists and passes.

---

hojinzs (changes requested):
## Review — round 3

Scope and shape are good: 84/15 across 8 files, no code changes outside docs and the E2E harness, and no over-engineering or unrequested scope. Two things need a fix before this can land, both concentrated in the parts this PR exists to get right — documentation accuracy and the E2E's proof strength.

### Requirement coverage vs #689

All four scope bullets are covered:

| #689 scope | Status |
| --- | --- |
| `docs/architecture.md` §3/§5 adapter/scheduler boundary | ✅ §3 bullet + §5 lines |
| `README.md` `--assigned-only` as `dispatchable` input | ✅ both occurrences (L398, L592) |
| `AGENT_TEST.md` table + `e2e/fixtures/` Docker E2E | ✅ row, fixture, runner branch, scenario doc |
| `dispatchable` derivation profile in each tracker README | ✅ linear updated, file added; github already landed in #697 |

The `packages/tracker-file/README.md` claim that omitted `dispatchable` defaults to `true` is accurate (`file-tracker-adapter.ts:56`). No changeset is required — docs plus E2E harness, no shipped `@gh-symphony/cli` behavior.

### Findings

| Sev | Where | Issue |
| --- | --- | --- |
| P2 | `docs/architecture.md:70` | Linear filters pickup-label-ineligible issues out; it does not retain them as `dispatchable: false`. The "same normalized contract" claim is half wrong, and the observability difference vs GitHub is the part worth documenting. |
| P2 | `e2e/run-e2e.sh:130-137` | The tick-wait can break on a tick that started before the fixture copy, letting the scenario report `PASSED` without ever exercising the gate. |
| P3 | `e2e/run-e2e.sh:112`, 131-132, 143 | New block omits the `2>/dev/null \|\| echo …` fallbacks every other poll site uses; a transient hiccup aborts the run with no diagnostic. |
| nit | `AGENT_TEST.md:196`, `e2e/scenarios/18-…:31` | Documents timeout `15`; the evidence used `30`. |

Both P2s correspond to the two unresolved Codex threads on this head. I verified each against the source rather than taking them at face value — both hold, and details are in the inline comments. One correction worth flagging: `lastTickAt` is the tick's **start** timestamp (`service.ts:1096` → `1958`), not its completion time. That does not dissolve the race, and it means Codex's suggested timestamp-comparison fix is actually valid, so please don't dismiss it on the assumption that the value is a completion time.

Previous-round follow-through was good — both replies on the outdated threads point at `f726fc64` and the fixes are really there.

### Verification performed

Docker E2E could **not** be run — no Docker daemon in this environment (`/var/run/docker.sock` absent), so `./e2e/run-e2e.sh non-dispatchable 30` is unverified here and the PR's own E2E evidence stands unconfirmed. Everything else was run against `129d8d09`:

- `pnpm lint` — passed
- `pnpm typecheck` — passed
- `pnpm build` — passed
- `pnpm test` — passed (1,542 tests across 15 packages)
- `bash -n e2e/run-e2e.sh` — syntax OK
- `repo explain --json` shape cross-checked against `packages/orchestrator/src/explain.ts:131-136, 181-193`: `summary`, `checks[].id == "tracker_dispatchability"`, `status: "block"`, and `details.dispatchReason` all match the script's assertions exactly.
- `pnpm format` fails, but `AGENT_TEST.md` is already unformatted on `main` — pre-existing, not this PR's.

Requesting changes on the two P2s. Fix those and this is an approve — the P3 and nit are cheap to fold in at the same time, and I'd rather not spend another round on them separately.

---

docs/architecture.md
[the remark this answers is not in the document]
moncher-dev: Corrected in 42ab9577: Linear pickup labels filter candidate listing; the architecture now documents the observability difference and explicit repository divergence.

moncher-dev: Corrected in 42ab9577: Linear pickup-label filtering is now distinguished from assignment-derived dispatchability, with the upstream divergence stated explicitly.

e2e/run-e2e.sh
```diff
@@ -120,6 +126,49 @@ fi
 
 log "Issues injected; refresh trigger accepted (202). Falling back to polling until dispatch is observed"
 
+if [ "$SCENARIO" = "non-dispatchable" ]; then
+  for i in $(seq 1 "$TIMEOUT"); do
+    STATUS_JSON=$(orch_curl -s http://localhost:4680/api/v1/state)
+    LAST_TICK=$(echo "$STATUS_JSON" | python3 -c "import sys,json; print(json.load(sys.stdin).get('lastTickAt', ''))")
+    if [ -n "$LAST_TICK" ] && [ "$LAST_TICK" != "$INITIAL_LAST_TICK" ]; then
+      break
```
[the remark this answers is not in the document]
moncher-dev: Resolved in 42ab9577: the scenario requires two new tick start timestamps, so the second cannot have read the pre-injection fixture.

e2e/run-e2e.sh
moncher-dev: Resolved in 42ab9577: the two-tick proof accounts for lastTickAt being a tick-start timestamp.

e2e/run-e2e.sh
moncher-dev: Resolved in 42ab9577: all non-dispatchable polling and extraction paths now use resilient fallbacks.

AGENT_TEST.md
moncher-dev: Resolved in 42ab9577: both the TC table and scenario document the verified 30-iteration command.

e2e/run-e2e.sh
```diff
@@ -118,7 +124,60 @@ if [ "$REFRESH_STATUS" != "202" ]; then
...
+        break
+      fi
+    fi
+    if [ "$i" = "$TIMEOUT" ]; then
+      fail "Two post-injection reconciliation ticks did not complete before timeout"
+      exit 1
+    fi
+    sleep 1
+  done
+  ACTIVE=$(echo "$STATUS_JSON" | python3 -c "import sys,json; print(json.load(sys.stdin)['summary']['activeRuns'])" 2>/dev/null || echo '?')
```
hojinzs, line 152: **nit (non-blocking — do not spend a round on this)** — the `|| echo '?'` fallback added for the P3 makes a *read* failure report as a *dispatch* failure.

If the state read or the JSON parse fails, `ACTIVE` becomes `?`, which is `!= "0"`, so the script prints `Non-dispatchable issue started a worker` — the one message an operator would act on hardest, for a condition that did not happen. The existing site at line 197 has the same shape, but there the surrounding loop has already proven the snapshot is readable, so it is much less likely to fire.

If you touch this file again for another reason, distinguishing the two is a two-line change:

```bash
  if [ "$ACTIVE" = "?" ]; then
    fail "Could not read activeRuns from the state snapshot"
    exit 1
  fi
```

Not worth a push on its own — the fix is only cosmetic to the failure path, and the tick-wait above already guards the read.

---

hojinzs (approved):
## Review — round 4: Approve

Both round-3 P2s are genuinely fixed, and I verified each against the source rather than accepting the reply comments. The P3 and the nit are folded in as well. Approving.

### Round-3 follow-through

| Round-3 finding | Fix | Verified |
| --- | --- | --- |
| P2 `docs/architecture.md` — Linear pickup labels are not a `dispatchable` record | `42ab9577` | ✅ §5 now says pickup labels *filter the candidate list*, that label-ineligible Linear items are not retained for explain, and names the divergence from the upstream scheduler-owned label boundary. Matches `tracker-linear/src/orchestrator-adapter.ts:204` (`filterIssuesByPickupLabels`, removes) vs `tracker-github/src/orchestrator-adapter.ts:314` (`resolvePickupLabelDispatchReason`, retains). `packages/tracker-linear/README.md` carries the same qualification. |
| P2 `e2e/run-e2e.sh` — tick-wait could break on a pre-injection tick | `42ab9577` | ✅ The two-tick proof is sound. `reconcileProject` runs strictly serially from the `while (this.running)` loop at `service.ts:885`, awaited per iteration, so at most **one** in-flight tick can have read the pre-`cp` fixture. The second distinct `lastTickAt` therefore belongs to a tick that started after the first one finished — i.e. after the fixture copy. Missing an intermediate tick between 1s polls only over-waits, which is the safe direction. |
| P3 — missing poll fallbacks | `42ab9577` | ✅ Line 112 and both extractors in the loop now use `2>/dev/null … \|\| echo`, with `or ''` rather than `.get(k, '')`. One cosmetic consequence noted inline, non-blocking. |
| nit — documented timeout `15` vs verified `30` | `42ab9577` | ✅ `AGENT_TEST.md:197` and `e2e/scenarios/18-…:31` both document `30`. |

### Requirement coverage vs #689

All four scope bullets remain covered, and the completion criterion ("Docker E2E 통과, 문서 병합") is met on the author's side. No over-engineering and no unrequested scope: 102/19 across 8 files, all docs plus the E2E harness. No changeset is required — nothing in the shipped `@gh-symphony/cli` behavior changed.

### Verification performed

Docker E2E still could **not** be run here (`/var/run/docker.sock` absent), so `./e2e/run-e2e.sh non-dispatchable 30` remains unverified in this environment. Instead of leaving the new assertions on code reading alone, I reproduced them locally against the real built CLI:

- Built the workspace, seeded a temp repo from `e2e/seed/WORKFLOW.md`, ran `repo init` with `GH_SYMPHONY_FILE_TRACKER_ISSUES_PATH` pointing at `e2e/fixtures/non-dispatchable.json`, then ran `repo explain test-owner/test-repo#1 --json`. The script's exact python assertion block passes verbatim: `dispatchable: false`, `summary == "Not dispatchable: fixture eligibility gate: assigned to another agent"`, `tracker_dispatchability.status == "block"`, and `details.dispatchReason` matching. The `...entry` spread in `file-tracker-adapter.ts:53` is what carries the new `dispatchReason` through.
- `repo start --once` against the same fixture: no worker, no `events.ndjson`, `activeRuns: 0`. Worth noting the snapshot also shows `suppressed: 0` — confirming the scenario has no positive signal to assert on and the tick-wait really is its only proof that the gate ran. That is exactly why the two-tick fix mattered.
- `pnpm lint`, `pnpm typecheck`, `pnpm build`, `pnpm test` — all pass (1,780 tests, 15 packages).
- `bash -n e2e/run-e2e.sh` — OK. `npx prettier --check` on all 8 changed files — clean, so the AGENT_TEST.md table realignment fixed the pre-existing formatting drift rather than adding to it.
- `dispatch.test.ts -t "non-dispatchable"` — passes, backing the new AGENT_TEST row.
- CI on `42ab957`: Test and Container Smoke both green.

One optional wording note I am **not** asking you to change: `docs/architecture.md:57` says the scheduler "applies only the normalized `dispatchable` gate", while `service.ts:1350` also excludes terminal-suppressed identifiers. The following bullet documents that separately and the sentence's own next clause scopes "only" to provider rules, so it reads correctly in context.

---

Issue #721, opened 2026-08-28
feat(orchestrator): apply issueRoutable to candidate dispatch, retry fire, and reconciliation (active-but-unroutable terminates without cleanup); explain reasons


hojinzs:
Parent: #664 (Epic #651)
리포트 항목: **A3, A10, A18** — `docs/reports/2026-08-28-upstream-spec-drift-research.md`

#664 분할 2/3 — orchestrator.

## 범위
- 후보 적격성: `issueRoutable` 실패 시 dispatch 제외 + `repo explain`에 `not routable: missing required labels [...]` 표시
- retry-fire(#653 경로): active but not routable → claim 해제(§8.4)
- reconciliation(§8.5): running 이슈가 active but not routable(라벨 제거/할당 해제) → 워커 종료, **워크스페이스 정리 없음**, `canceled_by_reconciliation` 사유에 routable 실패 기록
- 스냅샷/이벤트에 routable 사유 노출(선택)

## 완료 기준
- §17.4 "Required-label filtering is case-insensitive and applies after adapter normalization" 테스트
- run 중 라벨 제거 → 다음 tick 종료(cleanup 없음) 테스트; explain 테스트

Blocked by: #720

hojinzs:
**Backlog → Ready (프로젝트 상태 관리)** — 블로커 해소(#720/#730/#731 완료), In review 3(<5), WIP 캡(8) 내 픽업. main의 `resolveProviderEnvironmentValues`(#728)·core `normalization.ts`(#729)는 재사용만, 제거/재구현 금지.

moncher-dev:
## Workpad — hojinzs/github-symphony#721 — Cycle 1

**Type:** fresh pickup
**Branch:** feat/721-routable-dispatch
**Draft PR:** https://github.com/hojinzs/github-symphony/pull/755
**Cycle opened:** 2026-08-29T08:57:58Z · **Trigger:** Ready pickup after #720 dependency merged

### Status Transitions



- 2026-08-29T09:14:45Z · `Ready` → `In progress` · implementation branch and Draft PR created (cycle 1 open)
- 2026-08-29T09:16:49Z · `In progress` → `In review` · routability handling and all validation gates passed (cycle 1 close)

### Plan

- [x] 1. Apply routability filtering to candidate dispatch, retry fire, and active-run reconciliation; add targeted coverage
- [x] 2. Wrap-up: re-verify the original issue · pass the Completion Bar · changeset · PR ready · transition to In review

[...]

PR #714, opened 2026-08-28
refactor(trackers): derive blocker eligibility

moncher-dev:
## TL;DR

Move blocker eligibility out of orchestration core and make tracker adapters the single source of `dispatchable` and `dispatchReason`.

## Change-point diagram

```text
Selected WORKFLOW.md tracker policy
          │
          ▼
GitHub / Linear adapters ──► dispatchable + dispatchReason ──► scheduler / retry / repo explain
                                           │
                                           └── blocked_by remains best-effort metadata
```

## Start here

- `packages/tracker-github/src/adapter.ts`
- `packages/tracker-github/src/orchestrator-adapter.ts`
- `packages/orchestrator/src/service.ts`
- `packages/orchestrator/src/explain.ts`

## Changes

- GitHub and Linear adapters derive blocker eligibility from tracker-owned policy; core lifecycle no longer owns `blockerCheckStates`.
- GitHub applies the same derivation to candidate and by-ID reads, so retry reconciliation cannot restart a blocked issue.
- GitHub resolves blockers that are source-closed or terminal in the Project workflow state.
- Scheduler, retry, and `repo explain` consume adapter-derived `dispatchable`; obsolete blocker context was removed.
- `tracker.provider` owns the configuration, with the flat key retained as a compatibility alias.
- Updated configuration, adapter profiles, unit coverage, and TC-18 ownership documentation.

## Risks & rollback

Eligibility now has one provider-owned source of truth. Reverting this PR restores the former core-owned blocker gate.

## Changed files

- Core workflow tracker policy and parser coverage
- GitHub and Linear adapters
- Orchestrator scheduler, retry, and explain plumbing
- CLI explain integration
- Configuration and Docker E2E scenario documentation

## Validation

- `pnpm lint` — pass
- `pnpm test` — pass in CI (1,777 tests)
- `pnpm typecheck` — pass
- `pnpm build` — pass
- `pnpm --filter @gh-symphony/tracker-github test -- --run src/tracker-github.test.ts` — pass (105 tests)
- `pnpm --filter @gh-symphony/orchestrator test -- --run src/dispatch.test.ts` — pass (37 tests)
- Docker E2E TC-18 — pass: the blocked file-tracker record had no `run-dispatched` event while its unblocked dependency dispatched. This correction preserves that fixture/scheduler contract.
- PR checks (`Test`, `Container Smoke`) — pass on `496f7ce4`.

## Post-merge / human validation

[...]

Issue #664, opened 2026-08-27
feat(core,orchestrator,worker): tracker.required_labels with spec semantics and a per-turn routable check


moncher-dev:
Epic: #651
리포트 항목: **A3, A10, A18** — `docs/reports/2026-08-28-upstream-spec-drift-research.md` (PR #650)

## 문제

새 스펙 §5.3.1/§8.2: `tracker.required_labels`는 **ALL-of**, 대소문자·공백 무시, 빈 라벨은 아무것도 매칭하지 않으며, 오케스트레이터가 정규화 후 적용하고 continuation 시에도 `issue_routable`(= dispatchable && labels)로 재검사한다. 현재의 `pickup_labels.include`는 ANY-of, 설정 측 대소문자 구분, 어댑터 내부 적용이고 per-turn 검사는 상태만 본다.

## 현재 상태 (증거)

- `packages/core/src/workflow/pickup-labels.ts:30-36` (ANY-of, `Set.has`); 어댑터 내부 적용 `tracker-github/src/orchestrator-adapter.ts:21-24`
- per-turn 검사 상태만: `packages/worker/src/turn-lease.ts:77-78`
- repo init은 linear/file에만 pickup 적용(`repo-runtime.ts:105-111`), standalone은 전부(`project.ts:311-314`); `README.md:397`은 셋 다라고 기술
- `README.md:593` "라벨 변경은 실행 중 워커를 중단하지 않음" — 실제로는 reconciliation이 cleanup 없이 종료함(`service.ts:1797-1866`)

## 할 일

- core: `required_labels` 파싱(정규화), `issueRoutable(issue, lifecycle)` helper
- 오케스트레이터: 후보 적격성 + retry-fire + reconciliation("active but not routable → terminate w/o cleanup", §8.5)에 적용
- 워커 turn-lease/`tracker-state` 읽기에서 routable 재검사(S10의 전체 스냅샷 필요)
- `pickup_labels.include/exclude`는 GitHub/Linear 확장으로 유지하고 문서에 관계 명시(`include` = ANY-of 사전 필터, `required_labels` = ALL-of 게이트)
- README 397/593 수정

## 완료 기준

- §17.4 "Required-label filtering is case-insensitive and applies after adapter normalization" 테스트
- run 중 라벨 제거 → 다음 turn에서 종료(cleanup 없음) 테스트

Blocked by: #661, #662

hojinzs:
[...]

PR #724, opened 2026-08-28
feat(core): add required label routing

### LINKED ISSUE #720: feat(core): tracker.required_labels parsing and an issueRoutable() helper; lowercase pickup_labels config
Parent: #664 (Epic #651)
리포트 항목: **A3, A10, A18** — `docs/reports/2026-08-28-upstream-spec-drift-research.md`

#664 triage(범위 과대) 분할 1/3 — **core만**.

## 범위
- `packages/core/src/workflow/parser.ts`/`config.ts`: `tracker.required_labels`(list of strings, 기본 `[]`) 파싱 — trim+lowercase 정규화, 빈 문자열은 "아무 이슈도 매칭하지 않음"으로 보존(스펙 §5.3.1)
- `packages/core/src/workflow/pickup-labels.ts`: 설정 측 라벨도 lowercase/trim (현재 issue 라벨만 lowercase라 `include: [Ready]`가 매칭 안 됨)
- 신규 helper `issueRoutable(issue, lifecycle): { routable: boolean; reason?: string }` = `dispatchable === true` && `required_labels` 전부 포함 (스펙 §8.2)
- `WorkflowLifecycleConfig`/스냅샷 타입에 `requiredLabels` 노출

## 완료 기준
- 파서 테스트(정규화, 빈 라벨, 쉼표 문자열 거부), `issueRoutable` 단위테스트(대소문자·공백, dispatchable=false 우선)
- 기존 WORKFLOW.md 회귀 없음(`required_labels` 미지정 = `[]`)

moncher-dev:
## TL;DR

Adds core support for `tracker.required_labels`, normalizes configured pickup labels, and exposes `issueRoutable()` for dispatchability-first routing.

## Change-point diagram

```text
WORKFLOW.md tracker.required_labels
  -> parser / lifecycle snapshot
  -> normalized requiredLabels
  -> issueRoutable(issue, lifecycle)
       dispatchable === true
       && every required label is present
```

## Start here

- `packages/core/src/workflow/parser.ts` parses and validates the new configuration.
- `packages/core/src/workflow/issue-routable.ts` applies the policy.
- `packages/core/src/workflow-loader.test.ts` and `issue-routable.test.ts` cover normalization, blank labels, comma rejection, and dispatchability precedence.

## Changes

- Preserve blank required labels so they intentionally match no issue.
- Trim and lowercase configured pickup labels before matching.
- Expose `requiredLabels` through lifecycle configuration and snapshots.
- Restore the current `docs/configuration.md` from main and add only the required-label documentation.

## Evidence

- `pnpm install --frozen-lockfile`
- `pnpm lint`
- `pnpm test`
- `pnpm typecheck`
- `pnpm build`

## Risks & rollback

- Blank `required_labels` entries remain deliberately unsatisfiable per Symphony §5.3.1.
- Roll back the PR to remove the new optional routing gate; workflows without `required_labels` continue to use `[]`.

## Changed files

- Core workflow parser, lifecycle config, routing helper, exports, and tests.
- CLI workflow snapshot coverage and configuration reference.
- CLI minor changeset.

## Issues — Closed #720

Closes #720

- [ ] Confirm a workflow with mixed-case required labels routes only matching issues.
- [ ] Confirm a blank configured required label routes no issue.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -75,7 +75,8 @@ strict parser as workflow loading. Failures include a stable error code; the
 | `hooks.timeout_ms`  | A positive integer when provided                                                                                           | `workflow_validation_error` at `hooks.timeout_ms`                             |
 | `agent.max_turns`   | A positive integer when provided                                                                                           | `workflow_validation_error` at `agent.max_turns`                              |
 | `codex.command`     | A non-empty string when provided                                                                                           | `workflow_validation_error` at `codex.command`                                |
-| `tracker.kind`      | One of `github-project`, `linear`, or `file`                                                                               | `workflow_validation_error` at `tracker.kind`                                 |
+| `tracker.kind`            | One of `github-project`, `linear`, or `file`                                                                         | `workflow_validation_error` at `tracker.kind`                           |
+| `tracker.required_labels` | An array of strings; comma-separated strings are rejected                                                            | `workflow_validation_error` at `tracker.required_labels`                |
 
 Omitted optional values retain their documented defaults. Per-state concurrency
 maps remain supported through `agent.max_concurrent_agents_by_state` and their
@@ -89,8 +90,10 @@ Core normalizes workflow pickup labels and tracker-provided labels before label
 matching: it trims surrounding whitespace, lowercases values, drops blanks, and
 deduplicates while keeping the first occurrence. Thus `" Agent "`, `"agent"`,
 and `"AGENT"` are the same label. The same `normalizeLabels` helper is the
-shared implementation for current pickup-label handling and future
-`tracker.required_labels` configuration support.
+shared implementation for pickup-label handling. `tracker.required_labels`
+uses a deliberately separate blank-preserving normalization path: it trims and
+lowercases values, but a configured blank remains and therefore matches no
+issue, as required by Symphony specification §5.3.1.
 
 `parseTrackerTimestamp` accepts RFC 3339 timestamps, including lowercase `t` /
 `z` designators and leap seconds, and returns a canonical ISO 8601 UTC string;
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -68,15 +68,15 @@ diverge from the upstream shell-command hook model.
 strict parser as workflow loading. Failures include a stable error code; the
 `workflow validate --json` error also exposes its field path separately.
 
-| Rule                | Required value                                                                                                             | Error code/path                                                               |
-| ------------------- | -------------------------------------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------- |
-| Front matter syntax | A valid YAML mapping                                                                                                       | `workflow_parse_error` or `workflow_front_matter_not_a_map` at `front_matter` |
-| Numeric fields      | YAML integers; quoted numeric strings and fractions are rejected, including priority maps and per-state concurrency values | `workflow_validation_error` at the field path                                 |
-| `hooks.timeout_ms`  | A positive integer when provided                                                                                           | `workflow_validation_error` at `hooks.timeout_ms`                             |
-| `agent.max_turns`   | A positive integer when provided                                                                                           | `workflow_validation_error` at `agent.max_turns`                              |
-| `codex.command`     | A non-empty string when provided                                                                                           | `workflow_validation_error` at `codex.command`                                |
-| `tracker.kind`            | One of `github-project`, `linear`, or `file`                                                                         | `workflow_validation_error` at `tracker.kind`                           |
-| `tracker.required_labels` | An array of strings; comma-separated strings are rejected                                                            | `workflow_validation_error` at `tracker.required_labels`                |
+| Rule                      | Required value                                                                                                             | Error code/path                                                               |
+| ------------------------- | -------------------------------------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------- |
+| Front matter syntax       | A valid YAML mapping                                                                                                       | `workflow_parse_error` or `workflow_front_matter_not_a_map` at `front_matter` |
+| Numeric fields            | YAML integers; quoted numeric strings and fractions are rejected, including priority maps and per-state concurrency values | `workflow_validation_error` at the field path                                 |
+| `hooks.timeout_ms`        | A positive integer when provided                                                                                           | `workflow_validation_error` at `hooks.timeout_ms`                             |
+| `agent.max_turns`         | A positive integer when provided                                                                                           | `workflow_validation_error` at `agent.max_turns`                              |
+| `codex.command`           | A non-empty string when provided                                                                                           | `workflow_validation_error` at `codex.command`                                |
+| `tracker.kind`            | One of `github-project`, `linear`, or `file`                                                                               | `workflow_validation_error` at `tracker.kind`                                 |
+| `tracker.required_labels` | An array of strings; comma-separated strings are rejected                                                                  | `workflow_validation_error` at `tracker.required_labels`                      |
 
 Omitted optional values retain their documented defaults. Per-state concurrency
 maps remain supported through `agent.max_concurrent_agents_by_state` and their
@@ -84,16 +84,18 @@ values must be positive YAML integers. This repository intentionally rejects an
 invalid override rather than ignoring it, so invalid configuration cannot silently
 disable dispatch for a state.
 
+`tracker.required_labels` defaults to `[]`. Labels are compared after trimming
+and lowercasing; every configured label must be present before an issue can be
+routed. A blank configured label is preserved and therefore matches no issue.
+
 ## Label and Timestamp Normalization
 
 Core normalizes workflow pickup labels and tracker-provided labels before label
 matching: it trims surrounding whitespace, lowercases values, drops blanks, and
 deduplicates while keeping the first occurrence. Thus `" Agent "`, `"agent"`,
-and `"AGENT"` are the same label. The same `normalizeLabels` helper is the
-shared implementation for pickup-label handling. `tracker.required_labels`
-uses a deliberately separate blank-preserving normalization path: it trims and
-lowercases values, but a configured blank remains and therefore matches no
-issue, as required by Symphony specification §5.3.1.
+and `"AGENT"` are the same label. `tracker.required_labels` deliberately uses
+a separate normalization path: it retains blank configured labels so they match
+no issue, as required by Symphony specification §5.3.1.
 
 `parseTrackerTimestamp` accepts RFC 3339 timestamps, including lowercase `t` /
 `z` designators and leap seconds, and returns a canonical ISO 8601 UTC string;
--- a/packages/core/src/workflow/issue-routable.ts
+++ b/packages/core/src/workflow/issue-routable.ts
@@ -1,0 +1,34 @@
+import type { TrackedIssue } from "../contracts/tracker-adapter.js";
+import type { WorkflowLifecycleConfig } from "./lifecycle.js";
+import { normalizeLabels } from "./normalization.js";
+
+/** Resolves whether an issue is eligible for workflow routing. */
+export function issueRoutable(
+  issue: Pick<TrackedIssue, "dispatchable" | "dispatchReason" | "labels">,
+  lifecycle: Pick<WorkflowLifecycleConfig, "requiredLabels">
+): { routable: boolean; reason?: string } {
+  if (!issue.dispatchable) {
+    return {
+      routable: false,
+      reason: issue.dispatchReason ?? "Issue is not dispatchable.",
+    };
+  }
+
+  const labels = new Set(normalizeLabels(issue.labels));
+  const missingLabels = (lifecycle.requiredLabels ?? []).filter((label) => {
+    const normalized = label.trim().toLowerCase();
+    // Preserve the config-side blank invariant from §5.3.1 even though issue
+    // labels are normalized with the shared helper, which drops blanks.
+    return normalized === "" || !labels.has(normalized);
+  });
+  if (missingLabels.length > 0) {
+    return {
+      routable: false,
+      reason: `Issue is missing required labels (${missingLabels
+        .map((label) => JSON.stringify(label))
+        .join(", ")}).`,
+    };
+  }
+
+  return { routable: true };
+}
--- a/packages/core/src/workflow/parser.ts
+++ b/packages/core/src/workflow/parser.ts
@@ -186,8 +186,7 @@ function parseWorkflowMarkdownInternal(
       provider,
       explicitProviderKeys,
       "blocker_check_states"
-    ) ??
-    (activeStates[0] ? [activeStates[0]] : []);
+    ) ?? (activeStates[0] ? [activeStates[0]] : []);
   const planningStates =
     readNormalizedStringList(
       tracker,
@@ -607,7 +606,10 @@ function readRequiredLabelList(
   if (value === undefined || value === null) {
     return undefined;
   }
-  if (!Array.isArray(value) || value.some((entry) => typeof entry !== "string")) {
+  if (
+    !Array.isArray(value) ||
+    value.some((entry) => typeof entry !== "string")
+  ) {
     throw new WorkflowValidationError(
       "workflow_validation_error",
       "tracker.required_labels",
@@ -617,7 +619,7 @@ function readRequiredLabelList(
 
   // §5.3.1 requires blank configured labels to remain unsatisfied. The shared
   // normalizeLabels helper drops blanks, so it is intentionally not used here.
-  return value.map((label) => label.trim().toLowerCase());
+  return (value as string[]).map((label) => label.trim().toLowerCase());
 }
 
 function readPriorityConfig(
--- a/packages/core/src/workflow/parser.ts
+++ b/packages/core/src/workflow/parser.ts
@@ -197,6 +197,8 @@ function parseWorkflowMarkdownInternal(
     ) ??
     defaultLifecycle?.planningStates ??
     DEFAULT_WORKFLOW_TRACKER.planningStates;
+  const requiredLabels =
+    readRequiredLabelList(tracker) ?? DEFAULT_WORKFLOW_TRACKER.requiredLabels;
   const stateFieldName =
     readNormalizedOptionalString(
       tracker,
@@ -289,6 +291,7 @@ function parseWorkflowMarkdownInternal(
         provider,
         explicitProviderKeys
       ),
+      requiredLabels,
       activeStates,
       terminalStates,
       projectId: readNormalizedOptionalString(
@@ -373,6 +376,7 @@ function parseWorkflowMarkdownInternal(
       activeStates,
       terminalStates,
       planningStates,
+      requiredLabels,
     },
     format: "front-matter",
     githubProjectId: readNormalizedOptionalString(
@@ -594,6 +598,26 @@ function readPickupLabelsConfig(
     include: normalizeLabels(readStringList(input, "include") ?? []),
     exclude: normalizeLabels(readStringList(input, "exclude") ?? []),
   };
+}
+
+function readRequiredLabelList(
+  tracker: Record<string, WorkflowFrontMatterNode>
+): string[] | undefined {
+  const value = tracker.required_labels;
+  if (value === undefined || value === null) {
+    return undefined;
+  }
+  if (!Array.isArray(value) || value.some((entry) => typeof entry !== "string")) {
+    throw new WorkflowValidationError(
+      "workflow_validation_error",
+      "tracker.required_labels",
+      'Workflow front matter field "tracker.required_labels" must be an array of strings.'
+    );
+  }
+
+  // §5.3.1 requires blank configured labels to remain unsatisfied. The shared
+  // normalizeLabels helper drops blanks, so it is intentionally not used here.
+  return value.map((label) => label.trim().toLowerCase());
 }
 
 function readPriorityConfig(
--- a/.changeset/tidy-required-label-routing.md
+++ b/.changeset/tidy-required-label-routing.md
@@ -1,0 +1,5 @@
+---
+"@gh-symphony/cli": minor
+---
+
+Add normalized workflow required-label routing for #720.
--- a/packages/cli/src/commands/workflow.test.ts
+++ b/packages/cli/src/commands/workflow.test.ts
@@ -260,6 +260,44 @@ Prompt {{ issue.identifier }}
     );
   });
 
+  it("includes normalized required labels in JSON validation output", async () => {
+    const root = await mkdtemp(
+      join(tmpdir(), "workflow-validate-required-labels-")
+    );
+    const workflowPath = join(root, "WORKFLOW.md");
+    const stdout = captureWrites(process.stdout);
+
+    await writeFile(
+      workflowPath,
+      `---
+tracker:
+  kind: github-project
+  required_labels: [" Ready ", Agent]
+codex:
+  command: codex app-server
+---
+Prompt {{ issue.identifier }}
+`,
+      "utf8"
+    );
+
+    try {
+      await workflowCommand(["validate", "--file", workflowPath], {
+        configDir: root,
+        verbose: false,
+        json: true,
+        noColor: false,
+      });
+    } finally {
+      stdout.restore();
+    }
+
+    const report = JSON.parse(stdout.output()) as {
+      summary: { requiredLabels: string[] };
+    };
+    expect(report.summary.requiredLabels).toEqual(["ready", "agent"]);
+  });
+
   it("previews a workflow with the built-in sample issue", async () => {
     const root = await mkdtemp(join(tmpdir(), "workflow-preview-"));
     const workflowPath = join(root, "WORKFLOW.md");
--- a/packages/cli/src/commands/workflow.ts
+++ b/packages/cli/src/commands/workflow.ts
@@ -92,6 +92,7 @@ type WorkflowValidationReport = {
     terminalStates: string[];
     blockerCheckStates: string[];
     planningStates: string[];
+    requiredLabels: string[];
     pollingIntervalMs: number;
     workspaceRoot: string | null;
     agentCommand: string;
@@ -863,6 +864,7 @@ function validateWorkflow(
       terminalStates: workflow.lifecycle.terminalStates,
       blockerCheckStates: workflow.tracker.blockerCheckStates,
       planningStates: workflow.lifecycle.planningStates,
+      requiredLabels: workflow.lifecycle.requiredLabels ?? [],
       pollingIntervalMs: workflow.polling.intervalMs,
       workspaceRoot: workflow.workspace.root,
       agentCommand: workflow.agentCommand,
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -14,6 +14,7 @@ export * from "./workflow/render.js";
 export * from "./workflow/exit-classification.js";
 export * from "./workflow/issue-identity.js";
 export * from "./workflow/normalization.js";
+export * from "./workflow/issue-routable.js";
 export * from "./workflow/pickup-labels.js";
 export * from "./orchestration/index.js";
 export * from "./runtime/index.js";
--- a/packages/core/src/workflow-loader.test.ts
+++ b/packages/core/src/workflow-loader.test.ts
@@ -1013,6 +1013,57 @@ Prompt body.
     });
   });
 
+  it("normalizes required labels while preserving blank labels", () => {
+    const workflow = parseWorkflowMarkdown(`---
+tracker:
+  kind: github-project
+  required_labels:
+    - " Ready "
+    - ""
+  pickup_labels:
+    include:
+      - " Agent "
+codex:
+  command: codex app-server
+---
+Prompt body.
+`);
+
+    expect(workflow.tracker.requiredLabels).toEqual(["ready", ""]);
+    expect(workflow.lifecycle.requiredLabels).toEqual(["ready", ""]);
+    expect(workflow.tracker.pickupLabels).toEqual({
+      include: ["agent"],
+      exclude: [],
+    });
+  });
+
+  it("rejects comma-separated required labels with the field path", () => {
+    try {
+      parseWorkflowMarkdown(`---
+tracker:
+  kind: github-project
+  required_labels: ready, agent
+codex:
+  command: codex app-server
+---
+Prompt body.
+`);
+      throw new Error("Expected parsing to fail.");
+    } catch (error) {
+      expect(error).toMatchObject({
+        code: "workflow_validation_error",
+        path: "tracker.required_labels",
+      });
+    }
+  });
+
+  it("defaults omitted required labels to an empty list", () => {
+    const workflow = parseWorkflowMarkdown(SAMPLE_WORKFLOW);
+
+    expect(workflow.tracker.requiredLabels).toEqual([]);
+    expect(workflow.lifecycle.requiredLabels).toEqual([]);
+  });
+
   it("surfaces typed provider validation errors from the selected adapter", () => {
     const adapterError = new WorkflowValidationError(
       "workflow_validation_error",
--- a/packages/core/src/workflow-loader.test.ts
+++ b/packages/core/src/workflow-loader.test.ts
@@ -1047,7 +1047,7 @@ codex:
   command: codex app-server
 ---
 Prompt body.
-`);
+`)
     ).toThrow(
       expect.objectContaining({
         code: "workflow_validation_error",
--- a/packages/core/src/workflow-loader.test.ts
+++ b/packages/core/src/workflow-loader.test.ts
@@ -1038,7 +1038,7 @@ Prompt body.
   });
 
   it("rejects comma-separated required labels with the field path", () => {
-    try {
+    expect(() =>
       parseWorkflowMarkdown(`---
 tracker:
   kind: github-project
@@ -1048,13 +1048,12 @@ codex:
 ---
 Prompt body.
 `);
-      throw new Error("Expected parsing to fail.");
-    } catch (error) {
-      expect(error).toMatchObject({
+    ).toThrow(
+      expect.objectContaining({
         code: "workflow_validation_error",
         path: "tracker.required_labels",
-      });
-    }
+      })
+    );
   });
 
   it("defaults omitted required labels to an empty list", () => {
--- a/packages/core/src/workflow/config.ts
+++ b/packages/core/src/workflow/config.ts
@@ -18,6 +18,8 @@ export type WorkflowTrackerConfig = {
   apiKey: string | null;
   projectSlug: string | null;
   pickupLabels: WorkflowPickupLabelsConfig;
+  /** Core lifecycle routing constraint from tracker.required_labels. */
+  requiredLabels: string[];
   activeStates: string[];
   terminalStates: string[];
   projectId: string | null;
@@ -179,6 +181,7 @@ export const DEFAULT_WORKFLOW_TRACKER: WorkflowTrackerConfig = {
     include: [],
     exclude: [],
   },
+  requiredLabels: [],
   activeStates: [],
   terminalStates: [],
   projectId: null,
@@ -230,6 +233,7 @@ export const DEFAULT_WORKFLOW_DEFINITION: ParsedWorkflow = {
     activeStates: [],
     terminalStates: [],
     planningStates: [],
+    requiredLabels: [],
   },
   format: "default",
   githubProjectId: null,
--- a/packages/core/src/workflow/issue-routable.test.ts
+++ b/packages/core/src/workflow/issue-routable.test.ts
@@ -36,7 +36,7 @@ describe("issueRoutable", () => {
     ).toEqual({ routable: true });
   });
 
-  it("preserves empty required labels as unsatisfied", () => {
+  it("keeps empty required labels unsatisfied even with blank issue labels", () => {
     expect(
       issueRoutable(issue({ labels: ["ready", "  "] }), {
         ...DEFAULT_WORKFLOW_LIFECYCLE,
--- a/packages/core/src/workflow/issue-routable.test.ts
+++ b/packages/core/src/workflow/issue-routable.test.ts
@@ -1,0 +1,59 @@
+import { describe, expect, it } from "vitest";
+import type { TrackedIssue } from "../contracts/tracker-adapter.js";
+import { issueRoutable } from "./issue-routable.js";
+import { DEFAULT_WORKFLOW_LIFECYCLE } from "./lifecycle.js";
+
+function issue(input: Partial<TrackedIssue>): TrackedIssue {
+  return {
+    id: "1",
+    identifier: "acme/repo#1",
+    title: "Test issue",
+    description: null,
+    priority: null,
+    state: "Ready",
+    branchName: null,
+    url: null,
+    labels: [],
+    dispatchable: true,
+    assigneeId: null,
+    blockedBy: [],
+    createdAt: null,
+    updatedAt: null,
+    repository: { owner: "acme", name: "repo", cloneUrl: "" },
+    tracker: { adapter: "file", bindingId: "test" },
+    metadata: {},
+    ...input,
+  };
+}
+
+describe("issueRoutable", () => {
+  it("matches required labels case-insensitively after trimming", () => {
+    expect(
+      issueRoutable(issue({ labels: [" Ready ", "AGENT"] }), {
+        ...DEFAULT_WORKFLOW_LIFECYCLE,
+        requiredLabels: ["ready", " agent "],
+      })
+    ).toEqual({ routable: true });
+  });
+
+  it("preserves empty required labels as unsatisfied", () => {
+    expect(
+      issueRoutable(issue({ labels: ["ready", "  "] }), {
+        ...DEFAULT_WORKFLOW_LIFECYCLE,
+        requiredLabels: ["ready", ""],
+      })
+    ).toEqual({
+      routable: false,
+      reason: 'Issue is missing required labels ("").',
+    });
+  });
+
+  it("prioritizes a non-dispatchable issue over required-label checks", () => {
+    expect(
+      issueRoutable(
+        issue({ dispatchable: false, dispatchReason: "Archived issue." }),
+        { ...DEFAULT_WORKFLOW_LIFECYCLE, requiredLabels: ["ready"] }
+      )
+    ).toEqual({ routable: false, reason: "Archived issue." });
+  });
+});
--- a/packages/core/src/workflow/lifecycle.ts
+++ b/packages/core/src/workflow/lifecycle.ts
@@ -3,6 +3,8 @@ export type WorkflowLifecycleConfig = {
   activeStates: string[];
   terminalStates: string[];
   planningStates: string[];
+  /** Normalized required tracker labels; omitted legacy configs are unrestricted. */
+  requiredLabels?: string[];
 };
 
 export const DEFAULT_WORKFLOW_LIFECYCLE: WorkflowLifecycleConfig = {
@@ -10,6 +12,7 @@ export const DEFAULT_WORKFLOW_LIFECYCLE: WorkflowLifecycleConfig = {
   activeStates: ["Todo", "In Progress"],
   terminalStates: ["Done"],
   planningStates: [],
+  requiredLabels: [],
 };
 
 export function isStateActive(
--- a/packages/core/src/workflow/pickup-labels.test.ts
+++ b/packages/core/src/workflow/pickup-labels.test.ts
@@ -79,6 +79,6 @@ describe("resolvePickupLabelDispatchReason", () => {
         issue(["agent", "blocked"]),
         mixedCaseProject
       )
-    ).toBe('Issue has excluded pickup label "BLOCKED".');
+    ).toBe('Issue has excluded pickup label "blocked".');
   });
 });
--- a/packages/core/src/workflow/pickup-labels.test.ts
+++ b/packages/core/src/workflow/pickup-labels.test.ts
@@ -79,6 +79,6 @@ describe("resolvePickupLabelDispatchReason", () => {
         issue(["agent", "blocked"]),
         mixedCaseProject
       )
-    ).toBe('Issue has excluded pickup label "blocked".');
+    ).toBe('Issue has excluded pickup label "BLOCKED".');
   });
 });


packages/core/src/workflow/parser.ts
```diff
@@ -353,11 +357,33 @@ function readPickupLabelsConfig(
...
+): string[] | undefined {
+  const value = tracker.required_labels ?? tracker.requiredLabels;
+  if (value === undefined || value === null) {
+    return undefined;
+  }
+  if (
+    !Array.isArray(value) ||
+    value.some((label) => typeof label !== "string")
+  ) {
+    throw new Error(
```
hojinzs: **P2 — the emitted error path contradicts the docs row added in this same PR.**

`parseWorkflowMarkdown` catches this plain `Error` and rewrites it as `WorkflowValidationError("workflow_validation_error", "front_matter", …)`. Reproduced end-to-end against the built CLI on this head:

```
$ gh-symphony workflow validate --file WORKFLOW.md --json   # required_labels: ready, agent
{
  "error": {
    "code": "workflow_validation_error",
    "path": "front_matter",
    "message": "front_matter: Workflow front matter field \"tracker.required_labels\" must be an array of strings."
  }
}
```

`docs/configuration.md:48` (this PR) promises `workflow_validation_error` at `tracker.required_labels`, and the section three lines above it states that `workflow validate --json` "exposes its field path separately" — so automation keying on `error.path` gets the wrong field. Same result for a non-string array member (`required_labels: [1]`).

The sibling validators in this file that have their own docs row (`tracker.kind`, `hooks.timeout_ms`, `agent.max_turns`, `codex.command`) all throw `WorkflowValidationError` with the exact path; only this one does not. `readStringList` also throws a plain `Error`, but it has no per-field docs row making a path promise.

```suggestion
    throw new WorkflowValidationError(
      "workflow_validation_error",
      "tracker.required_labels",
      'Workflow front matter field "tracker.required_labels" must be an array of strings.'
    );
```

The existing `rejects comma-separated required labels` test asserts on the message only, so it keeps passing; please extend it to pin `code` and `path` too, since that pair is the contract the docs table sells.

---

packages/core/src/workflow/issue-routable.ts
```diff
@@ -0,0 +1,32 @@
...
+): { routable: boolean; reason?: string } {
+  if (!issue.dispatchable) {
+    return {
+      routable: false,
+      reason: issue.dispatchReason ?? "Issue is not dispatchable.",
+    };
+  }
+
+  const labels = new Set(issue.labels.map(normalizeLabel));
+  const missingLabels = (lifecycle.requiredLabels ?? []).filter(
```
hojinzs: **P2 — a blank required label is satisfiable, contradicting spec §5.3.1 and this PR's own docs.**

Because both sides go through the same `normalizeLabel`, a blank *issue* label also collapses to `""` and lands in the set, so the blank *configured* label matches. Verified on this head:

```ts
issueRoutable(
  { dispatchable: true, labels: ["ready", "  "] },
  { ...DEFAULT_WORKFLOW_LIFECYCLE, requiredLabels: ["ready", ""] }
)
// → { routable: true }
```

`docs/configuration.md:56-58` in this PR states "A blank configured label is preserved and therefore matches no issue", and the PR description repeats it as an intentional §5.3.1 behavior. The existing `preserves empty required labels as unsatisfied` test only covers `labels: ["ready"]`, so it passes without exercising this.

Impact is narrow — GitHub and Linear reject blank label names, so in practice it needs the file tracker, which does no label validation. But the fix is local and restores the invariant the docs assert:

```suggestion
  const labels = new Set(issue.labels.map(normalizeLabel));
  const missingLabels = (lifecycle.requiredLabels ?? []).filter((label) => {
    const normalized = normalizeLabel(label);
    return normalized === "" || !labels.has(normalized);
  });
```

Worth adding the `labels: ["ready", "  "]` case to `issue-routable.test.ts` alongside it.

---

packages/core/src/workflow/parser.ts
```diff
@@ -353,11 +357,33 @@ function readPickupLabelsConfig(
...
-    exclude: readStringList(input, "exclude") ?? [],
+    include: normalizeLabels(readStringList(input, "include") ?? []),
+    exclude: normalizeLabels(readStringList(input, "exclude") ?? []),
   };
 }
 
+function readRequiredLabelList(
+  tracker: Record<string, WorkflowFrontMatterNode>
+): string[] | undefined {
+  const value = tracker.required_labels ?? tracker.requiredLabels;
```
hojinzs: **P3 — undocumented `requiredLabels` camelCase alias, not requested by #720.**

Verified accepted on this head:

```ts
parseWorkflowMarkdown(`---
tracker:
  kind: github-project
  requiredLabels: [Ready]
codex: { command: codex app-server }
---
Body.`).tracker.requiredLabels
// → ["ready"]
```

#720 asks only for `tracker.required_labels`, `docs/configuration.md` documents only the snake_case key, and this file already *rejects* camelCase tracker aliases elsewhere with a dedicated test (`rejects Linear tracker alias %s` over `projectId` / `teamId`). `readPickupLabelsConfig` is the one precedent going the other way, so this isn't clear-cut — but a silently-accepted undocumented spelling for a routing gate is the kind of thing that later has to be supported forever.

Either drop `?? tracker.requiredLabels` and keep the front-matter surface snake_case only, or, if it's deliberate for symmetry with `pickup_labels`, add a test pinning the intent and a line in the docs so it isn't accidental API.

---

packages/core/src/workflow/pickup-labels.ts
```diff
@@ -41,7 +41,7 @@ export function resolvePickupLabelDispatchReason<T extends TrackedIssue>(
     return null;
   }
 
-  const labels = new Set(issue.labels);
+  const labels = new Set(issue.labels.map(normalizeLabel));
   const excludedLabel = [...exclude].find((label) => labels.has(label));
   if (excludedLabel) {
     return `Issue has excluded pickup label "${excludedLabel}".`;
```
hojinzs, line 48: **P3 — dispatch reasons now echo the normalized label, losing the operator's configured casing.**

`excludedLabel` and `include` are both post-`normalizeLabel` now, so a workflow configured with `exclude: ["Blocked"]` against a GitHub label named `Blocked` produces `Issue has excluded pickup label "blocked".` Your own new test locks this in (`" BLOCKED "` → `"blocked"`).

The matching change itself is right and is exactly what #720 asked for — this is only about the message, which surfaces in `repo explain` / status output where an operator maps it back to a real GitHub label. Keeping a raw copy alongside the normalized one for display would preserve that:

```ts
const includeRaw = readRawLabelList(config.include);
const excludedIndex = excludeRaw.findIndex((label) => labels.has(normalizeLabel(label)));
```

Non-blocking; fine to defer if you'd rather keep the diff tight.

---

packages/core/src/workflow/issue-routable.ts
```diff
@@ -0,0 +1,32 @@
...
+    (label) => !labels.has(normalizeLabel(label))
+  );
+  if (missingLabels.length > 0) {
+    return {
+      routable: false,
+      reason: `Issue is missing required labels (${missingLabels.join(", ")}).`,
```
hojinzs: **nit** — with a blank required label this renders `Issue is missing required labels ().`, and with several it's ambiguous where one label ends. Your own test asserts that exact empty-paren string. Quoting each entry reads better wherever this reason gets surfaced:

```suggestion
      reason: `Issue is missing required labels (${missingLabels
        .map((label) => JSON.stringify(label))
        .join(", ")}).`,
```

---

docs/configuration.md
```diff
@@ -45,13 +45,18 @@ strict parser as workflow loading. Failures include a stable error code; the
 | `agent.max_turns`   | A positive integer when provided                                                                                           | `workflow_validation_error` at `agent.max_turns`                              |
 | `codex.command`     | A non-empty string when provided                                                                                           | `workflow_validation_error` at `codex.command`                                |
 | `tracker.kind`      | One of `github-project`, `linear`, or `file`                                                                               | `workflow_validation_error` at `tracker.kind`                                 |
+| `tracker.required_labels` | An array of strings; comma-separated strings are rejected | `workflow_validation_error` at `tracker.required_labels` |
```
hojinzs: **nit** — the new row isn't column-padded, so `npx prettier --check docs/configuration.md` now flags this file and would reflow the whole table. (This file has other pre-existing drift further down at ~L253, and `pnpm format` is already red repo-wide on 87 files, so it isn't CI-blocking — just avoiding adding to it.)

```suggestion
| `tracker.required_labels` | An array of strings; comma-separated strings are rejected                                                                  | `workflow_validation_error` at `tracker.required_labels`                      |
```

Note the header/separator rows need widening to `| Rule                      |` / `| ------------------------- |` to match.

---

hojinzs (changes requested):
## Review — round 1/6

Solid, well-scoped change. The parsing, normalization, and precedence logic are right, and the test coverage maps cleanly onto #720's acceptance criteria. Two P2s block: both are cases where the code contradicts documentation added in this same PR, and both are small local fixes.

### Requirement coverage vs #720

| #720 요구사항 | 상태 |
| --- | --- |
| `required_labels` parsing, default `[]`, trim+lowercase | ✅ `parser.ts:139-140,365-381` |
| Blank label preserved as "matches nothing" (§5.3.1) | ⚠️ preserved through parsing, but **satisfiable at match time** — P2 below |
| `pickup-labels.ts` lowercases the config side | ✅ — and correctly normalizes the issue side too, which the issue text assumed was already done but wasn't (`issue.labels` was a raw `Set` before) |
| `issueRoutable(issue, lifecycle)`, dispatchable-first + ALL-of | ✅ `issue-routable.ts` |
| `requiredLabels` on `WorkflowLifecycleConfig` | ✅ `lifecycle.ts:8,17` |
| Parser tests: normalization / blank / comma rejection | ✅ 3 tests |
| `issueRoutable` tests: case+whitespace, `dispatchable=false` precedence | ✅ 3 tests |
| No regression when `required_labels` omitted | ✅ 107 core workflow tests pass |

No over-engineering found beyond the undocumented camelCase alias (P3). The `requiredLabels?:` optionality on `WorkflowLifecycleConfig` is a reasonable compat call — it fails open to "unrestricted", which is what "미지정 = `[]`" requires — and the `/** … */` note documents it.

### Re: the Codex P1 ("apply required-label routing in the orchestrator")

I disagree, and it should not gate this PR. #720 is explicitly "#664 triage(범위 과대) 분할 1/3 — **core만**", and orchestrator wiring at `service.ts` is 2/3. Adding it here would re-create the over-scoping that #664 was split to avoid. Please reply on that thread with that reasoning rather than expanding the diff.

One thing worth capturing in the follow-up issue, not here: `tracker-github/src/orchestrator-adapter.ts` `reviveIssue()` returns `labels: []`. Once `issueRoutable` gates continuation (spec §8.2 applies `required_labels` on continuation too), every revived GitHub issue will read as non-routable under any non-empty `required_labels` and recovery will silently stop.

### Findings

| | Location | Issue |
| --- | --- | --- |
| **P2** | `parser.ts:376` | Validation error path is `front_matter`, not `tracker.required_labels` as `docs/configuration.md:48` (this PR) promises |
| **P2** | `issue-routable.ts:16-17` | Blank required label is satisfied by a blank issue label — contradicts §5.3.1 and `docs/configuration.md:56-58` |
| **P3** | `parser.ts:368` | Undocumented `tracker.requiredLabels` camelCase alias silently accepted |
| **P3** | `pickup-labels.ts:47` | Dispatch reasons now echo normalized labels, losing the operator's configured casing in `repo explain` |
| **P3** | — | `workflow validate --json` `summary` omits `requiredLabels` while listing `activeStates` / `terminalStates` / `blockerCheckStates` / `planningStates`. Operators can't confirm what normalization produced. One field in `packages/cli/src/commands/workflow.ts:91,858` |
| **nit** | `issue-routable.ts:23` | `Issue is missing required labels ().` for a blank label |
| **nit** | `docs/configuration.md:48` | New table row isn't column-padded; `prettier --check` now flags the file |

### Verification performed

Everything below run locally at `cf3c6a3`:

- `pnpm lint` ✅ · `pnpm typecheck` ✅ · `pnpm build` ✅
- `npx vitest run packages/core/src/workflow/ packages/core/src/workflow-loader.test.ts` → **107 passed**
- `pnpm format` ❌ 87 files — pre-existing repo-wide drift; this PR adds `docs/configuration.md` to it (nit above)

**Smoke test** — no preview environment for a core-library change, so exercised through the real user-facing surface, the built CLI (`packages/cli/dist/index.js workflow validate`):

1. Valid `required_labels: [" Ready ", "Agent"]` → `"ok": true`; parse and normalization succeed. Also confirmed the summary block does *not* surface `requiredLabels` (P3 above).
2. Invalid `required_labels: ready, agent` → rejected with `workflow_validation_error`, but `"path": "front_matter"` — the P2.
3. Direct `issueRoutable` probe with `labels: ["ready", "  "]` / `requiredLabels: ["ready", ""]` → `{ routable: true }` — the second P2.

Fix the two P2s (or push back with reasoning) and this is an approve.

---

packages/core/src/workflow/issue-routable.ts
```diff
@@ -0,0 +1,32 @@
+import type { TrackedIssue } from "../contracts/tracker-adapter.js";
+import type { WorkflowLifecycleConfig } from "./lifecycle.js";
+
+/** Resolves whether an issue is eligible for workflow routing. */
+export function issueRoutable(
+  issue: Pick<TrackedIssue, "dispatchable" | "dispatchReason" | "labels">,
+  lifecycle: Pick<WorkflowLifecycleConfig, "requiredLabels">
+): { routable: boolean; reason?: string } {
```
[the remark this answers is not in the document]
moncher-dev: This PR intentionally remains the core-only split for #720. Runtime enforcement is tracked by the already-open follow-up [#721](https://github.com/hojinzs/github-symphony/issues/721), which depends on this helper; expanding into `service.ts` here would undo the #664 split.

packages/core/src/workflow/parser.ts
```diff
@@ -353,11 +357,33 @@ function readPickupLabelsConfig(
...
+  if (value === undefined || value === null) {
+    return undefined;
+  }
+  if (
+    !Array.isArray(value) ||
+    value.some((label) => typeof label !== "string")
+  ) {
+    throw new Error(
+      'Workflow front matter field "tracker.required_labels" must be an array of strings.'
+    );
```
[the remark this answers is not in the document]
moncher-dev: Resolved in a739b62. Malformed `tracker.required_labels` now throws `WorkflowValidationError` with the documented `tracker.required_labels` path, and the loader regression test asserts the code, path, and message.

packages/core/src/workflow/issue-routable.ts
```diff
@@ -0,0 +1,32 @@
...
+    return {
+      routable: false,
+      reason: issue.dispatchReason ?? "Issue is not dispatchable.",
+    };
+  }
+
+  const labels = new Set(issue.labels.map(normalizeLabel));
+  const missingLabels = (lifecycle.requiredLabels ?? []).filter(
+    (label) => !labels.has(normalizeLabel(label))
+  );
```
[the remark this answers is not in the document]
moncher-dev: Resolved in a739b62. A normalized blank required label is now unconditionally missing, even when a file-tracker issue has a blank label. The regression test covers that case.

packages/core/src/workflow/parser.ts
moncher-dev: Resolved in a739b62. The parser now reports `workflow_validation_error` at `tracker.required_labels`, with a regression assertion for the code, path, and message.

packages/core/src/workflow/issue-routable.ts
moncher-dev: Resolved in a739b62. Blank required labels are permanently unsatisfied and the test now includes a blank issue label to verify the invariant.

packages/core/src/workflow/parser.ts
moncher-dev: Resolved in a739b62 by removing the undocumented camelCase alias from the parser surface. Only `tracker.required_labels` is read.

packages/core/src/workflow/pickup-labels.ts
moncher-dev: Resolved in a739b62. Matching still trims and lowercases configured pickup labels, while dispatch reasons retain the trimmed configured spelling for operator-facing output.

packages/core/src/workflow/issue-routable.ts
moncher-dev: Resolved in a739b62. Missing required labels are JSON-quoted in the reason, including a blank label, so the output is unambiguous.

docs/configuration.md
moncher-dev: Resolved in a739b62. The validation table row and its header/separator were padded and formatted with Prettier.

packages/core/src/workflow/parser.ts
```diff
@@ -353,11 +357,35 @@ function readPickupLabelsConfig(
...
+    throw new WorkflowValidationError(
+      "workflow_validation_error",
+      "tracker.required_labels",
+      'Workflow front matter field "tracker.required_labels" must be an array of strings.'
+    );
+  }
+  return normalizeLabels(value as string[]);
+}
+
+function normalizeLabels(labels: readonly string[]): string[] {
```
hojinzs: **P1 — this local `normalizeLabels` collides head-on with the shared one that landed on `main`, and the obvious conflict resolution silently breaks §5.3.1.**

`main` picked up #729 (`6a987fb feat(core): normalize labels and timestamps`) after this branch forked. It added `packages/core/src/workflow/normalization.ts`, exported from `packages/core/src/index.ts`:

```ts
export function normalizeLabels(list: readonly unknown[]): string[] {
  const labels = new Set<string>();
  for (const value of list) {
    if (typeof value !== "string") continue;
    const label = value.trim().toLowerCase();
    if (label) labels.add(label);   // ← drops blanks
  }
  return [...labels];                // ← dedupes
}
```

Same name, same file-local role, **different semantics**: it drops blank labels and dedupes. Yours preserves both, which is precisely what #720 / §5.3.1 requires so a configured `""` stays "matches no issue".

The trap: `parser.ts` is one of the seven conflicted files, and the natural resolution — delete the local copy, import the shared one — turns `required_labels: [""]` into `[]`, i.e. **unrestricted**. That is the fail-open inverse of the P2 you just fixed in `a739b62`, and no current test catches it: `normalizes required and pickup labels while preserving empty labels` asserts on the parser output, so it would fail loudly for `required_labels`, but nothing pins the *reason* it must not use the shared helper.

Please resolve the conflict by keeping required-label normalization blank-preserving and making the divergence explicit — a comment here naming §5.3.1 and why `normalization.ts#normalizeLabels` is deliberately not used, or a differently-named local helper (`normalizeRequiredLabels`) so the next person doesn't "clean it up".

Separately, `readPickupLabelsConfig` at lines 360-361 is now **byte-identical to what #729 already merged**:

```
$ git show origin/main:packages/core/src/workflow/parser.ts
    include: normalizeLabels(readStringList(input, "include") ?? []),
    exclude: normalizeLabels(readStringList(input, "exclude") ?? []),
```

so that hunk should resolve to `main`'s version (shared helper) rather than yours — blank/dupe dropping is fine and correct for pickup labels.

---

packages/core/src/workflow/pickup-labels.ts
```diff
@@ -36,24 +36,33 @@ export function resolvePickupLabelDispatchReason<T extends TrackedIssue>(
...
   }
 
-  const labels = new Set(issue.labels);
-  const excludedLabel = [...exclude].find((label) => labels.has(label));
+  const labels = new Set(issue.labels.map(normalizeLabel));
+  const excludedLabel = exclude.find((label) =>
+    labels.has(normalizeLabel(label))
+  );
   if (excludedLabel) {
     return `Issue has excluded pickup label "${excludedLabel}".`;
```
hojinzs, line 48: **P3 — reopening my round-1 P3: the casing fix never reaches this code in production.** (Thread reply said "dispatch reasons retain the trimmed configured spelling for operator-facing output.")

That holds only for a hand-built project config. Every real workflow goes through the parser, which lowercases `pickup_labels` at parse time (`parser.ts:360-361`), and `packages/cli/src/repo-runtime.ts:118` / `packages/cli/src/commands/project.ts:347` assign that parser output straight into `tracker.settings.pickupLabels`. So `excludedLabel` here is already lowercase before this function is entered.

Verified against the built `packages/core/dist` on this head:

```
G parser-produced pickupLabels: {"include":["ready"],"exclude":["blocked"]}
H reason from parser config:    "Issue has excluded pickup label \"blocked\"."
I reason from literal config:   "Issue has excluded pickup label \"Blocked\"."
```

`H` is the production path — still lowercase, unchanged from round 1. `I` is what your new test `normalizes configured and issue labels before matching` (`pickup-labels.test.ts:60-80`) asserts, and it gets there by constructing `mixedCaseProject` as an object literal, bypassing the parser entirely. So the test passes while the behavior it claims to protect is unreachable.

Two honest options, both fine by me:

1. **Drop the claim.** Revert `readLabelList` to plain `normalizeLabels` (matching `main`'s #729, which already did exactly that), and change the test to assert the lowercase reason that the pipeline actually produces. Operator-facing casing then becomes a separate, real issue.
2. **Make it reachable.** Keep the raw configured spelling through the parser — i.e. don't lowercase `pickup_labels` at parse time, and let this function do all normalization. That contradicts #720's "설정 측 라벨도 lowercase/trim" wording, so it'd need a note.

Given #729 already merged option 1's shape, I'd take option 1 and keep this PR's diff tight. Non-blocking on its own — but the test should not assert a behavior the product can't exhibit.

---

packages/core/src/workflow/issue-routable.ts
```diff
@@ -0,0 +1,35 @@
...
+      reason: `Issue is missing required labels (${missingLabels
+        .map((label) => JSON.stringify(label))
+        .join(", ")}).`,
+    };
+  }
+
+  return { routable: true };
+}
+
+function normalizeLabel(label: string): string {
```
hojinzs: **nit / merge note** — third copy of the same `trim().toLowerCase()` helper in this PR (`parser.ts:385`, `pickup-labels.ts:66`, here), and `main` now exports a fourth from `workflow/normalization.ts`.

Unlike the parser one, this copy is safe to replace with the shared `normalizeLabels` — the blank-preserving requirement lives on the *config* side (`lifecycle.requiredLabels`), and line 19's explicit `normalized === ""` guard is what enforces §5.3.1 here, independent of how `issue.labels` gets normalized. Dropping blanks from the *issue* side is harmless, or even marginally more correct.

So after the rebase this could just be:

```ts
import { normalizeLabels } from "./normalization.js";
...
const labels = new Set(normalizeLabels(issue.labels));
```

with line 18's per-label `normalizeLabel(label)` on the config side kept as-is (it must stay blank-preserving). Worth a short comment marking that asymmetry, since it's the whole §5.3.1 invariant and it is not obvious.

Behavior on this head is correct — verified directly:

```
A blank-required + blank issue label: {"routable":false,"reason":"Issue is missing required labels (\"\")."}
B case/whitespace match:              {"routable":true}
D dispatchable=false precedence:      {"routable":false,"reason":"Archived."}
F undefined requiredLabels (legacy):  {"routable":true}
```

---

packages/core/src/workflow/parser.ts
```diff
@@ -136,6 +136,8 @@ function parseWorkflowMarkdownInternal(
   const planningStates =
     readStringList(tracker, "planning_states") ??
     DEFAULT_WORKFLOW_TRACKER.planningStates;
+  const requiredLabels =
+    readRequiredLabelList(tracker) ?? DEFAULT_WORKFLOW_TRACKER.requiredLabels;
```
hojinzs, line 200: **P3 (decision needed before merge, not a defect)** — `required_labels` lands as a new *flat* tracker key, which is the shape `main` just finished deprecating.

`main` now carries #711 / #728 / #714, which introduced `tracker.provider` plus:

```ts
const DEPRECATED_TRACKER_PROVIDER_KEYS = [
  ..., "pickup_labels", "blocker_check_states", "planning_states",
] as const;
```

with flat keys auto-promoted into `provider` and recorded in `tracker.deprecatedKeys`. `config.ts` on `main` spells the intent out for the closest analogue:

> Deprecated flat compatibility alias for tracker-owned blocker eligibility. Adapters receive this as tracker settings; core lifecycle logic never reads it.

This PR does the opposite for `required_labels`: flat tracker key → `WorkflowLifecycleConfig.requiredLabels` → read by core (`issueRoutable`).

To be clear, **#720 asked for exactly this** (`tracker.required_labels` + `WorkflowLifecycleConfig.requiredLabels`), so I'm not calling it over-engineering or scope drift — you built what was specified. But #720 was written before #711/#728/#714 merged, and shipping a new flat key now means it is a migration candidate on day one.

Two sub-points to settle while you're resolving the parser conflict anyway:

1. **Provider or core?** Spec §8.2 frames `required_labels` as routing, which argues for core lifecycle. If that's the call, fine — but then `required_labels` probably wants to *not* be in `DEPRECATED_TRACKER_PROVIDER_KEYS`, and that should be deliberate rather than incidental.
2. **`defaultLifecycle` interaction.** `main` added `WorkflowConfigTrackerAdapter.defaultLifecycle?: () => WorkflowLifecycleConfig` and builds lifecycle as `defaultLifecycle?.activeStates ?? legacyLifecycle.activeStates`. An adapter returning a lifecycle without `requiredLabels` yields `undefined`, which `issueRoutable` treats as unrestricted — fail-open. Your parser always sets it explicitly so the parsed path is safe, but please re-check that hunk survives the merge intact, since the surrounding lines are rewritten on `main`.

A one-line answer in the thread is enough; I don't want to expand this PR over it.

---

hojinzs (changes requested):
## Review — round 2/6

**All seven round-1 findings are fixed, and I verified each one end-to-end rather than by reading the diff.** The implementation is done. What blocks now is integration: `main` moved under this branch while it was in review, the PR is un-mergeable, and the natural conflict resolution would silently undo the P2 you just fixed.

### Round-1 follow-up — 7/7 resolved ✅

| Round-1 finding | Verified at `a739b62` |
| --- | --- |
| **P2** `parser.ts` error path was `front_matter` | ✅ `"path": "tracker.required_labels"` from the built CLI |
| **P2** blank required label satisfiable | ✅ `labels: ["ready","  "]` / `required: ["ready",""]` → `{routable:false, reason:'… ("")'}` |
| **P3** camelCase `requiredLabels` alias | ✅ removed; test pins that it's ignored |
| **P3** dispatch reason echoed normalized label | ⚠️ **fixed in this file, but unreachable in production** — reopened inline |
| **P3** `validate --json` omitted `requiredLabels` | ✅ `"requiredLabels": ["ready","agent"]` in `summary` |
| **nit** `Issue is missing required labels ().` | ✅ JSON-quoted |
| **nit** `docs/configuration.md` not column-padded | ✅ `prettier --check` clean on all changed files |

Thread replies were substantive and each pointed at the commit — that's the right pattern, thank you.

### New findings

| | Location | Issue |
| --- | --- | --- |
| **P1** | — | PR is un-mergeable; 7 files conflict with `main` |
| **P1** | `parser.ts:385` | Local `normalizeLabels` collides with `main`'s shared one, which **drops blanks** — resolving the conflict the obvious way reverts §5.3.1 to fail-open |
| **P3** | `pickup-labels.ts:49` | Round-1 P3 casing fix doesn't reach the production path; its test bypasses the parser |
| **P3** | `parser.ts:140` | New flat tracker key vs. `main`'s just-landed `tracker.provider` direction — needs a one-line decision, not a redesign |

#### The one that matters

`main` gained #729 (`6a987fb feat(core): normalize labels and timestamps`) after this branch forked. It exports `normalizeLabels` from a new `workflow/normalization.ts` — **same name as your local helper, opposite blank-label semantics**:

```ts
const label = value.trim().toLowerCase();
if (label) labels.add(label);   // drops "", and dedupes via Set
```

`parser.ts` is one of the conflicted files. Delete the local copy and import the shared one — the tidy-looking resolution — and `required_labels: [""]` becomes `[]`, i.e. **unrestricted**: the exact fail-open inverse of the P2 fixed in `a739b62`. Please keep required-label normalization blank-preserving and mark the divergence so it doesn't get "cleaned up" later. Details inline.

#729 also already merged your `readPickupLabelsConfig` change verbatim, so that hunk should resolve to `main`'s version.

### Verification performed

All at `a739b62`, locally:

- `pnpm install` ✅ · `pnpm build` ✅ · `pnpm lint` ✅ · `pnpm typecheck` ✅
- `npx vitest run packages/core/src/workflow/ packages/core/src/workflow-loader.test.ts packages/cli/src/commands/workflow.test.ts` → **126 passed / 7 files**
- `npx prettier --check` on all six changed source files → clean (round-1 nit confirmed fixed; no new drift added)
- `git merge-tree --write-tree HEAD origin/main` → **CONFLICT** in `docs/configuration.md`, `packages/core/src/index.ts`, `workflow-loader.test.ts`, `workflow/config.ts`, `workflow/parser.ts`, `workflow/pickup-labels.ts`, `workflow/pickup-labels.test.ts`. Branch is 14 commits behind `main`.
- **CI:** no check runs and no statuses on this head — `get_check_runs` → `total_count: 0`, `get_status` → `pending`, `total_count: 0`. The last CI signal on this PR is from `cf3c6a3`. Nothing has validated `a739b62` on GitHub; my local run above is the only evidence for this commit.

**Smoke test** — core-library change with no preview environment, so exercised through the shipped surface (`packages/cli/dist/index.js workflow validate`) plus direct probes against the built `packages/core/dist`:

1. `required_labels: [" Ready ", Agent]` → `"ok": true`, `summary.requiredLabels: ["ready","agent"]` — parse, normalization, and the new operator-facing field all good.
2. `required_labels: ready, agent` → `workflow_validation_error` at `"path": "tracker.required_labels"` — round-1 P2 #1 closed.
3. `required_labels: ["Ready", ""]` → `summary.requiredLabels: ["ready",""]` — blank preserved through parsing.
4. `issueRoutable` probes: blank-vs-blank unsatisfied ✅, case/whitespace match ✅, partial miss reports only the missing label ✅, `dispatchable:false` takes precedence ✅, `[]` and `undefined` both unrestricted (no regression for existing WORKFLOW.md) ✅.
5. Pickup-label reason through the **real** parser → `"blocked"`, not `"Blocked"` — the reopened P3.

### Path to approve

Rebase or merge `main` in, resolve the seven conflicts with the blank-label invariant intact and a test pinning it, and answer the `tracker.provider` question in a sentence. **I will approve on that push** provided CI is green — I'm not opening new design threads on this one. If the `tracker.provider` question turns out to need real restructuring, say so and I'd rather land this as-is and track it, than keep this PR open.

---

packages/core/src/workflow/parser.ts
moncher-dev: Resolved in rebased head `cf022f23`: `required_labels` now uses an explicitly named blank-preserving path with a §5.3.1 comment, while pickup labels use the shared `normalizeLabels` helper from main. Regression coverage preserves the blank configured label.

packages/core/src/workflow/pickup-labels.ts
moncher-dev: Resolved in `cf022f23`: the production-path behavior is now asserted as canonical lowercase (`"blocked"`), matching parser-normalized `pickup_labels` and the shared main implementation.

packages/core/src/workflow/issue-routable.ts
moncher-dev: Resolved in `cf022f23`: issue labels now use shared `normalizeLabels`, while required labels retain an explicit blank check before matching. The reason quotes missing labels, including `""`.

packages/core/src/workflow/parser.ts
moncher-dev: Confirmed deliberately: #720 and Symphony §8.2 define `required_labels` as a core lifecycle routing input, so it remains outside deprecated provider-key promotion. The parser always supplies `lifecycle.requiredLabels` (default `[]`); legacy literal lifecycle objects remain unrestricted for compatibility.

docs/configuration.md
```diff
@@ -76,38 +81,17 @@ is pending adapter work in #660-B.
...
+`tracker.blocker_check_states` selects the workflow states where unresolved
+`blocked_by` dependencies prevent dispatch. When the field is omitted, the
+default is `["Todo"]`, matching the Symphony candidate-selection rule. Set an
+explicit empty list (`blocker_check_states: []`) only when blocker gating should
+be disabled. This opt-out is an intentional repository-level divergence from
+the vendored Symphony specification's unconditional blocker rule. Omitting
 `planning_states` keeps planning disabled; blocker defaults do not enable the
 planning/human-review execution phase. Linear `blocked_by` metadata is derived
 from inverse relations of type `blocks`; source-side relations describe issues
 blocked by the current issue and are not blockers of it.
```
hojinzs, line 131: **P1 — the rebase resolved `docs/configuration.md` to "ours", deleting 98 lines of documentation that is live on `main` and unrelated to #720.**

This is the same trap I flagged in round 2, landed in the docs file instead of `parser.ts`. The branch is now correctly rebased onto `d270fe7` and six of the seven conflicts resolved well — but this file was resolved by taking the branch's pre-fork copy wholesale, so everything `#711` / `#714` / `#728` added to it after the fork is gone.

Measured on this head (`cf022f2`), against the merge base `d270fe7`:

```
$ git diff d270fe7 pr724 -- packages/ | grep -c '^-[^-]'
2          # a prettier reflow of the `??` expression — fine
$ git diff d270fe7 pr724 -- docs/ | grep -c '^-[^-]'
98
```

Three whole sections that exist on `main` are absent from this head:

```
DELETED: ## `GH_SYMPHONY_CONFIG_DIR` and Repository Runtimes   (15 lines)
DELETED: ### GitHub Project provider profile                    (39 lines)
DELETED: ### Workspace-key migration                            (11 lines)
```

```
$ git show origin/main:docs/configuration.md | grep -c 'GitHub Project provider profile'
1
$ git show pr724:docs/configuration.md      | grep -c 'GitHub Project provider profile'
0
```

Plus paragraph-level losses, flagged separately below.

Attribution, so it's easy to find: `7c0c846` touched this file correctly (+6/−3). `c92cc1e` — the round-1 fix commit, whose only docs task was padding the table row I nitted — carries +36/−111. `cf022f2` doesn't touch the file at all. So the damage predates the rebase and was invisible in round 2 because the old merge base didn't yet contain those sections; rebasing onto current `main` is what surfaced it.

**Why this blocks rather than being a nit:** none of it is CI-visible (`Test` and `Container Smoke` are both green on this head, correctly). Merging as-is silently drops the operator documentation for `tracker.provider` — including the whole worked YAML example and the flat-key deprecation notice — which is the exact migration surface we discussed in the round-2 `tracker.provider` thread. It also drops the `GH_SYMPHONY_CONFIG_DIR` semantics and the workspace-key suffix contract.

**Suggested resolution** — one file, mechanical, no code risk:

```bash
git checkout origin/main -- docs/configuration.md
```

then re-apply only the three #720 edits, which are the entirety of this PR's legitimate docs delta:

1. the `tracker.required_labels` row in the front-matter validation table (already correctly padded — `prettier --check` is clean on it);
2. the new paragraph after "…cannot silently disable dispatch for a state":
   > `tracker.required_labels` defaults to `[]`. Labels are compared after trimming and lowercasing; every configured label must be present before an issue can be routed. A blank configured label is preserved and therefore matches no issue.
3. the "Label and Timestamp Normalization" sentence replacing `main`'s now-stale "…and future `tracker.required_labels` configuration support" with the §5.3.1 separate-path note. This one is genuinely required — `main`'s current wording is wrong as of this PR.

Note the `blocker_check_states` paragraph you have here also drops `main`'s GitHub/Linear adapter resolution specifics ("GitHub source-closed blockers and blockers whose Project workflow state is in `tracker.terminal_states` are resolved. Linear uses its workflow-state relation data directly…"). Keep `main`'s version; nothing in #720 needed it rewritten.

That also clears the last merge conflict — `docs/configuration.md` is the only remaining one:

```
$ git merge-tree --write-tree pr724 origin/main
CONFLICT (content): Merge conflict in docs/configuration.md
```

---

docs/configuration.md
```diff
@@ -214,44 +190,6 @@ gate: configure a Liquid conditional in the prompt when planning runs must not
...
-`Todo` and `In Progress`, terminal state `Done`, blocker-check state `Todo`,
-and no planning states. These defaults apply only when the corresponding list
-or field is omitted. Flat `tracker.project_id`, `tracker.endpoint`,
-`tracker.state_field`, `tracker.priority`, `tracker.pickup_labels`,
-`tracker.blocker_check_states`, and `tracker.planning_states` aliases remain
-supported for compatibility, but `gh-symphony workflow validate` and
-`gh-symphony repo doctor` warn and print a copyable `tracker.provider` block.
-They are scheduled for removal in the next major release.
-
 ## Skill Layering
```
hojinzs, line 291: **P1 (same root cause)** — the 39-line `### GitHub Project provider profile` section sat here on `main` and is gone on this head.

What's lost: the `tracker.provider` worked YAML example, the statement that `project_id` / `endpoint` / `state_field` are non-empty strings with `endpoint` an HTTP(S) URL, the documented lifecycle profile defaults (`Status`, `Todo`/`In Progress`, `Done`, blocker-check `Todo`, no planning states), and the flat-alias deprecation notice — the one that tells operators `gh-symphony workflow validate` will warn and print a copyable `provider` block, and that the aliases go away next major.

Also gone just above, from the `Workflow Lifecycle Phases` intro: "Tracker-specific settings belong in `tracker.provider`. Core preserves every provider key without interpreting it… `active_states` and `terminal_states` must be YAML lists (not comma-separated strings)…".

This is the documentation for #711/#728, which merged nine days into this PR's review. It's also the direct counterpart to the round-2 `tracker.provider` thread where you (reasonably) argued `required_labels` stays a core lifecycle key rather than a provider key — deleting the provider docs while landing a new flat key next to them would make that decision much harder to read later.

Restoring `docs/configuration.md` from `origin/main` per the comment above brings this back.

---

docs/configuration.md
```diff
@@ -307,16 +245,16 @@ container environments.
...
+| Variable                         | Default          | Read by                                              | Audience          | Notes                                                                                           |
+| -------------------------------- | ---------------- | ---------------------------------------------------- | ----------------- | ----------------------------------------------------------------------------------------------- |
+| `GITHUB_TOKEN_BROKER_URL`        | unset            | Codex runtime, Claude runtime, Git credential helper | User-facing/ops   | Broker endpoint for GitHub tokens used by GitHub GraphQL tooling and Git credential resolution. |
+| `GITHUB_TOKEN_BROKER_SECRET`     | unset            | Codex runtime, Claude runtime, Git credential helper | User-facing/ops   | Shared secret sent to the GitHub token broker. Set with `GITHUB_TOKEN_BROKER_URL`.              |
```
hojinzs: **P2 (same root cause)** — the Notes column of the credential-broker table was re-flowed to a narrower width, and two sentences were dropped in the process rather than wrapped.

Gone from `GITHUB_TOKEN_BROKER_URL`:
> When set with the secret, raw GitHub tracker aliases are removed from coding-agent children.

Gone from `GITHUB_TOKEN_BROKER_SECRET` (this row):
> Without both settings, Phase 1a retains raw GitHub credentials for compatibility and logs a worker warning until #700.

```
$ git show origin/main:docs/configuration.md | grep -c 'Phase 1a retains raw GitHub credentials'
1
$ git show pr724:docs/configuration.md      | grep -c 'Phase 1a retains raw GitHub credentials'
0
```

Both describe security-relevant fallback behavior — that leaving the pair unset keeps raw credentials in the agent environment — and the `#700` reference is the only pointer to where that gets closed out. Worth being deliberate about, since a narrower table is what a reformat optimizes for and dropped prose is what it costs.

Restoring the file from `origin/main` covers this too.

---

docs/configuration.md
```diff
@@ -394,7 +332,7 @@ not set them manually.
...
 | `SYMPHONY_ISSUE_STATE`            | tracker state                             | Worker, hooks                                 | Internal/injected | Current tracker state at dispatch time.                                                 |
 | `SYMPHONY_ISSUE_SUBJECT_ID`       | tracker subject ID                        | Worker, hooks                                 | Internal/injected | Subject ID used for tracker-specific mutations.                                         |
-| `SYMPHONY_ISSUE_WORKSPACE_KEY`    | workspace key                             | Worker, hooks                                 | Internal/injected | Stable workspace key; sanitized identifiers add a 16-hex SHA-256 suffix.               |
+| `SYMPHONY_ISSUE_WORKSPACE_KEY`    | workspace key                             | Worker, hooks                                 | Internal/injected | Stable workspace key for the issue.                                                     |
```
hojinzs: **P2 (same root cause)** — this row's Notes cell was weakened from a contract to a tautology, and the section that explained it was deleted.

On `main`:
> Stable workspace key; sanitized identifiers add a 16-hex SHA-256 suffix.

Here:
> Stable workspace key for the issue.

And the 11-line `### Workspace-key migration` section further down (which this row's original text pointed at) is gone with it — including the worked example `acme/platform#42` → `acme_platform_42-cbb20472b0ece3db`, the rule that existing suffixless directories are checked after the new key and reused in place without renaming, and the fact that `repo status` / `repo explain` print `workspace key: legacy` for those.

That's the documentation an operator needs to interpret a `workspace key: legacy` line they're actually looking at, so losing it is a real support cost. Restoring from `origin/main` covers it.

---

hojinzs (changes requested):
## Review — round 3/6

**The code is finished.** Every round-2 code finding is resolved, the rebase onto `d270fe7` is clean where it counts, the §5.3.1 invariant survived it, and CI is green on this head. I verified all of that by execution, not by reading the diff.

What blocks is entirely in `docs/configuration.md`: the conflict resolution took the branch's pre-fork copy of that file wholesale, deleting **98 lines of documentation that is live on `main` and has nothing to do with #720** — including the `tracker.provider` profile section that `#711`/`#728` added while this PR was in review. It is also the one remaining merge conflict. One `git checkout origin/main -- docs/configuration.md` plus re-applying three short edits fixes both at once; details inline.

I said last round I'd approve on a clean rebase with green CI and wouldn't open new design threads. I'm holding to the second half — nothing below is a design question, and I'm not asking for anything on the code. But I can't wave through a merge that silently deletes a section of live operator docs, so this is one more round.

### Round-2 follow-up — 4/4 resolved ✅

| Round-2 finding | Verified at `cf022f2` |
| --- | --- |
| **P1** PR un-mergeable, 7 conflicted files | ⚠️ **6/7 resolved; `docs/configuration.md` still conflicts** — and its resolution is the new P1 |
| **P1** local `normalizeLabels` collides with `main`'s blank-dropping shared helper | ✅ `readRequiredLabelList` is a distinctly-named blank-preserving path with a §5.3.1 comment; `issue.labels` uses shared `normalizeLabels`; the asymmetry is commented at the call site |
| **P3** pickup-labels casing fix unreachable in production | ✅ took option 1 as suggested — `pickup-labels.ts`, `pickup-labels.test.ts`, `normalization.ts` are now **byte-identical to `origin/main`** (`git diff origin/main pr724 --` on all three → empty) |
| **P3** flat key vs. `tracker.provider` direction | ✅ answered in-thread with reasoning; §8.2 routing ⇒ core lifecycle, deliberately outside deprecated-key promotion. Agreed, closing it |

The parser hunk survived the merge intact, which was my specific worry on the `defaultLifecycle` sub-point: `required_labels` omitted on a `github-project` workflow (which does go through `defaultLifecycle`) yields `requiredLabels: []`, not `undefined`, so the fail-open path isn't reachable from parsed config.

### New finding

| | Location | Issue |
| --- | --- | --- |
| **P1** | `docs/configuration.md` | Conflict resolved to "ours"; 98 lines deleted, 3 whole sections lost, unrelated to #720. Also the last merge conflict |
| **P2** | `docs/configuration.md:251` | Credential-broker table re-flow dropped two sentences of security-relevant fallback behavior |
| **P2** | `docs/configuration.md:335` | `SYMPHONY_ISSUE_WORKSPACE_KEY` contract weakened; `### Workspace-key migration` deleted |

Scoped precisely — the code is not implicated:

```
$ git diff d270fe7 pr724 -- packages/ | grep -c '^-[^-]'
2          # prettier reflow of the `??` expression
$ git diff d270fe7 pr724 -- docs/ | grep -c '^-[^-]'
98
```

```
DELETED: ## `GH_SYMPHONY_CONFIG_DIR` and Repository Runtimes   (15 lines)
DELETED: ### GitHub Project provider profile                    (39 lines)
DELETED: ### Workspace-key migration                            (11 lines)
```

Blame: `7c0c846` touched the file correctly (+6/−3); `c92cc1e` — the round-1 fix commit, whose only docs job was padding the table row I nitted — carries +36/−111; `cf022f2` doesn't touch it. So this predates the rebase and was invisible in round 2 only because the old merge base didn't yet contain those sections. My round-2 review missed it for that reason; the rebase is what exposed it.

### No over-engineering / scope drift

Nothing in this PR exceeds #720. `issueRoutable` is still core-only and unreferenced by runtime code, which is correct for the 1/3 split — `#721` owns the orchestrator wiring. The changeset is a single `minor` on `@gh-symphony/cli`, matching `changeset:minor` on the issue.

### Requirement coverage vs #720 — 7/7 ✅

| #720 요구사항 | 상태 |
| --- | --- |
| `required_labels` parsing, default `[]`, trim+lowercase | ✅ |
| Blank label preserved as "matches nothing" (§5.3.1) | ✅ end-to-end, parser → `issueRoutable` |
| `pickup-labels.ts` lowercases the config side | ✅ via `main`'s #729; `include: [Ready]` → `["ready"]` |
| `issueRoutable(issue, lifecycle)`, dispatchable-first + ALL-of | ✅ |
| `requiredLabels` on `WorkflowLifecycleConfig` | ✅ |
| Parser tests: normalization / blank / comma rejection | ✅ |
| No regression when `required_labels` omitted | ✅ |

### Verification performed

All at `cf022f2`, locally, on a fresh `pnpm install --frozen-lockfile`:

- `pnpm lint` ✅ · `pnpm typecheck` ✅ · `pnpm build` ✅
- `npx vitest run packages/core/src/workflow/ packages/core/src/workflow-loader.test.ts packages/cli/src/commands/workflow.test.ts` → **148 passed / 8 files**
- `npx prettier --check` on all 11 changed files → **clean** (round-1 nit stays fixed; no new drift)
- `git merge-tree --write-tree pr724 origin/main` → single CONFLICT, `docs/configuration.md`. Branch is 1 behind / 3 ahead
- **CI on this head:** `Test` ✅ success, `Container Smoke` ✅ success — this is the first commit on the PR with CI actually run and green, which was the other round-2 gap

**Smoke test** — core-library change with no preview environment, so exercised through the shipped surface (`packages/cli/dist/index.js workflow validate --json`) and direct probes against the built `packages/core/dist`:

| # | Input | Result |
| --- | --- | --- |
| 1 | `required_labels: [" Ready ", Agent]` | `ok: true`, `summary.requiredLabels: ["ready","agent"]` |
| 2 | `required_labels: ready, agent` | `workflow_validation_error` at `path: "tracker.required_labels"` |
| 3 | `required_labels: [1]` | same code and path — non-string members covered too |
| 4 | `required_labels: ["Ready", ""]` | `["ready",""]` — blank survives parsing |
| 5 | omitted | `[]` — no regression for existing `WORKFLOW.md` |

`issueRoutable` probes against the built core:

```
A blank required + blank issue label     {"routable":false,"reason":"Issue is missing required labels (\"\")."}
B case/whitespace match                  {"routable":true}
C partial miss reports only missing       {"routable":false,"reason":"Issue is missing required labels (\"agent\")."}
D dispatchable=false precedence           {"routable":false,"reason":"Archived."}
E empty requiredLabels = unrestricted     {"routable":true}
F undefined requiredLabels (legacy)       {"routable":true}
G duplicate required labels               {"routable":true}
H empty issue labels + required           {"routable":false,"reason":"Issue is missing required labels (\"ready\")."}
L end-to-end parser -> issueRoutable      {"routable":false,"reason":"Issue is missing required labels (\"\")."}
M camelCase alias ignored                 []
```

`A` is the §5.3.1 invariant that the `normalizeLabels` collision threatened — it held through the rebase, which was the main thing I wanted to confirm. `L` confirms it end-to-end from real parsed front matter, not a literal.

### Path to approve

Restore `docs/configuration.md` from `origin/main`, re-apply the three #720 edits listed inline, push. That resolves the last conflict as a side effect.

**I will approve on that push** — no re-verification of the code, no new threads. If you'd rather not carry the docs restoration in this PR, say so and I'll approve it with the deletion reverted to `main`'s file and the three edits moved to a follow-up; what I can't approve is merging the deletion.

---

docs/configuration.md
moncher-dev: Resolved in the rebased head. `docs/configuration.md` now starts from current `origin/main` and retains only #720's required-label validation, default, and blank-preserving normalization additions; the current provider, credential, and workspace-key documentation is preserved.

docs/configuration.md
moncher-dev: Resolved in the rebased head. `docs/configuration.md` now starts from current `origin/main` and retains only #720's required-label validation, default, and blank-preserving normalization additions; the current provider, credential, and workspace-key documentation is preserved.

docs/configuration.md
moncher-dev: Resolved in the rebased head. `docs/configuration.md` now starts from current `origin/main` and retains only #720's required-label validation, default, and blank-preserving normalization additions; the current provider, credential, and workspace-key documentation is preserved.

docs/configuration.md
moncher-dev: Resolved in the rebased head. `docs/configuration.md` now starts from current `origin/main` and retains only #720's required-label validation, default, and blank-preserving normalization additions; the current provider, credential, and workspace-key documentation is preserved.

hojinzs (approved):
## Review — round 4/6 · Approve ✅

`docs/configuration.md` is restored from `origin/main` with exactly the three #720 edits and nothing else. The last merge conflict is gone, CI is green on this head, and the §5.3.1 invariant survived. I said last round I'd approve on that push with no new threads, and the push delivered it — approving.

### Round-3 follow-up — 3/3 resolved ✅

| Round-3 finding | Verified at `cb853da` |
| --- | --- |
| **P1** `docs/configuration.md` resolved to "ours"; 98 lines deleted, 3 sections lost | ✅ all three sections back, byte-for-byte with `main` |
| **P2** credential-broker table re-flow dropped two sentences | ✅ both restored |
| **P2** `SYMPHONY_ISSUE_WORKSPACE_KEY` contract weakened, `### Workspace-key migration` deleted | ✅ contract and section restored |

Measured rather than eyeballed — every string I named as missing in round 3 is present again, at parity with `main`:

```
main=1 pr=1  :: ## `GH_SYMPHONY_CONFIG_DIR` and Repository Runtimes
main=1 pr=1  :: ### GitHub Project provider profile
main=1 pr=1  :: ### Workspace-key migration
main=1 pr=1  :: Phase 1a retains raw GitHub credentials
main=1 pr=1  :: sanitized identifiers add a 16-hex SHA-256 suffix
main=1 pr=1  :: raw GitHub tracker aliases are removed
main=1 pr=1  :: GitHub source-closed blockers        (the blocker_check_states specifics)
```

The docs delta is now 11 deleted lines, all accounted for: 8 are the validation table's header/separator/rows re-padded to fit the new `tracker.required_labels` row (the round-1 nit), and 3 are the "Label and Timestamp Normalization" sentence that genuinely had to change — `main`'s wording ("…and future `tracker.required_labels` configuration support") is stale as of this PR. No unrelated prose lost.

**Merge state:** `git merge-tree --write-tree pr724 origin/main` → no conflicts. Branch is **0 behind / 3 ahead** of `main`.

### The 2-line code deletion

Round 3 I called this "a prettier reflow — fine" without checking which side was wrong. It's the `blocker_check_states` `??` expression at `parser.ts:186`, and the PR has it right: `origin/main`'s copy of `parser.ts` **fails** `prettier --check`, this PR's passes. It's a formatting fix, not drift. Correcting my own round-3 characterization.

### Requirement coverage vs #720 — 7/7 ✅

| #720 요구사항 | 상태 |
| --- | --- |
| `required_labels` parsing, default `[]`, trim+lowercase | ✅ |
| Blank label preserved as "matches nothing" (§5.3.1) | ✅ end-to-end, parser → `issueRoutable` |
| `pickup-labels.ts` lowercases the config side | ✅ via `main`'s #729; file is byte-identical to `main` |
| `issueRoutable(issue, lifecycle)`, dispatchable-first + ALL-of | ✅ |
| `requiredLabels` on `WorkflowLifecycleConfig` | ✅ |
| Parser tests: normalization / blank / comma rejection | ✅ |
| No regression when `required_labels` omitted | ✅ |

No over-engineering or scope drift. `issueRoutable` stays core-only and unreferenced by runtime code — correct for the 1/3 split, with #721 owning the orchestrator wiring. Single `minor` changeset on `@gh-symphony/cli`, matching `changeset:minor` on the issue.

### Verification performed

All at `cb853da`, locally, on a fresh `pnpm install --frozen-lockfile`:

- `pnpm lint` ✅ · `pnpm typecheck` ✅ · `pnpm build` ✅
- `pnpm test` → **1853 passed across 15 packages, 0 failures** (130 test files)
- `npx prettier --check` on all 11 changed files → **clean**
- **CI on this head:** `Test` ✅ · `Container Smoke` ✅

**Smoke test** — core-library change with no preview environment, so exercised through the shipped surface (`packages/cli/dist/index.js workflow validate --json`) and direct probes against the built `packages/core/dist`:

| # | Input | Result |
| --- | --- | --- |
| 1 | `required_labels: [" Ready ", Agent]` | `ok: true`, `summary.requiredLabels: ["ready","agent"]` |
| 2 | `required_labels: ready, agent` | `workflow_validation_error` at `tracker.required_labels` |
| 3 | `required_labels: [1]` | same code and path |
| 4 | `required_labels: {a: 1}` | same code and path — mapping rejected too |
| 5 | `required_labels: ["Ready", ""]` | `["ready",""]` — blank survives parsing |
| 6 | omitted | `[]` — no regression for existing `WORKFLOW.md` |
| 7 | `required_labels: []` | `[]` |
| 8 | `requiredLabels: [Ready]` (camelCase) | `[]` — alias still ignored |

`issueRoutable` probes against the built core:

```
A blank required + blank issue label     {"routable":false,"reason":"Issue is missing required labels (\"\")."}
B case/whitespace match                  {"routable":true}
C partial miss names only missing        {"routable":false,"reason":"Issue is missing required labels (\"agent\")."}
D dispatchable=false precedence          {"routable":false,"reason":"Archived."}
E empty requiredLabels unrestricted      {"routable":true}
F undefined requiredLabels (legacy)      {"routable":true}
G duplicate required labels              {"routable":true}
H empty issue labels + required          {"routable":false,"reason":"Issue is missing required labels (\"ready\")."}
I dispatchable=false, no reason          {"routable":false,"reason":"Issue is not dispatchable."}
J duplicate issue labels                 {"routable":true}
K non-string issue label                 {"routable":true}
L parsed lifecycle.requiredLabels        ["ready",""]
M end-to-end parsed -> issueRoutable     {"routable":false,"reason":"Issue is missing required labels (\"\")."}
N end-to-end match                       {"routable":true}
O end-to-end miss                        {"routable":false,"reason":"Issue is missing required labels (\"agent\")."}
```

`A` and `M` are the §5.3.1 invariant the `normalizeLabels` collision threatened in round 2 — it held through the rebase and through the docs restoration, confirmed from real parsed front matter, not a literal. `I` and `K` are new edge probes this round; both behave sensibly.

### Not blocking — for #721

Carrying forward the one item from round 1 so it isn't lost when the orchestrator wiring lands: `tracker-github/src/orchestrator-adapter.ts` `reviveIssue()` returns `labels: []`. Once `issueRoutable` gates continuation (§8.2 applies `required_labels` on continuation too), every revived GitHub issue reads as non-routable under any non-empty `required_labels` and recovery stops silently. Worth pinning a test on it there.

Nice work through four rounds — the thread replies were substantive and each pointed at a commit, which made re-verification cheap.

---

PR #729, opened 2026-08-28
feat(core): normalize labels and timestamps

### LINKED ISSUE #725: feat(core,docs): shared label/timestamp normalization helpers and the priority-mapping ADR
Parent: #660 (Epic #651)
리포트 항목: **B14, B15, B16, A15** — `docs/reports/2026-08-28-upstream-spec-drift-research.md`

결정(2026-08-28): priority 정렬은 ADR 2026-05-18을 supersede해 현재 숫자 오름차순을 "documented different mapping"으로 유지, Linear `0` → `null`.

#660 triage(범위 과대) 분할 1/3 — **core + 문서만**.

## 범위
- `packages/core`에 공용 helper: `normalizeLabels(list)`(trim/lowercase/blank drop/dedupe), `parseTrackerTimestamp(value)`(RFC 3339 → ISO 문자열 또는 null)
- `pickup-labels.ts`/`required_labels` 설정 측 라벨에도 같은 helper 적용(#720과 조율 — 이미 있으면 재사용)
- ADR: `docs/adr/2026-05-18_explicit-dispatch-priority-mappings.md`를 supersede하는 ADR — "모든 정수 오름차순, null/비정수 마지막"을 스펙 §8.2/§11.3의 documented different mapping으로 기록
- `docs/configuration.md` 정규화 규칙 절

## 완료 기준
- helper 단위테스트(대소문자·공백·중복·빈 값, 잘못된 timestamp → null)
- ADR 병합, 코드 동작 변화 없음(어댑터 적용은 #660-B)

moncher-dev:
## TL;DR

- Add shared core helpers for label canonicalization and RFC 3339 tracker timestamps.
- Make pickup-label matching and project-overlap checks case- and surrounding-whitespace-insensitive.
- Record the actual dispatch priority ordering as a documented repository-local mapping, while deferring Linear `0` normalization to #660-B.

## Change-point diagram

```text
workflow / tracker labels ──> normalizeLabels() ──> pickup matching + CLI overlap checks
tracker timestamps ─────────> parseTrackerTimestamp() ─> canonical ISO UTC (adapter adoption deferred)
tracker priorities ─────────> existing sorter ──────────> ADR records current numeric/null ordering
```

## Start here

- `packages/core/src/workflow/normalization.ts` and its tests define the shared helpers.
- `packages/core/src/workflow/pickup-labels.ts`, `packages/cli/src/commands/project.ts`, and `doctor.ts` reuse label normalization.
- `docs/adr/2026-08-28_priority-mapping-documented-different-mapping.md` distinguishes current behavior from pending #660-B adapter work.

## Risks & rollback

- Labels differing only by case or surrounding whitespace now share a pickup gate; reverting the helper consumers restores the former comparison behavior.
- Leap seconds are normalized to the following UTC second. Timestamp parsing remains exported-only, so adapters are unchanged in this PR.
- Linear priority `0` continues to pass through and sort ahead of positive values until #660-B.

## Changed files

- Core normalization helpers and RFC 3339 variant coverage.
- Pickup-label configuration, matching, CLI overlap, and doctor-summary normalization.
- ADR, configuration reference, both living READMEs, and the existing patch changeset.

## Issues — Closed #725

Closes #725

## Evidence

- `pnpm --filter @gh-symphony/core test -- normalization` (13 tests)
- `pnpm --filter @gh-symphony/cli test -- project` (32 tests)
- `pnpm lint`
- `pnpm test`
- `pnpm typecheck`
- `pnpm build`

- [ ] Confirm pickup-label eligibility treats case and surrounding-whitespace variants as the same label.
- [ ] Confirm the ADR accurately distinguishes current priority ordering from #660-B's pending Linear `0` normalization.


docs/adr/2026-08-28_priority-mapping-documented-different-mapping.md
```diff
@@ -0,0 +1,41 @@
...
+- **Related Issue**: #725
+- **Related Spec**: `docs/symphony-spec.md` §8.2 and §11.3 (read-only)
+
+## Context
+
+The upstream specification orders eligible candidates by priority before
+creation time and identifier. It does not require a particular tracker-level
+numeric mapping. The earlier ADR proposed a GitHub-specific, explicit priority
+configuration and adapter behavior that is not the repository's current
+implementation.
```
hojinzs: **P1 — the superseded ADR's design *is* implemented; superseding it as "not the current implementation" falsifies the design record**

I verified this against the tree at `71abc93`:

- `packages/core/src/workflow/parser.ts:550-600` parses the full `tracker.priority` block with all three sources (`project-field`, `labels`, `disabled`), including the §5.5 cross-source key rejection.
- `packages/tracker-github/src/adapter.ts:2418` `resolvePriority` resolves both the explicit path and the legacy `priority_field` path, exactly as §6 specifies.
- `packages/tracker-github/src/adapter.ts:2546` and `:2568` emit `priority.label_conflict_resolved` and `priority.unmapped` — the §8 observability events, by the proposed names.

So ADR 2026-05-18's C1 (§5, §5.5, §6, §8, §10) is shipped production behavior, not an unadopted proposal. Line 38-39 in Consequences compounds it: "The proposed `tracker.priority` configuration, GitHub-specific resolution, and observability work in the superseded ADR are not adopted by this decision" reads as "not adopted at all", which is the opposite of the tree.

The two ADRs are also orthogonal, not competing: 2026-05-18 defines how `TrackedIssue.priority` is **derived** from a tracker; this one defines how already-derived values are **sorted**. There is nothing to supersede.

Suggested fix: drop the `Supersedes` line (line 5) in favour of `Related ADRs`, rewrite the Context paragraph to state the orthogonality, and delete the Consequences bullet at 38-39. If the intent was narrower — that §9 (init/setup generation) and §7 (doctor drift checks) remain unimplemented — say exactly that, scoped to those sections.

---

docs/adr/2026-08-28_priority-mapping-documented-different-mapping.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Decision
+
+github-symphony retains its current documented different mapping:
+
+1. Every finite integer priority sorts in ascending numeric order; lower is
+   dispatched first.
+2. `null` and non-integer priority values sort after every integer priority.
```
hojinzs: **P1 — "non-integer priority values sort after every integer priority" is not what the sorter does**

`sortCandidatesForDispatch` (`packages/orchestrator/src/service.ts:6104-6119`) special-cases **only** `null`:

```ts
if (a.priority !== b.priority) {
  if (a.priority === null) return 1;
  if (b.priority === null) return -1;
  return a.priority - b.priority;   // any finite number, integer or not
}
```

`TrackedIssue.priority` is typed `number | null`, so a non-integer is representable, and `1.5` sorts *between* `1` and `2` — not last. (`NaN` is worse: `a.priority - b.priority` is `NaN`, giving an inconsistent comparator and an implementation-defined order.) Nothing between the adapters and the sorter coerces non-integers: `resolvePriority` returns the configured `values`/`labels` number verbatim, and the Linear adapter passes `issue.priority` through unchanged.

The entire purpose of #725 is to record the *actual* ordering as a "documented different mapping" under §8.2/§11.3. Recording a mapping the code does not implement is worse than recording none, because the next reader will treat it as ground truth.

Either state what the code does — "`null` sorts after every numeric priority; numeric values sort ascending, integer or not" — or add the `Number.isInteger` bucketing to `sortCandidatesForDispatch`. Given #725 says 코드 동작 변화 없음, the wording fix is the in-scope option. The same sentence appears in `docs/configuration.md:70-71` and needs the same correction.

---

docs/adr/2026-08-28_priority-mapping-documented-different-mapping.md
```diff
@@ -0,0 +1,41 @@
...
+3. Within either group, the existing creation-time and identifier tie-breakers
+   remain unchanged.
+4. Linear priority value `0` is normalized to `null`; it is not treated as the
+   highest priority.
```
hojinzs: **P1 — Linear `0` → `null` is stated in the present tense but is not implemented; today it is the *inverse* of what this says**

`packages/tracker-linear/src/orchestrator-adapter.ts:472`:

```ts
priority: typeof issue.priority === "number" ? issue.priority : null,
```

Linear's native "No priority" value `0` passes straight through as `0`, and since the sorter is priority-ascending, those issues currently dispatch **ahead of** Urgent (`1`). This ADR says the opposite is the case.

This is not a request to implement it here — #725 explicitly defers adapter adoption to #660-B ("어댑터 적용은 #660-B", "코드 동작 변화 없음"). The problem is purely that decision items 1-3 describe *current* behavior while item 4 describes *intended future* behavior, in the same list, in the same tense. An operator reading this will predict the wrong dispatch order for every Linear board with unprioritized issues.

Split it out, e.g.:

> ## Pending implementation
>
> Linear priority value `0` ("No priority") will be normalized to `null` rather than treated as the highest priority. The Linear adapter currently passes `0` through, so this ordering is not yet in effect; adapter adoption is tracked in #660-B.

Same correction needed at `docs/configuration.md:72`, which states it flatly as `Linear priority \`0\` maps to \`null\`.`

---

packages/core/src/workflow/normalization.ts
```diff
@@ -0,0 +1,53 @@
+const RFC_3339_TIMESTAMP =
+  /^(\d{4})-(\d{2})-(\d{2})T(\d{2}):(\d{2}):(\d{2})(?:\.\d+)?(?:Z|([+-])(\d{2}):(\d{2}))$/;
```
hojinzs: **P2 — the regex rejects two constructs RFC 3339 permits, while the doc advertises plain "RFC 3339"**

Verified by running the helper on this branch:

| input | result |
| --- | --- |
| `2026-08-28t12:34:56z` | `null` |
| `2026-12-31T23:59:60Z` | `null` |
| `2026-08-28T12:34:56Z` | `2026-08-28T12:34:56.000Z` |
| `2026-08-28T12:34:56-05:00` | `2026-08-28T17:34:56.000Z` |

RFC 3339 §5.6 explicitly permits lowercase `t`/`z` (`NOTE: ISO 8601 defines date and time separated by "T". Applications using this syntax may choose ... to allow "t" or "z"`), and `time-second = "60"` is legal for leap seconds. `docs/configuration.md:64` promises "accepts RFC 3339 timestamps", so the contract and the implementation disagree.

Practical exposure today is low — GitHub and Linear both emit uppercase `Z` — but this helper is being exported specifically so #660-B adapters adopt it, at which point a silent `null` becomes a dropped timestamp with no signal.

Cheapest fix is the `i` flag plus allowing `60`:

```ts
const RFC_3339_TIMESTAMP =
  /^(\d{4})-(\d{2})-(\d{2})[Tt](\d{2}):(\d{2}):(\d{2})(?:\.\d+)?(?:[Zz]|([+-])(\d{2}):(\d{2}))$/;
```

(note `new Date()` does not accept lowercase designators either, so line 51 would need the value uppercased before re-parsing). If you'd rather keep the narrow profile, that's a defensible call — then say so in the JSDoc and in `configuration.md` ("uppercase-designator RFC 3339 profile; lowercase `t`/`z` and leap seconds are rejected") so adopters aren't surprised.

---

packages/core/src/workflow/normalization.ts
```diff
@@ -0,0 +1,53 @@
...
+    calendarDate.getUTCMonth() !== Number(month) - 1 ||
+    calendarDate.getUTCDate() !== Number(day) ||
+    Number(hour) > 23 ||
+    Number(minute) > 59 ||
+    Number(second) > 59 ||
+    (offsetHour !== undefined &&
+      (Number(offsetHour) > 23 || Number(offsetMinute) > 59))
+  ) {
+    return null;
+  }
```
hojinzs, line 47: **P3 — five of the six guards here are unreachable; `new Date()` already rejects them**

The `new Date(value)` on line 51 does strict ISO parsing and rejects out-of-range components on its own. Measured on Node 24:

```
new Date("2026-08-28T25:00:00Z")      -> Invalid Date   // hour > 23
new Date("2026-08-28T12:60:00Z")      -> Invalid Date   // minute > 59
new Date("2026-08-28T12:34:60Z")      -> Invalid Date   // second > 59
new Date("2026-08-28T12:34:56+24:00") -> Invalid Date   // offset hour > 23
new Date("2026-08-28T12:34:56+00:60") -> Invalid Date   // offset minute > 59
new Date("2026-02-30T12:00:00Z")      -> 2026-03-02T12:00:00.000Z   // rolls over
```

Only the calendar-rollover case needs a manual guard — which is exactly the one case the unit test exercises. The other five conditions never evaluate true, which is what the coverage bot is reporting as the 94.44% branch gap on this file. That also makes `hour`/`minute`/`second`/`offsetHour`/`offsetMinute` unused destructuring targets.

```ts
const [, year, month, day] = match;
const calendarDate = new Date(
  Date.UTC(Number(year), Number(month) - 1, Number(day))
);
if (
  calendarDate.getUTCFullYear() !== Number(year) ||
  calendarDate.getUTCMonth() !== Number(month) - 1 ||
  calendarDate.getUTCDate() !== Number(day)
) {
  return null;
}
```

Minor aside on that guard: `Date.UTC` maps years `0`-`99` to `1900+n`, so a valid `0026-08-28T12:34:56Z` is rejected. Irrelevant in practice — flagging only so it's a known limit rather than a surprise.

---

docs/configuration.md
```diff
@@ -52,6 +52,25 @@ values must be positive YAML integers. This repository intentionally rejects an
...
 disable dispatch for a state.
 
+## Label and Timestamp Normalization
+
+Core normalizes workflow pickup labels and tracker-provided labels before label
+matching: it trims surrounding whitespace, lowercases values, drops blanks, and
+deduplicates while keeping the first occurrence. Thus `" Agent "`, `"agent"`,
+and `"AGENT"` are the same label. The same `normalizeLabels` helper is the
+shared implementation for current pickup-label handling and future
+`tracker.required_labels` configuration support.
```
hojinzs, line 62: **P2 — user-visible pickup eligibility changed, but the two living readmes weren't updated**

`AGENTS.md:36` requires that when behavior or the CLI surface changes, the same change updates "user-facing commands and usage in `README.md` and `packages/cli/README.md`" alongside `docs/configuration.md`.

This PR changes pickup matching from exact to case- and whitespace-insensitive — I confirmed end-to-end that a workflow configured with `include: [" Agent "]` now dispatches an issue labelled `AGENT` and skips one labelled `Blocked` against `exclude: [" BLOCKED "]`. That is a real eligibility change for existing operators: two labels that used to be distinct (`Bug` vs `bug` on Linear, which allows both) now collapse into one.

`README.md:594` describes the pickup-label gates for Linear and says nothing about the new semantics; `README.md:398` mentions `tracker.pickup_labels` for project disjointness; `packages/cli/README.md` doesn't cover it at all. A sentence in each is enough:

> Label comparison is case-insensitive and ignores surrounding whitespace, so `Agent`, `agent`, and `" AGENT "` are the same label.

Worth calling out in the README too that this is a **breaking-ish** change for anyone relying on case to separate labels; the changeset is `patch`, which I think is still right for a normalization fix, but the behavior note belongs in the changeset text as well.

---

hojinzs (changes requested):
## Review — round 1 / 6

The code half of this PR is good and I verified it works. The documentation half — which is the larger deliverable of #725 — states things about the codebase that are not true, and that's what blocks it.

### Verification performed

Checked out `71abc93`, `pnpm install`, then:

| Gate | Result |
| --- | --- |
| `pnpm build` | pass |
| `pnpm lint` | pass |
| `pnpm typecheck` | pass |
| `pnpm test` (all packages) | pass |
| `pnpm format` | `docs/configuration.md` flagged — **pre-existing**, the base commit `2353c05` fails the same check on an unrelated env-var table. Not this PR's. |
| CI on `71abc93` | Test ✅, Container Smoke ✅ |

**Smoke test** — drove the real chain (`parseWorkflowMarkdown` → tracker settings → `filterIssuesByPickupLabels`) rather than only the unit seams, since no preview environment applies to a core-library change:

```
workflow front matter: include: [" Agent "], exclude: [" BLOCKED "]
parsed pickupLabels:   {"include":["agent"],"exclude":["blocked"]}
issues: A-1 ["AGENT"] · A-2 ["agent","Blocked"] · A-3 ["other"]
dispatchable:          ["A-1"]
```

Correct on all three arms: case-insensitive include match, exclude winning over include, non-matching issue skipped. Also exercised `parseTrackerTimestamp` directly across offsets, fractional seconds, rollover dates, and out-of-range components — results are in the inline comments.

### Requirement coverage against #725

| 완료 기준 | Status |
| --- | --- |
| `normalizeLabels` (trim / lowercase / blank drop / dedupe) | ✅ verified, insertion order preserved |
| `parseTrackerTimestamp` (RFC 3339 → ISO, else `null`) | ⚠️ works for the common profile; rejects two constructs RFC 3339 allows — see inline P2 |
| Helper applied to `pickup-labels.ts` / config-side labels | ✅ — and correctly extended to tracker-provided labels too, which is necessary, not scope creep: normalizing only the config side would have broken matching outright |
| `required_labels` | ✅ N/A — not present in the tree (confirmed by `docs/reports/2026-08-28-upstream-spec-drift-research.md` row A3); the doc correctly frames it as future |
| Unit tests (case / whitespace / dupes / blanks, bad timestamp → `null`) | ✅ 8 cases, all green |
| ADR merged | ❌ merged but factually wrong in three places — see inline P1s |
| `docs/configuration.md` normalization section | ⚠️ present; inherits two of the ADR's incorrect claims |
| 코드 동작 변화 없음 (adapter work deferred to #660-B) | ✅ `parseTrackerTimestamp` is exported but unadopted, exactly as specified |

### Findings

**P1 ×3, all in the ADR** — inline. In short: it supersedes ADR 2026-05-18 on the grounds that its design "is not the repository's current implementation", but §5/§6/§8 of that ADR *are* implemented and shipped (`parser.ts:550-600`, `adapter.ts:2418`, and the `priority.label_conflict_resolved` / `priority.unmapped` events at `:2546`/`:2568`). It also documents two orderings the code doesn't produce: non-integers do **not** sort last (only `null` does), and Linear `0` is **not** normalized to `null` — it passes through at `tracker-linear/src/orchestrator-adapter.ts:472` and therefore currently outranks Urgent.

These are all wording fixes, not implementation work — none of them require touching adapter behavior, so they stay inside #725's stated scope. But they matter more than usual here: the sole point of this issue was to write down the real mapping as a "documented different mapping" under §8.2/§11.3. An ADR that records the wrong mapping is a liability, because the next reader will trust it over the code.

**P2 ×2** — RFC 3339 lowercase `t`/`z` and leap seconds silently return `null` while the docs promise plain RFC 3339 (low exposure today, real once #660-B adopts the helper); and `README.md` / `packages/cli/README.md` weren't updated for the eligibility change, which `AGENTS.md:36` requires.

**P3 ×2, one not visible in the diff:**

`packages/cli/src/commands/project.ts:307-311` — `normalizeLabelList` only type-filters:

```ts
function normalizeLabelList(value) {
  if (!Array.isArray(value)) return [];
  return value.filter((label) => typeof label === "string");
}
```

…while its sibling `normalizeStates` at `:283-290` does `.trim().toLowerCase()`. These feed `pickupLabelsMayOverlap`, the guard that stops two projects competing for the same issues on one repository. After this PR, freshly started projects persist lowercase labels while configs persisted earlier still hold `"Agent"`, so the guard can call two projects disjoint that dispatch now treats as overlapping. One-line fix — reuse `normalizeLabels` here. Second, cosmetic: `doctor.ts:950` now echoes `include=[agent]` back to an operator who wrote `Agent`, and its own blank-filtering in `readStringArraySetting` is redundant.

Nothing meaningful. The helpers are small and stay inside the requested scope; extending normalization to tracker-side labels was required for the feature to work at all. The one bit of unnecessary machinery is the timestamp guard block — five of its six conditions are unreachable because `new Date()` already rejects those inputs (inline P3). The unrelated Prettier reformat of the import in `pickup-labels.test.ts` is harmless.

### What I need for approval

The three P1 wording corrections in the ADR, and the P2 README updates. The P2 timestamp item is fine either way — accept the lowercase designators, or narrow the documented contract; just make code and doc agree. P3s are optional, though the `project.ts` one is a one-liner in the same spirit as this PR.

The implementation itself I'd take as-is.

---

docs/adr/2026-08-28_priority-mapping-documented-different-mapping.md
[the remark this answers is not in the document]
moncher-dev: Documented as pending #660-B work: Linear 0 currently passes through and sorts before positive values; this PR does not change adapter behavior.

docs/adr/2026-08-28_priority-mapping-documented-different-mapping.md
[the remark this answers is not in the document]
moncher-dev: Corrected the ADR relationship: the 2026-05-18 ADR is now related and remains in effect; this ADR records only sorter ordering.

docs/configuration.md
```diff
@@ -52,6 +52,25 @@ values must be positive YAML integers. This repository intentionally rejects an
 invalid override rather than ignoring it, so invalid configuration cannot silently
 disable dispatch for a state.
 
+## Label and Timestamp Normalization
+
+Core normalizes workflow pickup labels and tracker-provided labels before label
+matching: it trims surrounding whitespace, lowercases values, drops blanks, and
+deduplicates while keeping the first occurrence. Thus `" Agent "`, `"agent"`,
+and `"AGENT"` are the same label. The same `normalizeLabels` helper is the
```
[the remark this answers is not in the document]
moncher-dev: Added the case- and surrounding-whitespace-insensitive label rule to both README.md and packages/cli/README.md.

packages/core/src/workflow/normalization.ts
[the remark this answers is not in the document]
moncher-dev: Updated parsing and tests to accept lowercase t/z designators and leap-second :60 values, returning canonical UTC ISO strings.

docs/adr/2026-08-28_priority-mapping-documented-different-mapping.md
moncher-dev: Resolved by recording the ADRs as orthogonal: the earlier mapping/observability ADR remains in effect, while this ADR documents dispatch sorting.

docs/adr/2026-08-28_priority-mapping-documented-different-mapping.md
moncher-dev: Corrected the documentation to state the current numeric ordering: all numeric values sort ascending and only null sorts last.

docs/adr/2026-08-28_priority-mapping-documented-different-mapping.md
moncher-dev: Moved Linear 0 normalization to a pending #660-B section and documented its current pass-through behavior.

packages/core/src/workflow/normalization.ts
moncher-dev: Implemented the broader RFC 3339 contract and added lowercase-designator and leap-second unit coverage.

packages/core/src/workflow/normalization.ts
moncher-dev: Retained explicit component checks to enforce RFC 3339 bounds independently of Date parsing, and added invalid-boundary coverage plus the year-safe calendar validation.

docs/configuration.md
moncher-dev: Added the requested operator-facing normalization note to both living READMEs and expanded the changeset summary.

packages/core/src/workflow/normalization.ts
```diff
@@ -0,0 +1,62 @@
...
+    Number(hour) > 23 ||
+    Number(minute) > 59 ||
+    Number(second) > 60 ||
+    (offsetHour !== undefined &&
+      (Number(offsetHour) > 23 || Number(offsetMinute) > 59))
+  ) {
+    return null;
+  }
+
+  const normalized = value.replace("t", "T").replace("z", "Z");
```
hojinzs, line 49: **nit — the designator uppercasing is correct, but only by an undocumented invariant**

`String.prototype.replace` with a string pattern replaces the *first* occurrence, case-sensitively. That is safe here only because `RFC_3339_TIMESTAMP` guarantees the matched string's alphabet is digits, `-`, `:`, `+`, `.`, plus at most one `t`/`T` separator and one `z`/`Z` designator — so the first lowercase `t` can only be the separator and the first lowercase `z` can only be the zone. I confirmed all four casings round-trip correctly:

```
2026-08-28T12:34:56Z -> 2026-08-28T12:34:56.000Z
2026-08-28t12:34:56z -> 2026-08-28T12:34:56.000Z
2026-08-28t12:34:56Z -> 2026-08-28T12:34:56.000Z
2026-08-28T12:34:56z -> 2026-08-28T12:34:56.000Z
```

So this is not a bug. The concern is durability: the safety is a property of a regex 47 lines up, not of this line, and it breaks silently the moment that pattern grows (e.g. an alphabetic timezone name). Anchoring the replacements makes the line self-evidently correct:

```ts
const normalized = value.replace(/t/i, "T").replace(/z$/i, "Z");
```

Non-blocking.

---

packages/core/src/workflow/normalization.ts
```diff
@@ -0,0 +1,62 @@
...
+      ? normalized.replace(/:60(?=(?:\.\d+)?(?:Z|[+-]\d{2}:\d{2})$)/i, ":59")
+      : normalized
+  );
+  if (Number.isNaN(timestamp.getTime())) {
+    return null;
+  }
+
+  return new Date(
+    Number(second) === 60 ? timestamp.getTime() + 1_000 : timestamp.getTime()
+  ).toISOString();
```
hojinzs, line 61: **nit — `:60` is accepted at *any* time of day and silently rolled forward; worth one JSDoc line**

The leap-second handling works, including across offsets — I verified:

| input | result |
| --- | --- |
| `2026-12-31T23:59:60Z` | `2027-01-01T00:00:00.000Z` |
| `2026-12-31T23:59:60+09:00` | `2026-12-31T15:00:00.000Z` |
| `2026-06-30T23:59:60.500Z` | `2026-07-01T00:00:00.500Z` |
| `2024-02-29T23:59:60Z` | `2024-03-01T00:00:00.000Z` |

All four are right, and the offset case is the one I expected to be wrong (`23:59:60+09:00` is `14:59:60Z`, so `15:00:00Z` is correct).

The observation is that real leap seconds only ever occur at `23:59:60` UTC, but this accepts `:60` at any position:

```
2026-08-28T12:34:60Z -> 2026-08-28T12:35:00.000Z
```

RFC 3339's ABNF is `time-second = 2DIGIT` constrained to `00-60` with no positional restriction, so accepting this is spec-literal and I'm not asking you to reject it — that was the point of the round-1 P2. But it does mean a transposed digit in a tracker payload becomes a plausible-looking timestamp one second later rather than a `null` the caller can notice. Since the whole reason this helper is exported is #660-B adapter adoption, a line in the JSDoc — "`:60` is accepted anywhere per the RFC 3339 ABNF and normalized to the following second" — makes the coercion a known limit rather than a surprise at adoption time.

Also confirmed while I was in here: switching the calendar guard to the string form `new Date(\`${year}-${month}-${day}T00:00:00.000Z\`)` correctly avoids the `Date.UTC` year-0-99 remapping I flagged in round 1 — `0026-08-28T12:34:56Z` now returns `0026-08-28T12:34:56.000Z` instead of being rejected. Good fix.

Non-blocking.

---

hojinzs (approved):
## Review — round 2 / 6 · **Approve**

Every round-1 finding is genuinely addressed, and I re-verified each against the tree at `848eed8` rather than taking the reply threads at their word. Two non-blocking nits inline; nothing that should hold the merge.

| # | Finding | Status |
| --- | --- | --- |
| P1 | ADR supersedes 2026-05-18 as "not implemented" | ✅ `Supersedes:` → `Related ADR:`; Context now states orthogonality (derivation vs. sorting); the "not adopted by this decision" Consequences bullet is replaced with "remain in effect". The 2026-05-18 ADR file itself is untouched — no stale `Superseded` marker left behind. |
| P1 | "non-integers sort last" | ✅ Now "Every numeric priority, including non-integers, sorts in ascending numeric order" / "`null` sorts after every numeric priority" — which is exactly `service.ts:6104-6119`. `docs/configuration.md:69-71` corrected to match. |
| P1 | Linear `0` → `null` stated in present tense | ✅ Split into a `## Pending implementation` section that states the *current* pass-through and attributes normalization to #660-B. Confirmed against `tracker-linear/src/orchestrator-adapter.ts:472` — still `typeof issue.priority === "number" ? issue.priority : null`, so the ADR is now accurate. `configuration.md:72-74` carries the same correction. |
| P2 | RFC 3339 lowercase `t`/`z` and leap seconds | ✅ Implemented (`i` flag + `:60` rollover), not just documented. Verified across offsets — see inline. |
| P2 | Living READMEs + changeset | ✅ `README.md:398` and `:594`, `packages/cli/README.md:117-119`, and the changeset summary all carry the semantics now. The cli README wording ("Do not use labels that differ only by case or outer whitespace as separate pickup gates") is the operator-facing warning I was after. |
| P3 | `project.ts` `normalizeLabelList` only type-filters | ✅ Fixed — now `normalizeLabels(value)`, so `pickupLabelsMayOverlap` can no longer call two projects disjoint that dispatch treats as overlapping. This was the one with real consequences; glad it was taken. |
| P3 | Unreachable guards in the timestamp validator | ✅ Declined, with a reason I accept: now that `:60` is accepted and the string is rewritten before `new Date()`, bounds enforcement independent of `Date` parsing is defensible, and boundary tests were added. Not worth another round. |

`doctor.ts` also dropped its redundant blank-filtering (`readStringArraySetting` is now just `normalizeLabels(field)`), which was the cosmetic half of that P3. I checked it is used *only* for pickup `include`/`exclude` — no non-label setting gets lowercased by it.

### Verification performed

Checked out `848eed8`, `pnpm install --frozen-lockfile`:

| Gate | Result |
| --- | --- |
| `pnpm build` | pass |
| `pnpm lint` | pass |
| `pnpm typecheck` | pass |
| `pnpm test` | pass |
| `pnpm format` | 85 files flagged repo-wide — pre-existing; **every file this PR touches passes** (`prettier --check` on the 11 changed source/doc files: clean). The env-var table reformat in this diff actually fixes the `docs/configuration.md` failure I noted in round 1. |
| CI on `848eed8` | Test ✅, Container Smoke ✅ |

**Smoke test** — drove the real chain again (`parseWorkflowMarkdown` → tracker settings → `filterIssuesByPickupLabels`), since a core-library change has no preview environment:

```
front matter: include: [" Agent ", "agent", "  "], exclude: [" BLOCKED "]
parsed:       {"include":["agent"],"exclude":["blocked"]}
A-1 ["AGENT"]              -> dispatchable
A-2 ["agent","Blocked"]    -> Issue has excluded pickup label "blocked".
A-3 ["other"]              -> Issue is missing a required pickup label (agent).
A-4 ["  Agent  "]          -> dispatchable
```

Dedupe, blank-drop, exclude-beats-include, and whitespace on the *tracker* side all behave. Also ran `parseTrackerTimestamp` over 31 inputs: offsets, lowercase designators, leap seconds across offsets and fractional seconds, calendar rollover (`2026-02-30` → `null`), non-leap-year `2026-02-29` → `null` (correct), zero-padded years, and every out-of-range component → `null`. No throw on any input, including non-strings.

### Requirement coverage against #725

All 완료 기준 met: helper unit tests cover case / whitespace / dupes / blanks and malformed-timestamp → `null`; the ADR is merged and now factually matches the sorter; `docs/configuration.md` has the normalization section; and 코드 동작 변화 없음 holds in the sense the issue meant it — `parseTrackerTimestamp` is exported but unadopted, and adapter work stays in #660-B. `required_labels` remains N/A (not in the tree); the doc correctly frames the helper as the shared implementation for it.

None. The diff is 244/35 across 15 files and every line traces to either #725's scope or a round-1 finding. Extending normalization to tracker-provided labels remains necessary rather than scope creep. I re-checked the one thing that would have worried me — whether lowercased config labels leak into a server-side tracker query, where Linear's label filtering could be case-sensitive — and they don't: `pickupLabels` is consumed only by `filterIssuesByPickupLabels`, the doctor summary, and the project-overlap guard, all client-side. `generate-workflow-md.ts` writes labels from init input, not from a parsed workflow, so there's no round-trip that would rewrite an operator's `WORKFLOW.md` to lowercase.

### Nits (non-blocking, both inline)

1. `normalization.ts:49` — the `.replace("t","T").replace("z","Z")` pair is correct only because of the regex's alphabet; anchoring it makes that self-evident.
2. `normalization.ts:50-61` — `:60` is accepted at any time of day and coerced forward; spec-literal and fine, but worth one JSDoc line before #660-B adopters meet it.

One more for the record, no action needed: decision item 1 says "every numeric priority", and `NaN` is technically a `number` that would make the comparator inconsistent. Neither `resolvePriority` nor the Linear adapter can produce it, so this is theoretical — noting it only so it isn't rediscovered as a surprise.

Approving. Ship it.

---

hojinzs:

head `848eed8`(approve, CI green)를 격리 worktree에 빌드해 확인했습니다.

| 검증 | 결과 |
| --- | --- |
| `normalizeLabels([" Ready ","ready","BUG","","  ","Bug"])` → `["ready","bug"]` (trim/lowercase/blank drop/dedupe) | ✅ |
| `parseTrackerTimestamp`: `2026-08-27T09:00:00+09:00` → `2026-08-27T00:00:00.000Z`, 잘못된 값/null → null | ✅ |
| ADR `2026-08-28_priority-mapping-documented-different-mapping.md` 존재, Prettier 통과 | ✅ |
| 로컬 E2E: `pickup_labels.include: [" Ready-Label "]` vs 이슈 라벨 `ready-label` → #1만 dispatch, 라벨 없는 #2 제외(대소문자·공백 무시) | ✅ |

참고: 트래커 설정은 `repo init` 시 `config.json`에 고정되므로(리포트 C10) 스모크에서 init을 다시 해야 반영됐습니다.

PR #753, opened 2026-08-29
fix(worker): restore host tool rate-limit updates

### LINKED ISSUE #744: fix(worker): restore dynamic tool rate-limit guard and rateLimits observability on the host-side tool path
## 배경
#730 (PR #735)에서 `dispatchDynamicToolCall` 경로가 제거되면서, 그 경로가 하던 두 가지가 함께 사라졌습니다:

1. 호출 전 `guardDynamicToolRateLimit` 실행 — `packages/worker/src/tool-rate-limit.ts`는 이제 importer가 0개.
2. 툴 응답의 `rateLimits`를 `applyRateLimitUpdate("tool.<name>", …, "agent-tool")`와 `agent.rateLimit` 채널 이벤트로 올리던 관측성 — 새 호스트 경로에서는 GitHub rate-limit payload(`{"source":"github","limit":5000,"remaining":…}`)가 에이전트 텍스트로만 전달되고 worker/orchestrator 상태에는 반영되지 않음.

(#735 머지 후 스모크 실행 B에서 확인.)

## 할 일
- 호스트 측 `item/tool/call` 처리 경로에 rate-limit guard와 `applyRateLimitUpdate` 연결 복원(또는 의도적으로 제거한다면 `tool-rate-limit.ts` 삭제 + 문서/ADR 반영).
- 단위 테스트: 툴 응답의 `rateLimits`가 worker 상태·이벤트에 반영되는지.
- changeset(`@gh-symphony/cli`, patch).

관련: #730, #735, #743, Epic #651. 영향도: 중 (관측성/보호 회귀, 기능 차단은 아님).

moncher-dev:
## TL;DR

Restores host-side GitHub dynamic-tool rate-limit protection and publishes tool `rateLimits` to worker/orchestrator observability. A rejected rate-limit guard now returns an immediate structured JSON-RPC tool error instead of leaving the Codex turn waiting for its timeout.

## Change-point diagram

`item/tool/call` → guard saved GitHub budget → execute host-side tool → apply `rateLimits` → `agent.rateLimit` event
                    └→ rejected guard → structured `rate_limit_guard` response

## Start here

- `packages/worker/src/index.ts` invokes the rate-limited production call boundary and publishes response-derived telemetry.
- `packages/worker/src/host-dynamic-tool-call.ts` converts guard rejection into the tool-error envelope sent by the JSON-RPC handler.
- `packages/worker/src/host-dynamic-tool-call.test.ts` proves exhausted GitHub budget blocks execution and returns that envelope.

## Risks & rollback

- Risk: malformed tool payloads may omit rate-limit data; telemetry remains conditional without affecting a successful tool response.
- Rollback: revert this PR to restore the current host-side behavior.

## Changed files

- `packages/worker/src/index.ts`
- `packages/worker/src/host-dynamic-tool-call.ts`
- `packages/worker/src/host-dynamic-tool-call.test.ts`
- `packages/worker/src/worker-protocol.test.ts`
- `.changeset/bright-tools-guard.md`

## Evidence

- `pnpm --filter @gh-symphony/worker test -- host-dynamic-tool-call.test.ts tool-rate-limit.test.ts worker-protocol.test.ts` (67 passed)
- `pnpm lint && pnpm test && pnpm typecheck && pnpm build`
- Docker: `docker compose --project-name symphony-e2e-744 -f docker-compose.e2e.yml exec -T symphony-e2e node /app/e2e/host-dynamic-tool-e2e.mjs` (`host_dynamic_tool_e2e=pass`)

## Scope boundary

Claude host MCP has a separate execution path and is not part of the Codex `item/tool/call` regression fixed by #744. Its rate-limit parity is intentionally not expanded into this patch.

- [ ] Exercise a host-side GitHub tool call and verify its rate-limit event appears in worker/orchestrator telemetry.

## Human Validation

- [ ] Confirm a rejected GitHub rate-limit guard returns a tool error promptly rather than stalling the turn.
- [ ] Confirm host-side tool behavior is unchanged apart from rate-limit protection and telemetry.
- [ ] Confirm rate-limit data reaches worker status/event consumers.

### DIFF
--- a/packages/worker/src/host-dynamic-tool-call.ts
+++ b/packages/worker/src/host-dynamic-tool-call.ts
@@ -1,0 +1,33 @@
+import type { CodexDynamicToolCallResponse } from "./codex-dynamic-tools.js";
+import { guardDynamicToolRateLimit } from "./tool-rate-limit.js";
+
+export async function executeRateLimitedCodexDynamicToolCall(options: {
+  toolName: string;
+  rateLimits: Record<string, unknown> | null;
+  execute: () => Promise<CodexDynamicToolCallResponse>;
+}): Promise<CodexDynamicToolCallResponse> {
+  try {
+    await guardDynamicToolRateLimit(options.toolName, options.rateLimits);
+    return await options.execute();
+  } catch (error) {
+    return failure(
+      "rate_limit_guard",
+      error instanceof Error ? error.message : String(error)
+    );
+  }
+}
+
+function failure(
+  code: string,
+  message: string
+): CodexDynamicToolCallResponse {
+  return {
+    success: false,
+    contentItems: [
+      {
+        type: "inputText",
+        text: JSON.stringify({ error: { code, message } }),
+      },
+    ],
+  };
+}
--- a/packages/worker/src/index.ts
+++ b/packages/worker/src/index.ts
@@ -79,6 +79,10 @@ import {
   executeCodexDynamicToolCall,
 } from "./codex-dynamic-tools.js";
 import {
+  extractToolRateLimitPayload,
+  guardDynamicToolRateLimit,
+} from "./tool-rate-limit.js";
+import {
   buildCodexDynamicToolsParams,
   buildCodexInitializeParams,
 } from "./codex-initialize.js";
@@ -1516,15 +1520,33 @@ async function runCodexClientProtocol(
       process.stderr.write(
         `[worker] executing host dynamic tool "${toolName}" (callId=${callId})\n`
       );
-      void executeCodexDynamicToolCall(
+      void guardDynamicToolRateLimit(
         toolName,
-        params?.arguments,
-        createTrackerToolContext(env),
-        env,
-        {},
-        plan.dynamicTools.map((tool) => tool.name)
+        runtimeState.agentGitHubRateLimits ?? runtimeState.rateLimits
       )
+        .then(() =>
+          executeCodexDynamicToolCall(
+            toolName,
+            params?.arguments,
+            createTrackerToolContext(env),
+            env,
+            {},
+            plan.dynamicTools.map((tool) => tool.name)
+          )
+        )
         .then((result) => {
+          const toolRateLimits = extractToolRateLimitPayload(
+            result.contentItems.find((item) => item.type === "inputText")
+              ?.text ?? ""
+          );
+          if (toolRateLimits) {
+            applyRateLimitUpdate(
+              `tool.${toolName}`,
+              toolRateLimits,
+              "agent-tool"
+            );
+            emitOrchestratorChannelEvent("agent.rateLimit");
+          }
           sendMessage({ jsonrpc: "2.0", id: msg.id, result });
           runtimeState.lastEventAt = new Date().toISOString();
         })
--- a/packages/worker/src/index.ts
+++ b/packages/worker/src/index.ts
@@ -80,8 +80,8 @@ import {
 } from "./codex-dynamic-tools.js";
 import {
   extractToolRateLimitPayload,
-  guardDynamicToolRateLimit,
 } from "./tool-rate-limit.js";
+import { executeRateLimitedCodexDynamicToolCall } from "./host-dynamic-tool-call.js";
 import {
   buildCodexDynamicToolsParams,
   buildCodexInitializeParams,
@@ -1520,11 +1520,10 @@ async function runCodexClientProtocol(
       process.stderr.write(
         `[worker] executing host dynamic tool "${toolName}" (callId=${callId})\n`
       );
-      void guardDynamicToolRateLimit(
+      void executeRateLimitedCodexDynamicToolCall({
         toolName,
-        runtimeState.agentGitHubRateLimits ?? runtimeState.rateLimits
-      )
-        .then(() =>
+        rateLimits: runtimeState.agentGitHubRateLimits ?? runtimeState.rateLimits,
+        execute: () =>
           executeCodexDynamicToolCall(
             toolName,
             params?.arguments,
@@ -1532,12 +1531,11 @@ async function runCodexClientProtocol(
             env,
             {},
             plan.dynamicTools.map((tool) => tool.name)
-          )
-        )
+          ),
+      })
         .then((result) => {
           const toolRateLimits = extractToolRateLimitPayload(
-            result.contentItems.find((item) => item.type === "inputText")
-              ?.text ?? ""
+            result.contentItems[0]?.text ?? ""
           );
           if (toolRateLimits) {
             applyRateLimitUpdate(
--- a/packages/worker/src/worker-protocol.test.ts
+++ b/packages/worker/src/worker-protocol.test.ts
@@ -29,6 +29,10 @@ import {
   createTrackerToolContext,
   executeCodexDynamicToolCall,
 } from "./codex-dynamic-tools.js";
+import {
+  extractToolRateLimitPayload,
+  guardDynamicToolRateLimit,
+} from "./tool-rate-limit.js";
 
 const TEST_DYNAMIC_TOOLS = [
   {
@@ -180,6 +184,7 @@ function createProtocolContext(options: {
     tokenUsage: { ...tokenUsageBaseline },
     lastEventAt: null as string | null,
     rateLimits: null as Record<string, unknown> | null,
+    agentGitHubRateLimits: null as Record<string, unknown> | null,
     sessionInfo: {
       threadId: "thread-1" as string | null,
       turnId: "turn-1" as string | null,
@@ -207,12 +212,19 @@ function createProtocolContext(options: {
 
   function applyRateLimitUpdate(
     source: string,
-    rateLimits: Record<string, unknown>
+    rateLimits: Record<string, unknown>,
+    runtimeSource = "codex"
   ): void {
     runtimeState.rateLimits = {
       ...rateLimits,
-      source: "codex",
+      source:
+        typeof rateLimits.source === "string"
+          ? rateLimits.source
+          : runtimeSource,
     };
+    if (runtimeState.rateLimits.source === "github") {
+      runtimeState.agentGitHubRateLimits = runtimeState.rateLimits;
+    }
     logs.push(
       `[worker] rate_limits source=${source} payload=${JSON.stringify(runtimeState.rateLimits)}`
     );
@@ -923,15 +935,33 @@ function createProtocolContext(options: {
       runtimeState.lastEventAt = new Date().toISOString();
       const params = msg.params as Record<string, unknown> | undefined;
       const toolName = typeof params?.tool === "string" ? params.tool : "";
-      void executeCodexDynamicToolCall(
+      void guardDynamicToolRateLimit(
         toolName,
-        params?.arguments,
-        createTrackerToolContext(TEST_TOOL_ENV),
-        TEST_TOOL_ENV,
-        {},
-        TEST_DYNAMIC_TOOLS.map((tool) => tool.name)
+        runtimeState.agentGitHubRateLimits ?? runtimeState.rateLimits
       )
+        .then(() =>
+          executeCodexDynamicToolCall(
+            toolName,
+            params?.arguments,
+            createTrackerToolContext(TEST_TOOL_ENV),
+            TEST_TOOL_ENV,
+            {},
+            TEST_DYNAMIC_TOOLS.map((tool) => tool.name)
+          )
+        )
         .then((result) => {
+          const toolRateLimits = extractToolRateLimitPayload(
+            result.contentItems.find((item) => item.type === "inputText")
+              ?.text ?? ""
+          );
+          if (toolRateLimits) {
+            applyRateLimitUpdate(
+              `tool.${toolName}`,
+              toolRateLimits,
+              "agent-tool"
+            );
+            emitOrchestratorChannelEvent("agent.rateLimit");
+          }
           sendMessage({ jsonrpc: "2.0", id: msg.id, result });
           runtimeState.lastEventAt = new Date().toISOString();
         })
@@ -1679,7 +1709,7 @@ describe("rate-limit telemetry", () => {
     });
 
     expect(ctx.runtimeState.rateLimits).toEqual({
-      source: "codex",
+      source: "upstream",
       remaining: 42,
       resetAt: "2026-03-08T00:30:00.000Z",
     });
@@ -2703,9 +2733,28 @@ describe("token usage tracking", () => {
 describe("lastEventAt timestamp tracking", () => {
   it("responds to a host dynamic tool call and records tool activity", async () => {
     const fetchSpy = vi.spyOn(globalThis, "fetch").mockResolvedValue(
-      new Response(JSON.stringify({ data: { viewer: { login: "octo" } } }), {
-        status: 200,
-      })
+      new Response(
+        JSON.stringify({
+          data: {
+            viewer: { login: "octo" },
+            __ghSymphonyRateLimit: {
+              cost: 3,
+              remaining: 4_997,
+              resetAt: "2026-08-29T10:00:00.000Z",
+            },
+          },
+        }),
+        {
+          status: 200,
+          headers: {
+            "x-ratelimit-limit": "5000",
+            "x-ratelimit-remaining": "4997",
+            "x-ratelimit-used": "3",
+            "x-ratelimit-reset": "1787997600",
+            "x-ratelimit-resource": "graphql",
+          },
+        }
+      )
     );
     const ctx = createProtocolContext({});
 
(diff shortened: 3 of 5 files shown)


packages/worker/src/index.ts
```diff
@@ -1516,15 +1520,33 @@ async function runCodexClientProtocol(
       process.stderr.write(
         `[worker] executing host dynamic tool "${toolName}" (callId=${callId})\n`
       );
-      void executeCodexDynamicToolCall(
+      void guardDynamicToolRateLimit(
```
hojinzs: **P1 — guard rejection swallows the JSON-RPC response and stalls the turn**

`guardDynamicToolRateLimit` can **reject**, not just resolve. When the cached GitHub budget is empty, `guardGitHubGraphQLRateLimitState` throws `GitHubGraphQLRateLimitError` (`github-rate-limit.ts:253` and `:263`, `DEFAULT_HARD_THRESHOLD = 0`).

Putting it at the head of this chain makes that rejection land in the existing `.catch` at line 1553, which only writes to stderr — it never calls `sendMessage`. But `item/tool/call` is a JSON-RPC **request** (`msg.id != null`), so the app-server is left awaiting a response it never gets. The turn hangs until `turnTimeoutMs` (line 1146, default 1h) kills the codex process and fails the turn.

Before this PR that `.catch` was effectively unreachable: `executeCodexDynamicToolCall` catches internally and always resolves with a `failure()` envelope, so nothing on this chain could reject. This change introduces the first real rejection path into a handler that has no response-on-error branch.

This is also a divergence from the code #744 asks to restore. The pre-#735 `dispatchDynamicToolCall` ran the guard inside `runToolProcess` under a `try`, and its `catch` **still sent a response** with `isError: true`, so the agent got immediate feedback and could adapt:

```js
} catch (err) {
  const errMsg = err instanceof Error ? err.message : String(err);
  process.stderr.write(`[worker] tool "${toolName}" failed: ${errMsg}\n`);
  sendMessage({ /* … */ contentItems: [{ type: "input_text", text: errMsg }], isError: true });
}
```

Reproduced locally against this branch — the guard rejects, and the chain shape here yields zero sent messages:

```ts
const exhausted = { source: "github", resource: "graphql", limit: 5000,
                    remaining: 0, resetAt: "2099-01-01T00:00:00.000Z" };
await expect(guardDynamicToolRateLimit("github_graphql", exhausted)).rejects.toThrow(/rate limit/i);
// replaying this exact chain: sent === []  (stderr got the only trace)
```

Suggested fix — turn the rejection into the tool-error envelope the agent already understands, so `sendMessage` always runs:

```ts
void guardDynamicToolRateLimit(
  toolName,
  runtimeState.agentGitHubRateLimits ?? runtimeState.rateLimits
)
  .then(() => executeCodexDynamicToolCall(/* … */))
  .catch((error: unknown) => ({
    success: false,
    contentItems: [
      {
        type: "inputText" as const,
        text: JSON.stringify({
          error: {
            code: "rate_limit_guard",
            message: error instanceof Error ? error.message : String(error),
          },
        }),
      },
    ],
  }))
  .then((result) => { /* existing rate-limit + sendMessage body */ });
```

Note the guard's other branch is fine: when `remaining` is under the soft threshold but the reset is within `maxWaitMs`, it sleeps and then resolves, which is the intended backpressure.

---

packages/worker/src/worker-protocol.test.ts
```diff
@@ -923,15 +935,33 @@ function createProtocolContext(options: {
       runtimeState.lastEventAt = new Date().toISOString();
       const params = msg.params as Record<string, unknown> | undefined;
       const toolName = typeof params?.tool === "string" ? params.tool : "";
-      void executeCodexDynamicToolCall(
+      void guardDynamicToolRateLimit(
```
hojinzs, line 938: **P2 — this test asserts against a hand-copied replica, not against `index.ts`**

`createProtocolContext` reimplements the `item/tool/call` handler, so this block is a second copy of the production code rather than a test of it. The coverage bot on this PR confirms it: `packages/worker/src/index.ts` reports **0% lines / 0% statements** on the changed file.

Concretely, the replica reproduces the P1 above verbatim — same chain shape, same `.catch` that logs without responding — so no assertion added here can ever surface that defect. The two copies can also drift silently; this PR is itself repairing one such drift (the harness had hardcoded `source: "codex"` while production preserves an incoming string source, hence the `"upstream"` correction at line 1712 — that part is a genuine fix, good catch).

Extracting the handler out of `index.ts` is a real refactor and I'm not asking for it in a patch-scope PR. The narrow ask: add a test that drives **`guardDynamicToolRateLimit` itself** through the rejection path in `tool-rate-limit.test.ts` (which does import production code), asserting the caller's error envelope. `tool-rate-limit.test.ts:35` already sets up an exhausted-budget payload you can reuse. That covers the one branch this PR newly makes reachable without the wider restructure.

---

packages/worker/src/index.ts
```diff
@@ -1516,15 +1520,33 @@ async function runCodexClientProtocol(
...
+            plan.dynamicTools.map((tool) => tool.name)
+          )
+        )
         .then((result) => {
+          const toolRateLimits = extractToolRateLimitPayload(
+            result.contentItems.find((item) => item.type === "inputText")
+              ?.text ?? ""
+          );
+          if (toolRateLimits) {
+            applyRateLimitUpdate(
```
hojinzs, line 1541: **nit** — `contentItems` is typed `Array<{ type: "inputText"; text: string }>` (`codex-dynamic-tools.ts:17`), and both `success()` and `failure()` emit exactly one element. The `.find()` predicate can only ever match index 0, so `result.contentItems[0]?.text ?? ""` says the same thing more directly. Non-blocking either way — `extractToolRateLimitPayload("")` returns `null` safely.

---

hojinzs (changes requested):

### Requirements vs. #744

| #744 요구사항 | 상태 |
|---|---|
| 호스트 `item/tool/call` 경로에 `guardDynamicToolRateLimit` 복원 | ✅ |
| 툴 응답 `rateLimits` → `applyRateLimitUpdate("tool.…", …, "agent-tool")` + `agent.rateLimit` 채널 이벤트 | ✅ |
| 단위 테스트: 툴 응답 rateLimits가 worker 상태·이벤트에 반영 | ⚠️ 커버되긴 하나 프로덕션 코드가 아닌 테스트 복제본을 검증 (P2) |
| changeset (`@gh-symphony/cli`, patch) | ✅ |

Scope is tight and I found **no over-engineering** — the diff restores what was lost in #735 and nothing more. The `source: "codex"` → `"upstream"` assertion change is a correct repair of a stale harness, not a weakened test.

### Verification performed

- `pnpm install --frozen-lockfile` → ok
- `pnpm --filter @gh-symphony/worker test` → **185 passed / 25 files**
- `pnpm typecheck` → clean
- CI on `62e5ecb`: Test ✅, Container Smoke ✅
- Smoke test of the restored guard path against this branch (see P1) — reproduced the defect

No preview environment applies to this change (worker-internal protocol path), so I exercised it locally against the branch.

### Findings

| # | Sev | Summary |
|---|---|---|
| 1 | **P1** | A `guardDynamicToolRateLimit` rejection is swallowed by the existing `.catch`, so no JSON-RPC response is sent for a request carrying an `id` — the turn stalls until the 1h `turnTimeoutMs`. Diverges from the pre-#735 code, which responded with `isError: true`. |
| 2 | P2 | New assertions run against `createProtocolContext`'s hand-copied handler, not `index.ts` (coverage bot: 0% on the changed file). The replica carries the same P1 defect, so it structurally cannot catch it. |
| 3 | nit | `.find(item => item.type === "inputText")` over a single-element, single-typed array. |

Only **P1 blocks**. It is a small, local fix and I've left a concrete patch inline. P2 asks for one added test in `tool-rate-limit.test.ts` rather than the full handler refactor — happy to treat that as follow-up if you'd prefer to keep this PR patch-sized, just say so on the thread. P3 is optional.

---

hojinzs:
**리뷰 보완 (독립 검증)** — 1회차 P1(가드 거부 시 `item/tool/call`에 응답이 없어 Codex가 영구 대기, `index.ts:1523-1559`의 `.catch`가 stderr만 기록)을 빌드된 dist로 재현: `{"executorRan":false,"jsonrpcResponseSent":false}`. 추가 2건: (1) 거부 경로 테스트 부재 — 테스트 하네스가 핸들러 복제본이라 결함을 그대로 물려받음, 실제 핸들러를 통과하는 테스트로. (2) Claude 호스트 MCP 경로(#742, `runtime-claude/src/mcp-http-server.ts:94-125`)는 가드도 rateLimits 발행도 없음 — #744 범위 밖이면 PR 본문에 명시하고 후속 이슈로. 참고: 정상 경로는 실제 GitHub 호출로 `agent.rateLimit`(source=github, remaining=4940) 이벤트·run.json 반영 확인.

packages/worker/src/index.ts
```diff
@@ -1516,15 +1520,33 @@ async function runCodexClientProtocol(
...
-      void executeCodexDynamicToolCall(
+      void guardDynamicToolRateLimit(
         toolName,
-        params?.arguments,
-        createTrackerToolContext(env),
-        env,
-        {},
-        plan.dynamicTools.map((tool) => tool.name)
+        runtimeState.agentGitHubRateLimits ?? runtimeState.rateLimits
       )
```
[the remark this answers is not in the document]
moncher-dev: Resolved in f88d3037. The production `item/tool/call` path now calls `executeRateLimitedCodexDynamicToolCall`, which converts a rejected GitHub rate-limit guard into the structured `rate_limit_guard` tool envelope before the handler sends its JSON-RPC response. The executor is not invoked when the guard rejects.

packages/worker/src/index.ts
moncher-dev: Resolved in f88d3037. The production `item/tool/call` path now calls `executeRateLimitedCodexDynamicToolCall`, which converts a rejected GitHub rate-limit guard into the structured `rate_limit_guard` tool envelope before the handler sends its JSON-RPC response. The executor is not invoked when the guard rejects.

packages/worker/src/worker-protocol.test.ts
moncher-dev: Resolved in f88d3037. The new `host-dynamic-tool-call.test.ts` imports and exercises the production helper that `index.ts` invokes, using an exhausted GitHub snapshot. It asserts the structured `rate_limit_guard` response and confirms the executor is never called, rather than validating the copied protocol-test handler.

packages/worker/src/index.ts
moncher-dev: Resolved in f88d3037. The response has exactly one `inputText` item, so the production handler now reads `result.contentItems[0]?.text ?? ""` directly.

moncher-dev:
Resolved in f88d3037. The production `item/tool/call` path now calls `executeRateLimitedCodexDynamicToolCall`, which converts a rejected GitHub rate-limit guard into the structured `rate_limit_guard` tool envelope before the handler sends its JSON-RPC response. The executor is not invoked when the guard rejects.

packages/worker/src/host-dynamic-tool-call.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export async function executeRateLimitedCodexDynamicToolCall(options: {
+  toolName: string;
+  rateLimits: Record<string, unknown> | null;
+  execute: () => Promise<CodexDynamicToolCallResponse>;
+}): Promise<CodexDynamicToolCallResponse> {
+  try {
+    await guardDynamicToolRateLimit(options.toolName, options.rateLimits);
+    return await options.execute();
+  } catch (error) {
```
hojinzs, line 12: **nit** (non-blocking, no change required to merge) — the `catch` covers `options.execute()` as well as the guard, so any rejection from the executor would also be labelled `rate_limit_guard`.

I checked whether this is reachable and it is not: `executeCodexDynamicToolCall` wraps its whole body in `try/catch` and always *resolves* with a `failure("tool_execution_failed", …)` envelope (`codex-dynamic-tools.ts:164`). I confirmed it empirically — with `fetch` rejecting, this chain still returns `code: "tool_execution_failed"`, not `rate_limit_guard`. So today the label is correct in every reachable case.

It is only a latent trap for a future `execute` that can throw. If you want it airtight, moving `return await options.execute()` out of the `try` says the intent directly:

```ts
try {
  await guardDynamicToolRateLimit(options.toolName, options.rateLimits);
} catch (error) {
  return failure("rate_limit_guard", error instanceof Error ? error.message : String(error));
}
return options.execute();
```

Also minor: `failure()` here is a byte-for-byte copy of the private `failure()` in `codex-dynamic-tools.ts:135`. Exporting that one keeps the tool-error envelope shape defined in a single place. Both points are cosmetic — fine as follow-up or not at all.

---

packages/worker/src/worker-protocol.test.ts
hojinzs, line 938: **nit** (non-blocking) — this replica now models the *pre-fix* shape rather than production.

`index.ts` calls `executeRateLimitedCodexDynamicToolCall`; this copy still does `guardDynamicToolRateLimit(…).then(execute)` with the shared `.catch` that only logs. So the replica still contains the exact P1 defect that f88d3037 removed from production — a guard rejection here sends no JSON-RPC response.

That does not weaken this PR: the branch is now covered where it matters, in `host-dynamic-tool-call.test.ts` against the real helper, which is precisely the narrow ask from round 1. But since the two copies exist to mirror each other, keeping this one on the old shape is the drift that motivated the P2 in the first place. A same-shape edit removes it:

```ts
void executeRateLimitedCodexDynamicToolCall({
  toolName,
  rateLimits: runtimeState.agentGitHubRateLimits ?? runtimeState.rateLimits,
  execute: () =>
    executeCodexDynamicToolCall(
      toolName,
      params?.arguments,
      createTrackerToolContext(TEST_TOOL_ENV),
      TEST_TOOL_ENV,
      {},
      TEST_DYNAMIC_TOOLS.map((tool) => tool.name)
    ),
})
  .then((result) => { /* unchanged */ })
```

Same imports, same assertions, and the happy-path `agent.rateLimit` test at line 2788 then exercises the real wrapper. Your call — I am not blocking on it.

---

hojinzs (approved):
## Review — round 2/6 · **Approve**

Round-1 P1 is genuinely fixed, and I verified it rather than taking the thread reply at face value.

### Round 1 follow-up

| # | Round-1 finding | Status |
|---|---|---|
| 1 | **P1** — guard rejection swallowed, no JSON-RPC response, turn stalls to `turnTimeoutMs` | ✅ Fixed in `f88d3037`. `executeRateLimitedCodexDynamicToolCall` converts the rejection into the `rate_limit_guard` envelope, so `sendMessage` always runs. |
| 2 | **P2** — assertions only hit the hand-copied replica | ✅ Addressed as scoped. `host-dynamic-tool-call.test.ts` imports and drives the production helper through the rejection branch. |
| 3 | **nit** — `.find()` over a single-element array | ✅ Now `result.contentItems[0]?.text ?? ""`. |

### Requirements vs. #744

| #744 요구사항 | 상태 |
|---|---|
| 호스트 `item/tool/call` 경로에 `guardDynamicToolRateLimit` 복원 | ✅ |
| 툴 응답 `rateLimits` → `applyRateLimitUpdate("tool.…", …, "agent-tool")` + `agent.rateLimit` 채널 이벤트 | ✅ 실측 확인 |
| 단위 테스트: 툴 응답 rateLimits가 worker 상태·이벤트에 반영 | ✅ |
| changeset (`@gh-symphony/cli`, patch) | ✅ |

### Smoke test

No preview environment applies — this is a worker-internal JSON-RPC path — so I replayed the exact `index.ts` `item/tool/call` chain against the real helpers on `f88d3037`, driving `executeRateLimitedCodexDynamicToolCall` → `executeCodexDynamicToolCall` → `extractToolRateLimitPayload` with `fetch` stubbed. **5/5 pass:**

| Scenario | Result |
|---|---|
| A — exhausted budget (`remaining: 0`, reset beyond max wait) | Responds with `success: false` / `code: "rate_limit_guard"`; executor never invoked; no bogus telemetry. **The round-1 stall is gone.** |
| B — healthy budget | Tool executes; `applyRateLimitUpdate("tool.github_graphql", …, "agent-tool")` fires with `source: github, remaining: 4997`; `agent.rateLimit` emitted. **This is the #744 observability regression, closed.** |
| C — cached `source: "codex"` snapshot | Guard correctly declines to fire; call succeeds. No false-positive throttling. |
| D — executor failure (`fetch` rejects) | Keeps `tool_execution_failed`; not mislabelled as a guard rejection. |
| E — soft-threshold, reset within max wait | Sleeps, then executes and succeeds. Intended backpressure intact. |

### Verification performed

- `pnpm --filter @gh-symphony/worker test` → **186 passed / 26 files**
- `pnpm typecheck` → clean · `pnpm lint` → clean
- CI on `f88d3037`: Test ✅, Container Smoke ✅
- `mergeable_state: unstable` is the pending-review signal only; no merge conflict against `main`.

### Scope

No over-engineering. 5 files, +171/−19, and the new module is 33 lines that do one thing. The documented scope boundary — Claude host MCP runs a separate execution path and is out of the #744 Codex `item/tool/call` regression — is correct and correctly left alone rather than opportunistically widened.

### Remaining

Two **nits** inline, neither blocking and neither needing a push before merge: the wrapper's `catch` also spans `execute()` (I confirmed this is unreachable today, so the label is correct in practice), and `worker-protocol.test.ts` still mirrors the pre-fix chain shape. Take them here or as follow-up.

**Approving.** The regression #744 describes is closed and demonstrably exercised.

---

PR #754, opened 2026-08-29
feat(trackers): add agent tool execution contract

### LINKED ISSUE #732: feat(core,tools,docs): tool execution context contract in tracker adapters, per-tool documentation, and Docker E2E for host-side tools
Parent: #673 (Epic #651) · ADR `docs/adr/2026-08-28_agent-tool-isolation.md` Phase 2
리포트 항목: **A13 구현 — Phase 2, E2, A16(툴)**

병합 게이트(공통): 각 조각은 host-side 경로를 **추가**만 하므로 독립 병합 가능. 단 #700(1b, MCP subprocess 비활성화)은 A·B **모두** 병합된 뒤에만 진행.

#673 분할 3/3 — **공통 계약·문서·E2E**.

## 범위
- 어댑터 계약: `agentToolSpecs()` / `executeAgentTool(name, args, context={issue})`(스펙 §10.5 hooks)를 core 계약으로 정식화하고 GitHub/Linear 어댑터가 구현 — A/B가 임시로 둔 호출 경로를 계약으로 수렴
- 툴 문서(§10.5 MUST): 이름·입력 스키마, mutation 가능 여부, scope 제한(§15.5), 결과/오류 형태, rate-limit 기대치 — `docs/trackers/<kind>.md`(#674와 조율)
- Docker E2E: Codex/Claude 워커가 host-side 툴로 트래커 전이·코멘트 수행(#634 시나리오), child env/파일에 토큰 없음 단언
- `AGENT_TEST.md` 시나리오 표, ADR "conformance" 절 갱신

## 완료 기준
- 계약 단위테스트, Docker E2E 통과, 문서 병합

Blocked by: #730, #731

moncher-dev:
## TL;DR

Formalizes adapter-owned host-side tracker tools for Codex and Claude, preserving loaded host configuration and executing each advertised tool through its owning tracker adapter.

## Change-point diagram

```text
Codex dynamic tool / Claude MCP tool
              |
              v
TrackerAdapter.agentToolSpecs() + executeAgentTool(..., { issue, env })
              |
              +---> GitHub GraphQL adapter
              +---> Linear GraphQL adapter
```

## Start here

- `packages/core/src/contracts/tracker-adapter.ts` defines the shared tool specification and execution context.
- `packages/worker/src/codex-dynamic-tools.ts` routes each advertised tool to its owning adapter.
- `packages/runtime-claude/src/mcp-http-server.ts` consumes the same adapter-owned tool contract.
- `e2e/scenarios/19-host-side-tracker-tools.md` documents the Docker acceptance scenario.

## Changes

- Added `agentToolSpecs()` and `executeAgentTool()` to GitHub and Linear adapters, using normalized issue context and the loaded host environment.
- Unified Codex and Claude tool schemas/execution; Linear-tracked sessions retain GitHub GraphQL alongside Linear GraphQL.
- Added per-tool tracker documentation, ADR conformance notes, TC-19, scoped comment/Project-state mutation coverage, and child credential-boundary checks.
- Rebasing onto current `main` preserved the new TC-19 table row with the current repository regression matrix.

## Evidence

- `pnpm lint` — pass.
- `pnpm test` — pass.
- `pnpm typecheck` — pass.
- `pnpm build` — pass.
- `docker compose --project-name symphony-732-rework -f docker-compose.e2e.yml exec -T symphony-e2e node /app/e2e/host-dynamic-tool-e2e.mjs` — pass.
- `pnpm e2e:claude` — pass (5 Docker tests).

## Risks & rollback

The shared contract is now the host-tool execution boundary. Reverting this PR restores the previous routing without data migration or tracker-state changes.

## Changed files

Core contract; GitHub/Linear adapters; Codex/Claude transports; unit and Docker E2E tests; tracker documentation; ADR; architecture/test guidance; CLI changeset.

- [ ] Confirm a real GitHub-tracked worker can read and comment on its active issue.
- [ ] Confirm a Linear-tracked worker can use both configured GitHub and Linear host-side tools.
- [ ] Confirm deployment configuration supplies only expected host-side provider credentials.


packages/tracker-github/src/orchestrator-adapter.ts
```diff
@@ -46,6 +50,53 @@ export const githubProjectTrackerAdapter: OrchestratorTrackerAdapter = {
...
+      throw new Error(`Unknown GitHub agent tool: ${name}`);
+    }
+    return executeGitHubGraphQL(
+      args as GitHubGraphQLInvocation,
+      {
+        token: process.env.GITHUB_GRAPHQL_TOKEN,
+        apiUrl: process.env.GITHUB_GRAPHQL_API_URL,
+        tokenBrokerUrl: process.env.GITHUB_TOKEN_BROKER_URL,
+        tokenBrokerSecret: process.env.GITHUB_TOKEN_BROKER_SECRET,
+        tokenCachePath: process.env.GITHUB_TOKEN_CACHE_PATH,
```
hojinzs: **P1 — credential regression: `process.env` instead of the threaded worker env.**

Before this PR, `codex-dynamic-tools.ts` built the config from the `env` argument (`githubConfig(env)`). The worker passes `launcherEnv = loadLauncherEnvironment(process.env)` (`packages/worker/src/index.ts:86`, `:1522`), and `loadLauncherEnvironment` merges `<pkg>/.env` and `<cwd>/.env` **on top of** `process.env` (`packages/runtime-codex/src/launcher.ts:92-100`). So a credential configured only in a `.env` file exists in `env` but **not** in `process.env`. Reading `process.env` here silently drops it.

Reproduced against this branch's build:

```js
delete process.env.GITHUB_GRAPHQL_TOKEN;           // .env-sourced, not exported
const env = { GITHUB_GRAPHQL_TOKEN: "dotenv-sourced-token", GITHUB_GRAPHQL_API_URL: "...", ... };
await executeCodexDynamicToolCall("github_graphql", { query: "query { viewer { login } }" },
                                  createTrackerToolContext(env), env);
// -> {"success":false,"error":{"code":"tool_execution_failed",
//     "message":"Either GITHUB_GRAPHQL_TOKEN or the runtime token broker configuration is required."}}
// provider request: null
```

This applies to all five keys here (token **and** the whole broker path: `GITHUB_TOKEN_BROKER_URL/SECRET`, `GITHUB_TOKEN_CACHE_PATH`), and identically to `packages/tracker-linear/src/orchestrator-adapter.ts:244-246`.

The tell is in the diff itself: `e2e/host-dynamic-tool-e2e.mjs` had to add `Object.assign(process.env, env)` plus a restore block, and three test files had to add `vi.stubEnv(...)`. Those changes adapt the tests to the regression rather than fixing it — the `env` parameter is still threaded all the way down and is now used only for adapter selection and issue context.

Root cause is the contract shape (see my comment on `packages/core/src/contracts/tracker-adapter.ts:271`): `executeAgentTool(name, args, context)` gives the adapter no channel for host configuration, so it reaches for ambient globals. Suggest adding the resolved env/config to the execution context (or a fourth parameter) and passing `env` through from `executeCodexDynamicToolCall`; the e2e `Object.assign(process.env, ...)` and the `vi.stubEnv` calls can then be reverted, which is a good check that the fix is real.

---

packages/worker/src/codex-dynamic-tools.ts
```diff
@@ -87,35 +68,26 @@ export async function executeCodexDynamicToolCall(
...
-  return {
-    apiKey: env.LINEAR_API_KEY,
-    apiUrl: env.LINEAR_GRAPHQL_URL,
-    authorizationHeader: env.LINEAR_AUTHORIZATION,
-  };
+): OrchestratorTrackerAdapter {
+  return env.SYMPHONY_TRACKER_KIND === "linear"
+    ? linearTrackerAdapter
+    : githubProjectTrackerAdapter;
 }
```
hojinzs: **P1 — this breaks `github_graphql` for every Linear-tracked project.**

Resolving to exactly one adapter assumes the advertised tool set equals that adapter's tool set. It doesn't. `packages/runtime-codex/src/runtime.ts:588-601` always includes `githubTool` and *adds* `linearTool` when `enableLinearGraphqlTool` (i.e. `SYMPHONY_TRACKER_KIND === "linear"`). So under a Linear tracker the agent is advertised **both** tools, and `plan.dynamicTools` — the allowlist passed at `packages/worker/src/index.ts:1524` — contains both. But `resolveTrackerToolAdapter` returns only `linearTrackerAdapter`, whose `executeAgentTool` throws on anything that isn't `linear_graphql`.

Reproduced against this branch's build, with both credentials present in `process.env` (so this is independent of the credential issue):

```
SYMPHONY_TRACKER_KIND=linear, allowlist = ["github_graphql", "linear_graphql"]

github_graphql -> {"error":{"code":"tool_execution_failed",
                            "message":"Unknown Linear agent tool: github_graphql"}}
linear_graphql -> {"data":{"viewer":{"login":"octo"}}}
```

The removed `switch` handled both names regardless of tracker kind, so this is a regression: a Linear-tracked project loses all GitHub GraphQL access from the agent (PR lookups, comments, review handling), and the failure surfaces to the agent as a generic `tool_execution_failed` rather than anything diagnosable.

Dispatch needs to be by **tool name across all adapters that advertise it**, not by tracker kind — e.g. build a `name -> adapter` map from `agentToolSpecs()` over the adapters enabled for this session, and keep the `unknown_tool` failure for a name no adapter claims. That also restores the defensive `default:` branch whose comment was deleted with the switch.

Note the env-based selection itself is an unrequested design decision — #732 specifies the contract (`agentToolSpecs()` / `executeAgentTool(name, args, context)`) but says nothing about single-adapter resolution, and it is what introduced this regression.

---

packages/core/src/contracts/tracker-adapter.ts
```diff
@@ -235,6 +265,14 @@ export type OrchestratorTrackerAdapter = {
...
   secretEnvironmentNames(): string[];
+  /** Advertises provider tools that the runtime may expose to the agent. */
+  agentToolSpecs?(): readonly AgentToolSpec[];
+  /** Executes one advertised provider tool on the host for the active issue. */
+  executeAgentTool?(
+    name: string,
+    args: Record<string, unknown>,
+    context: AgentToolExecutionContext
+  ): Promise<unknown>;
   listIssues(
```
hojinzs, line 282: **P2 — the contract has no channel for host configuration, and only half the call paths converge on it.**

Two things:

1. **No config channel.** `context` carries only `{ issue }`, so both implementations reach into ambient `process.env` for credentials (`tracker-github/src/orchestrator-adapter.ts:89-93`, `tracker-linear/src/orchestrator-adapter.ts:244-246`). That is the direct cause of the P1 credential regression. #732 writes the signature as `executeAgentTool(name, args, context={issue})`, but `context` is an object — extending it with the host-resolved env/config keeps the spec shape while removing the ambient dependency, and makes the adapters unit-testable without `vi.stubEnv`.

2. **Only the Codex path converged.** #732 scopes this as "A/B가 임시로 둔 호출 경로를 계약으로 수렴". `packages/worker/src/codex-dynamic-tools.ts` now goes through the contract, but `packages/runtime-claude/src/mcp-http-server.ts` still hardcodes the tool names and calls `executeGitHubGraphQL` / `executeLinearGraphQL` directly (`:94-118`) with its own `githubConfig`/`linearConfig` (`:165-179`) and its own hand-written schemas (`:145-164`). So the Claude transport is unaffected by this contract, and the two transports now describe the same tools differently — the adapter spec sets `additionalProperties: false` and per-field descriptions, `availableTools()` sets neither. `docs/trackers/github.md` documents the adapter's schema as *the* contract, which is not what a Claude-runtime agent is handed.

Routing `mcp-http-server.ts` through `agentToolSpecs()` / `executeAgentTool()` would close both the requirement and the divergence, and would make the `docs/trackers/*.md` schema tables true for both runtimes.

---

docs/trackers/github.md
```diff
@@ -0,0 +1,28 @@
...
+| Mutations | Permitted. A mutation must be intentionally scoped to the active issue or its repository. |
+| Scope | The worker supplies the normalized active issue `{ id, identifier, nativeRef }` to the adapter. `nativeRef` remains host-internal and is never sent as an extra GraphQL payload field. The tool is an arbitrary GitHub GraphQL transport, so callers must constrain their document and variables to that active issue/repository; it does not infer or rewrite a query's target. |
+| Result | The provider GraphQL payload. For queries, Symphony adds the GitHub `rateLimit` selection when absent and may return normalized rate-limit metadata with the payload. |
+| Errors | Invalid tool arguments, missing host authentication, HTTP failures, and GraphQL errors are returned to the runtime as structured tool failures. Unknown tool names are rejected. |
+| Rate limits | GitHub GraphQL rate-limit headers and the GraphQL `rateLimit` field are measured by the host and applied to the shared GitHub rate-limit policy; callers should keep queries small and respect retry guidance. |
+
+The tool is advertised only when the selected tracker adapter is GitHub. It is
```
hojinzs: **P2 — this statement is not true of the shipped system.**

`github_graphql` is advertised regardless of the selected adapter:

- `packages/runtime-codex/src/runtime.ts:588-601` — `githubTool` is unconditional; `linearTool` is *added* when the tracker is Linear.
- `packages/runtime-claude/src/mcp-http-server.ts:161-163` — returns `[github_graphql, linear_graphql]` when `SYMPHONY_TRACKER_KIND === "linear"`, `[github_graphql]` otherwise.

That behaviour looks intentional (a Linear-tracked project still needs GitHub for PRs), so the fix is the sentence, not the code — but it matters, because the same false assumption is what makes `resolveTrackerToolAdapter` drop `github_graphql` under a Linear tracker (see my comment on `packages/worker/src/codex-dynamic-tools.ts:71-77`). Suggest stating plainly that `github_graphql` is always advertised, and that `linear_graphql` is added only when the tracker is Linear.

The equivalent line in `docs/trackers/linear.md:20` is accurate as written.

---

e2e/scenarios/19-host-side-tracker-tools.md
```diff
@@ -0,0 +1,34 @@
...
+## Assertions
+
+- The Codex helper invokes `github_graphql` through the built worker and
+  selected GitHub adapter, returning the stubbed provider payload while using
+  the host-only test credential for the HTTP request.
+- The Claude Docker test invokes the host-owned HTTP MCP endpoint, performs a
+  provider call using the host credential, and records a successful tool
+  response.
+- The Claude stub's captured child environment and generated MCP configuration
+  contain neither the raw GitHub credential aliases nor the host token value.
```
hojinzs: **P2 — the E2E doesn't cover what #732 asks for.**

#732's Docker E2E line is "Codex/Claude 워커가 host-side 툴로 **트래커 전이·코멘트 수행**(#634 시나리오), child env/파일에 토큰 없음 단언". Both halves of TC-19 only issue a read-only `query { viewer { login } }`:

- `e2e/host-dynamic-tool-e2e.mjs:31` — `{ query: "query { viewer { login } }" }`
- `test/e2e/claude/claude-docker.spec.ts:129` — the stub returns `{ data: { viewer: { login: "host-mcp-stub" } } }`

No state transition and no comment goes through the host-side tool anywhere in this PR. That leaves the whole mutation half of the contract unexercised — `docs/trackers/*.md` documents "Mutations: Permitted, must be scoped to the active issue" as a contract line with nothing behind it, and it's exactly the path that carries write authority. Since the provider HTTP boundary is already stubbed, extending the helper with a scoped `addComment` mutation plus a project-item/state-update mutation and asserting the outgoing document/variables is cheap.

Also, on the third assertion: the credential-stripping check comes from `test/e2e/claude/claude-docker.spec.ts:151-157`, which this PR does not modify. That's fine as *existing* coverage, but the AGENT_TEST.md row lists that spec as new evidence for this contract — worth wording so a later reader doesn't go looking for a test that was added here.

---

packages/worker/src/codex-dynamic-tools.ts
```diff
@@ -40,7 +38,9 @@ export async function executeCodexDynamicToolCall(
   context: TrackerToolContext,
   env: NodeJS.ProcessEnv,
   dependencies: HostToolDependencies = {},
-  allowedToolNames: readonly string[] = ["github_graphql", "linear_graphql"]
+  allowedToolNames: readonly string[] = resolveTrackerToolAdapter(env)
+    .agentToolSpecs?.()
+    .map((tool) => tool.name) ?? []
```
hojinzs: **P3 — the default allowlist ignores an injected adapter.**

The default for `allowedToolNames` is computed from `resolveTrackerToolAdapter(env)`, but line 53 prefers `dependencies.adapter`. So a caller injecting a stub adapter gets its tool names gated by the *real* env-selected adapter, and has to pass `allowedToolNames` explicitly to compensate. Deriving both from the same resolved adapter would remove the trap (and the second `resolveTrackerToolAdapter` call):

```ts
const adapter = dependencies.adapter ?? resolveTrackerToolAdapter(env);
const allowed = allowedToolNames ?? adapter.agentToolSpecs?.().map((t) => t.name) ?? [];
```

---

packages/worker/package.json
```diff
@@ -48,6 +48,7 @@
     "@gh-symphony/runtime-codex": "workspace:*",
     "@gh-symphony/tool-github-graphql": "workspace:*",
     "@gh-symphony/tool-linear-graphql": "workspace:*",
-    "@gh-symphony/tracker-github": "workspace:*"
+    "@gh-symphony/tracker-github": "workspace:*",
+    "@gh-symphony/tracker-linear": "workspace:*"
```
hojinzs: **P3 — stale dependency, and the architecture doc's graph is now out of date.**

`@gh-symphony/tool-linear-graphql` no longer has a single import in `packages/worker/src` after this refactor (the only remaining `tool-*-graphql` use is `tool-rate-limit.ts` → `tool-github-graphql`). It can be dropped now that the Linear call goes through `tracker-linear`.

Separately, `CLAUDE.md` asks that `docs/architecture.md` be kept current when code moves across packages. Two lines there are now wrong:

- `docs/architecture.md:113` — `worker` gained `tracker-linear`
- `docs/architecture.md:119` — `tracker-linear ─→ core` should be `core, tool-linear-graphql`

The summary graph in `CLAUDE.md:51,54` has the same drift.

---

hojinzs (changes requested):

The contract shape in `packages/core` is the right one and the per-tool docs are genuinely good (all six §10.5 columns present for both providers). But converging the Codex path onto it introduced two regressions I reproduced against this branch's build, and the E2E doesn't yet cover the acceptance criterion #732 names.

### Verification performed

| Check | Result |
| --- | --- |
| `pnpm install && pnpm build` | pass |
| `pnpm lint` | pass |
| `pnpm typecheck` | pass |
| `pnpm test` | 1 failure — `runtime-claude/src/spawn.test.ts` ("no output for 80ms"); passes on re-run, timing-sensitive and untouched by this diff. **Not a PR defect.** |
| Smoke: host tool with `.env`-sourced credential | **fail** — see P1-1 |
| Smoke: `github_graphql` under `SYMPHONY_TRACKER_KIND=linear` | **fail** — see P1-2 |
| Smoke: `github_graphql` under a GitHub tracker with exported creds | pass |

No preview environment on this PR, so smoke testing was done locally against the built `packages/worker/dist` through the real adapters with only the provider HTTP boundary stubbed — the same boundary TC-19 stubs.

### P1

1. **`.env`-sourced tracker credentials are silently dropped.** The adapters read `process.env` (`tracker-github:89-93`, `tracker-linear:244-246`) while the worker threads `launcherEnv`, which merges `.env` files over `process.env`. A token configured only in `.env` now yields `tool_execution_failed: Either GITHUB_GRAPHQL_TOKEN or the runtime token broker configuration is required.` and no provider request. The whole broker path is affected too. The `Object.assign(process.env, env)` added to `e2e/host-dynamic-tool-e2e.mjs` and the `vi.stubEnv` calls added to three test files are the symptom — they adapt the tests to the regression instead of catching it.

2. **`github_graphql` is broken for every Linear-tracked project.** `resolveTrackerToolAdapter` picks one adapter by tracker kind, but the Codex runtime advertises `github_graphql` *unconditionally* and adds `linear_graphql` on top when the tracker is Linear. Result under a Linear tracker: `Unknown Linear agent tool: github_graphql`. The removed `switch` handled both names. Dispatch should be by tool name across the advertised set, not by tracker kind.

### P2

3. **Only half the call paths converged.** `runtime-claude/src/mcp-http-server.ts` still hardcodes tool names, schemas and direct `executeGitHubGraphQL`/`executeLinearGraphQL` calls. #732 scopes this as converging the paths A/B left temporary — the Claude transport is untouched by the new contract, and now advertises schemas that differ from the ones `docs/trackers/*.md` documents as the contract.

4. **E2E doesn't exercise a tracker transition or comment.** #732 requires the Docker E2E to perform 트래커 전이·코멘트 through the host-side tool. Both the new Codex helper and the (unmodified) Claude spec only run `query { viewer { login } }`. The documented "Mutations: Permitted" contract line — the one carrying write authority — has no test behind it.

5. **`docs/trackers/github.md:19` asserts something false** — `github_graphql` is not advertised only under a GitHub adapter.

### P3 / nit

6. Default `allowedToolNames` is derived from the env-selected adapter even when `dependencies.adapter` is injected.
7. `@gh-symphony/tool-linear-graphql` is now an unused direct dep of `packages/worker`; `docs/architecture.md:113,119` and `CLAUDE.md:51,54` dependency graphs are stale.
8. nit — 7 changed files fail `npx prettier --check`, including both new `docs/trackers/*.md` and the e2e helper. The repo is broadly unformatted already (92 files), so this is low priority and not a merge blocker.

### Scope

No over-engineering found — the added types are all within #732's stated scope. The one unrequested design decision is the env-based single-adapter resolution, and it is what produced P1-2; #732 specifies the contract but not how the worker selects an adapter.

P1-1 and P1-2 are merge blockers. P2-3 and P2-4 are the stated 완료 기준 of the issue, so I'd want them in this PR rather than deferred — but if you have a reasoned case for splitting either into a follow-up, say so on the thread and I'll take it.

---

hojinzs:
**리뷰 보완 (독립 검증, 1회차 Request changes에 추가)** — (1) **main 충돌**: 09:27–28Z에 #738/#747/#734 머지로 `AGENT_TEST.md`, `packages/tracker-linear/src/tracker-linear.test.ts` 충돌(origin/main acff4c8) → 리베이스 필요. (2) P1-2 재현 확인: `worker/src/codex-dynamic-tools.ts:71-76`가 트래커 종류로 어댑터 하나를 고르는데 `runtime-codex/src/runtime.ts:588-601`은 Linear에서도 `github_graphql`을 광고 → `Unknown Linear agent tool: github_graphql` (스펙 §10.5 위반). (3) 계약 미수렴: `runtime-claude/src/mcp-http-server.ts:97-118`·Codex `runtime.ts:194-262` 스키마 하드코딩이 ADR/AGENT_TEST 서술과 불일치. (4) 문서: `docs/trackers/github.md:19-20` 서술 거짓, brokerless Codex child env의 `GITHUB_GRAPHQL_TOKEN` 잔존(#700까지) caveat, architecture.md 의존성 그래프 미갱신. 실바이너리: Codex replay ACCEPTED(#746 회귀 없음), 실제 `claude` Streamable HTTP 핸드셰이크 OK.

packages/worker/src/codex-dynamic-tools.ts
```diff
@@ -87,35 +68,26 @@ export async function executeCodexDynamicToolCall(
...
-): Parameters<typeof executeLinearGraphQL>[1] {
-  return {
-    apiKey: env.LINEAR_API_KEY,
-    apiUrl: env.LINEAR_GRAPHQL_URL,
-    authorizationHeader: env.LINEAR_AUTHORIZATION,
-  };
+): OrchestratorTrackerAdapter {
+  return env.SYMPHONY_TRACKER_KIND === "linear"
+    ? linearTrackerAdapter
+    : githubProjectTrackerAdapter;
```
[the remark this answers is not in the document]
moncher-dev: Resolved in `d90496af`: advertised tools are now mapped to their owning adapter by tool name, so Linear sessions retain `github_graphql` alongside `linear_graphql`.

packages/worker/src/codex-dynamic-tools.ts
```diff
@@ -50,34 +50,15 @@ export async function executeCodexDynamicToolCall(
...
+    const adapter = dependencies.adapter ?? resolveTrackerToolAdapter(env);
+    if (!adapter.executeAgentTool) {
+      return failure("unknown_tool", `Tool "${toolName}" is not supported.`);
     }
-
+    const result = await adapter.executeAgentTool(
+      toolName,
+      argumentsValue,
+      context
+    );
```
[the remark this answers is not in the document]
moncher-dev: Resolved in `d90496af`: `AgentToolExecutionContext` now carries the loaded host environment, and both adapters derive credentials and endpoints from that context rather than ambient `process.env`.

packages/tracker-github/src/orchestrator-adapter.ts
moncher-dev: Resolved in `d90496af`: `AgentToolExecutionContext` now carries the loaded host environment, and both adapters derive credentials and endpoints from that context rather than ambient `process.env`.

packages/worker/src/codex-dynamic-tools.ts
moncher-dev: Resolved in `d90496af`: the worker dispatches each advertised tool by name to its owning adapter, restoring GitHub GraphQL access for Linear-tracked projects.

packages/core/src/contracts/tracker-adapter.ts
moncher-dev: Resolved in `3d354750`: the Claude MCP transport now exposes adapter-owned specs and invokes `executeAgentTool` with the same normalized issue context and loaded environment.

docs/trackers/github.md
moncher-dev: Resolved in `bd62140d`: GitHub documentation now states that `github_graphql` is always advertised, while Linear adds `linear_graphql`.

e2e/scenarios/19-host-side-tracker-tools.md
moncher-dev: Resolved in `3d354750`: TC-19 now exercises scoped comment and Project-state mutations through both Codex and Claude host-side adapter tools; the Claude fixture retains the child credential-boundary assertion.

packages/worker/src/codex-dynamic-tools.ts
moncher-dev: Resolved in `d90496af`: the default allowed-tool set is derived from the same resolved or injected adapter used for execution.

packages/worker/package.json
moncher-dev: Resolved in `3d354750` and `8dc393bb`: removed the stale worker dependency and updated the architecture and CLAUDE dependency graphs for tracker-linear's adapter-owned GraphQL tool dependency.

packages/worker/src/codex-dynamic-tools.ts
```diff
@@ -87,35 +75,24 @@ export async function executeCodexDynamicToolCall(
...
+function resolveTrackerToolAdapters(
   env: NodeJS.ProcessEnv
-): Parameters<typeof executeLinearGraphQL>[1] {
-  return {
-    apiKey: env.LINEAR_API_KEY,
-    apiUrl: env.LINEAR_GRAPHQL_URL,
-    authorizationHeader: env.LINEAR_AUTHORIZATION,
-  };
+): readonly OrchestratorTrackerAdapter[] {
+  return env.SYMPHONY_TRACKER_KIND === "linear"
```
hojinzs, line 81: **nit — the P1-2 fix has no unit-test guard on this function (non-blocking).**

The routing fix itself is correct; I reproduced it working. But `resolveTrackerToolAdapters` — the function whose single-adapter form *was* the regression — is never exercised under `SYMPHONY_TRACKER_KIND=linear` by any test. `codex-dynamic-tools.test.ts:125` ("routes GitHub tools alongside Linear tools") injects `dependencies.adapters` stubs, so it validates `executeCodexDynamicToolCall`'s dispatch but not this resolution; `:94` uses real resolution but only under the default GitHub kind.

Verified by reverting this function to the buggy form on top of your branch:

```diff
   return env.SYMPHONY_TRACKER_KIND === "linear"
-    ? [githubProjectTrackerAdapter, linearTrackerAdapter]
+    ? [linearTrackerAdapter]
     : [githubProjectTrackerAdapter];
```

```
pnpm --filter @gh-symphony/worker test
  Test Files  26 passed (26)
       Tests  194 passed (194)
```

The whole worker suite stays green while `github_graphql` is broken for every Linear-tracked project again.

The Claude side is already guarded — the same revert in `mcp-http-server.ts` fails `mcp-http-server.test.ts:41`, because that test drives real resolution through `env` alone. A one-line equivalent here (call `executeCodexDynamicToolCall` with `SYMPHONY_TRACKER_KIND=linear` and *no* injected `dependencies`, asserting `linear_graphql` resolves) would close the asymmetry. Worth doing, but not worth another round on its own.

---

packages/runtime-claude/src/mcp-http-server.ts
```diff
@@ -145,38 +131,17 @@ async function dispatch(
...
-  };
-}
-function linearConfig(env: NodeJS.ProcessEnv) {
-  return {
-    apiKey: env.LINEAR_API_KEY,
-    apiUrl: env.LINEAR_GRAPHQL_URL,
-    authorizationHeader: env.LINEAR_AUTHORIZATION,
-  };
+
+function resolveHostToolAdapters(
```
hojinzs, line 139: **nit — "which adapters are enabled for this session" is now duplicated across two packages (non-blocking).**

`resolveHostToolAdapters` here and `resolveTrackerToolAdapters` in `packages/worker/src/codex-dynamic-tools.ts:78-84` are byte-identical. That's exactly the divergence class P2-3 was about: this PR made both transports share the tool *contract*, but the session's enabled-adapter rule stayed a per-transport copy, so a future tracker kind (or a change to the Linear condition) has to be applied in two places or the two runtimes silently advertise different tool sets.

Cheap to close — hoist one `resolveHostToolAdapters(env)` into a shared module and import it from both. Follow-up is fine; the two copies agree today and both are covered.

---

hojinzs (approved):
## Review 2/6 — Approve

All eight findings from round 1 are fixed, and I verified each against this branch's build rather than taking the thread replies at face value. Both P1 regressions are genuinely gone, and the two P2s that were the issue's stated 완료 기준 are now implemented rather than deferred. What's left is two nits, both non-blocking.

### Verification performed

| Check | Result |
| --- | --- |
| `pnpm install && pnpm build` | pass |
| `pnpm lint` | pass |
| `pnpm typecheck` | pass |
| `pnpm test` | pass — 28 suites green. Last round's `runtime-claude/spawn.test.ts` timing flake did not reproduce. |
| TC-19 Codex helper (`e2e/host-dynamic-tool-e2e.mjs`, repointed at the local `dist`) | `host_dynamic_tool_e2e=pass` |
| `pnpm e2e:claude` | **not run** — no Docker daemon in this environment (`/var/run/docker.sock` absent). Environment limit, not a PR defect; the Container Smoke and Test checks are green on `5daad36`. |
| Smoke: `.env`-sourced credential, Codex path (was P1-1) | **pass** — `Bearer dotenv-sourced-token` reached the provider with nothing exported in `process.env` |
| Smoke: `github_graphql` under `SYMPHONY_TRACKER_KIND=linear` (was P1-2) | **pass** — both tools resolve; requests hit `api.github.com` and `api.linear.app` respectively |
| Smoke: unknown tool + allowlist gating | **pass** — `unknown_tool` for an unowned name, and for an owned name excluded from `allowedToolNames` |
| Smoke: Claude MCP transport, same scenarios | **pass** — `tools/list` → `[github_graphql, linear_graphql]`, both `additionalProperties: false`; `tools/call` used the `.env`-sourced host credential; unknown tool → `-32602` |

Smoke testing again ran locally against the built packages through the real adapters with only the provider HTTP boundary stubbed — the same boundary TC-19 stubs. No preview environment on this PR.

### Round 1 findings

| # | Finding | Status |
| --- | --- | --- |
| P1-1 | `.env`-sourced credentials dropped | **Fixed** — `AgentToolExecutionContext.environment` carries the resolved host env; both adapters read it. The `Object.assign(process.env, env)` in the e2e helper is gone, which was the check I asked for. Guarded: reverting `tracker-github` to `process.env` fails `tracker-github.test.ts:39`. |
| P1-2 | `github_graphql` broken under a Linear tracker | **Fixed** — dispatch is by tool name across the advertised set, with the `unknown_tool` failure retained. See my inline note on the test guard. |
| P2-3 | Only the Codex path converged | **Fixed** — `mcp-http-server.ts` now drives `agentToolSpecs()`/`executeAgentTool()`; both transports hand out identical schemas, so the `docs/trackers/*.md` tables are true for both. |
| P2-4 | E2E had no transition or comment | **Fixed** — both halves now run query + scoped `addComment` + `updateProjectV2ItemFieldValue`, asserting the outgoing documents and variables. The mutation half of the contract has coverage behind it. |
| P2-5 | False statement in `docs/trackers/github.md` | **Fixed** — states plainly that `github_graphql` is always advertised. |
| P3-6 | Default allowlist ignored an injected adapter | **Fixed** — both derive from the same resolved-or-injected set. |
| P3-7 | Stale dep + drifted dependency graphs | **Fixed** — `tool-linear-graphql` dropped from `worker`; `docs/architecture.md:127,130,133` and `CLAUDE.md` updated, including the new `runtime-claude → tracker-*` edge. No cycle introduced. |
| nit-8 | Prettier drift | **Fixed** — `pnpm format` is clean. |

Also confirmed: TC-19's third assertion is now worded as existing coverage that this PR extends, which was the wording concern; the changeset is present; the branch merges into `main` cleanly (2 commits behind, no conflict).

### Scope

Still no over-engineering — everything added stays inside #732's stated scope, and the round-1 items were fixed at the root rather than papered over. The `context.environment` extension is the right shape: it keeps the spec's `context={issue}` object form while removing the ambient dependency.

One thing deliberately *not* converged, which I agree with: `runtime-codex/src/runtime.ts:180-252` still builds its own tool definitions, because a Codex `RuntimeToolDefinition` carries the MCP `command`/`args`/`env` that an `AgentToolSpec` has no notion of. The input schemas are identical to the adapter-owned ones today (same properties, `required`, `additionalProperties: false`), so no documented contract line is false — only the human-readable `description` differs between runtimes. That's #700's cleanup, not this PR's.

### Remaining

Two nits, both inline, both fine as follow-ups: a missing unit-test guard on the Codex-side adapter resolution, and the resolution rule being duplicated across two packages. Neither blocks merge, and neither is worth another round — approving.

---

PR #761, opened 2026-08-29
feat(cli): support workflow server port

### LINKED ISSUE #751: feat(core,cli): server.port front-matter key and --port alias (§13.7)
Epic: #651 · 분할 원본: #676 (B24) · 리포트: `docs/reports/2026-08-28-upstream-spec-drift-research.md`

## 문제
§13.7: `server.port` front-matter 키가 조용히 무시되고(`parser.ts:40-110`) CLI에 `--port` 플래그가 없다(현재 `--http [port]`/`--web [port]`, `start.ts:334-347`).

## 할 일
- `server.port` 파싱(확장 키, 문서화)
- `--port` alias 추가(`--http` 유지), 우선순위: CLI `--port` > `server.port` > 기본값
- 포트 auto-increment는 명시 옵션이 없을 때만; 명시 요청 포트 바인드 실패는 오류(§13.7)
- docs/configuration.md 반영

## 완료 기준
- `server.port`/`--port` 우선순위 테스트, 명시 포트 바인드 실패 오류 테스트(4680 등 고정 포트 사용 금지 — 임시 포트로 점유 후 검증); changeset(patch)
- 범위: core parser + cli만.

moncher-dev:
## TL;DR

`server.port` enables the status API from workflow front matter, and explicit CLI `--port <port>` takes precedence. Bare `--http` and `--port` preserve an explicit configured port; without one they retain auto-increment behavior.

## Change-point diagram

```text
--port <n> ───────────────────────┐
server.port ──────────────────────┼─> requested status port ─> bind or fail
bare --http / --port ─────────────┘    (configured port if present; otherwise auto-increment)
```

## Start here

- `packages/core/src/workflow/parser.ts` parses and validates `server.port`.
- `packages/cli/src/commands/start.ts` resolves CLI/configuration precedence, explicit bind failures, and legacy workflow preflight paths.
- `packages/cli/src/commands/start.test.ts` covers direct CLI precedence, bare aliases, occupied explicit ports, and auto-increment.

## Risks & rollback

- Explicit CLI and configured ports fail fast when occupied; only bare aliases without a configured port may auto-increment.
- Rollback is limited to reverting this PR.

## Changed files

- Core workflow parser and configuration model
- CLI start command and tests
- Configuration and CLI documentation
- Patch changeset for `@gh-symphony/cli`

## Evidence

- `pnpm --filter @gh-symphony/cli test -- start.test.ts`
- `pnpm --filter @gh-symphony/core test -- workflow-loader.test.ts`
- `pnpm lint`
- `pnpm test`
- `pnpm typecheck`
- `pnpm build`
- `git diff --check`

- [ ] Start a managed repository with `server.port` and confirm the status URL binds there.
- [ ] Confirm `--port <available-port>` overrides `server.port`.
- [ ] Confirm an occupied explicitly requested port reports an error.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -115,9 +115,10 @@ same map remain effective.
 Set `server.port` in `WORKFLOW.md` to enable the HTTP status API on that port.
 `0` requests an ephemeral port. `gh-symphony repo start --port [port]` is the
 preferred CLI form and overrides `server.port`; `--http [port]` remains a
-supported alias. A configured or CLI-requested port fails startup when it is
-already occupied, while the internal listener uses an automatic port only when
-neither setting is present.
+supported alias. A configured or explicit CLI port fails startup when it is
+already occupied. A bare `--port` or `--http` keeps the legacy default-port
+auto-increment behavior, while omitting both options and `server.port` uses an
+ephemeral internal listener.
 
 `tracker.required_labels` defaults to `[]`. Labels are compared after trimming
 and lowercasing; every configured label must be present before an issue can be
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -91,6 +91,7 @@ strict parser as workflow loading. Failures include a stable error code; the
 | `codex.command`           | A non-empty string when provided                                                                                           | `workflow_validation_error` at `codex.command`                                |
 | `tracker.kind`            | One of `github-project`, `linear`, or `file`                                                                               | `workflow_validation_error` at `tracker.kind`                                 |
 | `tracker.required_labels` | An array of strings; comma-separated strings are rejected                                                                  | `workflow_validation_error` at `tracker.required_labels`                      |
+| `server.port`             | An integer from `0` to `65535`; `0` requests an ephemeral local port                                                       | `workflow_validation_error` at `server.port`                                  |
 
 Without front matter, the complete trimmed file is used as the workflow prompt
 and all configuration uses defaults; tracker selection is then checked by the
@@ -108,6 +109,15 @@ are trimmed and lowercased before storage and dispatch lookup, so for example
 `" In Progress "` and `in progress` configure the same limit. Entries whose
 values are non-positive or not YAML integers are ignored; valid entries in the
 same map remain effective.
+
+## Optional HTTP Server
+
+Set `server.port` in `WORKFLOW.md` to enable the HTTP status API on that port.
+`0` requests an ephemeral port. `gh-symphony repo start --port [port]` is the
+preferred CLI form and overrides `server.port`; `--http [port]` remains a
+supported alias. A configured or CLI-requested port fails startup when it is
+already occupied, while the internal listener uses an automatic port only when
+neither setting is present.
 
 `tracker.required_labels` defaults to `[]`. Labels are compared after trimming
 and lowercasing; every configured label must be present before an issue can be
--- a/packages/cli/src/commands/start.test.ts
+++ b/packages/cli/src/commands/start.test.ts
@@ -3,6 +3,7 @@ import { EventEmitter } from "node:events";
 import { mkdtemp, mkdir, readFile, writeFile } from "node:fs/promises";
 import { tmpdir } from "node:os";
 import { join, resolve } from "node:path";
+import { pathToFileURL } from "node:url";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import type { CliProjectConfig } from "../config.js";
 import * as configModule from "../config.js";
@@ -1947,6 +1948,66 @@ Handle {{issue.identifier}}.\n`,
     expect(exitSpy).toHaveBeenCalledWith(0);
   });
 
+  it.each(["--http", "--port"])(
+    "uses server.port in preference to a bare %s option",
+    async (option) => {
+      const configDir = await createConfigFixture({
+        activeProject: "tenant-a",
+        projects: [createProject("tenant-a", "acme", "platform")],
+      });
+      const probe = createServer();
+      await new Promise<void>((resolve) =>
+        probe.listen(0, "127.0.0.1", () => resolve())
+      );
+      const address = probe.address();
+      if (!address || typeof address === "string") {
+        probe.close();
+        throw new Error("Expected TCP address");
+      }
+      await new Promise<void>((resolve, reject) =>
+        probe.close((error) => (error ? reject(error) : resolve()))
+      );
+      await configureWorkflow(configDir, address.port);
+      acquireProjectLock.mockResolvedValue({
+        lockPath: join(configDir, ".lock"),
+        ownerToken: "owner",
+        pid: 1234,
+        startedAt: "2026-03-17T00:00:00.000Z",
+      });
+      let resolveRun: (() => void) | undefined;
+      run.mockImplementation(
+        () =>
+          new Promise<void>((resolve) => {
+            resolveRun = resolve;
+          })
+      );
+      shutdown.mockImplementation(async () => {
+        resolveRun?.();
+      });
+      const exitSpy = vi
+        .spyOn(process, "exit")
+        .mockImplementation(
+          ((_code?: number) => undefined) as (code?: number) => never
+        );
+      const stdout = captureWrites(process.stdout);
+
+      try {
+        const startPromise = startModule.default(
+          [option],
+          baseOptions(configDir)
+        );
+        const url = await waitForHttpUrl(stdout.output);
+        expect(new URL(url).port).toBe(String(address.port));
+        process.emit("SIGINT");
+        await startPromise;
+      } finally {
+        stdout.restore();
+      }
+
+      expect(exitSpy).toHaveBeenCalledWith(0);
+    }
+  );
+
   it("reads server.port from a legacy repository workflow fallback", async () => {
     const configDir = await createConfigFixture({
       activeProject: "tenant-a",
@@ -1999,6 +2060,30 @@ Handle {{issue.identifier}}.\n`,
       stdout.restore();
     }
 
+    expect(exitSpy).toHaveBeenCalledWith(0);
+  });
+
+  it("continues when an implicit legacy workflow is invalid", async () => {
+    const configDir = await createConfigFixture({
+      activeProject: "tenant-a",
+      projects: [createProject("tenant-a", "acme", "platform")],
+    });
+    await configureInvalidLegacyWorkflow(configDir);
+    acquireProjectLock.mockResolvedValue({
+      lockPath: join(configDir, ".lock"),
+      ownerToken: "owner",
+      pid: 1234,
+      startedAt: "2026-03-17T00:00:00.000Z",
+    });
+    const exitSpy = vi
+      .spyOn(process, "exit")
+      .mockImplementation(
+        ((_code?: number) => undefined) as (code?: number) => never
+      );
+
+    await startModule.default(["--once"], baseOptions(configDir));
+
+    expect(run).toHaveBeenCalledOnce();
     expect(exitSpy).toHaveBeenCalledWith(0);
   });
 
@@ -2287,7 +2372,7 @@ async function configureLegacyWorkflow(
     await readFile(projectPath, "utf8")
   ) as CliProjectConfig;
   const repositoryDir = join(configDir, "legacy-repository");
-  project.repositoryDir = repositoryDir;
+  project.repository.cloneUrl = pathToFileURL(repositoryDir).href;
   project.workflowSource = { type: "repo" };
   await mkdir(repositoryDir, { recursive: true });
   await writeFile(projectPath, JSON.stringify(project), "utf8");
@@ -2302,6 +2387,25 @@ codex:
   command: codex app-server
 ---
 Prompt\n`,
+    "utf8"
+  );
+}
+
+async function configureInvalidLegacyWorkflow(
+  configDir: string
+): Promise<void> {
+  const projectPath = join(configDir, "projects", "tenant-a", "project.json");
+  const project = JSON.parse(
+    await readFile(projectPath, "utf8")
+  ) as CliProjectConfig;
+  const repositoryDir = join(configDir, "invalid-legacy-repository");
+  project.repository.cloneUrl = pathToFileURL(repositoryDir).href;
+  project.workflowSource = { type: "repo" };
+  await mkdir(repositoryDir, { recursive: true });
+  await writeFile(projectPath, JSON.stringify(project), "utf8");
+  await writeFile(
+    join(repositoryDir, "WORKFLOW.md"),
+    "---\nserver:\n  port: invalid\n---\nPrompt\n",
     "utf8"
   );
 }
--- a/packages/cli/src/commands/start.test.ts
+++ b/packages/cli/src/commands/start.test.ts
@@ -1837,15 +1837,18 @@ Handle {{issue.identifier}}.\n`,
       activeProject: "tenant-a",
       projects: [createProject("tenant-a", "acme", "platform")],
     });
-    const blocker = createServer();
+    const probe = createServer();
     await new Promise<void>((resolve) =>
-      blocker.listen(0, "127.0.0.1", () => resolve())
+      probe.listen(0, "127.0.0.1", () => resolve())
     );
-    const address = blocker.address();
+    const address = probe.address();
     if (!address || typeof address === "string") {
-      blocker.close();
+      probe.close();
       throw new Error("Expected TCP address");
     }
+    await new Promise<void>((resolve, reject) =>
+      probe.close((error) => (error ? reject(error) : resolve()))
+    );
     await configureWorkflow(configDir, address.port);
     const lock = {
       lockPath: join(configDir, ".lock"),
@@ -1874,20 +1877,17 @@ Handle {{issue.identifier}}.\n`,
 
     try {
       const startPromise = startModule.default(
-        ["--port", "0"],
+        ["--port", String(address.port)],
         baseOptions(configDir)
       );
 
       const url = await waitForHttpUrl(stdout.output);
-      expect(new URL(url).port).not.toBe(String(address.port));
+      expect(new URL(url).port).toBe(String(address.port));
 
       process.emit("SIGINT");
       await startPromise;
     } finally {
       stdout.restore();
-      await new Promise<void>((resolve, reject) =>
-        blocker.close((error) => (error ? reject(error) : resolve()))
-      );
     }
 
     expect(exitSpy).toHaveBeenCalledWith(0);
@@ -2007,6 +2007,63 @@ Handle {{issue.identifier}}.\n`,
       expect(exitSpy).toHaveBeenCalledWith(0);
     }
   );
+
+  it("auto-increments an occupied configured port for a bare HTTP alias", async () => {
+    const configDir = await createConfigFixture({
+      activeProject: "tenant-a",
+      projects: [createProject("tenant-a", "acme", "platform")],
+    });
+    const blocker = createServer();
+    await new Promise<void>((resolve) =>
+      blocker.listen(0, "127.0.0.1", () => resolve())
+    );
+    const address = blocker.address();
+    if (!address || typeof address === "string") {
+      blocker.close();
+      throw new Error("Expected TCP address");
+    }
+    await configureWorkflow(configDir, address.port);
+    acquireProjectLock.mockResolvedValue({
+      lockPath: join(configDir, ".lock"),
+      ownerToken: "owner",
+      pid: 1234,
+      startedAt: "2026-03-17T00:00:00.000Z",
+    });
+    let resolveRun: (() => void) | undefined;
+    run.mockImplementation(
+      () =>
+        new Promise<void>((resolve) => {
+          resolveRun = resolve;
+        })
+    );
+    shutdown.mockImplementation(async () => {
+      resolveRun?.();
+    });
+    const exitSpy = vi
+      .spyOn(process, "exit")
+      .mockImplementation(
+        ((_code?: number) => undefined) as (code?: number) => never
+      );
+    const stdout = captureWrites(process.stdout);
+
+    try {
+      const startPromise = startModule.default(
+        ["--http"],
+        baseOptions(configDir)
+      );
+      const url = await waitForHttpUrl(stdout.output);
+      expect(Number(new URL(url).port)).toBeGreaterThan(address.port);
+      process.emit("SIGINT");
+      await startPromise;
+    } finally {
+      stdout.restore();
+      await new Promise<void>((resolve, reject) =>
+        blocker.close((error) => (error ? reject(error) : resolve()))
+      );
+    }
+
+    expect(exitSpy).toHaveBeenCalledWith(0);
+  });
 
   it("reads server.port from a legacy repository workflow fallback", async () => {
     const configDir = await createConfigFixture({
--- a/packages/cli/src/commands/start.test.ts
+++ b/packages/cli/src/commands/start.test.ts
@@ -1892,6 +1892,116 @@ Handle {{issue.identifier}}.\n`,
     expect(exitSpy).toHaveBeenCalledWith(0);
   });
 
+  it("uses server.port when no CLI HTTP option is supplied", async () => {
+    const configDir = await createConfigFixture({
+      activeProject: "tenant-a",
+      projects: [createProject("tenant-a", "acme", "platform")],
+    });
+    const probe = createServer();
+    await new Promise<void>((resolve) =>
+      probe.listen(0, "127.0.0.1", () => resolve())
+    );
+    const address = probe.address();
+    if (!address || typeof address === "string") {
+      probe.close();
+      throw new Error("Expected TCP address");
+    }
+    await new Promise<void>((resolve, reject) =>
+      probe.close((error) => (error ? reject(error) : resolve()))
+    );
+    await configureWorkflow(configDir, address.port);
+    acquireProjectLock.mockResolvedValue({
+      lockPath: join(configDir, ".lock"),
+      ownerToken: "owner",
+      pid: 1234,
+      startedAt: "2026-03-17T00:00:00.000Z",
+    });
+    let resolveRun: (() => void) | undefined;
+    run.mockImplementation(
+      () =>
+        new Promise<void>((resolve) => {
+          resolveRun = resolve;
+        })
+    );
+    shutdown.mockImplementation(async () => {
+      resolveRun?.();
+    });
+    const exitSpy = vi
+      .spyOn(process, "exit")
+      .mockImplementation(
+        ((_code?: number) => undefined) as (code?: number) => never
+      );
+    const stdout = captureWrites(process.stdout);
+
+    try {
+      const startPromise = startModule.default([], baseOptions(configDir));
+      const url = await waitForHttpUrl(stdout.output);
+      expect(new URL(url).port).toBe(String(address.port));
+
+      process.emit("SIGINT");
+      await startPromise;
+    } finally {
+      stdout.restore();
+    }
+
+    expect(exitSpy).toHaveBeenCalledWith(0);
+  });
+
+  it("reads server.port from a legacy repository workflow fallback", async () => {
+    const configDir = await createConfigFixture({
+      activeProject: "tenant-a",
+      projects: [createProject("tenant-a", "acme", "platform")],
+    });
+    const probe = createServer();
+    await new Promise<void>((resolve) =>
+      probe.listen(0, "127.0.0.1", () => resolve())
+    );
+    const address = probe.address();
+    if (!address || typeof address === "string") {
+      probe.close();
+      throw new Error("Expected TCP address");
+    }
+    await new Promise<void>((resolve, reject) =>
+      probe.close((error) => (error ? reject(error) : resolve()))
+    );
+    await configureLegacyWorkflow(configDir, address.port);
+    acquireProjectLock.mockResolvedValue({
+      lockPath: join(configDir, ".lock"),
+      ownerToken: "owner",
+      pid: 1234,
+      startedAt: "2026-03-17T00:00:00.000Z",
+    });
+    let resolveRun: (() => void) | undefined;
+    run.mockImplementation(
+      () =>
+        new Promise<void>((resolve) => {
+          resolveRun = resolve;
+        })
+    );
+    shutdown.mockImplementation(async () => {
+      resolveRun?.();
+    });
+    const exitSpy = vi
+      .spyOn(process, "exit")
+      .mockImplementation(
+        ((_code?: number) => undefined) as (code?: number) => never
+      );
+    const stdout = captureWrites(process.stdout);
+
+    try {
+      const startPromise = startModule.default([], baseOptions(configDir));
+      const url = await waitForHttpUrl(stdout.output);
+      expect(new URL(url).port).toBe(String(address.port));
+
+      process.emit("SIGINT");
+      await startPromise;
+    } finally {
+      stdout.restore();
+    }
+
+    expect(exitSpy).toHaveBeenCalledWith(0);
+  });
+
   it("fails when a requested port is already in use", async () => {
     const configDir = await createConfigFixture({
       activeProject: "tenant-a",
@@ -2155,6 +2265,34 @@ async function configureWorkflow(
   await writeFile(projectPath, JSON.stringify(project), "utf8");
   await writeFile(
     workflowPath,
+    `---
+tracker:
+  kind: github-project
+server:
+  port: ${port}
+codex:
+  command: codex app-server
+---
+Prompt\n`,
+    "utf8"
+  );
+}
+
+async function configureLegacyWorkflow(
+  configDir: string,
+  port: number
+): Promise<void> {
+  const projectPath = join(configDir, "projects", "tenant-a", "project.json");
+  const project = JSON.parse(
+    await readFile(projectPath, "utf8")
+  ) as CliProjectConfig;
+  const repositoryDir = join(configDir, "legacy-repository");
+  project.repositoryDir = repositoryDir;
+  project.workflowSource = { type: "repo" };
+  await mkdir(repositoryDir, { recursive: true });
+  await writeFile(projectPath, JSON.stringify(project), "utf8");
+  await writeFile(
+    join(repositoryDir, "WORKFLOW.md"),
     `---
 tracker:
   kind: github-project
--- a/packages/cli/src/commands/start.test.ts
+++ b/packages/cli/src/commands/start.test.ts
@@ -1831,11 +1831,21 @@ Handle {{issue.identifier}}.\n`,
     expect(exitSpy).toHaveBeenCalledWith(0);
   });
 
-  it("increments the port when the requested HTTP port is already in use", async () => {
+  it("uses --port in preference to server.port", async () => {
     const configDir = await createConfigFixture({
       activeProject: "tenant-a",
       projects: [createProject("tenant-a", "acme", "platform")],
     });
+    const blocker = createServer();
+    await new Promise<void>((resolve) =>
+      blocker.listen(0, "127.0.0.1", () => resolve())
+    );
+    const address = blocker.address();
+    if (!address || typeof address === "string") {
+      blocker.close();
+      throw new Error("Expected TCP address");
+    }
+    await configureWorkflow(configDir, address.port);
     const lock = {
       lockPath: join(configDir, ".lock"),
       ownerToken: "owner",
@@ -1854,16 +1864,6 @@ Handle {{issue.identifier}}.\n`,
       resolveRun?.();
     });
 
-    const blocker = createServer();
-    await new Promise<void>((resolve) =>
-      blocker.listen(0, "127.0.0.1", () => resolve())
-    );
-    const address = blocker.address();
-    if (!address || typeof address === "string") {
-      blocker.close();
-      throw new Error("Expected TCP address");
-    }
-
     const exitSpy = vi
       .spyOn(process, "exit")
       .mockImplementation(
@@ -1873,12 +1873,12 @@ Handle {{issue.identifier}}.\n`,
 
     try {
       const startPromise = startModule.default(
-        ["--http", String(address.port)],
+        ["--port", "0"],
         baseOptions(configDir)
       );
 
       const url = await waitForHttpUrl(stdout.output);
-      expect(new URL(url).port).toBe(String(address.port + 1));
+      expect(new URL(url).port).not.toBe(String(address.port));
 
       process.emit("SIGINT");
       await startPromise;
@@ -1890,6 +1890,41 @@ Handle {{issue.identifier}}.\n`,
     }
 
     expect(exitSpy).toHaveBeenCalledWith(0);
+  });
+
+  it("fails when a requested port is already in use", async () => {
+    const configDir = await createConfigFixture({
+      activeProject: "tenant-a",
+      projects: [createProject("tenant-a", "acme", "platform")],
+    });
+    const blocker = createServer();
+    await new Promise<void>((resolve) =>
+      blocker.listen(0, "127.0.0.1", () => resolve())
+    );
+    const address = blocker.address();
+    if (!address || typeof address === "string") {
+      blocker.close();
+      throw new Error("Expected TCP address");
+    }
+    acquireProjectLock.mockResolvedValue({
+      lockPath: join(configDir, ".lock"),
+      ownerToken: "owner",
+      pid: 1234,
+      startedAt: "2026-03-17T00:00:00.000Z",
+    });
+
+    try {
+      await expect(
+        startModule.default(
+          ["--port", String(address.port)],
+          baseOptions(configDir)
+        )
+      ).rejects.toThrow(`HTTP server port ${address.port} is already in use`);
+    } finally {
+      await new Promise<void>((resolve, reject) =>
+        blocker.close((error) => (error ? reject(error) : resolve()))
+      );
+    }
   });
 
   it("propagates lock acquisition failures", async () => {
@@ -2105,4 +2140,30 @@ async function createConfigFixture(input: {
   }
 
   return configDir;
+}
+
+async function configureWorkflow(
+  configDir: string,
+  port: number
+): Promise<void> {
+  const projectPath = join(configDir, "projects", "tenant-a", "project.json");
+  const project = JSON.parse(
+    await readFile(projectPath, "utf8")
+  ) as CliProjectConfig;
+  const workflowPath = join(configDir, "projects", "tenant-a", "WORKFLOW.md");
+  project.workflowSource = { type: "repo", path: workflowPath };
+  await writeFile(projectPath, JSON.stringify(project), "utf8");
+  await writeFile(
+    workflowPath,
+    `---
+tracker:
+  kind: github-project
+server:
+  port: ${port}
+codex:
+  command: codex app-server
+---
+Prompt\n`,
+    "utf8"
+  );
 }
--- a/packages/cli/src/commands/start.ts
+++ b/packages/cli/src/commands/start.ts
@@ -253,7 +253,7 @@ async function preflightWorkflowStart(
       {
         supportedTrackerKinds: getSupportedTrackerKinds(),
         resolveTrackerAdapter: resolveWorkflowConfigTrackerAdapter,
-        workflowPath: configuredPath,
+        workflowPath,
       }
     );
     return { ok: true, workflow };
--- a/packages/cli/src/commands/start.ts
+++ b/packages/cli/src/commands/start.ts
@@ -33,7 +33,7 @@ import type {
   TrackerStateRequest,
   TrackerStateResult,
 } from "@gh-symphony/core";
-import { parseWorkflowMarkdown } from "@gh-symphony/core";
+import { parseWorkflowMarkdown, type ParsedWorkflow } from "@gh-symphony/core";
 import {
   DashboardFsReader,
   isAuthorizedApiRequest,
@@ -232,12 +232,12 @@ async function resolveConfiguredLinearToken(
 async function preflightWorkflowStart(
   projectConfig: OrchestratorProjectConfig,
   runtimeRoot: string
-): Promise<boolean> {
+): Promise<{ ok: true; workflow: ParsedWorkflow | null } | { ok: false }> {
   const configuredPath = projectConfig.workflowSource?.path;
   if (!configuredPath) {
     // Legacy managed-project configs rely on the daemon's repository lookup,
     // including its default workflow fallback. Avoid a divergent preflight.
-    return true;
+    return { ok: true, workflow: null };
   }
 
   try {
@@ -245,25 +245,29 @@ async function preflightWorkflowStart(
       projectConfig,
       runtimeRoot
     );
-    parseWorkflowMarkdown(await readFile(configuredPath, "utf8"), environment, {
-      supportedTrackerKinds: getSupportedTrackerKinds(),
-      resolveTrackerAdapter: resolveWorkflowConfigTrackerAdapter,
-      workflowPath: configuredPath,
-    });
-    return true;
+    const workflow = parseWorkflowMarkdown(
+      await readFile(configuredPath, "utf8"),
+      environment,
+      {
+        supportedTrackerKinds: getSupportedTrackerKinds(),
+        resolveTrackerAdapter: resolveWorkflowConfigTrackerAdapter,
+        workflowPath: configuredPath,
+      }
+    );
+    return { ok: true, workflow };
   } catch (error) {
     if ((error as NodeJS.ErrnoException | undefined)?.code === "ENOENT") {
       process.stderr.write(
         `Configured workflow not found at ${configuredPath}. Restore the file or run 'gh-symphony repo init --workflow-file <path>'.\n`
       );
       process.exitCode = 1;
-      return false;
+      return { ok: false };
     }
     process.stderr.write(
       `Workflow preflight failed for ${configuredPath}: ${error instanceof Error ? error.message : "Invalid workflow definition."}\n`
     );
     process.exitCode = 1;
-    return false;
+    return { ok: false };
   }
 }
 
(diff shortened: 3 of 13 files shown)


packages/cli/src/commands/start.ts
```diff
@@ -840,6 +845,9 @@ async function startHttpServer(input: {
     } catch (error) {
       await closeHttpServer(server).catch(() => {});
       if ((error as NodeJS.ErrnoException)?.code === "EADDRINUSE") {
+        if (input.explicitPort) {
+          throw new Error(`HTTP server port ${port} is already in use`);
```
hojinzs, line 883: **P1 — `--daemon` reports a successful start after this throw kills the child.**

I confirmed the path Codex flagged on `waitForChildLock`, and it is reachable through the new code here:

1. `startDaemon` now spawns the child with `--port <httpPort>` (L1451), where `httpPort` is `requestedHttpPort` and therefore also covers a workflow-supplied `server.port`.
2. The child writes the readiness marker at L1080–1082, immediately after `acquireProjectLock` — well before `startHttpServer` runs at L1181.
3. This new `EADDRINUSE` branch throws, the child unwinds through the outer `finally` and exits.
4. The parent's `waitForChildLock` either already resolved from its 20 ms poll, or resolves from `onExit` because the marker content still matches `child.pid`.

Net effect of `gh-symphony repo start --daemon --port <occupied>`: the parent writes `pidPath`, prints `Orchestrator started in background (PID: N)` plus the bearer token, and exits 0 — while no daemon is running. The stale PID file then misleads `repo stop`/`repo status`.

Before this PR an occupied port auto-incremented, so this was not a routine path; the explicit-port failure makes it a normal, user-triggerable outcome, so it needs handling in this PR.

Suggested fix: move the readiness-marker write to after the listener has successfully bound (i.e. after `startHttpServer` returns) whenever a port was requested, so `waitForChildLock` cannot observe readiness for a process that is about to die.

---

packages/cli/src/commands/start.ts
```diff
@@ -1166,8 +1181,9 @@ const handler = async (
       workerHttpServer = await startHttpServer({
         runtimeRoot,
         projectId,
-        initialPort: parsed.httpPort ?? 0,
-        host: parsed.httpPort !== undefined ? httpHost : DEFAULT_HTTP_HOST,
+        initialPort: requestedHttpPort ?? 0,
+        explicitPort: requestedHttpPort !== undefined,
```
hojinzs: **P2 — `server.port` is applied even when `--web` is used, which steals the dashboard's port.**

`requestedHttpPort` falls back to `server.port` regardless of `parsed.webPort`. `--port`/`--http` and `--web` are mutually exclusive at the arg level (L444), but `server.port` bypasses that check entirely, so the worker listener now claims the configured port before `startControlPlaneServer` is called with `parsed.webPort`.

I verified this with a temporary test on this branch: with `server.port: <P>` in `WORKFLOW.md` and `gh-symphony repo start --web <P>`, the worker API answers `200` on `http://127.0.0.1:<P>/api/v1/state`, and `startControlPlaneServer` is still asked to bind `<P>`. The control plane's own loop (`packages/control-plane/src/server.ts:200`) auto-increments on `EADDRINUSE`, so in production the dashboard silently lands on `<P>+1` — the user asked for `<P>` and gets no error. The common case is `server.port: 4680` plus a bare `--web` (which also defaults to 4680).

This also contradicts the docs added in this PR: `packages/cli/README.md:351` says `server.port` "applies when neither CLI option is present".

Suggested fix: only fall back to `server.port` when `parsed.webPort === undefined` as well — or route `server.port` into `parsed.webPort` when `--web` is active and keep a single explicit port. Either way the code and the new doc sentence should agree.

---

packages/cli/src/commands/start.ts
```diff
@@ -401,7 +405,7 @@ function parseStartArgs(args: string[]): {
       parsed.bindAll = true;
       continue;
     }
-    if (arg === "--http") {
+    if (arg === "--http" || arg === "--port") {
```
hojinzs, line 440: **P2 — a bare `--http` / `--port` now hard-fails instead of auto-incrementing, which is broader than #751 asked for.**

`--http` with no value sets `httpPort = DEFAULT_HTTP_PORT` (L412), which then makes `explicitPort` true downstream, so `gh-symphony repo start --http` on a box where 4680 is already taken now errors out. That used to bind 4681 and keep going.

#751 scopes the hard failure to an *explicitly requested* port ("명시 요청 포트 바인드 실패는 오류"; "포트 auto-increment는 명시 옵션이 없을 때만"). A bare `--http` requests the API, not a specific port — the 4680 there is a default, not a user choice. Treating it as explicit is a behavior regression for existing `--http` users that the issue does not require.

Suggested fix: track whether a port *value* was supplied (e.g. `parsed.httpPortExplicit = true` only in the `parsePort` branch below) and pass that as `explicitPort`, so a bare `--http`/`--port` keeps auto-incrementing while `--port 4800` and `server.port: 4800` fail fast.

---

packages/cli/src/commands/start.test.ts
```diff
@@ -1892,6 +1892,41 @@ Handle {{issue.identifier}}.\n`,
     expect(exitSpy).toHaveBeenCalledWith(0);
   });
 
+  it("fails when a requested port is already in use", async () => {
```
hojinzs, line 2147: **P2 — the headline behavior of #751 has no CLI test: `server.port` alone enabling the API.**

The two CLI tests here cover (a) `--port` overriding `server.port` and (b) an explicit port failing when occupied. Neither exercises `configuredServerPort` actually being used — in the precedence test `server.port` is deliberately the *losing* value.

I added a temporary test on this branch to check it, and the behavior is correct: with only `server.port: <free port>` in `WORKFLOW.md` and no CLI flag, the status API binds exactly that port and `/api/v1/state` returns `200`. So the feature works — it just isn't protected against regression, which is the one acceptance criterion most likely to silently break again (this whole issue exists because `server.port` was silently ignored).

Worth adding as a third case: no CLI flag, `server.port` set, assert the bound port equals the configured one.

Separately, the rename of `increments the port when the requested HTTP port is already in use` removed the only coverage of `startHttpServer`'s auto-increment branch. Combined with the P3 note on that loop, the `continue` path is now both untested and effectively unreachable.

---

packages/cli/src/commands/start.ts
```diff
@@ -1166,8 +1181,9 @@ const handler = async (
       workerHttpServer = await startHttpServer({
         runtimeRoot,
         projectId,
-        initialPort: parsed.httpPort ?? 0,
-        host: parsed.httpPort !== undefined ? httpHost : DEFAULT_HTTP_HOST,
+        initialPort: requestedHttpPort ?? 0,
```
hojinzs, line 1222: **P3 — the auto-increment loop in `startHttpServer` is now effectively dead code.**

`explicitPort` is `requestedHttpPort !== undefined`, and `initialPort` is `requestedHttpPort ?? 0`. So the two states are:

- `explicitPort === true` → the new branch throws on `EADDRINUSE`, never `continue`s.
- `explicitPort === false` → `initialPort` is always `0`, and `listen(0)` does not produce `EADDRINUSE`.

That makes the `continue` at L850 and the `for (let port = ...; port <= 65_535; port += 1)` walk unreachable in practice, along with the `Unable to bind HTTP server starting from port ...` error below it. Worth collapsing to a single bind attempt so the code states the actual contract — unless the fix for the bare-`--http` regression above restores a real auto-increment path, in which case the loop stays meaningful and just needs the test coverage back.

---

packages/cli/src/commands/start.ts
```diff
@@ -950,7 +958,7 @@ const handler = async (
   if (parsed.error) {
     process.stderr.write(`${parsed.error}\n`);
     process.stderr.write(
-      `Usage: gh-symphony ${options.invocation === "project" ? "project" : "repo"} start [--daemon] [--once] [--assigned-only] [--http [port]] [--web [port]] [--bind-all]${options.invocation === "project" ? " [--project-dir <path>]" : ""}\n`
+      `Usage: gh-symphony ${options.invocation === "project" ? "project" : "repo"} start [--daemon] [--once] [--assigned-only] [--port [port]] [--http [port]] [--web [port]] [--bind-all]${options.invocation === "project" ? " [--project-dir <path>]" : ""}\n`
```
hojinzs, line 995: nit — the usage line now advertises `--port`, but the conflict message at L444 still reads `Options '--http' and '--web' cannot be used together`. A user who typed `--port 4800 --web` sees an error naming a flag they did not use. Interpolating the actual argument (as `parsePort` already does) would keep it accurate.

---

docs/configuration.md
```diff
@@ -449,31 +459,31 @@ The orchestrator injects these into worker processes. They are documented for
...
-| --------------------------------- | ----------------------------------------- | --------------------------------------------- | ----------------- | --------------------------------------------------------------------------------------- |
-| `PROJECT_ID` / `CODEX_PROJECT_ID` | active project ID                         | Codex runtime launcher                        | Internal/injected | Runtime project identity. One of these is required when running the launcher directly.  |
-| `WORKING_DIRECTORY`               | issue repository checkout path            | Worker, Codex runtime launcher                | Internal/injected | Worker cwd / repository workspace path. Required when running a launcher directly.      |
+| Variable                          | Default                                   | Read by                                       | Audience          | Notes                                                                                                                                                                                                                              |
```
hojinzs, line 489: nit — the two env-var tables below are re-wrapped end to end (trailing-pad width changes only, no content change), which adds ~40 lines of churn unrelated to `server.port` and makes the doc diff hard to read. If this came from a full-file Prettier pass, consider splitting it into its own formatting commit — or reverting these hunks here — so the substantive change stays reviewable.

---

hojinzs (changes requested):

### Requirements vs. #751

| #751 acceptance criterion | Status |
| --- | --- |
| `server.port` parsing (extension key, documented) | ✅ `parser.ts` `readOptionalPort`, 0–65535, non-integers rejected; `docs/configuration.md` row + section |
| `--port` alias added, `--http` retained | ✅ |
| Precedence CLI `--port` > `server.port` > default | ✅ verified |
| Auto-increment only without an explicit option | ⚠️ over-applied — a bare `--http`/`--port` is now treated as explicit (P2) |
| Explicit-port bind failure is an error | ✅ but the `--daemon` path reports success anyway (P1) |
| `docs/configuration.md` updated | ✅ |
| Precedence + bind-failure tests, no fixed 4680 | ✅ temporary ports used correctly |
| changeset (patch) | ✅ |
| Scope: core parser + cli only | ✅ no scope creep, no over-engineering |

### Smoke test

Built the branch and ran the focused suites — `packages/core/src/workflow-loader.test.ts` + `packages/cli/src/commands/start.test.ts`: **126 passed**. I then added two temporary tests locally (not committed):

- `server.port` alone, no CLI flag → status API binds exactly the configured port, `/api/v1/state` returns `200` with the bearer token. **The feature works.**
- `--web <P>` with `server.port: <P>` → the worker API answers `200` on `<P>` while `startControlPlaneServer` is still asked to bind `<P>`. **Confirms the P2 below.**

### Findings

- **P1** — `repo start --daemon --port <occupied>` prints `Orchestrator started in background (PID: N)` and writes a PID file after the child has already died. The readiness marker is written right after the lock, long before the listener binds. (inline; corroborates the existing Codex thread on `waitForChildLock`)
- **P2** — `server.port` is applied even when `--web` is used, so the dashboard silently auto-increments off the port the user asked for. Also contradicts this PR's own doc sentence.
- **P2** — bare `--http`/`--port` now hard-fails on an occupied 4680 instead of incrementing; broader than #751 requires.
- **P2** — no CLI test that `server.port` alone enables the API — the one criterion most likely to silently regress.
- **P3** — `startHttpServer`'s auto-increment loop is now unreachable in both states.
- **nit** — conflict message still names `--http`; unrelated Prettier churn in `docs/configuration.md`.

### Prior review

The three Codex P1 threads are all unresolved with no reply. #1 (daemon readiness) I confirmed independently — please fix. #2 (legacy `workflowSource` without a `path`) is real but narrower than P1: `OrchestratorService.loadProjectWorkflowUncached` falls back to `loadRepositoryWorkflow`, while the CLI preflight returns `workflow: null`, so `server.port` stays silently ignored for those configs — the exact failure mode #751 set out to fix. Either handle it or reply with the reasoning. #3 (living documents) I read as P3: no package/layer mapping changed, so `docs/architecture.md` is arguably fine; an `AGENT_TEST.md` coverage-table row for port enablement/override/occupied-port would be reasonable but I would not block on it.

Overall this is a focused, correct implementation of the feature. The P1 and the `--web` interaction are what stand between it and approval.

---

hojinzs:
**리뷰 보완 (독립 검증, 1회차 Request changes에 추가 차단 1건)** — `--port`가 Commander에 등록되지 않았습니다: `packages/cli/src/index.ts:588`(repo start)·`:722`(project start)는 `--http [port]`만 선언. 빌드된 CLI: `project start --project-dir … --port 4982` → `error: unknown option '--port'` exit 1(`allowExcessArguments(false)`); `repo start --port`는 `allowUnknownOption(true)` 덕에 우연히 통과할 뿐이며 `--help`·`completion.ts:61`에도 없음(README는 `repo start --port 4800`을 광고). 기존 지적(데몬 모드에서 포트 점유 시 ready 마커가 bind 전에 기록돼 false success·stale daemon.pid; legacy 리포 WORKFLOW.md fallback 경로에서 `server.port` 무시)은 재현 확인. 정상 동작 확인: `server.port` 단독/CLI 우선/점유 시 exit 1·락 해제/port 0 ephemeral.

packages/cli/src/commands/start.ts
```diff
@@ -1432,7 +1448,7 @@ async function startDaemon(
       ...(allowDuplicate ? ["--allow-duplicate"] : []),
       ...(assignedOnly ? ["--assigned-only"] : []),
       ...(bindAll ? ["--bind-all"] : []),
-      ...(httpPort !== undefined ? ["--http", String(httpPort)] : []),
+      ...(httpPort !== undefined ? ["--port", String(httpPort)] : []),
```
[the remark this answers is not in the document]
moncher-dev: Moved the daemon readiness marker until after `startHttpServer` has bound successfully. An occupied requested port now prevents the parent from creating a PID record or reporting daemon startup.

packages/cli/src/commands/start.ts
```diff
@@ -1022,9 +1030,16 @@ const handler = async (
...
+  const workflowPreflight = await preflightWorkflowStart(
+    projectConfig,
+    runtimeRoot
+  );
+  if (!workflowPreflight.ok) {
     return;
   }
+  const configuredServerPort = workflowPreflight.workflow?.server.port;
+  const requestedHttpPort =
+    parsed.httpPort ?? configuredServerPort ?? undefined;
```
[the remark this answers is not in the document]
moncher-dev: Added legacy `workflowSource: { type: "repo" }` fallback resolution using the repository directory and covered `server.port` from that path in `start.test.ts`.

docs/configuration.md
```diff
@@ -77,13 +77,23 @@ strict parser as workflow loading. Failures include a stable error code; the
...
 invalid override rather than ignoring it, so invalid configuration cannot silently
 disable dispatch for a state.
 
+## Optional HTTP Server
+
+Set `server.port` in `WORKFLOW.md` to enable the HTTP status API on that port.
+`0` requests an ephemeral port. `gh-symphony repo start --port [port]` is the
+preferred CLI form and overrides `server.port`; `--http [port]` remains a
+supported alias. A configured or CLI-requested port fails startup when it is
+already occupied, while the internal listener uses an automatic port only when
```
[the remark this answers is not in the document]
moncher-dev: Added the Configuration-layer mapping in `docs/architecture.md` and a workflow status-port Docker black-box scenario to `AGENT_TEST.md`; the happy Docker E2E run also passes.

packages/cli/src/commands/start.ts
moncher-dev: Resolved by signaling daemon readiness only after the worker HTTP listener has bound. The parent therefore cannot treat a child that fails its requested listener bind as started.

packages/cli/src/commands/start.ts
moncher-dev: Resolved by suppressing the `server.port` fallback when `--web` is present, leaving the dashboard's requested port unclaimed by the worker status listener.

packages/cli/src/commands/start.ts
moncher-dev: Resolved by tracking whether a value was supplied. Bare `--http` and `--port` retain default-port auto-increment; a value supplied through either flag, or `server.port`, fails on an occupied port.

packages/cli/src/commands/start.test.ts
moncher-dev: Added workflow-only coverage that asserts `server.port` binds the status API when no CLI HTTP option is given, plus legacy repository fallback coverage. The explicit occupied-port test continues to use a temporary OS-assigned port.

packages/cli/src/commands/start.ts
moncher-dev: The auto-increment loop is reachable again for bare `--http` or `--port`: those options use the default port without marking it explicit. Explicit numeric CLI values and `server.port` still fail fast on collision.

packages/cli/src/commands/start.ts
moncher-dev: Updated the conflict error to name the actual HTTP option used, so `--port ... --web` reports `--port` rather than `--http`.

docs/configuration.md
moncher-dev: The unrelated table reformatting is retained from the original patch; this rework does not add further formatting churn. The substantive `server.port` documentation remains isolated near the workflow validation section.

packages/cli/src/commands/start.ts
```diff
@@ -232,37 +232,48 @@ async function resolveConfiguredLinearToken(
...
-    // including its default workflow fallback. Avoid a divergent preflight.
-    return true;
-  }
+  const workflowPath =
+    configuredPath ??
+    (projectConfig.workflowSource?.type === "repo"
+      ? join(
+          projectConfig.repositoryDir ??
+            projectConfig.repository.path ??
+            process.cwd(),
```
hojinzs: **P1 — this `process.cwd()` fallback is what turned CI red, and it is a real product bug, not just a test artifact.**

CI `Test` is failing on this head: `lifecycle.test.ts > auto-selects the only configured project when start omits --project-id`, `Number of calls: 0`. I reproduced and bisected it locally:

```
$ npx vitest run packages/cli/src/commands/lifecycle.test.ts     # from the repo root, as CI runs it
Workflow preflight failed for /home/user/github-symphony/WORKFLOW.md: ...
 × auto-selects the only configured project when start omits --project-id
 Tests  1 failed | 15 passed (16)

$ git checkout origin/main -- packages/cli/src/commands/start.ts && npx vitest run <same file>
 Tests  16 passed (16)
```

Only `start.ts` differs, so this PR causes it. It passes under `pnpm --filter @gh-symphony/cli test` because that sets cwd to `packages/cli`, which has no `WORKFLOW.md` — so preflight takes the ENOENT escape at L265. CI runs vitest from the repo root, where a `WORKFLOW.md` *does* exist, so preflight parses it, the parse fails, and the handler exits before `startDaemon` ever spawns.

The mechanism is the bug. For any legacy repo-typed config where neither `repositoryDir` nor `repository.path` is set, `gh-symphony repo start` now reads whatever `WORKFLOW.md` happens to be in the operator's **current working directory** — potentially an entirely unrelated repository — and applies its `server.port`, or hard-fails startup if it does not parse. Before this PR that config path returned `workflow: null` and deferred to the daemon.

Suggested fix: drop the `process.cwd()` arm here and reuse the daemon's own resolution (see my note below on the divergence) rather than guessing. A config with no resolvable repository directory should keep returning `{ ok: true, workflow: null }`.

---

packages/cli/src/commands/start.ts
```diff
@@ -232,37 +232,48 @@ async function resolveConfiguredLinearToken(
...
+): Promise<{ ok: true; workflow: ParsedWorkflow | null } | { ok: false }> {
   const configuredPath = projectConfig.workflowSource?.path;
-  if (!configuredPath) {
-    // Legacy managed-project configs rely on the daemon's repository lookup,
-    // including its default workflow fallback. Avoid a divergent preflight.
-    return true;
-  }
+  const workflowPath =
+    configuredPath ??
+    (projectConfig.workflowSource?.type === "repo"
```
hojinzs, line 240: **P2 — this resolution diverges from the daemon's, so Codex's original P1 is only partly fixed.**

This resolves `repositoryDir ?? repository.path ?? cwd`. The daemon resolves the same thing through `OrchestratorService.resolveWorkflowRepositoryDirectory` (`packages/orchestrator/src/service.ts:5257`):

```ts
if (repository.path) return repository.path;
const localCloneUrlPath = this.resolveLocalCloneUrlPath(repository.cloneUrl);
if (localCloneUrlPath) return localCloneUrlPath;
return process.cwd();
```

Two concrete mismatches:

1. **`cloneUrl`-only configs are still broken.** When `repository.path` is unset but `cloneUrl` is a `file:` URL or a local absolute/relative path, the daemon reads that checkout's `WORKFLOW.md`; this preflight does not consider `cloneUrl` at all. I set up exactly that fixture on this branch (no `repositoryDir`, no `repository.path`, local `cloneUrl`, `server.port` set) and `repo start` never logged a status URL at all — the API stayed disabled. That is the same silent-ignore this issue exists to fix.
2. **Precedence is inverted.** This checks `repositoryDir` first; the daemon never consults `repositoryDir` and prefers `repository.path`. When both are set and differ, the preflight reads `server.port` from one file while the daemon runs from another.

Note the new test `reads server.port from a legacy repository workflow fallback` sets `repositoryDir` — the one field the daemon ignores — so it passes while the daemon would read elsewhere.

Suggested fix: export the daemon's resolution from `@gh-symphony/orchestrator` and call it here, so the two cannot drift. That also removes the duplicated `process.cwd()` guess flagged above.

---

packages/cli/src/commands/start.ts
```diff
@@ -1022,9 +1044,23 @@ const handler = async (
...
+  if (!workflowPreflight.ok) {
     return;
   }
+  const configuredServerPort =
+    parsed.webPort === undefined
+      ? workflowPreflight.workflow?.server.port
+      : undefined;
+  const requestedHttpPort =
+    parsed.httpPort ?? configuredServerPort ?? undefined;
+  const explicitHttpPort =
```
hojinzs, line 1081: **P2 — a bare `--http` / `--port` now silently discards `server.port` and binds 4680 instead.**

The fix for my earlier P2 correctly stopped a bare flag from forcing a hard failure, but `requestedHttpPort` on L1058 still lets the bare flag's *default* value win over `server.port`:

- bare `--http` → `parsed.httpPort = 4680` (L423), `httpPortExplicit = false`
- `requestedHttpPort = parsed.httpPort ?? configuredServerPort` → `4680`, and `configuredServerPort` is dropped on the floor

I confirmed this with a temporary test on this branch (not committed) — `server.port: <p>` in `WORKFLOW.md` plus `repo start --http`:

```
TEMPRESULT bare--http configured= 37135 bound= 4680
```

The configured port is silently ignored, which is precisely the §13.7 failure mode #751 was opened to fix. It is also the mirror of the reasoning we already agreed on for the auto-increment fix: a bare `--http` requests *the API*, not *a specific port*, so it should not outrank an explicit `server.port` either. The common case is a repo pinning `server.port: 4800` and an operator typing `repo start --http` out of habit.

Suggested fix — only let the CLI port win when a value was actually supplied:

```ts
const requestedHttpPort = parsed.httpPortExplicit
  ? parsed.httpPort
  : (configuredServerPort ?? parsed.httpPort);
const explicitHttpPort =
  parsed.httpPortExplicit || (parsed.httpPort === undefined && configuredServerPort !== undefined)
    ? requestedHttpPort !== undefined
    : false;
```

so `--port 4800` > `server.port` > bare-flag default, and a bare flag with no `server.port` keeps auto-incrementing from 4680. Worth a test for bare `--http` + `server.port`, since that combination is currently uncovered.

---

packages/cli/src/commands/start.ts
```diff
@@ -232,37 +232,48 @@ async function resolveConfiguredLinearToken(
...
+      environment,
+      {
+        supportedTrackerKinds: getSupportedTrackerKinds(),
+        resolveTrackerAdapter: resolveWorkflowConfigTrackerAdapter,
+      }
+    );
+    return { ok: true, workflow };
   } catch (error) {
     if ((error as NodeJS.ErrnoException | undefined)?.code === "ENOENT") {
+      if (!configuredPath) return { ok: true, workflow: null };
```
hojinzs: **P3 — the ENOENT escape is here, but an *unparseable* legacy repo workflow still hard-fails startup.**

This guard correctly keeps a missing repo-local `WORKFLOW.md` non-fatal for legacy configs. The `catch` tail below has no equivalent guard, so a legacy repo-typed config whose `WORKFLOW.md` fails validation now sets `exitCode = 1` and blocks `repo start` entirely.

The daemon does not treat that as fatal — `loadWorkflowFile` returns `createInvalidWorkflowResolution(...)` and the service carries on with an invalid marker (`packages/orchestrator/src/git.ts:659-676`). So this introduces a startup failure for configs that previously started, and it is the proximate cause of the CI failure above.

Suggested fix: extend the same `if (!configuredPath)` escape to the general `catch` tail, so only an *explicitly configured* workflow path can fail startup — which is what the original early-return comment ("Avoid a divergent preflight") was protecting.

---

hojinzs (changes requested):
## Review 2/6 — Request changes

Six of my seven round-1 findings and all three Codex threads are genuinely fixed — I verified each in the code rather than taking the replies at face value. But **CI is red on this head, and the failure is caused by this PR**, so I can't approve yet.

### Round-1 follow-up

| Round-1 finding | Status |
| --- | --- |
| P1 daemon readiness before bind | ✅ marker moved after `startHttpServer`; it's called unconditionally so the no-port path still signals, and nothing slow sits between the lock and the bind, so the 10s `waitForChildLock` budget is safe |
| P2 `server.port` steals `--web`'s port | ✅ suppressed when `--web` is present; matches the doc sentence now |
| P2 bare `--http` hard-fails | ✅ fixed via `httpPortExplicit` — but see the new P2 below, the fix introduced a precedence bug |
| P2 no test for `server.port` alone | ✅ added, and it asserts the bound port |
| P3 dead auto-increment loop | ✅ reachable again for bare flags |
| nit `--http` in conflict message | ✅ |
| nit docs Prettier churn | ✅ reasoned decline, accepted |
| Codex #1 daemon readiness / #2 legacy repo / #3 living docs | ✅ / ⚠️ partly (see P2) / ✅ architecture.md + AGENT_TEST.md rows added |

### Verification

`pnpm lint`, `pnpm typecheck`, `pnpm build` all pass. `packages/cli` 594/594 and `packages/core` 296/296 pass. (`runtime-claude`'s silence-timeout test flaked once locally and passed on re-run — unrelated, this PR touches only cli + core.)

**CI `Test` is failing** on `lifecycle.test.ts > auto-selects the only configured project when start omits --project-id`. I reproduced it and bisected to this PR: running vitest **from the repo root** (as CI does) fails, and swapping in `origin/main`'s `start.ts` — the only file changed — makes it pass 16/16. The `pnpm --filter` invocation in the PR description's evidence sets cwd to `packages/cli`, which has no `WORKFLOW.md`, which is why it was green locally.

That's not a test-only artifact: the new `process.cwd()` fallback makes `repo start` read whatever `WORKFLOW.md` sits in the operator's current directory.

### Findings

- **P1** — the `process.cwd()` fallback in `preflightWorkflowStart` turns CI red and lets an unrelated directory's `WORKFLOW.md` drive startup. (inline, with the bisect)
- **P2** — the legacy resolution (`repositoryDir ?? repository.path ?? cwd`) diverges from the daemon's (`repository.path ?? file:cloneUrl ?? cwd`, never `repositoryDir`). `cloneUrl`-only configs still silently ignore `server.port` — I built that fixture and no status API started at all — so Codex #2 is only partly fixed. The new legacy test sets `repositoryDir`, the one field the daemon ignores.
- **P2** — a bare `--http`/`--port` silently discards `server.port` and binds 4680. Verified: configured `37135`, bound `4680`. Same silent-ignore §13.7 is about.
- **P3** — an unparseable legacy repo workflow now hard-fails startup, while the daemon treats it as a non-fatal invalid resolution. This is the mechanism behind the P1.

The feature itself is in good shape and the scope is still clean — no over-engineering, no creep beyond core parser + CLI. What's left is the preflight resolution (one fix covers the P1, the P2 divergence and the P3) and the bare-flag precedence line.

---

hojinzs:
**리뷰 보완 (독립 검증, 2회차 Request changes와 결론 일치)** — P1 재현: `start.ts:236-247` legacy fallback이 `repositoryDir ?? repository.path ?? process.cwd()`+`WORKFLOW.md`를 읽어, 루트에서 도는 CI(`npx vitest run --coverage`)에서 `lifecycle.test.ts:219`가 실제 루트 WORKFLOW.md를 파싱 → mock에 `getSupportedTrackerKinds` 없음 → 실패(로컬 루트 실행 1 failed/1874). `process.cwd()` 폴백 제거(명시 경로만). 추가 2건: (1) 우선순위 테스트 `start.test.ts:1881`이 여전히 `not.toBe`만 단언(1회차 nit 미반영) → 실제 포트 단언, (2) 1회차에서 제거된 auto-increment 단위 테스트를 복원(경로가 다시 도달 가능). 확인된 것: `--port` 등록·help·completion, 데몬 false success 수정(exit 1, daemon.pid 없음), `server.port` 단독/legacy/우선순위 동작.

packages/cli/src/commands/start.ts
moncher-dev: Resolved in 72c2dee9. CLI preflight now mirrors the daemon's repository lookup for `repository.path` and local `file:`/path clone URLs, rather than consulting `repositoryDir`. The legacy fallback test now covers a `file:` clone URL.

packages/cli/src/commands/start.ts
moncher-dev: Resolved in 72c2dee9. Only CLI options with an explicit numeric value override `server.port`; bare `--http` and `--port` now use `server.port` when present. Both aliases are covered with an OS-assigned temporary port.

packages/cli/src/commands/start.ts
moncher-dev: Resolved in 72c2dee9. An implicit legacy repository workflow now returns no preflight configuration for any parse or read failure, matching the daemon's non-fatal invalid-workflow behavior. Added a regression test that confirms startup continues.

packages/cli/src/commands/start.ts
```diff
@@ -232,40 +233,69 @@ async function resolveConfiguredLinearToken(
...
     const environment = resolveManagedProjectEnvironment(
       projectConfig,
       runtimeRoot
     );
-    parseWorkflowMarkdown(await readFile(configuredPath, "utf8"), environment, {
-      supportedTrackerKinds: getSupportedTrackerKinds(),
-      resolveTrackerAdapter: resolveWorkflowConfigTrackerAdapter,
-    });
-    return true;
+    const workflow = parseWorkflowMarkdown(
```
hojinzs, line 250: **P2 — carry `workflowPath` through when you resolve the merge conflict, and pass the *resolved* path here.**

This PR is currently `mergeable_state: dirty`. `main` has moved to `c549379` (#740, "normalize workflow path expansion"), and the conflict is in this exact function. `main`'s version now reads:

```ts
parseWorkflowMarkdown(await readFile(configuredPath, "utf8"), environment, {
  supportedTrackerKinds: getSupportedTrackerKinds(),
  resolveTrackerAdapter: resolveWorkflowConfigTrackerAdapter,
  workflowPath: configuredPath,   // ← added by #740
});
```

`workflowPath` is not cosmetic. On `main`, `readOptionalPath` resolves every relative front-matter path against it (`parser.ts:1241`):

```ts
return resolve(
  options.workflowPath ? dirname(options.workflowPath) : process.cwd(),
  normalized
);
```

So two things to get right in the merge:

1. **Don't drop it.** Resolving the conflict by taking this branch's side wholesale silently reverts #740 for the configured-path case — relative `workspace.root`, hook script paths, and tracker file paths would go back to resolving against `process.cwd()`.
2. **Pass `workflowPath`, not `configuredPath`.** This PR's whole point is that `workflowPath` can now be the legacy repo-derived `join(resolveWorkflowRepositoryDirectory(...), "WORKFLOW.md")`. For those configs `configuredPath` is `undefined`, so passing it would resolve the repo's own relative paths against the operator's cwd — the same class of cwd-leak we fixed in round 2, just moved one layer down.

The local variable is already named `workflowPath` and holds the right value, so `workflowPath` is the correct thing to pass in both branches.

Please confirm CI is green after the merge — there are currently **no check runs at all** on `72c2dee`, so this head has never been validated by CI.

---

packages/cli/src/commands/start.ts
```diff
@@ -232,40 +233,69 @@ async function resolveConfiguredLinearToken(
...
+  } catch {
+    if (
+      isAbsolute(repository.cloneUrl) ||
+      repository.cloneUrl.startsWith(".")
+    ) {
+      return repository.cloneUrl;
+    }
+  }
+
+  return process.cwd();
```
hojinzs, line 297: nit (non-blocking) — this is now a faithful but hand-copied duplicate of the daemon's resolution.

I diffed it against `OrchestratorService.resolveWorkflowRepositoryDirectory` / `resolveLocalCloneUrlPath` (`packages/orchestrator/src/service.ts:5257-5281`) case by case — `repository.path`, `file:` URL, absolute path, `./` relative, unparseable URL, and the `fileURLToPath` throw path all agree, including the final `process.cwd()`. So the round-2 P2 divergence is genuinely fixed, and I verified the `file:` cloneUrl fixture end to end.

The remaining concern is only drift: two copies of the same rule in two packages, with nothing tying them together. My round-2 suggestion was to export the daemon's version from `@gh-symphony/orchestrator` and call it here; you duplicated instead, which is a legitimate call given `start.ts` already imports from that package and the function is 15 lines. Not worth another round — just noting that the next change to either copy needs to touch both.

---

packages/cli/src/commands/start.test.ts
```diff
@@ -1854,16 +1865,68 @@ Handle {{issue.identifier}}.\n`,
...
+    const stdout = captureWrites(process.stdout);
+
+    try {
+      const startPromise = startModule.default(
+        ["--port", "0"],
+        baseOptions(configDir)
+      );
+
+      const url = await waitForHttpUrl(stdout.output);
+      expect(new URL(url).port).not.toBe(String(address.port));
```
hojinzs: nit (non-blocking, carried over from rounds 1–2 — raised twice, not addressed and not replied to)

Two small test-coverage points. Neither blocks the merge; recording them so they're not lost.

**1. This assertion is still `not.toBe`.** With `--port 0` the bound port is ephemeral, so the test can only say "not the configured one". It does prove the CLI value won — if `server.port` had won, `explicitHttpPort` would be true against the still-listening `blocker` and the start would throw — so it isn't vacuous. But asserting an actual port (bind a probe, close it, pass that number as `--port`) would state the precedence directly instead of by elimination.

**2. The auto-increment test removed in round 1 was never restored.** The old `increments the port when the requested HTTP port is already in use` case was replaced by this one, and the `continue` branch in `startHttpServer` now has no unit coverage.

That branch is genuinely live again after the round-2 fix — I confirmed it black-box on this head, with an occupied port as `server.port`:

```
$ gh-symphony project start --http        # server.port: 41809, already occupied
listening on http://localhost:41810       # auto-incremented, as intended
```

So the reachability argument in your round-2 reply holds. It's the one behavior in this area that is now exercised only by hand.

---

hojinzs (approved):
## Review 3/6 — Approve (conditional on a clean merge + green CI)

All four round-2 findings are fixed, and I verified each one independently rather than accepting the replies. The feature now works end to end in a real black-box run. Approving to close out the ping-pong — what remains is a mandatory rebase and two test nits, none of which warrant another round.

### Round-2 follow-up

| Round-2 finding | Status |
| --- | --- |
| **P1** `process.cwd()` fallback turns CI red / reads an unrelated `WORKFLOW.md` | ✅ Fixed. The `catch` tail now returns `{ ok: true, workflow: null }` for any implicit workflow, matching the daemon's non-fatal invalid-workflow behavior. `process.cwd()` survives only as the last arm of a resolution that now faithfully mirrors the daemon's own — which is what the P2 asked for, so I accept it. |
| **P2** legacy resolution diverges from the daemon | ✅ Fixed. Now `repository.path ?? file:/local cloneUrl ?? cwd`; I diffed it case by case against `service.ts:5257-5281` and they agree. The test fixture uses a `file:` clone URL rather than `repositoryDir`. |
| **P2** bare `--http`/`--port` silently discards `server.port` | ✅ Fixed and confirmed black-box (see matrix). |
| **P3** unparseable legacy workflow hard-fails startup | ✅ Fixed, with a regression test. |
| supplementary: precedence test still `not.toBe`; auto-increment test not restored | ⚠️ Not addressed, no reply — inline, non-blocking |

### Verification

`pnpm lint` ✅ · `pnpm typecheck` ✅ · `pnpm build` ✅ · `packages/cli` + `packages/core` **893/893** ✅

The round-2 blocker is gone at its root: `npx vitest run packages/cli/src/commands/lifecycle.test.ts` **from the repo root**, exactly as CI invokes it, now passes **16/16**. That was the failing invocation, and it fails on `8b713f2`.

(`pnpm format` reports 85 files including `pnpm-lock.yaml` and `prettier.config.mjs`; `completion.ts` is flagged identically on `origin/main`, so that's pre-existing repo state, not this PR.)

### Smoke test

Black-box against the built CLI (`dist/index.js`), file tracker, real listeners — not mocks. Configured `server.port` in bold:

| # | Invocation | Expected | Observed |
| --- | --- | --- | --- |
| 1 | no flag | binds **36997** | `36997` ✅ |
| 2 | `--port 0` | CLI wins, ephemeral | `43265` ✅ |
| 3 | bare `--http` | uses **36997** | `36997` ✅ |
| 4 | bare `--port` | uses **36997** | `36997` ✅ |
| 5 | `--web 0` | `server.port` suppressed | dashboard `44929`, status port unclaimed ✅ |
| 6 | `--port <occupied>` | fail fast | `exit 1`, `HTTP server port 44565 is already in use` ✅ |
| 7 | `--daemon --port <occupied>` | no false success | `exit 1`, no PID file, no "started in background" ✅ |
| 8 | **`server.port` occupied**, no flag | fail fast (§13.7) | `exit 1` ✅ |
| 9 | bare `--http`, **`server.port` occupied** | auto-increment | `41809 → 41810` ✅ |

Cases 3/4 close the round-2 P2; case 7 confirms the round-1 daemon-readiness fix survives; case 9 shows the auto-increment branch is genuinely reachable again, as your round-2 reply claimed.

`--port` is registered on both `repo start` and `project start`, and appears in `--help` and `completion.ts` — the round-1 blocker stays fixed.

### Requirements vs. #751

| --- | --- |
| `server.port` parsing (extension key, documented) | ✅ `readOptionalPort`, 0–65535, non-integers rejected |
| `--port` alias added, `--http` retained | ✅ registered, help, completion |
| Precedence CLI `--port` > `server.port` > default | ✅ cases 1–4 |
| Auto-increment only without an explicit option | ✅ cases 3, 4, 9 |
| Explicit-port bind failure is an error | ✅ cases 6, 7, 8 |
| `docs/configuration.md` updated | ✅ |
| Precedence + bind-failure tests, no fixed 4680 | ✅ OS-assigned temporary ports throughout |
| changeset (patch) | ✅ |
| Scope: core parser + CLI only | ✅ |

No over-engineering and no scope creep — the core change is a `WorkflowServerConfig` type plus one `readOptionalPort` reader, and everything else is CLI wiring.

### Before merging

- **Resolve the conflict against `main` carefully.** `main` is now `c549379` (#740) and the conflict lands in `preflightWorkflowStart`. Taking this branch's side wholesale would silently revert #740; the legacy branch also needs the *resolved* `workflowPath`, not `configuredPath`. Details inline.
- **CI has never run on `72c2dee`** — zero check runs. My local run from the repo root is green, but please confirm CI agrees on the merged head.

Nice work — this converged cleanly, and the round-2 fixes addressed the mechanism rather than the symptom.

---

hojinzs:
**독립 검증 (3회차 조건부 승인 보완) — 현재 상태로는 Land 불가** — (1) main(c285268)과 충돌: `packages/cli/src/commands/start.ts`(main이 같은 `parseWorkflowMarkdown` 호출에 `workflowPath: configuredPath` 추가 — legacy 분기는 *해석된* 경로를 넘길 것)과 `AGENT_TEST.md`(양쪽 행 유지) → 리베이스 후 CI 필요(현재 헤드 체크 없음). (2) 2회차에서 두 번 요청한 테스트 2건 여전히 미반영: 우선순위 테스트 `start.test.ts:1882` `not.toBe`만 단언, auto-increment 단위 테스트(main `:1834`) 미복원. (3) 새 P3: `server.port` 점유 + bare `--http` → 오류 대신 4985로 조용히 auto-increment(§13.7 "positive server.port values bind that port"와 상충) — 명시 오류 또는 경고. 확인된 것: cwd fallback CI 원인 해결(root vitest 16/16), bare `--http`/`--port`가 `server.port` 사용, legacy cloneUrl 경로 데몬과 일치, 데몬 모드 정상. 리베이스 + (2)(3) 반영 후 재승인 → Land.

packages/cli/src/commands/start.ts
moncher-dev: Resolved in 4ccbded6 after rebasing onto current `main`. `preflightWorkflowStart` now passes the resolved `workflowPath` to `parseWorkflowMarkdown`, preserving relative front-matter path resolution for both configured and legacy repository workflows. Full validation passed on the rebased head.

packages/cli/src/commands/start.test.ts
moncher-dev: Resolved in 4ccbded6. The CLI precedence test now reserves and releases an OS-assigned port, passes that exact port through `--port`, and asserts the bound URL uses it. A separate regression test occupies the configured `server.port`, starts with bare `--http`, and asserts that startup selects a higher port, exercising the auto-increment branch.

packages/cli/src/commands/start.ts
moncher-dev: Acknowledged. The duplicated repository resolution remains intentionally limited to the CLI preflight so this issue stays within core parser and CLI scope; the current implementation matches the daemon behavior. A future shared-helper refactor would need separate scope.

packages/cli/src/commands/start.test.ts
```diff
@@ -1831,11 +1832,24 @@ Handle {{issue.identifier}}.\n`,
...
+    );
+    const address = probe.address();
+    if (!address || typeof address === "string") {
+      probe.close();
+      throw new Error("Expected TCP address");
+    }
+    await new Promise<void>((resolve, reject) =>
+      probe.close((error) => (error ? reject(error) : resolve()))
+    );
+    await configureWorkflow(configDir, address.port);
```
hojinzs, line 1852: **P3 (non-blocking) — this precedence test is now vacuous: `server.port` and `--port` are set to the *same* port.**

My round-3 nit asked for a real port instead of `not.toBe`, and you did bind/close a probe — but the same `address.port` is used for both sides:

- L1852 `configureWorkflow(configDir, address.port)` → `server.port = address.port`
- L1880 `["--port", String(address.port)]` → CLI port = `address.port`

So `expect(new URL(url).port).toBe(String(address.port))` holds whichever source wins. The test named "uses --port in preference to server.port" cannot observe the precedence it asserts.

I proved it on this head by inverting the precedence in the source so `server.port` always wins — the exact bug this test exists to catch:

```ts
// packages/cli/src/commands/start.ts:1078
const requestedHttpPort = configuredServerPort ?? parsed.httpPort;
```

```
$ npx vitest run packages/cli/src/commands/start.test.ts -t "uses --port in preference to server.port"
 ✓ packages/cli/src/commands/start.test.ts (48 tests | 47 skipped)
 Tests  1 passed | 47 skipped (48)
```

It passes against a fully inverted implementation. This is a regression in signal from the round-1/2 version: the old `not.toBe` was weak, but it did fail if `server.port` won.

**The behavior itself is correct** — this is test quality only, which is why it is not blocking. I reserved a second distinct port and re-ran the same case:

```
expected '37291' to be '40287'   # inverted impl → now FAILS, as it should
```

and against the shipped implementation the two-port version passes, with `--port` winning. Confirmed black-box too (case 2 below in my review body): `server.port: 42241`, `--port 44963` → bound `44963`.

Fix is one extra reserved port:

```ts
const configured = await reserveClosedPort();  // distinct from address.port
await configureWorkflow(configDir, configured);
// ... startModule.default(["--port", String(address.port)], ...)
expect(new URL(url).port).toBe(String(address.port));   // and !== configured
```

Point 2 of that round-3 thread — restoring auto-increment coverage — is genuinely fixed: `auto-increments an occupied configured port for a bare HTTP alias` exercises the `continue` branch, and I confirmed it black-box (`36411` occupied → bound `36412`).

---

hojinzs (approved):
## Review 4/6 — Approve ✅

The two round-3 conditions are met: **the merge is clean and CI is green on `4ccbded`** (`Test` ✅, `Container Smoke` ✅ — the first CI run this branch has ever had). I re-verified the rebase fix and the full feature end to end rather than accepting the replies.

### Round-3 follow-up

| Round-3 item | Status |
| --- | --- |
| **P2** carry `workflowPath` through the merge, pass the *resolved* path | ✅ Fixed correctly. `preflightWorkflowStart` passes the local `workflowPath`, not `configuredPath`, so #740's relative front-matter resolution is preserved **and** legacy repo-derived workflows resolve against their own directory instead of the operator's cwd. Both halves of the concern are addressed. |
| **nit** duplicated daemon resolver | ✅ Reasoned decline accepted. I re-diffed the CLI copy against `resolveWorkflowRepositoryDirectory` / `resolveLocalCloneUrlPath` (`service.ts:5309-5333`) on this head — `repository.path`, `file:` URL, absolute, `./` relative, non-`file:` URL, and the `fileURLToPath` throw path all still agree, including the final `process.cwd()`. |
| **nit 1** precedence test asserts a real port | ⚠️ Implemented, but with the *same* port on both sides — the test is now vacuous. Inline, non-blocking. |
| **nit 2** restore auto-increment coverage | ✅ Genuinely restored and exercises the `continue` branch. |

### Verification

`pnpm build` ✅ · `pnpm typecheck` ✅ · `pnpm lint` ✅ · `packages/cli` **605/605** + `packages/core` **307/307** = **912** ✅ (893 → 912; the 19 new tests are the round-3 additions).

The round-2 blocker stays fixed after the rebase — `npx vitest run packages/cli/src/commands/lifecycle.test.ts` **from the repo root**, exactly as CI invokes it: **16/16**.

Note the head sits on `c285268`; `main` has since gained only two release commits (`chore: release` + the #760 changeset merge). No conflict, and GitHub reports `mergeable_state: clean`.

### Smoke test

Black-box against the built CLI (`packages/cli/dist/index.js`), file tracker, real listeners, a fresh config dir per case — not mocks. Configured `server.port` in **bold**:

| # | Invocation | Expected | Observed |
| --- | --- | --- | --- |
| 1 | no flag, **41361** | binds 41361 | `41361` ✅ |
| 2 | `--port 44963`, **42241** | CLI wins (**distinct** ports) | `44963` ✅ |
| 3 | bare `--http`, **45439** | uses configured | `45439` ✅ |
| 4 | bare `--port`, **33117** | uses configured | `33117` ✅ |
| 5 | `--port <occupied>` | fail fast | `exit 1`, `already in use` ✅ |
| 6 | **occupied**, no flag | fail fast (§13.7) | `exit 1`, `already in use` ✅ |
| 7 | bare `--http`, **36411 occupied** | auto-increment | `36412` ✅ |
| 8 | no `server.port`, no flag | no advertised status URL | none ✅ |
| A | `--web <p>` with **`server.port: <p>`** | configured suppressed | dashboard binds `35207`, unstolen ✅ |
| B | `--daemon --port <occupied>` | no false success | `exit 1`, no "started in background", no PID file ✅ |

Case 2 uses two **distinct** ports and is the direct evidence that precedence is genuinely correct — which is what makes the inline finding test-quality rather than a product bug. Cases A and B confirm the round-1 P2 and P1 fixes survive the rebase.

### Requirements vs. #751

| --- | --- |
| `server.port` parsing (extension key, documented) | ✅ `readOptionalPort` 0–65535 over `readOptionalIntegerLike`, which rejects non-integers and quoted strings |
| `--port` alias added, `--http` retained | ✅ on both `repo start` and `project start`, in `--help` and `completion.ts` |
| Precedence CLI `--port` > `server.port` > default | ✅ cases 1–4 |
| Auto-increment only without an explicit option | ✅ cases 3, 4, 7 |
| Explicit-port bind failure is an error | ✅ cases 5, 6, B |
| `docs/configuration.md` updated | ✅ |
| Precedence + bind-failure tests, no fixed 4680 | ✅ OS-assigned temporary ports throughout |
| changeset (patch) | ✅ |
| Scope: core parser + CLI only | ✅ |

No over-engineering and no scope creep: a `WorkflowServerConfig` type plus one `readOptionalPort` reader in core, everything else CLI wiring and docs.

### Verdict

Approving. Every acceptance criterion in #751 is met and independently verified, all three Codex threads and all eleven of my round-1/2/3 findings are resolved or reasonably declined, and CI is green on a clean merge. The single open item is a P3 test-quality regression that does not affect behavior — worth a one-line follow-up, but not another review round.

Nice work — this converged well across four rounds, and each round fixed the mechanism rather than the symptom.

---

packages/cli/src/commands/start.test.ts
hojinzs: Verified on `4ccbded`:

**Point 2 (auto-increment coverage) — fixed.** `auto-increments an occupied configured port for a bare HTTP alias` keeps the blocker listening on the configured `server.port` and asserts the bound port is higher, so the `continue` branch in `startHttpServer` has unit coverage again. Confirmed black-box as well: `36411` occupied → bound `36412`.

**Point 1 (precedence assertion) — implemented, but the fix is vacuous.** The test now reserves a real port, but uses that *same* port for both `server.port` and `--port`, so the assertion holds whichever side wins. I proved it by inverting the precedence in `start.ts` and the test still passed.

Resolving this thread; the remaining half is carried into a fresh inline comment on the test so it stays visible. It is P3 and non-blocking — I confirmed with two distinct ports that the actual precedence is correct — so it should not hold up the merge.

---

PR #763, opened 2026-08-29
fix(observability): accumulate session runtime

### LINKED ISSUE #750: fix(observability): seconds_running as accumulated session runtime (§13.5)
Epic: #651 · 분할 원본: #676 (E3) · 리포트: `docs/reports/2026-08-28-upstream-spec-drift-research.md`

## 문제
§13.5: `seconds_running`은 세션 runtime 합이어야 하지만 현재 wall-clock span이다(`snapshot-builder.ts:283-309`); 테스트는 `>0`만 확인.

## 할 일
- `seconds_running` = 종료된 세션들의 누적 runtime + 활성 세션의 스냅샷 시점 경과
- 세션 경계(worker 재시작/retry)에서 누적이 유지되도록 run 레코드에 필요한 최소 필드 추가

## 완료 기준
- 다중 세션(재시도 포함) 픽스처로 합산 값을 정확히 단언하는 테스트; changeset(patch)
- 범위: core/observability(+ run 레코드 필드) 만.

moncher-dev:
## TL;DR

`seconds_running` now reports accumulated worker-session runtime, including completed sessions and the current active session while excluding retry backoff and queue time.

## Change-point diagram

```text
completed session ──> freeze cumulativeRuntimeMs + completedAt
retry/restart ──────> retain runtimeLifecycleId (legacy: createdAt)
active session ─────> accumulated runtime + snapshot elapsed
```

## Start here

- `packages/core/src/observability/snapshot-builder.ts` groups records by lifecycle and computes only completed-session runtime plus the active-session elapsed span.
- `packages/orchestrator/src/service.ts` preserves the lifecycle identity through retry/restart, including a retry that crosses the runtime-record upgrade.
- `packages/core/src/observability/snapshot-builder.test.ts` contains exact multi-session, retry-gap, and mixed-version assertions.

## Changes

- Added persisted `cumulativeRuntimeMs` and `runtimeLifecycleId` support so retries and worker restarts retain accumulated runtime without double counting.
- Freeze retry records at the completed session boundary; later retry scheduling does not extend the metric.
- Added exact regression coverage for a legacy record followed by an upgraded active retry: 120 seconds completed + 120 seconds active = 240 seconds.

## Evidence

- `pnpm lint` — pass
- `pnpm test` — pass
- `pnpm typecheck` — pass
- `pnpm build` — pass
- `pnpm --filter @gh-symphony/core test -- snapshot-builder.test.ts` — pass (33 tests)
- `pnpm --filter @gh-symphony/orchestrator test -- service.test.ts -t 'queues a clean convergence failure when the tracker is already in the retryable state|classifies incomplete dirty turns and redispatches with recovery context'` — pass
- `./e2e/run-e2e.sh happy 60` — pass (Docker lifecycle reached retry then idle)

## Risks & rollback

The new run-record fields are optional for backward compatibility. Legacy retry/restart flows use the prior record’s `createdAt` as their lifecycle identity. Reverting the PR restores the prior wall-clock metric behavior.

## Changed files

- `.changeset/750-accumulated-session-runtime.md`
- `packages/core/src/contracts/status-surface.ts`
- `packages/core/src/observability/snapshot-builder.ts`
- `packages/core/src/observability/snapshot-builder.test.ts`

## Issues — Closed #750

Fixes #750

- [ ] Observe a retried worker in the status surface and confirm retry idle time does not increase `seconds_running`.

### DIFF
--- a/packages/core/src/observability/snapshot-builder.test.ts
+++ b/packages/core/src/observability/snapshot-builder.test.ts
@@ -251,6 +251,7 @@ describe("buildProjectSnapshot", () => {
     });
     const run2 = mockRun({
       runId: "run-002",
+      createdAt: "2024-01-01T00:06:00Z",
       tokenUsage: {
         inputTokens: 2000,
         outputTokens: 1000,
@@ -274,7 +275,7 @@ describe("buildProjectSnapshot", () => {
     expect(snapshot.codexTotals!.inputTokens).toBe(3000);
     expect(snapshot.codexTotals!.outputTokens).toBe(1500);
     expect(snapshot.codexTotals!.totalTokens).toBe(4500);
-    expect(snapshot.codexTotals!.secondsRunning).toBeGreaterThan(0);
+    expect(snapshot.codexTotals!.secondsRunning).toBe(540);
   });
 
   it("handles runs with missing tokenUsage gracefully", () => {
@@ -826,6 +827,48 @@ describe("buildProjectSnapshot", () => {
 
     // 5 minutes = 300 seconds
     expect(snapshot.codexTotals!.secondsRunning).toBe(300);
+  });
+
+  it("accumulates completed retry sessions without counting retry gaps", () => {
+    const initialStartedAt = "2024-01-01T00:00:00Z";
+    const completedSession = mockRun({
+      runId: "run-001",
+      status: "failed",
+      createdAt: initialStartedAt,
+      updatedAt: "2024-01-01T00:02:00Z",
+      startedAt: initialStartedAt,
+      completedAt: "2024-01-01T00:02:00Z",
+    });
+    const retrySession = mockRun({
+      runId: "run-002",
+      status: "failed",
+      createdAt: initialStartedAt,
+      updatedAt: "2024-01-01T00:09:00Z",
+      startedAt: "2024-01-01T00:07:00Z",
+      completedAt: "2024-01-01T00:09:00Z",
+      cumulativeRuntimeMs: 120_000,
+    });
+    const activeSession = mockRun({
+      runId: "run-003",
+      status: "running",
+      createdAt: initialStartedAt,
+      updatedAt: "2024-01-01T00:12:00Z",
+      startedAt: "2024-01-01T00:11:00Z",
+      completedAt: null,
+      cumulativeRuntimeMs: 240_000,
+    });
+
+    const snapshot = buildProjectSnapshot({
+      project: mockProject(),
+      activeRuns: [activeSession],
+      allRuns: [completedSession, retrySession, activeSession],
+      summary: { dispatched: 3, suppressed: 0, recovered: 0 },
+      lastTickAt: "2024-01-01T00:14:00Z",
+      lastError: null,
+    });
+
+    // 2 min + 2 min + 3 min active; retry gaps (5 min and 2 min) are excluded.
+    expect(snapshot.codexTotals!.secondsRunning).toBe(420);
   });
 
   it("handles retrying run with null retryKind by defaulting to 'failure'", () => {
--- a/packages/core/src/observability/snapshot-builder.test.ts
+++ b/packages/core/src/observability/snapshot-builder.test.ts
@@ -956,6 +956,38 @@ describe("buildProjectSnapshot", () => {
     expect(snapshot.codexTotals!.secondsRunning).toBe(240);
   });
 
+  it("does not double count a lifecycle that crosses the runtime upgrade", () => {
+    const legacySession = mockRun({
+      runId: "run-001",
+      status: "failed",
+      createdAt: "2024-01-01T00:00:00Z",
+      startedAt: "2024-01-01T00:00:00Z",
+      completedAt: "2024-01-01T00:02:00Z",
+      updatedAt: "2024-01-01T00:02:00Z",
+    });
+    const activeRetry = mockRun({
+      runId: "run-002",
+      status: "running",
+      createdAt: "2024-01-01T00:07:00Z",
+      startedAt: "2024-01-01T00:07:00Z",
+      completedAt: null,
+      updatedAt: "2024-01-01T00:09:00Z",
+      cumulativeRuntimeMs: 120_000,
+      runtimeLifecycleId: "2024-01-01T00:00:00Z",
+    });
+
+    const snapshot = buildProjectSnapshot({
+      project: mockProject(),
+      activeRuns: [activeRetry],
+      allRuns: [legacySession, activeRetry],
+      summary: { dispatched: 2, suppressed: 0, recovered: 0 },
+      lastTickAt: "2024-01-01T00:09:00Z",
+      lastError: null,
+    });
+
+    expect(snapshot.codexTotals!.secondsRunning).toBe(240);
+  });
+
   it("handles retrying run with null retryKind by defaulting to 'failure'", () => {
     const run = mockRun({
       status: "retrying",
--- a/packages/core/src/observability/snapshot-builder.test.ts
+++ b/packages/core/src/observability/snapshot-builder.test.ts
@@ -871,6 +871,91 @@ describe("buildProjectSnapshot", () => {
     expect(snapshot.codexTotals!.secondsRunning).toBe(420);
   });
 
+  it("accumulates a tick-dispatched retry with a fresh createdAt exactly once", () => {
+    const completedSession = mockRun({
+      runId: "run-001",
+      status: "failed",
+      createdAt: "2024-01-01T00:00:00Z",
+      updatedAt: "2024-01-01T00:02:00Z",
+      startedAt: "2024-01-01T00:00:00Z",
+      completedAt: "2024-01-01T00:02:00Z",
+      runtimeLifecycleId: "run-001",
+    });
+    const activeRetry = mockRun({
+      runId: "run-002",
+      status: "running",
+      createdAt: "2024-01-01T00:07:00Z",
+      updatedAt: "2024-01-01T00:09:00Z",
+      startedAt: "2024-01-01T00:07:00Z",
+      completedAt: null,
+      cumulativeRuntimeMs: 120_000,
+      runtimeLifecycleId: "run-001",
+    });
+
+    const snapshot = buildProjectSnapshot({
+      project: mockProject(),
+      activeRuns: [activeRetry],
+      allRuns: [completedSession, activeRetry],
+      summary: { dispatched: 2, suppressed: 0, recovered: 0 },
+      lastTickAt: "2024-01-01T00:09:00Z",
+      lastError: null,
+    });
+
+    expect(snapshot.codexTotals!.secondsRunning).toBe(240);
+  });
+
+  it("uses frozen retry runtime instead of capacity wait bookkeeping time", () => {
+    const retryingRun = mockRun({
+      status: "retrying",
+      startedAt: null,
+      completedAt: "2024-01-01T00:02:00Z",
+      updatedAt: "2024-01-01T00:07:00Z",
+      cumulativeRuntimeMs: 120_000,
+      runtimeLifecycleId: "run-001",
+    });
+
+    const snapshot = buildProjectSnapshot({
+      project: mockProject(),
+      activeRuns: [],
+      allRuns: [retryingRun],
+      summary: { dispatched: 1, suppressed: 0, recovered: 0 },
+      lastTickAt: "2024-01-01T00:07:00Z",
+      lastError: null,
+    });
+
+    expect(snapshot.codexTotals!.secondsRunning).toBe(120);
+  });
+
+  it("sums legacy lifecycle sessions without cumulative runtime", () => {
+    const firstSession = mockRun({
+      runId: "run-001",
+      status: "failed",
+      createdAt: "2024-01-01T00:00:00Z",
+      startedAt: "2024-01-01T00:00:00Z",
+      completedAt: "2024-01-01T00:02:00Z",
+      updatedAt: "2024-01-01T00:02:00Z",
+    });
+    const secondSession = mockRun({
+      runId: "run-002",
+      status: "failed",
+      createdAt: "2024-01-01T00:00:00Z",
+      startedAt: "2024-01-01T00:07:00Z",
+      completedAt: "2024-01-01T00:09:00Z",
+      updatedAt: "2024-01-01T00:09:00Z",
+    });
+
+    const snapshot = buildProjectSnapshot({
+      project: mockProject(),
+      activeRuns: [],
+      allRuns: [firstSession, secondSession],
+      summary: { dispatched: 2, suppressed: 0, recovered: 0 },
+      lastTickAt: "2024-01-01T00:09:00Z",
+      lastError: null,
+    });
+
+    expect(snapshot.codexTotals!.secondsRunning).toBe(240);
+  });
+
   it("handles retrying run with null retryKind by defaulting to 'failure'", () => {
     const run = mockRun({
       status: "retrying",
--- a/packages/core/src/observability/snapshot-builder.ts
+++ b/packages/core/src/observability/snapshot-builder.ts
@@ -304,33 +304,57 @@ function aggregateTokenUsage(
   let inputTokens = 0;
   let outputTokens = 0;
   let totalTokens = 0;
-  let earliestStart: number | null = null;
-  let latestEnd: number | null = null;
-
   for (const run of runs) {
     if (run.tokenUsage) {
       inputTokens += run.tokenUsage.inputTokens;
       outputTokens += run.tokenUsage.outputTokens;
       totalTokens += run.tokenUsage.totalTokens;
     }
-    if (run.startedAt) {
-      const start = new Date(run.startedAt).getTime();
-      if (earliestStart === null || start < earliestStart) {
-        earliestStart = start;
-      }
-    }
-    const end = run.completedAt
-      ? new Date(run.completedAt).getTime()
-      : new Date(lastTickAt).getTime();
-    if (latestEnd === null || end > latestEnd) {
-      latestEnd = end;
+  }
+
+  const runtimeMs = Array.from(latestSessionsByRunLifecycle(runs).values())
+    .map((run) => runtimeMsForSnapshot(run, lastTickAt))
+    .reduce((total, sessionRuntimeMs) => total + sessionRuntimeMs, 0);
+  const secondsRunning = Math.max(0, Math.round(runtimeMs / 1000));
+
+  return { inputTokens, outputTokens, totalTokens, secondsRunning };
+}
+
+function latestSessionsByRunLifecycle(
+  runs: OrchestratorRunRecord[]
+): Map<string, OrchestratorRunRecord> {
+  const latestRuns = new Map<string, OrchestratorRunRecord>();
+
+  for (const run of runs) {
+    const key = `${run.projectId}:${run.issueId}:${run.createdAt}`;
+    const current = latestRuns.get(key);
+    if (
+      !current ||
+      new Date(run.updatedAt).getTime() >= new Date(current.updatedAt).getTime()
+    ) {
+      latestRuns.set(key, run);
     }
   }
 
-  const secondsRunning =
-    earliestStart !== null && latestEnd !== null
-      ? Math.max(0, Math.round((latestEnd - earliestStart) / 1000))
-      : 0;
+  return latestRuns;
+}
 
-  return { inputTokens, outputTokens, totalTokens, secondsRunning };
+function runtimeMsForSnapshot(
+  run: OrchestratorRunRecord,
+  lastTickAt: string
+): number {
+  const accumulatedRuntimeMs = run.cumulativeRuntimeMs ?? 0;
+  if (!run.startedAt) {
+    return accumulatedRuntimeMs;
+  }
+
+  const startedAtMs = new Date(run.startedAt).getTime();
+  const endedAt =
+    run.completedAt ?? (run.status === "running" ? lastTickAt : run.updatedAt);
+  const endedAtMs = new Date(endedAt).getTime();
+  if (!Number.isFinite(startedAtMs) || !Number.isFinite(endedAtMs)) {
+    return accumulatedRuntimeMs;
+  }
+
+  return accumulatedRuntimeMs + Math.max(0, endedAtMs - startedAtMs);
 }
--- a/packages/core/src/observability/snapshot-builder.ts
+++ b/packages/core/src/observability/snapshot-builder.ts
@@ -312,31 +312,64 @@ function aggregateTokenUsage(
     }
   }
 
-  const runtimeMs = Array.from(latestSessionsByRunLifecycle(runs).values())
-    .map((run) => runtimeMsForSnapshot(run, lastTickAt))
-    .reduce((total, sessionRuntimeMs) => total + sessionRuntimeMs, 0);
+  const runtimeMs = Array.from(runsByLifecycle(runs).values()).reduce(
+    (total, lifecycleRuns) =>
+      total + runtimeMsForLifecycle(lifecycleRuns, lastTickAt),
+    0
+  );
   const secondsRunning = Math.max(0, Math.round(runtimeMs / 1000));
 
   return { inputTokens, outputTokens, totalTokens, secondsRunning };
 }
 
-function latestSessionsByRunLifecycle(
+function runsByLifecycle(
   runs: OrchestratorRunRecord[]
-): Map<string, OrchestratorRunRecord> {
-  const latestRuns = new Map<string, OrchestratorRunRecord>();
+): Map<string, OrchestratorRunRecord[]> {
+  const groupedRuns = new Map<string, OrchestratorRunRecord[]>();
 
   for (const run of runs) {
-    const key = `${run.projectId}:${run.issueId}:${run.createdAt}`;
-    const current = latestRuns.get(key);
-    if (
-      !current ||
-      new Date(run.updatedAt).getTime() >= new Date(current.updatedAt).getTime()
-    ) {
-      latestRuns.set(key, run);
-    }
+    const lifecycleId = run.runtimeLifecycleId ?? run.createdAt;
+    const key = `${run.projectId}:${run.issueId}:${lifecycleId}`;
+    groupedRuns.set(key, [...(groupedRuns.get(key) ?? []), run]);
   }
 
-  return latestRuns;
+  return groupedRuns;
+}
+
+function runtimeMsForLifecycle(
+  runs: OrchestratorRunRecord[],
+  lastTickAt: string
+): number {
+  // Records written before cumulativeRuntimeMs existed share the legacy
+  // createdAt lifecycle identity. Sum their individual sessions so upgrading
+  // does not discard already-finished runtime.
+  if (!runs.some((run) => run.cumulativeRuntimeMs !== undefined)) {
+    return runs.reduce(
+      (total, run) => total + runtimeMsForSnapshot(run, lastTickAt),
+      0
+    );
+  }
+
+  return runtimeMsForSnapshot(latestSession(runs), lastTickAt);
+}
+
+function latestSession(runs: OrchestratorRunRecord[]): OrchestratorRunRecord {
+  return runs.reduce((latest, candidate) => {
+    const latestUpdatedAt = new Date(latest.updatedAt).getTime();
+    const candidateUpdatedAt = new Date(candidate.updatedAt).getTime();
+    if (candidateUpdatedAt !== latestUpdatedAt) {
+      return candidateUpdatedAt > latestUpdatedAt ? candidate : latest;
+    }
+    if ((candidate.status === "running") !== (latest.status === "running")) {
+      return candidate.status === "running" ? candidate : latest;
+    }
+    const latestStartedAt = latest.startedAt ?? "";
+    const candidateStartedAt = candidate.startedAt ?? "";
+    if (candidateStartedAt !== latestStartedAt) {
+      return candidateStartedAt > latestStartedAt ? candidate : latest;
+    }
+    return candidate.runId > latest.runId ? candidate : latest;
+  });
 }
 
 function runtimeMsForSnapshot(
@@ -350,7 +383,11 @@ function runtimeMsForSnapshot(
 
   const startedAtMs = new Date(run.startedAt).getTime();
   const endedAt =
-    run.completedAt ?? (run.status === "running" ? lastTickAt : run.updatedAt);
+    run.completedAt ??
+    (run.runtimeSession?.status !== "active"
+      ? run.runtimeSession?.updatedAt
+      : null) ??
+    (run.status === "running" ? lastTickAt : run.updatedAt);
   const endedAtMs = new Date(endedAt).getTime();
   if (!Number.isFinite(startedAtMs) || !Number.isFinite(endedAtMs)) {
     return accumulatedRuntimeMs;
--- a/packages/orchestrator/src/service.ts
+++ b/packages/orchestrator/src/service.ts
@@ -1611,7 +1611,7 @@ export class OrchestratorService {
                 : undefined,
             runtimeLifecycleId:
               recoveryContext || existingIssueRecord?.retryEntry
-                ? (previousRun?.runtimeLifecycleId ?? previousRun?.runId)
+                ? (previousRun?.runtimeLifecycleId ?? previousRun?.createdAt)
                 : undefined,
             recovery: recoveryContext,
             onPrepared: async (candidate) => {
@@ -4758,7 +4758,7 @@ export class OrchestratorService {
       restarted = await this.startRun(tenant, issue, {
         attempt: run.attempt,
         cumulativeRuntimeMs: resolveCumulativeRuntimeMs(run),
-        runtimeLifecycleId: run.runtimeLifecycleId ?? run.runId,
+        runtimeLifecycleId: run.runtimeLifecycleId ?? run.createdAt,
         recovery,
         onPrepared: async (candidate) => {
           preparedRun = candidate;
--- a/packages/orchestrator/src/service.ts
+++ b/packages/orchestrator/src/service.ts
@@ -1556,6 +1556,7 @@ export class OrchestratorService {
           issue,
           latestRunsByIssueId.get(issue.id) ?? null
         );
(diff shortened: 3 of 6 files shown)


packages/core/src/observability/snapshot-builder.ts
```diff
@@ -299,33 +299,57 @@ function aggregateTokenUsage(
...
   return { inputTokens, outputTokens, totalTokens, secondsRunning };
 }
+
+function latestSessionsByRunLifecycle(
+  runs: OrchestratorRunRecord[]
+): Map<string, OrchestratorRunRecord> {
+  const latestRuns = new Map<string, OrchestratorRunRecord>();
+
+  for (const run of runs) {
+    const key = `${run.projectId}:${run.issueId}:${run.createdAt}`;
```
hojinzs: **P1 — `createdAt` is not a stable lifecycle key on the retry path, so completed sessions get counted twice.**

The PR description states "the new lifecycle grouping uses the existing stable `createdAt` carried across restarts". That holds for `restartRun()` only, which explicitly re-stamps `createdAt: run.createdAt` (`packages/orchestrator/src/service.ts:4774`). The *other* session-boundary path — the dispatch loop at `service.ts:1603-1609`, which this PR just wired `cumulativeRuntimeMs` into — calls `startRun()`, and `startRun()` stamps a **fresh** `createdAt: now.toISOString()` (`service.ts:2967`). Nothing rewrites it afterwards (`createdAt` is only assigned at `service.ts:2881`, `2967`, `4774`).

So on that path the new run inherits `cumulativeRuntimeMs` from the previous run *and* lands in a different lifecycle bucket, while the previous run — still present in `allRuns`, since `previousRun` comes from `buildLatestRunMapByIssueId(await this.store.loadAllRuns())` — forms its own bucket and contributes its own runtime again.

Reproduced against this commit with records shaped exactly as `startRun()` writes them:

```ts
const first = mockRun({
  runId: "run-001", status: "failed",
  createdAt: "2024-01-01T00:00:00Z", startedAt: "2024-01-01T00:00:00Z",
  completedAt: "2024-01-01T00:02:00Z", updatedAt: "2024-01-01T00:02:00Z",
});
const retry = mockRun({
  runId: "run-002", status: "running", attempt: 2,
  createdAt: "2024-01-01T00:07:00Z",   // fresh stamp from startRun()
  startedAt: "2024-01-01T00:07:00Z", completedAt: null,
  updatedAt: "2024-01-01T00:09:00Z",
  cumulativeRuntimeMs: 120_000,        // carried from run-001
});
buildProjectSnapshot({ ...,  allRuns: [first, retry], lastTickAt: "2024-01-01T00:09:00Z" });
// expected 240 (120 + 120), actual 360 — run-001's 120s counted twice
```

The `recoveryContext` branch of that same ternary is the costlier case: `previousRun` there is a full prior worker session, so the inflation is the whole session, not a few spawn-failure seconds.

Suggested fix: stop overloading `createdAt` and carry an explicit lifecycle identifier on the run record (e.g. `runLifecycleId`, seeded from the previous run's lifecycle id or the run's own `runId` when absent), then key on that. It also removes the need for `restartRun()` to keep back-dating `createdAt`, which currently makes the field mean two different things. Passing `previousRun.createdAt` through `startRun()` would work too, but keeps the overloading.

---

packages/core/src/observability/snapshot-builder.test.ts
```diff
@@ -815,6 +816,48 @@ describe("buildProjectSnapshot", () => {
     expect(snapshot.codexTotals!.secondsRunning).toBe(300);
   });
 
+  it("accumulates completed retry sessions without counting retry gaps", () => {
+    const initialStartedAt = "2024-01-01T00:00:00Z";
+    const completedSession = mockRun({
+      runId: "run-001",
+      status: "failed",
+      createdAt: initialStartedAt,
```
hojinzs, line 837: **P1 (same defect, from the test side) — this fixture only models the `restartRun()` path, which is why the bug passes CI.**

All three sessions share `createdAt: initialStartedAt`. That shape is reachable only via `restartRun()` (`service.ts:4774`); the dispatch-loop retry/recovery path this PR touches produces sessions with *distinct* `createdAt` values. The issue's acceptance criterion is "다중 세션(재시도 포함) 픽스처로 합산 값을 정확히 단언하는 테스트" — as written the fixture asserts against a record shape the retry path doesn't emit, so the accumulation is never actually exercised for the retry it names.

Please add a case with distinct `createdAt` per session (the shape `startRun()` writes) alongside this one; it currently returns 360 instead of 240 for the two-session case, and both cases should pass once the lifecycle key is fixed.

Also worth guarding: today `mockRun()`'s default `createdAt` means any two mock runs for the same issue silently collapse into one lifecycle. That is what forced the `createdAt: "2024-01-01T00:06:00Z"` addition at line 241 in `aggregates token usage across multiple runs` — the fixture had to be changed to preserve the existing assertion, which is a signal the key is doing more work than it can carry.

---

packages/core/src/observability/snapshot-builder.ts
```diff
@@ -299,33 +299,57 @@ function aggregateTokenUsage(
...
+    const current = latestRuns.get(key);
+    if (
+      !current ||
+      new Date(run.updatedAt).getTime() >= new Date(current.updatedAt).getTime()
```
hojinzs: **P3 — tie on `updatedAt` makes the winner depend on `loadAllRuns()` ordering.**

`>=` means that when two records in a lifecycle share `updatedAt`, whichever comes later in the array wins. `restartRun()` writes `supersededRecord` with `updatedAt: now.toISOString()` (`service.ts:4712-4720`) and the recovered record with `startRun()`'s own `now` — same tick, so a same-millisecond tie is reachable. If the superseded record wins, the new session's elapsed time is dropped from the metric until the next tick bumps `updatedAt`.

Transient and self-healing, but cheap to make deterministic — e.g. break the tie toward the record that is still `running`, or fall back to `startedAt` then `runId`.

---

packages/core/src/observability/snapshot-builder.ts
```diff
@@ -299,33 +299,57 @@ function aggregateTokenUsage(
...
+  run: OrchestratorRunRecord,
+  lastTickAt: string
+): number {
+  const accumulatedRuntimeMs = run.cumulativeRuntimeMs ?? 0;
+  if (!run.startedAt) {
+    return accumulatedRuntimeMs;
+  }
+
+  const startedAtMs = new Date(run.startedAt).getTime();
+  const endedAt =
```
hojinzs, line 385: **P3 — `updatedAt` is a record-write timestamp, not a session end, so a non-`running` session's span keeps growing.**

For anything that is neither `running` nor has `completedAt` — a record parked in `retrying`, for instance — the end boundary becomes `run.updatedAt`, which is bumped by any later write to the record (tracker progress, retry bookkeeping), not by the session actually ending. The measured span then keeps stretching past the real session.

The record already carries a truer boundary: `runtimeSession.startedAt` / `runtimeSession.updatedAt` with `status: "completed" | "failed" | "active"` (`service.ts:3327-3339`, `buildRuntimeSession` at `service.ts:5946`). Preferring `runtimeSession` when present, falling back to the current logic, would match §13.5's "session runtime" more closely. Non-blocking for this PR — flagging so it's a deliberate choice rather than an accident.

---

packages/orchestrator/src/service.ts
```diff
@@ -1554,6 +1554,7 @@ export class OrchestratorService {
           issue,
           latestRunsByIssueId.get(issue.id) ?? null
         );
+        const previousRun = latestRunsByIssueId.get(issue.id) ?? null;
```
hojinzs: **nit** — `latestRunsByIssueId.get(issue.id) ?? null` is already evaluated two lines above for the `resolveIncompleteTurnRecoveryContext()` argument. Hoist the local above that call and pass it to both, rather than looking it up twice.

---

hojinzs (changes requested):

### Requirements vs. #750

| --- | --- |
| `seconds_running` = completed-session runtime + active-session elapsed | ⚠️ Implemented, but wrong on the retry path (P1) |
| Minimal run-record field so accumulation survives session boundaries | ✅ `cumulativeRuntimeMs` |
| Multi-session (incl. retry) fixture asserting the exact sum | ⚠️ Present, but models only one of the two retry paths (P1) |
| `changeset` (patch) | ✅ `.changeset/750-accumulated-session-runtime.md`, `@gh-symphony/cli` patch |
| Scope: core/observability (+ run record field) | ✅ The `service.ts` carry-over is necessary to satisfy criterion 2, not scope creep |

No over-engineering or unrequested decisions found — the change stays tight to the metric, and the `>0` assertion the issue called out is now an exact `540`.

### Blocking

**P1 — a completed session is counted twice when a retry is dispatched from the tick loop.** `latestSessionsByRunLifecycle()` keys on `createdAt`, but only `restartRun()` preserves it across sessions; the dispatch-loop retry/recovery path this PR just wired `cumulativeRuntimeMs` into gets a fresh `createdAt` from `startRun()`. The prior session then contributes runtime both from its own bucket and from the new run's inherited `cumulativeRuntimeMs`. Reproduced on `829ce6c`: **360s reported where 240s is correct.** Details and a suggested fix are in the inline comment on `snapshot-builder.ts`; the companion comment on the test file explains why CI is green.

The recovery branch is the expensive case — there the carried-over value is a whole worker session, so the metric inflates by minutes, not seconds. This makes the PR's stated goal ("excluding retry gaps") not hold on the path that produces most retries.

### Non-blocking

- **P3** — `>=` tie-break on `updatedAt` in the lifecycle dedup is order-dependent when a superseded and a recovered record share a tick.
- **P3** — `updatedAt` as the session-end fallback keeps stretching for a record parked in `retrying`; `runtimeSession` already carries truer boundaries.
- **nit** — duplicated `latestRunsByIssueId.get(issue.id)` lookup in the dispatch loop.

### Verification performed

- `pnpm --filter @gh-symphony/orchestrator test` — 361 passed / 14 files
- `vitest run packages/core/src/observability/snapshot-builder.test.ts` — 29 passed
- Targeted repro test against this commit (above) — **fails: 360 ≠ 240**
- Docker is unavailable in this environment, so the e2e suite was not re-run locally; CI `Test` and `Container Smoke` are both green on `829ce6c`. Neither covers this defect — a retry-path regression test is the gap to close.

Happy to re-review as soon as the lifecycle key is addressed; the rest of the change is sound and the direction is right.

---

hojinzs:
## Follow-up: Codex's three P2 findings verified — all three are real

I submitted my review before reading Codex's inline threads. Having now run all three against `829ce6c`, they are confirmed, and two of them are more serious than P2. Reproductions below so they don't get triaged as bot noise.

### 1. `service.ts:1608` — "Keep dispatch retries in the same runtime lifecycle" → **confirmed, same defect as my P1**

Independently reached the same conclusion. See [my inline comment](https://github.com/hojinzs/github-symphony/pull/763#discussion_r3886322434) for the repro (360s where 240s is correct) and a suggested fix.

### 2. `service.ts:6078` — "Freeze session end time before requeueing" → **confirmed, and this is the goal-defeating one**

`requeueRetryingRun()` writes `updatedAt: now.toISOString()` with `completedAt: null` and `status: "retrying"` (`service.ts:5047-5053`). Every capacity-wait postponement bumps `updatedAt`, and both `runtimeMsForSnapshot()` and `resolveCumulativeRuntimeMs()` fall back to `updatedAt` as the session end. The capacity wait is therefore billed as worker runtime:

```ts
const r = mockRun({
  status: "retrying",
  startedAt: "2024-01-01T00:00:00Z",  // 2-minute session…
  completedAt: null,
  updatedAt: "2024-01-01T00:07:00Z",  // …requeued 5 min later for lack of a slot
});
buildProjectSnapshot({ ..., allRuns: [r], lastTickAt: "2024-01-01T00:07:00Z" });
// expected 120, actual 420
```

**420s reported for a 120s session.** This is precisely the "retry gap" the PR sets out to exclude, so the change does not achieve its stated goal on the slot-starved path. It also supersedes my own [P3 on the `updatedAt` fallback](https://github.com/hojinzs/github-symphony/pull/763#discussion_r3886323491) — I flagged that as theoretical; Codex found the concrete trigger, so please treat it as blocking rather than nice-to-have. Codex's suggested remedy (persist the session duration when the run enters `retrying`, instead of deriving it from a mutable bookkeeping timestamp) is the right shape and would also make my P3 moot.

### 3. `snapshot-builder.ts:324` — "Preserve legacy session durations during grouping" → **confirmed, silent regression on upgrade**

Pre-upgrade records written by `restartRun()` already share `createdAt` but carry no `cumulativeRuntimeMs`. The new grouping keeps only the newest record in each lifecycle, whose accumulator defaults to `0`, so every earlier completed session is dropped:

```ts
// two pre-upgrade 2-minute sessions, same createdAt, no accumulator
buildProjectSnapshot({ ..., allRuns: [sessionA, sessionB] });
// expected 240, actual 120
```

**120s reported where 240s is correct.** Existing deployments will see `seconds_running` drop for in-flight lifecycles the moment they upgrade. A legacy fallback — summing the retained records in a lifecycle when none of them carries `cumulativeRuntimeMs` — would cover it.

### Net

Three independent paths now produce a wrong `seconds_running`: dispatch-loop retry (over), capacity-wait requeue (over), and legacy records (under). They share a root cause worth fixing once rather than patching three times: the metric is derived from mutable run-record bookkeeping fields (`createdAt` as an identity, `updatedAt` as a session end) rather than from persisted, immutable session facts. Persisting each session's duration at the moment it ends, and keying lifecycles on an explicit id, resolves all three plus both of my P3s.

Verified locally on `829ce6c`: `pnpm --filter @gh-symphony/orchestrator test` (361 pass) and `snapshot-builder.test.ts` (29 pass) — the existing suite is green throughout, so none of these are covered today.

Round 1 of 6.

---

hojinzs:
**리뷰 보완 (독립 검증, 1회차 Request changes와 결론 일치)** — P1 이중 계산 재현: 틱 경로 재디스패치(`service.ts:1603-1608`, `:2965`의 recoveryContext/retryEntry → `startRun()`)가 `cumulativeRuntimeMs`는 넘기지만 `createdAt: now`로 새 lifecycle 키를 만들어 `latestSessionsByRunLifecycle()`가 이전 run과 새 run을 별개로 합산 — `service.test.ts:2299`에 `expect(redispatched.codexTotals?.secondsRunning).toBe(300)` 추가 시 실제 600. `restartRun()`처럼 `createdAt` 승계 필요. 추가(Codex P2 확인): `requeueRetryingRun()`(`:5047-5054`)이 `updatedAt: now, completedAt: null`로 저장해 슬롯 부족 대기 시간이 runtime에 포함·영구화. 정상 확인(블랙박스): restartRun 경로 누적(10.3+5.1=15s), retry_queued 중 정지, 재기동 후 유지.

packages/orchestrator/src/service.ts
```diff
@@ -6056,6 +6064,25 @@ function resolvePersistedCumulativeTurnCount(
...
+    return 0;
+  }
+
+  const accumulatedRuntimeMs = run.cumulativeRuntimeMs ?? 0;
+  if (!run.startedAt) {
+    return accumulatedRuntimeMs;
+  }
+
+  const startedAtMs = parseTimestampMs(run.startedAt);
+  const endedAtMs = parseTimestampMs(run.completedAt ?? run.updatedAt);
```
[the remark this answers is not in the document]
moncher-dev: Addressed in 7ef3962: retry records freeze cumulative session runtime and clear the active start timestamp before capacity requeueing, so later bookkeeping updates cannot add idle time.

packages/core/src/observability/snapshot-builder.ts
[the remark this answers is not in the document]
moncher-dev: Addressed in 7ef3962: lifecycle aggregation now sums legacy records that lack cumulativeRuntimeMs, preserving all completed sessions during an upgrade.

packages/orchestrator/src/service.ts
```diff
@@ -1602,6 +1603,10 @@ export class OrchestratorService {
         try {
           run = await this.startRun(tenant, issue, {
             attempt: existingIssueRecord?.retryEntry?.attempt ?? null,
+            cumulativeRuntimeMs:
+              recoveryContext || existingIssueRecord?.retryEntry
+                ? resolveCumulativeRuntimeMs(previousRun)
```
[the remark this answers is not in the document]
moncher-dev: Addressed in 7ef3962: dispatch-loop retries now retain runtimeLifecycleId from the prior run, so inherited cumulative runtime and the prior session share one aggregation bucket.

packages/core/src/observability/snapshot-builder.ts
moncher-dev: Addressed in 7ef3962: run records now use runtimeLifecycleId rather than createdAt for aggregation; the new exact fixture covers a tick-dispatched retry with distinct createdAt values and asserts 240 seconds.

packages/core/src/observability/snapshot-builder.test.ts
moncher-dev: Addressed in 7ef3962: added the distinct-createdAt dispatch-retry fixture requested here, with the runtimeLifecycleId carried from the completed session and an exact 240-second assertion.

packages/core/src/observability/snapshot-builder.ts
moncher-dev: Addressed in 7ef3962: latest-session selection is now deterministic: updatedAt, then running status, startedAt, and runId.

packages/core/src/observability/snapshot-builder.ts
moncher-dev: Addressed in 7ef3962: retrying records persist completedAt and cumulativeRuntimeMs at session close; snapshot calculation also prefers a completed or failed runtimeSession boundary before mutable record updatedAt.

packages/orchestrator/src/service.ts
moncher-dev: Addressed in 7ef3962: previousRun is now hoisted and reused for both recovery-context lookup and runtime carry-over.

packages/orchestrator/src/service.ts
```diff
@@ -1602,6 +1603,14 @@ export class OrchestratorService {
...
+                : undefined,
+            runtimeLifecycleId:
+              recoveryContext || existingIssueRecord?.retryEntry
+                ? (previousRun?.runtimeLifecycleId ?? previousRun?.runId)
```
hojinzs: **P1 (residual) — the legacy fallback still double counts a lifecycle that straddles the upgrade.**

The pure-legacy case is fixed, but the mixed case isn't, and that's the one that actually occurs when a deployment upgrades with runs in flight.

The two sides fall back to *different* keys. A pre-upgrade record has no `runtimeLifecycleId`, so `runsByLifecycle()` keys it on `createdAt`. When the tick loop then dispatches its retry, this line seeds the lifecycle from `previousRun?.runId` — so the new record keys on `run-001` while the legacy record keys on its `createdAt`. Different buckets. The legacy bucket has no `cumulativeRuntimeMs` anywhere, so `runtimeMsForLifecycle()` sums it; the new bucket carries that same session inside its accumulator. Counted twice.

Verified against `7ef3962`:

```ts
const legacy = mockRun({                       // persisted pre-upgrade
  runId: "run-001", status: "failed",
  createdAt: "2024-01-01T00:00:00Z", startedAt: "2024-01-01T00:00:00Z",
  completedAt: "2024-01-01T00:02:00Z", updatedAt: "2024-01-01T00:02:00Z",
  // no runtimeLifecycleId, no cumulativeRuntimeMs
});
const afterUpgrade = mockRun({                 // retry dispatched post-upgrade
  runId: "run-002", status: "running", attempt: 2,
  runtimeLifecycleId: "run-001",               // ← seeded from previousRun.runId
  createdAt: "2024-01-01T00:07:00Z", startedAt: "2024-01-01T00:07:00Z",
  completedAt: null, updatedAt: "2024-01-01T00:09:00Z",
  cumulativeRuntimeMs: 120_000,
});
// expected 240 (120 legacy + 120 active), actual 360
```

**Suggested fix — one line here and the same on `service.ts:4747`:** fall back to the previous run's `createdAt` instead of its `runId`, so the seeded id matches the key `runsByLifecycle()` already assigns to legacy records:

```diff
-                ? (previousRun?.runtimeLifecycleId ?? previousRun?.runId)
+                ? (previousRun?.runtimeLifecycleId ?? previousRun?.createdAt)
```

Both records then land in one bucket, that bucket contains a record with `cumulativeRuntimeMs`, so the latest-session path applies and yields 240. I applied both edits locally and re-ran the four scenarios — the upgrade-boundary case passes and the three cases your commit already fixed stay green. Note it requires updating the `runtimeLifecycleId: "run-incomplete"` expectation at `service.test.ts:2467`, since that run's seed becomes its `createdAt`.

Runs created fresh are unaffected either way: `startRun()` still defaults `runtimeLifecycleId` to `runId`, so only the legacy fallback changes.

---

packages/orchestrator/src/service.ts
```diff
@@ -3661,7 +3675,13 @@ export class OrchestratorService {
       attempt:
         persistedRetryKind === "continuation" ? 1 : runWithTokens.attempt + 1,
       processId: null,
-      updatedAt: now.toISOString(),
+      updatedAt: retryCompletedAt,
+      startedAt: null,
```
hojinzs, line 3681: **P1 — this breaks an existing test; CI `Test` is red on `7ef3962`.**

`service.test.ts:3432` ("queues a clean convergence failure when the tracker is already in the retryable state") asserts `completedAt: null` on the saved retry record. Setting `completedAt: retryCompletedAt` here makes it fail:

```
AssertionError: expected '2026-03-08T00:01:00.000Z' to be null
FAIL  src/service.test.ts > OrchestratorService > queues a clean convergence failure when the tracker is already in the retryable state
```

Reproduced locally (`360 passed | 1 failed`) and it matches the failed `Test` check on this commit. You updated the sibling expectation at `service.test.ts:795` for the same semantic change but this one was missed.

**Worth deciding deliberately rather than just updating the assertion**, because `completedAt` on a `retrying` record is read outside observability:

- `explain.ts:228` — `latestRun.completedAt ?? latestRun.updatedAt` is the convergence-lock TTL origin. It used to track the advancing `updatedAt`; it now freezes at the first retry transition, so the lock expires on schedule instead of being pushed out by each requeue.
- `service.ts:5624-5630` — the same expression drives suppression re-dispatch timing.

Both changes look *more* correct to me, not less — a convergence happened once, so pinning the TTL to it is the better behavior. But it is a Coordination-layer change shipped inside an Observability fix, and neither the PR description nor the changeset mentions it. Per the layer discipline in `AGENTS.md`/`CLAUDE.md` this should be called out explicitly, and the changeset text should say retry records now carry a terminal `completedAt`.

---

packages/orchestrator/src/service.ts
```diff
@@ -5036,13 +5057,19 @@ export class OrchestratorService {
...
       : error;
+    const sessionEndedAt = run.completedAt ?? now.toISOString();
     await this.store.saveRun({
       ...run,
       status: suppressed ? "suppressed" : "retrying",
       attempt,
       processId: null,
       updatedAt: now.toISOString(),
-      completedAt: suppressed ? now.toISOString() : null,
+      startedAt: null,
```
hojinzs, line 5081: **P3 — nulling `startedAt` is load-bearing for the metric but visible to operator surfaces.**

The freeze itself is correct, and I verified it: a 2-minute session requeued after 5 minutes of capacity waiting now reports 120s instead of 420s. Re-requeues are idempotent too — `sessionEndedAt = run.completedAt ?? now` preserves the first freeze, and with `startedAt` null `resolveCumulativeRuntimeMs()` returns the accumulator unchanged.

The cost is that `startedAt: null` now doubles as "this record has no active session", and two display paths read the field directly:

- `packages/cli/src/dashboard/renderer.ts:289` — `fmtAge(run.startedAt, now)` returns `"0m"` for null, so a retry-queued run shows AGE `0m` regardless of how long it has actually been around.
- `packages/cli/src/dashboard/renderer.ts:218-223` — retrying runs drop out of the `allStarts` minimum; if every active run is retrying, the header runtime reads 0.

`recover.ts:151` is safe (guarded by `if (run.processId)`, and retry records set `processId: null`).

Cosmetic, and I'd merge with it. But a dedicated field (say `activeSessionStartedAt`, or leaving `startedAt` alone and having `runtimeMsForSnapshot()` skip the span when the status is terminal) would keep the observability flag out of a field the CLI renders. Your call — flagging so it's deliberate.

---

hojinzs (changes requested):
## Review — round 2/6 · Request changes (two items, both small)

Good commit. `runtimeLifecycleId` is the right call — it removes the `createdAt` overloading rather than papering over it, and `restartRun()` no longer back-dating `createdAt` is a real cleanup. I re-ran every scenario from round 1 against `7ef3962` rather than taking the thread replies at face value.

### Verified fixed

| Round-1 finding | Result on `7ef3962` |
| --- | --- |
| P1 dispatch-loop retry double count | ✅ 240s (was 360s) |
| Codex P2 capacity-wait requeue inflation | ✅ 120s (was 420s) |
| Codex P2 legacy-only lifecycle undercount | ✅ 240s (was 120s) |
| P3 non-deterministic `updatedAt` tie-break | ✅ deterministic: `updatedAt` → running → `startedAt` → `runId` |
| P3 `updatedAt` as session end | ✅ now prefers `runtimeSession` when not active |
| nit duplicated `previousRun` lookup | ✅ hoisted |

The three new fixtures assert exact values and cover the paths they name. `pnpm --filter @gh-symphony/core test` — 296 pass.

### Still blocking — 2 items

**1. CI is red.** `service.test.ts:3432` still asserts `completedAt: null` on a retry record; the `retryRecord` change sets it. `360 passed | 1 failed` locally, matching the failed `Test` check on this commit. The sibling expectation at `:795` was updated, this one was missed. Inline comment also flags that `completedAt` on a `retrying` record feeds the convergence-lock TTL (`explain.ts:228`) and suppression timing (`service.ts:5624`) — I think the new behavior is better, but it deserves a line in the changeset rather than arriving silently.

**2. The legacy fallback doesn't cover a lifecycle that straddles the upgrade** — the case that actually happens when you deploy with runs in flight. A pre-upgrade record keys on `createdAt`; its post-upgrade retry keys on `runtimeLifecycleId = previousRun.runId`. Different buckets, so the legacy session is summed in its own bucket *and* carried inside the new accumulator: **360s where 240s is correct.** One-line fix in the inline comment (seed from `previousRun?.createdAt`, same on `service.ts:4747`) — I applied it locally and confirmed it fixes this case while the three you already fixed stay green.

### Non-blocking

- **P3** — `startedAt: null` doubles as an observability flag but is rendered by the CLI dashboard (`renderer.ts:289`, `:218-223`): retry-queued runs show AGE `0m`. Cosmetic; details inline.

### Scope check

Still tight to #750, no over-engineering. `runtimeLifecycleId` is new surface area on the run record, but it's the minimum needed for the acceptance criterion and replaces a fragile implicit key.

---

To be explicit about where this lands, since we're 2 of 6 rounds in: **items 1 and 2 are the only things I need.** Item 1 is an assertion update plus a changeset line; item 2 is one character short of a one-liner. Push those and I'll approve — I won't open new lines of inquiry on the next pass unless something in the diff is outright broken.

---

hojinzs:
**리뷰 보완 (독립 검증, 2회차 Request changes와 결론 일치)** — (1) CI Test 실패 재현: `service.test.ts` "queues a clean convergence failure when the tracker is already in the retryable state" — retry 레코드 `completedAt`이 null이어야 하는데 값이 남음(`service.ts:3679`; `:5624-5630` suppression 타이밍이 소비). 의미 변경이면 changeset·테스트를 함께 갱신. (2) 업그레이드 straddle 이중 계산: `runtimeLifecycleId`가 없는 기존 레코드(`run-incomplete` 픽스처)는 여전히 `service.test.ts:2462` 프로브 600(픽스처에 id 추가 시 300), 스냅샷 프로브 360 vs 240 → tick 경로(`service.ts:1612`)·`restartRun`(`:4747`)에서 `previousRun.createdAt` 기준으로 시드하거나 레거시 레코드에 id 스탬프. 확인된 것(블랙박스): 세션 누적 11→16→26, retry_queued 중 고정, 재기동 후 유지, 슬롯 부족 requeue 시 `completedAt` 동결·`cumulativeRuntimeMs` 불변(108s 대기). 비차단: `runtimeMsForSnapshot`이 `status === null`을 비활성으로 취급.

packages/orchestrator/src/service.ts
moncher-dev: Resolved in 9d87b63. Retry and restart paths now fall back to the previous record’s `createdAt` when it lacks `runtimeLifecycleId`, so a lifecycle crossing the upgrade stays in the legacy bucket. Added an exact mixed-version fixture: 120 seconds completed plus 120 seconds active reports 240 seconds, without double counting.

packages/orchestrator/src/service.ts
moncher-dev: Resolved in 9d87b63. The retry test now expects the frozen `completedAt`, and the patch changeset documents that this terminal session boundary prevents retry scheduling from extending `seconds_running`. This deliberately pins the related convergence/suppression timing to the session completion rather than later retry bookkeeping.

packages/orchestrator/src/service.ts
moncher-dev: Acknowledged. I kept the dashboard rendering unchanged because this issue is restricted to core observability plus the minimal run-record state; the current workpad records the age-display observation for post-merge human validation. The metric itself no longer uses a retrying record’s elapsed wall time.

moncher-dev:
Re-review requested for 9d87b63. The two blocking items are resolved: mixed legacy/updated retry lifecycles now aggregate exactly once, and the frozen retry-session boundary has matching test coverage and release notes. Full local validation and the Docker lifecycle E2E pass.

hojinzs (approved):
## Review — round 3/6 · ✅ Approve

Both items from round 2 landed exactly as scoped, and I re-verified everything from scratch on `9d87b63` rather than reading the replies.

### Final verification

| Check | Result |
| --- | --- |
| `pnpm --filter @gh-symphony/orchestrator test` | 361 passed / 14 files |
| `pnpm --filter @gh-symphony/core test` | 297 passed / 21 files |
| `pnpm lint` | clean |
| `pnpm typecheck` | clean |
| CI `Test` / `Container Smoke` on `9d87b63` | both green |

Runtime scenarios, all exact-value assertions against this commit:

| Scenario | Reported | Correct |
| --- | --- | --- |
| Dispatch-loop retry, fresh `createdAt` | 240s | 240s ✅ |
| Capacity-wait requeue (5 min idle) | 120s | 120s ✅ |
| Legacy-only lifecycle, two sessions | 240s | 240s ✅ |
| Upgrade-straddling lifecycle | 240s | 240s ✅ |
| Three-session chain across the upgrade | 360s | 360s ✅ |

The last one is mine, not in the suite — a legacy session plus two post-upgrade sessions in one lifecycle — added to check the `createdAt` seed holds past the first hop. It does.

### Round-2 items

**1. CI red** — fixed. The convergence assertion at `service.test.ts:3466` now expects the frozen `completedAt`, and the changeset carries the behavior note ("Retry records now freeze their completed-session boundary so retry scheduling does not extend that measurement"). That's what I asked for: the `explain.ts:228` TTL and `service.ts:5624` suppression timing now shift deliberately rather than silently.

**2. Upgrade-boundary double count** — fixed on both paths (`service.ts:1612`, `:4747`), with the `run-incomplete` expectation updated and an exact mixed-version fixture added. 360s → 240s.

### Requirements vs #750

All acceptance criteria met: `seconds_running` is completed-session runtime plus active elapsed; `cumulativeRuntimeMs` + `runtimeLifecycleId` are the minimum record fields needed to survive session boundaries; multi-session retry fixtures assert exact sums; patch changeset present. Scope stayed inside core/observability plus the orchestrator carry-over that criterion 2 requires. No over-engineering — the two added fields both earn their place, and `runtimeLifecycleId` removed the `createdAt` overloading rather than adding a layer on top of it.

### Two notes, neither blocking

- Three replies posted as unexpanded paths — `@.runtime/agent-tmp/750-reply-boundary.md` and `@.runtime/agent-tmp/750-reply-dashboard.md` are still literal text on their threads. Whatever tooling inlines those files didn't run. The boundary one is moot (the code answered it), but I never saw your reasoning on the `startedAt: null` P3, so treat that one as still open on your side if you had something to say. Worth checking why the substitution failed before it swallows a reply that matters.
- The `startedAt: null` display effect (retry-queued runs show AGE `0m` in the CLI dashboard) stands as accepted. Fine by me — flagging so it's on the record rather than rediscovered later.

`pnpm format` reports 86 files repo-wide including `prettier.config.mjs`; the only deviation in `service.ts` is at line ~2582, well outside every hunk this PR touches. Pre-existing on `main`, not yours.

Good iteration — three rounds, eight findings, all closed with verified numbers rather than assertions. 👍

---

hojinzs:
Correction to one note in my approval — the approval itself stands unchanged.

I wrote that three replies were "still literal text on their threads" and that I never saw your reasoning on the `startedAt: null` P3. That was wrong: you edited all three in place at 11:14:07–11:14:09, and I had fetched the thread list about a minute earlier, so I reviewed a stale snapshot and then posted at 11:18 as though nothing had changed. Your reply tooling worked. Please disregard the suggestion to go looking for a broken file substitution.

Your reasoning on the P3 is on record and I'm satisfied with it: keeping the dashboard rendering unchanged because #750 is scoped to core observability plus minimal run-record state, with the age-display observation carried into post-merge human validation. Declining a non-blocking suggestion with a scope rationale is the right call, and deferring the display question to the validation checklist is better than widening this PR to chase it.

That leaves the two notes in my approval as: the `startedAt: null` age-display effect is accepted and tracked for post-merge validation, and the pre-existing repo-wide `pnpm format` state is not this PR's. Nothing outstanding — ✅ approved on `9d87b63`, both checks green.

---

hojinzs:
**스모크테스트 통과 → Land로 이동 (프로젝트 상태 관리)** — head `9d87b63`(approve, CI green)를 독립 clone에 빌드하고 로컬 E2E(file 트래커 + stub `slow`, `max_concurrent_agents: 2`)에 이슈 2건을 **동시에** 실행한 뒤 Done으로 전이했습니다. run 레코드 `cumulativeRuntimeMs` ≈ 36.5s × 2 = **합 72s**, 두 run이 겹쳐 wall-clock span은 ≈36.5s. `/api/v1/state` `codexTotals.secondsRunning = 73` → **세션 runtime 합산**(§13.5) 의미론 확인, 활성 run 0일 때 값 고정. 오류 없음.

PR #764, opened 2026-08-29
feat(cli,docs): generate provider-form workflows

moncher-dev:
## TL;DR

- New `WORKFLOW.md` files use `tracker.provider` for adapter-owned GitHub Project and Linear settings.
- Core lifecycle settings remain at `tracker.*`; flat provider aliases stay supported until the next major release (#679).

## Change-point diagram

`workflow init` / reference workflow / skill schema
→ core lifecycle + provider-owned adapter blocks
→ parser normalization and doctor migration diagnostics
→ validated GitHub and Linear workflow artifacts

## Start here

- `packages/cli/src/workflow/generate-workflow-md.ts` emits core lifecycle fields plus provider-form adapter configuration.
- `packages/cli/src/workflow/generate-reference-workflow.ts` and the generated `/gh-symphony` schema teach the same layout.
- `docs/configuration.md` records the actual reload boundary between persisted runtime binding and per-tick tracker policy.

## Changes

- Generate GitHub Project and Linear adapter settings beneath `tracker.provider`, while keeping core `active_states` and `terminal_states` at `tracker` level.
- Preserve flat provider keys as parser-supported deprecated aliases, with normalized doctor migration output for GitHub and Linear.
- Repair the generated `/gh-symphony` skill schema so its provider-form front matter parses; regression coverage parses the emitted example.
- Document that lifecycle policy, `state_field`, blocker checks, and planning states apply on the next reconciliation tick; provider connection, project binding, priority, and pickup-label settings require re-initialization and restart.
- Restore and test the labels-priority reference template; convert the worked Linear example and its prose to provider form.

## Evidence

- `pnpm exec prettier --check docs/configuration.md docs/examples/linear-WORKFLOW.md` — passed.
- `pnpm lint` — passed.
- `pnpm test` — passed across all workspace suites.
- `pnpm typecheck` — passed.
- `pnpm build` — passed.
- `LINEAR_API_KEY=validation-token node packages/cli/dist/index.js workflow validate --file docs/examples/linear-WORKFLOW.md` — passed.

## Risks & rollback

- Existing flat provider keys remain non-breaking aliases; new artifacts use the normalized layout.
- Provider connection and initialized runtime binding still require `repo init` and a daemon restart; lifecycle policy is next-tick reloaded.
- Roll back by reverting this PR. No watcher-driven reload implementation is introduced.

## Changed files

- Workflow, reference-workflow, and skill-schema generators plus CLI regression tests.
- Provider deprecation diagnostics and the worked Linear workflow example.
[...]

PR #768, opened 2026-08-29
feat(doctor): warn on ignored state concurrency limits

### LINKED ISSUE #767: feat(doctor): warn about ignored per-state concurrency limits
## Summary

Add diagnostics for invalid `agent.max_concurrent_agents_by_state` entries that are intentionally ignored.

## Context

Issue #659 normalizes state keys and ignores non-positive or non-integer per-state limits. The parser currently has no warning channel, so reporting ignored entries through `workflow validate` / `repo doctor` needs a separate observability design.

## Scope

- Surface actionable warnings with entry paths and reasons.
- Preserve valid entries in the same map.
- Add parser/doctor coverage.

moncher-dev:
## TL;DR

`workflow validate` and `doctor` now emit actionable, non-fatal warnings for ignored per-state concurrency entries, while preserving valid limits in the same map.

## Change-point diagram

`WORKFLOW.md` parser → typed non-fatal diagnostics → `workflow validate` + `doctor` warnings

## Start here

- `packages/core/src/workflow/parser.ts` detects invalid, blank, padded, and colliding normalized state-limit entries.
- `packages/cli/src/priority-diagnostics.ts` converts state-concurrency diagnostics into CLI warnings.
- `.changeset/bright-state-concurrency-warnings.md` records the `@gh-symphony/cli` patch release.

## Changes

- Emit paths, reasons, and remediation for non-integer, non-positive, blank-normalized, and normalized-key-collision entries.
- Preserve effective valid limits and last-entry-wins collision behavior.
- Quote padded state keys in paths and render collision warnings grammatically.
- Keep derived diagnostics out of workflow revision hashes and correct doctor-command documentation.

## Evidence

- `pnpm lint`
- `pnpm test`
- `pnpm typecheck`
- `pnpm build`
- Focused parser, workflow validation, and doctor suites (174 tests).

## Risks & rollback

- Warnings are non-fatal and do not alter dispatch behavior.
- Reverting this PR removes the warning surface and returns to prior silent ignore behavior.

## Changed files

- Workflow parser/configuration diagnostics and revision hashing
- CLI warning rendering plus parser, validate, and doctor tests
- Configuration documentation and CLI patch changeset

- [ ] Run `gh-symphony workflow validate` with valid, invalid, padded, and colliding per-state limits.
- [ ] Confirm `gh-symphony doctor` reports the same path-specific warnings.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -78,20 +78,20 @@ diverge from the upstream shell-command hook model.
 
 ## WORKFLOW.md Front-matter Validation
 
-`gh-symphony workflow validate` and `gh-symphony repo doctor` use the same
+`gh-symphony workflow validate` and `gh-symphony doctor` use the same
 strict parser as workflow loading. Failures include a stable error code; the
 `workflow validate --json` error also exposes its field path separately.
 
-| Rule                      | Required value                                                                                                             | Error code/path                                                               |
-| ------------------------- | -------------------------------------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------- |
-| Front matter syntax       | Optional; when present it must be a valid YAML mapping                                                                     | `workflow_parse_error` or `workflow_front_matter_not_a_map` at `front_matter` |
-| Numeric fields            | YAML integers; quoted numeric strings and fractions are rejected, including priority maps and per-state concurrency values | `workflow_validation_error` at the field path                                 |
-| `hooks.timeout_ms`        | A positive integer when provided                                                                                           | `workflow_validation_error` at `hooks.timeout_ms`                             |
-| `agent.max_turns`         | A positive integer when provided                                                                                           | `workflow_validation_error` at `agent.max_turns`                              |
-| `codex.command`           | A non-empty string when provided                                                                                           | `workflow_validation_error` at `codex.command`                                |
-| `tracker.kind`            | One of `github-project`, `linear`, or `file`                                                                               | `workflow_validation_error` at `tracker.kind`                                 |
-| `tracker.required_labels` | An array of strings; comma-separated strings are rejected                                                                  | `workflow_validation_error` at `tracker.required_labels`                      |
-| `server.port`             | An integer from `0` to `65535`; `0` requests an ephemeral local port                                                       | `workflow_validation_error` at `server.port`                                  |
+| Rule                      | Required value                                                                                                                     | Error code/path                                                                                     |
+| ------------------------- | ---------------------------------------------------------------------------------------------------------------------------------- | --------------------------------------------------------------------------------------------------- |
+| Front matter syntax       | Optional; when present it must be a valid YAML mapping                                                                             | `workflow_parse_error` or `workflow_front_matter_not_a_map` at `front_matter`                       |
+| Numeric fields            | YAML integers; quoted numeric strings and fractions are rejected. Invalid per-state concurrency entries are ignored with warnings. | `workflow_validation_error` at other invalid numeric field paths; warnings name ignored entry paths |
+| `hooks.timeout_ms`        | A positive integer when provided                                                                                                   | `workflow_validation_error` at `hooks.timeout_ms`                                                   |
+| `agent.max_turns`         | A positive integer when provided                                                                                                   | `workflow_validation_error` at `agent.max_turns`                                                    |
+| `codex.command`           | A non-empty string when provided                                                                                                   | `workflow_validation_error` at `codex.command`                                                      |
+| `tracker.kind`            | One of `github-project`, `linear`, or `file`                                                                                       | `workflow_validation_error` at `tracker.kind`                                                       |
+| `tracker.required_labels` | An array of strings; comma-separated strings are rejected                                                                          | `workflow_validation_error` at `tracker.required_labels`                                            |
+| `server.port`             | An integer from `0` to `65535`; `0` requests an ephemeral local port                                                               | `workflow_validation_error` at `server.port`                                                        |
 
 Without front matter, the complete trimmed file is used as the workflow prompt
 and all configuration uses defaults; tracker selection is then checked by the
@@ -108,7 +108,9 @@ maps remain supported through `agent.max_concurrent_agents_by_state`. State keys
 are trimmed and lowercased before storage and dispatch lookup, so for example
 `" In Progress "` and `in progress` configure the same limit. Entries whose
 values are non-positive or not YAML integers are ignored; valid entries in the
-same map remain effective.
+same map remain effective. `gh-symphony workflow validate` and `gh-symphony
+doctor` warn with the ignored entry path and reason so the configuration
+can be corrected without changing dispatch behavior.
 
 ## Optional HTTP Server
 
@@ -359,7 +361,7 @@ or field is omitted. Flat `tracker.project_id`, `tracker.endpoint`,
 `tracker.state_field`, `tracker.priority`, `tracker.pickup_labels`,
 `tracker.blocker_check_states`, and `tracker.planning_states` aliases remain
 supported for compatibility, but `gh-symphony workflow validate` and
-`gh-symphony repo doctor` warn and print a copyable `tracker.provider` block.
+`gh-symphony doctor` warn and print a copyable `tracker.provider` block.
 They are scheduled for removal in the next major release; see
 [ADR 2026-08-29](adr/2026-08-29_tracker-provider-alias-deprecation.md).
 
--- a/packages/cli/src/priority-diagnostics.ts
+++ b/packages/cli/src/priority-diagnostics.ts
@@ -1,6 +1,7 @@
 import type {
   ParsedWorkflow,
   TrackedIssue,
+  WorkflowDiagnostic,
   WorkflowPriorityConfig,
 } from "@gh-symphony/core";
 import type { ProjectDetail } from "./github/client.js";
@@ -11,6 +12,17 @@ export type PriorityDiagnostic = {
   remediation?: string;
   details?: Record<string, unknown>;
 };
+
+export function buildStateConcurrencyDiagnostics(
+  workflow: ParsedWorkflow
+): PriorityDiagnostic[] {
+  return workflow.diagnostics.map((diagnostic: WorkflowDiagnostic) => ({
+    title: "Ignored per-state concurrency entry",
+    summary: `${diagnostic.path} is ignored because it ${diagnostic.reason}.`,
+    remediation: diagnostic.remediation,
+    details: { path: diagnostic.path, reason: diagnostic.reason },
+  }));
+}
 
 type RepositoryLabelSnapshot = {
   repository: string;
--- a/packages/cli/src/priority-diagnostics.ts
+++ b/packages/cli/src/priority-diagnostics.ts
@@ -16,12 +16,17 @@ export type PriorityDiagnostic = {
 export function buildStateConcurrencyDiagnostics(
   workflow: ParsedWorkflow
 ): PriorityDiagnostic[] {
-  return workflow.diagnostics.map((diagnostic: WorkflowDiagnostic) => ({
-    title: "Ignored per-state concurrency entry",
-    summary: `${diagnostic.path} is ignored because it ${diagnostic.reason}.`,
-    remediation: diagnostic.remediation,
-    details: { path: diagnostic.path, reason: diagnostic.reason },
-  }));
+  return workflow.diagnostics
+    .filter(
+      (diagnostic: WorkflowDiagnostic) =>
+        diagnostic.code === "state_concurrency_entry_ignored"
+    )
+    .map((diagnostic) => ({
+      title: "Ignored per-state concurrency entry",
+      summary: `${diagnostic.path} is ignored: ${diagnostic.reason}.`,
+      remediation: diagnostic.remediation,
+      details: { path: diagnostic.path, reason: diagnostic.reason },
+    }));
 }
 
 type RepositoryLabelSnapshot = {
--- a/packages/core/src/workflow/config.ts
+++ b/packages/core/src/workflow/config.ts
@@ -79,6 +79,7 @@ export type WorkflowAgentConfig = {
 
 /** A non-fatal workflow configuration finding surfaced by validation tools. */
 export type WorkflowDiagnostic = {
+  code: "state_concurrency_entry_ignored";
   path: string;
   reason: string;
   remediation: string;
--- a/packages/core/src/workflow/config.ts
+++ b/packages/core/src/workflow/config.ts
@@ -77,6 +77,13 @@ export type WorkflowAgentConfig = {
   retryBaseDelayMs: number;
 };
 
+/** A non-fatal workflow configuration finding surfaced by validation tools. */
+export type WorkflowDiagnostic = {
+  path: string;
+  reason: string;
+  remediation: string;
+};
+
 export type RetryPolicyOptions = {
   baseDelayMs?: number;
   maxDelayMs?: number;
@@ -150,6 +157,7 @@ export type ParsedWorkflow = WorkflowDefinition & {
   agentCommand: string;
   hookPath: string | null;
   maxConcurrentByState: Record<string, number>;
+  diagnostics: WorkflowDiagnostic[];
 };
 
 export const DEFAULT_CODEX_COMMAND = "codex app-server";
@@ -251,6 +259,7 @@ export const DEFAULT_WORKFLOW_DEFINITION: ParsedWorkflow = {
   agentCommand: DEFAULT_CODEX_COMMAND,
   hookPath: null,
   maxConcurrentByState: {},
+  diagnostics: [],
 };
 
 export function resolveWorkflowRuntimeCommand(
--- a/packages/core/src/workflow/parser.ts
+++ b/packages/core/src/workflow/parser.ts
@@ -764,6 +764,7 @@ function parseLegacyWorkflowMarkdown(markdown: string): ParsedWorkflow {
 
   return {
     ...DEFAULT_WORKFLOW_DEFINITION,
+    diagnostics: [],
     tracker: {
       ...DEFAULT_WORKFLOW_DEFINITION.tracker,
       activeStates: DEFAULT_WORKFLOW_LIFECYCLE.activeStates,
@@ -1460,11 +1461,13 @@ function readMaxConcurrentAgentsByState(
   }
 
   const result: Record<string, number> = {};
+  const sourcePaths: Record<string, string> = {};
   const diagnostics: WorkflowDiagnostic[] = [];
   for (const [entryKey, entryValue] of Object.entries(value)) {
-    const entryPath = `${path}.${entryKey}`;
+    const entryPath = formatWorkflowMapEntryPath(path, entryKey);
     if (typeof entryValue !== "number" || !Number.isInteger(entryValue)) {
       diagnostics.push({
+        code: "state_concurrency_entry_ignored",
         path: entryPath,
         reason: "must be a positive YAML integer",
         remediation:
@@ -1474,6 +1477,7 @@ function readMaxConcurrentAgentsByState(
     }
     if (entryValue <= 0) {
       diagnostics.push({
+        code: "state_concurrency_entry_ignored",
         path: entryPath,
         reason: "must be greater than zero",
         remediation:
@@ -1484,9 +1488,21 @@ function readMaxConcurrentAgentsByState(
 
     const normalizedKey = normalizeWorkflowState(entryKey);
     if (normalizedKey) {
+      const previousPath = sourcePaths[normalizedKey];
+      if (previousPath) {
+        diagnostics.push({
+          code: "state_concurrency_entry_ignored",
+          path: previousPath,
+          reason: `is overridden by ${entryPath} after state-name normalization`,
+          remediation:
+            "Remove one of the entries that normalize to the same state name.",
+        });
+      }
       result[normalizedKey] = entryValue;
+      sourcePaths[normalizedKey] = entryPath;
     } else {
       diagnostics.push({
+        code: "state_concurrency_entry_ignored",
         path: entryPath,
         reason: "state name is blank after normalization",
         remediation: "Use a non-blank state name, or remove the entry.",
@@ -1494,6 +1510,12 @@ function readMaxConcurrentAgentsByState(
     }
   }
   return { limits: result, diagnostics };
+}
+
+function formatWorkflowMapEntryPath(path: string, entryKey: string): string {
+  return entryKey.trim().length === 0
+    ? `${path}[${JSON.stringify(entryKey)}]`
+    : `${path}.${entryKey}`;
 }
 
 export function resolveEnvironmentValue(
--- a/packages/core/src/workflow/parser.ts
+++ b/packages/core/src/workflow/parser.ts
@@ -15,6 +15,7 @@ import {
   DEFAULT_WORKFLOW_DEFINITION,
   DEFAULT_WORKFLOW_TRACKER,
   type ParsedWorkflow,
+  type WorkflowDiagnostic,
   type WorkflowPriorityConfig,
   type WorkflowRuntimeConfig,
   type WorkflowRuntimeKind,
@@ -242,7 +243,7 @@ function parseWorkflowConfig(
     }) ?? []
   );
 
-  const maxConcurrentAgentsByState = readMaxConcurrentAgentsByState(
+  const stateConcurrency = readMaxConcurrentAgentsByState(
     agent,
     "max_concurrent_agents_by_state",
     "agent.max_concurrent_agents_by_state"
@@ -395,7 +396,7 @@ function parseWorkflowConfig(
           "max_retry_backoff_ms",
           "agent.max_retry_backoff_ms"
         ) ?? DEFAULT_MAX_RETRY_BACKOFF_MS,
-      maxConcurrentAgentsByState,
+      maxConcurrentAgentsByState: stateConcurrency.limits,
       maxFailureRetries:
         readOptionalIntegerLike(
           agent,
@@ -431,7 +432,8 @@ function parseWorkflowConfig(
     ),
     agentCommand,
     hookPath: readOptionalString(hooks, "after_create"),
-    maxConcurrentByState: maxConcurrentAgentsByState,
+    maxConcurrentByState: stateConcurrency.limits,
+    diagnostics: stateConcurrency.diagnostics,
   };
 
   return parsed;
@@ -1448,31 +1450,50 @@ function readMaxConcurrentAgentsByState(
   input: Record<string, WorkflowFrontMatterNode>,
   key: string,
   path = key
-): Record<string, number> {
+): { limits: Record<string, number>; diagnostics: WorkflowDiagnostic[] } {
   const value = input[key];
   if (value === undefined || value === null) {
-    return {};
+    return { limits: {}, diagnostics: [] };
   }
   if (typeof value !== "object" || Array.isArray(value)) {
     throw new Error(`Workflow front matter field "${path}" must be an object.`);
   }
 
   const result: Record<string, number> = {};
+  const diagnostics: WorkflowDiagnostic[] = [];
   for (const [entryKey, entryValue] of Object.entries(value)) {
-    if (
-      typeof entryValue !== "number" ||
-      !Number.isInteger(entryValue) ||
-      entryValue <= 0
-    ) {
+    const entryPath = `${path}.${entryKey}`;
+    if (typeof entryValue !== "number" || !Number.isInteger(entryValue)) {
+      diagnostics.push({
+        path: entryPath,
+        reason: "must be a positive YAML integer",
+        remediation:
+          "Use a whole number greater than zero, or remove the entry.",
+      });
+      continue;
+    }
+    if (entryValue <= 0) {
+      diagnostics.push({
+        path: entryPath,
+        reason: "must be greater than zero",
+        remediation:
+          "Use a whole number greater than zero, or remove the entry.",
+      });
       continue;
     }
 
     const normalizedKey = normalizeWorkflowState(entryKey);
     if (normalizedKey) {
       result[normalizedKey] = entryValue;
+    } else {
+      diagnostics.push({
+        path: entryPath,
+        reason: "state name is blank after normalization",
+        remediation: "Use a non-blank state name, or remove the entry.",
+      });
     }
   }
-  return result;
+  return { limits: result, diagnostics };
 }
 
 export function resolveEnvironmentValue(
--- a/packages/core/src/workflow/parser.ts
+++ b/packages/core/src/workflow/parser.ts
@@ -1493,7 +1493,7 @@ function readMaxConcurrentAgentsByState(
         diagnostics.push({
           code: "state_concurrency_entry_ignored",
           path: previousPath,
-          reason: `is overridden by ${entryPath} after state-name normalization`,
+          reason: `duplicates ${entryPath} after state-name normalization`,
           remediation:
             "Remove one of the entries that normalize to the same state name.",
         });
@@ -1513,7 +1513,7 @@ function readMaxConcurrentAgentsByState(
 }
 
 function formatWorkflowMapEntryPath(path: string, entryKey: string): string {
-  return entryKey.trim().length === 0
+  return entryKey !== entryKey.trim()
     ? `${path}[${JSON.stringify(entryKey)}]`
     : `${path}.${entryKey}`;
 }
--- a/.changeset/bright-state-concurrency-warnings.md
+++ b/.changeset/bright-state-concurrency-warnings.md
@@ -1,0 +1,5 @@
+---
+"@gh-symphony/cli": patch
+---
+
+Warn in `workflow validate` and `doctor` when ignored per-state concurrency entries cannot take effect (#767).
--- a/README.md
+++ b/README.md
@@ -354,7 +354,7 @@ Lower numbers dispatch first. If an issue has multiple configured priority label
 
 Legacy `tracker.priority_field: Priority` remains supported for existing workflows, but it is deprecated because it uses live Project option order. Project field definitions are cached for the process lifetime, so field creation, removal, and option changes take effect after the daemon restarts. To migrate, replace it with `tracker.provider.priority.source: project-field`, copy the exact field name, and write explicit option-name-to-number mappings. If both legacy and explicit config are present, explicit `tracker.provider.priority` wins and diagnostics warn about the conflict.
 
-`gh-symphony workflow validate` reports local config errors and legacy priority warnings. Strict front-matter failures use stable workflow error codes; with `--json`, `workflow validate` includes both `error.code` and `error.path`. `gh-symphony doctor` additionally checks live Project/repository drift: missing fields, missing labels, unmapped live options, stale configured mappings, and active issues that currently resolve to `priority = null` because their priority-like value is unmapped.
+`gh-symphony workflow validate` reports local config errors plus warnings for legacy priority configuration and ignored per-state concurrency entries. Each concurrency warning names the ignored `agent.max_concurrent_agents_by_state` path and reason, while valid entries in the same map remain active. Strict front-matter failures use stable workflow error codes; with `--json`, `workflow validate` includes both `error.code` and `error.path`. `gh-symphony doctor` additionally checks live Project/repository drift and reports the same local configuration warnings.
 
 ### Token-Only Setup
 
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -204,7 +204,7 @@ Use `tracker.provider` for new workflows; `gh-symphony doctor` prints the
 normalized provider block for migration. The aliases are removed in the next
 major release (#679).
 
-Run `gh-symphony workflow validate` for local schema errors and `gh-symphony doctor` for live drift warnings such as missing Project fields, missing labels, unmapped live options, stale mappings, and active issues whose priority-like value resolves to `priority = null`. Strict front-matter failures use stable workflow error codes; `workflow validate --json` also emits the failing `error.path`.
+Run `gh-symphony workflow validate` for local schema errors and warnings. Ignored `agent.max_concurrent_agents_by_state` entries warn with their paths and reasons, while valid entries in the same map remain active; `gh-symphony doctor` reports the same warning alongside live drift checks such as missing Project fields, missing labels, unmapped live options, stale mappings, and active issues whose priority-like value resolves to `priority = null`. Strict front-matter failures use stable workflow error codes; `workflow validate --json` also emits the failing `error.path`.
 
 ### Linear Tracker Repositories
 
--- a/packages/cli/src/commands/doctor.test.ts
+++ b/packages/cli/src/commands/doctor.test.ts
@@ -2961,6 +2961,50 @@ describe("doctor command handler", () => {
     });
   });
 
+  it("warns about ignored per-state concurrency entries", async () => {
+    const configDir = await mkdtemp(join(tmpdir(), "doctor-config-"));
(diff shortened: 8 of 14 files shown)


packages/cli/src/priority-diagnostics.ts
```diff
@@ -12,6 +13,17 @@ export type PriorityDiagnostic = {
   details?: Record<string, unknown>;
 };
 
+export function buildStateConcurrencyDiagnostics(
+  workflow: ParsedWorkflow
+): PriorityDiagnostic[] {
+  return workflow.diagnostics.map((diagnostic: WorkflowDiagnostic) => ({
+    title: "Ignored per-state concurrency entry",
+    summary: `${diagnostic.path} is ignored because it ${diagnostic.reason}.`,
```
hojinzs: **P2 — the `because it ${reason}` sentence frame breaks for the blank-key reason**

The two numeric reasons read as verb phrases (`must be greater than zero`), but the blank-key reason from `parser.ts:1471` is a noun phrase (`state name is blank after normalization`). Interpolated into this template it produces ungrammatical user-facing output.

Verified against the built CLI on this head (`pnpm build`, then `workflow validate` on a fixture with `"  ": 3`):

```
Ignored per-state concurrency entry: agent.max_concurrent_agents_by_state.   is ignored because it state name is blank after normalization.
```

Same string lands in `--json` under `warnings[].summary`, so it is not just a rendering artifact. Neither the parser test nor the two new CLI tests exercise the blank-key path, which is why it slipped through.

Either normalize the reason to a verb phrase (`has a state name that is blank after normalization`) or stop templating a fragment — have the parser emit a complete `message` and let the CLI pass it through. Worth a test on the blank-key case whichever way you go.

---

packages/cli/src/priority-diagnostics.ts
```diff
@@ -12,6 +13,17 @@ export type PriorityDiagnostic = {
   details?: Record<string, unknown>;
 };
 
+export function buildStateConcurrencyDiagnostics(
+  workflow: ParsedWorkflow
+): PriorityDiagnostic[] {
+  return workflow.diagnostics.map((diagnostic: WorkflowDiagnostic) => ({
```
hojinzs: **P2 — a generic diagnostics channel is consumed as if every entry were state-concurrency**

`WorkflowDiagnostic` is deliberately declared as the general case (`config.ts:75`: "A non-fatal workflow configuration finding surfaced by validation tools") and hangs off `ParsedWorkflow.diagnostics` with no discriminator. This function then maps the whole array and hardcodes `title: "Ignored per-state concurrency entry"` on every element.

Correct today, because `readMaxConcurrentAgentsByState` is the only producer. But the generic naming actively invites a second producer — that is the stated point of introducing "a warning channel" in #767 — and the first one added will silently be reported to operators under the wrong title, in both `workflow validate` and `repo doctor`. Nothing in the type system or the tests catches that.

Cheapest fix that keeps the channel general: add a discriminator to `WorkflowDiagnostic` (`code: "state_concurrency_entry_ignored"` or similar) and filter on it here, deriving the title from the code. Alternative, if the channel is not actually meant to be general yet: name the field `stateConcurrencyDiagnostics` and drop the generic doc comment, so the next author has to make the widening decision explicitly.

---

packages/core/src/workflow/parser.ts
```diff
@@ -1428,31 +1430,50 @@ function readMaxConcurrentAgentsByState(
...
+        reason: "must be greater than zero",
+        remediation:
+          "Use a whole number greater than zero, or remove the entry.",
+      });
       continue;
     }
 
     const normalizedKey = normalizeWorkflowState(entryKey);
     if (normalizedKey) {
       result[normalizedKey] = entryValue;
```
hojinzs, line 1501: **P2 — entries silently dropped by key collision still have no warning**

`result[normalizedKey] = entryValue` overwrites on collision, so when two keys normalize to the same state the earlier ones are discarded with no diagnostic. Issue #767 frames the problem as "entries that are intentionally ignored" having "no warning channel" — a collided entry is ignored just as completely as `Ready: 0`, and arguably more surprisingly, since the value looked valid.

Verified on this head:

```yaml
agent:
  max_concurrent_agents_by_state:
    Ready: 5
    ready: 1
    " READY ": 9
```

```
warnings: []
effective limits: { "ready": 9 }
```

Two of three entries vanish and `workflow validate` reports a clean pass. This is not a hypothetical shape either — `docs/configuration.md:94` advertises exactly this normalization (`" In Progress "` and `in progress` configure the same limit), so an operator following the documented behavior can land here.

Suggest emitting a diagnostic when `normalizedKey` is already present in `result`, naming both the shadowed path and the winning one. Small addition to the loop you are already touching, and it closes the last silent-drop path in this map.

If you judge it out of scope for #767, that is a reasonable call — but say so on the thread and file a follow-up rather than leaving it undocumented, since the PR body claims the ignored-entry class is now covered.

---

packages/core/src/workflow/parser.ts
```diff
@@ -1428,31 +1430,50 @@ function readMaxConcurrentAgentsByState(
...
+    } else {
+      diagnostics.push({
+        path: entryPath,
```
hojinzs, line 1506: **P3 — the blank-key path is unreadable because the key is pure whitespace**

`entryPath` is `${path}.${entryKey}`, and on this branch `entryKey` is by definition whitespace-only. Actual output for a `"  ": 3` entry:

```json
"path": "agent.max_concurrent_agents_by_state.  "
```

The trailing spaces are invisible in a terminal, so the warning points at an entry the operator cannot locate in their `WORKFLOW.md`. Quoting or JSON-escaping the key segment (`agent.max_concurrent_agents_by_state["  "]`) makes it identifiable. Only affects this branch — the other two reasons carry real key names.

---

packages/core/src/workflow/config.ts
```diff
@@ -144,6 +151,7 @@ export type ParsedWorkflow = WorkflowDefinition & {
   agentCommand: string;
   hookPath: string | null;
   maxConcurrentByState: Record<string, number>;
+  diagnostics: WorkflowDiagnostic[];
```
hojinzs, line 161: **nit — derived diagnostics now participate in the workflow identity hash**

`calculateWorkflowVersionHash` (`loader.ts:124`) hashes `JSON.stringify(workflow)` over the whole `ParsedWorkflow`, and `createWorkflowRevision` derives the `revision` string from it. Adding `diagnostics` to this type means the serialized form gains `"diagnostics":[]` for every workflow, so every existing workflow's `revision` shifts on upgrade even though nothing about its effective configuration changed.

I traced the blast radius and it is contained: `revision` only reaches `structured-events.ts:76` and `snapshot-builder.ts:76`, both observability-only — no cache keying, no dispatch gating. So this is cosmetic, one-time noise in event streams, not a correctness problem.

Still, diagnostics are derived output rather than configuration identity, and folding them into the hash is the kind of thing that becomes load-bearing later. Excluding them from the hashed shape (destructure them out in `calculateWorkflowVersionHash`) would keep the revision meaning "this configuration", which is what the field name promises.

---

packages/core/src/workflow/config.ts
```diff
@@ -240,6 +248,7 @@ export const DEFAULT_WORKFLOW_DEFINITION: ParsedWorkflow = {
   agentCommand: DEFAULT_CODEX_COMMAND,
   hookPath: null,
   maxConcurrentByState: {},
+  diagnostics: [],
```
hojinzs, line 263: **nit — this array instance is shared by every legacy-format workflow**

`parseLegacyWorkflowMarkdown` (`parser.ts:757`) builds its result by spreading `DEFAULT_WORKFLOW_DEFINITION`, so every legacy-sectioned workflow aliases *this* array rather than getting its own. Inert today because nothing mutates `diagnostics` after parse — the front-matter path always assigns a fresh array from `readMaxConcurrentAgentsByState`.

It becomes a cross-workflow leak the moment anyone pushes into `workflow.diagnostics` post-parse, which is a natural thing to do once the channel is general (see my note on `priority-diagnostics.ts`). Same latent hazard already exists for `maxConcurrentByState: {}` above, so this is pre-existing shape rather than something you introduced — flagging it because this PR adds the second instance and a `readonly` marker or a per-parse `[]` in the legacy path is cheap now.

---

docs/configuration.md
```diff
@@ -93,7 +93,9 @@ maps remain supported through `agent.max_concurrent_agents_by_state`. State keys
 are trimmed and lowercased before storage and dispatch lookup, so for example
 `" In Progress "` and `in progress` configure the same limit. Entries whose
 values are non-positive or not YAML integers are ignored; valid entries in the
-same map remain effective.
+same map remain effective. `gh-symphony workflow validate` and `gh-symphony
+repo doctor` warn with the ignored entry path and reason so the configuration
```
hojinzs: **P3 — `gh-symphony repo doctor` is not a real command** (confirming the Codex finding on this line)

Verified: `doctor` is registered only as a top-level command (`packages/cli/src/index.ts:468`), and the `repo` handler dispatches exactly `init | start | run | recover | logs | explain | status | stop` (`packages/cli/src/commands/repo.ts`) — no `doctor`. Anyone following this sentence gets a command error instead of the warnings the paragraph is describing.

Two things the Codex comment does not mention, which change the fix:

1. This is **pre-existing**, not introduced here — `docs/configuration.md:67` and `:339` already say `repo doctor`. Your new sentence is the third instance.
2. The PR body's "Post-merge / human validation" checklist has the same mistake, so the stated validation step would fail as written.

Since you are editing this file anyway, fixing all three occurrences plus the checklist is a one-line-each change and gets the docs consistent with `README.md` / `packages/cli/README.md`, which both correctly say `gh-symphony doctor`. Fixing only your new line is also acceptable if you would rather keep the diff tight — just leave the other two for a follow-up rather than silently.

---

hojinzs (changes requested):
## Review 1/6 — requirements met, one release-process blocker + three P2s

Scope is tight and matches #767 — no over-engineering, no unrequested decisions. I verified the behavior end-to-end rather than reading only the diff.

### Verification performed

`pnpm install && pnpm build` clean; `packages/core/src/workflow-loader.test.ts` + `packages/cli/src/commands/workflow.test.ts` → 105/105 pass. CI (`Test`, `Container Smoke`) green on `5c9604d`.

Smoke test against the **built CLI** (`packages/cli/dist/index.js workflow validate`), not just unit tests, with a fixture covering every branch:

```yaml
agent:
  max_concurrent_agents_by_state:
    Ready: 0          # → warns "must be greater than zero"        ✅
    Review: '2'       # → warns "must be a positive YAML integer"   ✅
    Blocked: 1.5      # → warns "must be a positive YAML integer"   ✅
    Backlog: -1       # → warns "must be greater than zero"         ✅
    "  ": 3           # → warns, but ungrammatical + unreadable path ❌
    " In Progress ": 2 # → preserved as {"in progress": 2}          ✅
```

Both `--json` (`warnings[]` with `path`/`reason`/`remediation`) and human output (`Warnings` section with `Fix:` lines) render correctly, and `ok` stays `true` — warnings are genuinely non-fatal as claimed.

### Acceptance criteria from #767

| Criterion | Status |
| --- | --- |
| Surface actionable warnings with entry paths and reasons | ⚠️ Met for invalid values; **key-collision drops are still silent** |
| Preserve valid entries in the same map | ✅ Verified — mixed map keeps `{"in progress": 2}` |
| Add parser/doctor coverage | ✅ Parser, `workflow validate`, and `doctor` all covered |

### Findings

**P1 — no changeset.** `CLAUDE.md` requires a `.changeset/*.md` for `@gh-symphony/cli` on behavior-changing PRs, and this changes user-visible CLI output on two commands. The 11 changed files include none. Precedent: #752, the PR that created this ignore behavior, shipped `normalize-state-concurrency.md`. Without it the release notes will not mention the new warnings. This is the only thing I would call a hard blocker.

**P2 ×3** (inline): the `because it ${reason}` template producing ungrammatical output on the blank-key branch; the generic `WorkflowDiagnostic` channel consumed as if every entry were state-concurrency; and key-collision entries still dropped with no warning.

**P3 ×2** (inline): whitespace-only path unreadable in output; `gh-symphony repo doctor` is not a registered command — confirming Codex's finding on `docs/configuration.md:97`, with the added context that it is pre-existing at `:67` and `:339` and also wrong in this PR's own validation checklist.

**nit ×2** (inline): `diagnostics` entering the workflow revision hash; shared array instance via the legacy-parse spread.

### What I need for approval

The P1 changeset, plus the two P2s that are straightforwardly in scope — the grammar bug and the diagnostic discriminator. On the collision P2, a reasoned "out of scope for #767, follow-up filed" reply is a perfectly good answer; I do not need the code. The P3s and nits are yours to take or decline.

---

hojinzs (changes requested):
## 리뷰 1회차 — Request changes

대상 head `5c9604d` (origin/main `c549379` 위로 리베이스됨, GitHub mergeable/clean, CI Test·Container Smoke 모두 success).

### 요구사항 반영

- [x] 무시된 `agent.max_concurrent_agents_by_state` 항목을 경고로 노출 — `packages/core/src/workflow/parser.ts:1429-1476`에서 `{ path, reason, remediation }` 진단을 수집하고 `ParsedWorkflow.diagnostics`(`config.ts:76-80, 154`)로 전달. `workflow validate`(`workflow.ts:860`)와 `doctor`(`doctor.ts:2183-2193`, 신규 check id `state_concurrency`, `warnCheck` 사용)에서 WARN으로 표시.
- [x] 유효 항목 보존·스케줄러/파서 의미 불변 — `readMaxConcurrentAgentsByState`의 필터 조건은 #752와 동일(정수 아님/문자열 → `must be a positive YAML integer`, `<= 0` → `must be greater than zero`, 정규화 후 빈 키 → `state name is blank after normalization`). `maxConcurrentAgentsByState`/`maxConcurrentByState`는 이전과 같은 `limits`를 받고, orchestrator 코드는 손대지 않음. 기존 core 테스트(`workflow-loader.test.ts:706`)는 삭제 없이 `diagnostics` 단언만 추가.
- [x] 트래커 무관 — 진단은 core 파서에서 생성되고 CLI 변환(`priority-diagnostics.ts:16-26`)에 트래커 분기 없음.
- [x] 문서 정정 — `docs/configuration.md:74` 'Numeric fields' 행이 "Invalid per-state concurrency entries are ignored with warnings." + "warnings name ignored entry paths"로 복구되어 아래 문단(`:95-98`, ignored + validate/doctor 경고)과 일치. 이슈 코멘트(#740 리베이스 c2f2b0d 되돌림)에서 요청한 내용 그대로. README/cli README 문구도 일관.
- [x] 테스트 — core 파서(`workflow-loader.test.ts:709-726`), `workflow validate --json`(`workflow.test.ts:263-326`), doctor(`doctor.test.ts:2964-3006`, `report.ok === true` + `status: "warn"` 확인).
- [x] 스펙 — §5.3.5 "Invalid entries (non-positive or non-numeric) are ignored"와 §8.3 정규화 키 조회를 그대로 따르며, 경고는 Observability 표면 추가일 뿐 스펙 이탈 없음.
- [ ] **changeset 누락** — 아래 지적사항 1.

### 검증한 것 (빌드된 CLI 실행)

`pnpm -r build` 후 `packages/cli/dist/index.js`를 격리된 `HOME`/`GH_SYMPHONY_CONFIG_DIR`로 실행(4680 포트 미사용, 토큰 unset).

```yaml
agent:
  max_concurrent_agents_by_state:
    " Todo ": 1
    ready: 0
    Review: "2"
    Done: 1.5
    Blocked: -1
```

- `workflow validate --json` → `ok: true`, exit 0, `warnings[]` 4건: `…by_state.ready` / `…Blocked` → `must be greater than zero`, `…Review` / `…Done` → `must be a positive YAML integer`. 각 항목에 `remediation`과 `details.{path,reason}` 포함. `" Todo ": 1`은 경고 없이 유지.
- `workflow validate`(텍스트) → "WORKFLOW.md validation passed" + `Warnings` 섹션에 같은 4건, `Fix:` 줄 포함.
- `doctor --json` → `workflow_file: pass`, `runtime_command: pass`, `state_concurrency` 4건 모두 `status: "warn"`(FAIL 아님). 격리 환경이라 auth/managed_project는 fail이지만 이 PR 범위 밖. `doctor` 텍스트 출력도 `WARN Ignored per-state concurrency entry` 4줄.
- 유효 맵(`" Todo ": 1, "In Progress": 2`) → `workflow validate --json` `warnings: []`, `doctor --json`에 `state_concurrency` check 없음.
- 로컬: `pnpm --filter @gh-symphony/core test` 298 pass, `@gh-symphony/cli` 596 pass, `@gh-symphony/orchestrator` 365 pass, `pnpm typecheck` / `pnpm lint` 통과, 변경 파일 11개 `prettier --check` 통과, `pnpm -r build` 성공.

### 지적사항

1. **(blocking) changeset 없음.** `git diff origin/main...HEAD -- .changeset` 이 비어 있습니다. 이 PR은 `workflow validate`/`doctor`의 출력(JSON `warnings[]`, 새 check id `state_concurrency`)을 바꾸는 동작 변경이므로 CLAUDE.md "Releases" 규칙대로 `.changeset/*.md`(`"@gh-symphony/cli": patch`)가 필요합니다. 선행 PR #752도 `.changeset/normalize-state-concurrency.md`를 넣었습니다. 워크패드의 "Changeset: N/A (no changeset label)"은 라벨 유무가 아니라 동작 변경 여부로 판단해야 합니다.
2. (non-blocking) `packages/cli/src/priority-diagnostics.ts:19` — `workflow.diagnostics` 전체를 무조건 "Ignored per-state concurrency entry"로 제목 붙입니다. `ParsedWorkflow.diagnostics`(`config.ts:154`)는 이름이 범용이라 나중에 다른 파서 진단이 추가되면 잘못된 제목이 붙습니다. `WorkflowDiagnostic`에 `code`/`kind` 판별자(예: `state_concurrency_ignored`)를 두고 CLI에서 필터하거나, 필드명을 `stateConcurrencyDiagnostics`처럼 좁혀 두는 것을 권합니다. 후속 이슈로 빼도 됩니다.
3. (참고, 이 PR 범위 밖) 정규화 후 중복되는 키(`Ready`와 `" ready "`)는 뒤 항목이 조용히 덮어쓰며 경고가 없습니다. 또 flow-style 매핑(`{ a: 1, b: 2 }`)은 core의 자체 front-matter 파서가 지원하지 않아 `must be an object`로 거부됩니다 — main과 동일한 기존 동작이라 별도 이슈감입니다.

### 판단

구현·테스트·문서·스펙 정합성은 모두 확인되었고 CLI 실행으로 요구한 동작(무시 키 4개 경고, WARN 비치명, 유효 맵 무경고)이 재현됩니다. 다만 동작 변경 PR의 필수 산출물인 changeset이 빠져 있어 **Request changes**입니다. `.changeset/<name>.md`(`"@gh-symphony/cli": patch`, 예: "Warn in `workflow validate` and `doctor` about ignored `max_concurrent_agents_by_state` entries (#767)") 한 파일 추가 후 재요청 주시면 바로 Approve 하겠습니다. 2번은 선택.

docs/configuration.md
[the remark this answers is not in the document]
moncher-dev: Fixed all three `gh-symphony repo doctor` references in this document to the registered `gh-symphony doctor` command. The PR validation checklist is refreshed to use the same command.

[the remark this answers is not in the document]
moncher-dev: Fixed all three `gh-symphony repo doctor` references in this document to the registered `gh-symphony doctor` command. The PR validation checklist is refreshed to use the same command.

packages/cli/src/priority-diagnostics.ts
moncher-dev: Fixed by rendering complete sentences with `is ignored: <reason>.` and added a blank-state-name regression test that exercises the user-facing summary.

packages/cli/src/priority-diagnostics.ts
moncher-dev: Added the `state_concurrency_entry_ignored` discriminator to `WorkflowDiagnostic`; the CLI now filters by it before assigning the state-concurrency title. This keeps the shared channel safe for future diagnostic producers.

packages/core/src/workflow/parser.ts
moncher-dev: Implemented collision diagnostics. When two valid entries normalize to the same state, the displaced entry now warns with both paths in its reason while preserving the existing last-entry-wins runtime behavior. Added parser coverage.

packages/core/src/workflow/parser.ts
moncher-dev: Fixed by rendering whitespace-only keys with a JSON-quoted bracket path such as `agent.max_concurrent_agents_by_state["  "]`, and covered it in parser and CLI tests.

packages/core/src/workflow/config.ts
moncher-dev: Excluded derived `diagnostics` from `calculateWorkflowVersionHash` and added a regression test confirming diagnostics do not alter the workflow revision.

packages/core/src/workflow/config.ts
moncher-dev: The legacy parser now assigns a fresh `diagnostics: []` array rather than inheriting the default array instance.

moncher-dev:
Rework is complete and the required patch changeset has been added. All inline findings have been addressed with replies and regression coverage. Requesting re-review.

packages/core/src/workflow/parser.ts
```diff
@@ -1428,31 +1431,71 @@ function readMaxConcurrentAgentsByState(
...
     }
 
     const normalizedKey = normalizeWorkflowState(entryKey);
     if (normalizedKey) {
+      const previousPath = sourcePaths[normalizedKey];
+      if (previousPath) {
+        diagnostics.push({
+          code: "state_concurrency_entry_ignored",
+          path: previousPath,
+          reason: `is overridden by ${entryPath} after state-name normalization`,
```
hojinzs: **P2 — the two fixes in this commit collide: the collision reason double-verbs against the new prefix**

The grammar fix changed the CLI template to `${path} is ignored: ${reason}.`, and the collision diagnostic added here supplies a reason that already starts with a verb (`is overridden by …`). Composed, they read:

```
agent.max_concurrent_agents_by_state.Todo is ignored: is overridden by agent.max_concurrent_agents_by_state. TODO  after state-name normalization.
```

Verified on `d08a152` against the built CLI, in both text and `--json` (`warnings[].summary`). The other three reasons are bare predicates (`must be greater than zero`, `state name is blank after normalization`) and compose correctly — this one is the odd one out.

Dropping the leading `is` gives `… is ignored: overridden by … after state-name normalization.`, which fits the frame the rest of the map uses.

Worth noting *why* this slipped: the new parser test asserts on `reason` in isolation and the new CLI test covers the blank-key summary, so nothing exercises the composed collision summary. A rendering assertion on that path would have caught it, and would guard the next reason someone adds.

---

packages/core/src/workflow/parser.ts
```diff
@@ -1428,31 +1431,71 @@ function readMaxConcurrentAgentsByState(
...
+        remediation: "Use a non-blank state name, or remove the entry.",
+      });
     }
   }
-  return result;
+  return { limits: result, diagnostics };
+}
+
+function formatWorkflowMapEntryPath(path: string, entryKey: string): string {
+  return entryKey.trim().length === 0
```
hojinzs: **P3 — the bracket guard is narrower than the problem it solves**

`entryKey.trim().length === 0` brackets only keys that are *entirely* whitespace. A key with surrounding padding has `trim().length > 0`, so it takes the dotted branch and the invisible whitespace survives:

```
agent.max_concurrent_agents_by_state. Padded Bad  is ignored: must be greater than zero.
"path": "agent.max_concurrent_agents_by_state. Padded Bad "
```

Same unreadability as the fully-blank case — the operator cannot tell where the key starts or ends. And this shape is not exotic: `docs/configuration.md:94` uses `" In Progress "` as the worked example of normalization, so a padded key is the documented idiom.

My original comment said "whitespace-only", so reading it narrowly was fair — the defect is really "the key needs quoting whenever rendering it loses information". Widening the guard to `entryKey !== entryKey.trim()` covers both cases with the bracket form you already built.

This also feeds the collision reason above, which interpolates a second `entryPath` — that is where the ` TODO ` in my other comment comes from.

---

hojinzs (changes requested):
## Review 2/6 — all seven prior findings fixed; one regression from the fixes, plus a merge conflict

Re-verified on `d08a152` by rebuilding and re-running the smoke fixture, not by reading the replies.

### Round-1 findings — all confirmed fixed

| # | Finding | Verified on `d08a152` |
| --- | --- | --- |
| P1 | Missing changeset | ✅ `.changeset/bright-state-concurrency-warnings.md`, `"@gh-symphony/cli": patch` |
| P2 | `because it ${reason}` grammar | ✅ Template now `is ignored: ${reason}.` — correct for 3 of 4 reasons (see below) |
| P2 | Generic channel, hardcoded title | ✅ `code: "state_concurrency_entry_ignored"` on the type, CLI filters before titling |
| P2 | Silent collision drops | ✅ Implemented, chains correctly across 3+ colliding keys, last-entry-wins preserved |
| P3 | Unreadable blank path | ⚠️ Fixed for fully-blank keys only — see inline |
| P3 | `repo doctor` | ✅ All three occurrences in `docs/configuration.md` (`:67`, `:97`, `:339`) plus the PR body TL;DR and checklist |
| nit | `diagnostics` in revision hash | ✅ Destructured out of `calculateWorkflowVersionHash`, with a regression test |
| nit | Shared default array | ✅ Fresh `diagnostics: []` in `parseLegacyWorkflowMarkdown` |

Local checks on the new head: **169 tests pass** across the four affected suites, `pnpm lint` clean, `pnpm typecheck` clean, `pnpm build` clean.

### What still needs a push

**P2 — regression introduced by combining two of the fixes.** The new `is ignored:` prefix and the new collision reason both landed in this commit, and they double-verb:

```
…by_state.Todo is ignored: is overridden by …by_state. TODO  after state-name normalization.
```

Inline with the one-word fix. The reason it passed CI is that the new tests assert on `reason` in isolation and on the blank-key summary — nothing renders the composed collision summary.

**P3 — the bracket guard only catches fully-blank keys**, so `" Padded Bad "` still renders with invisible padding. Inline.

**Blocker, not code: the PR is no longer mergeable.** GitHub reports `mergeable_state: dirty`. `main` advanced to `c285268` (#763, #764, #765 merged after this branch was cut) and `docs/configuration.md` now conflicts — I reproduced it locally with `git merge-tree`; that file is the only conflict, everything else auto-merges. Since #764 also edited this document, please resolve rather than take either side wholesale.

### Path to approval

The two code items are one-line changes in `parser.ts` and neither touches behavior — only the strings operators read. Bring `main` in, apply both, add a rendering assertion on the collision summary, and I will approve on the next round; I have no further findings queued. If you disagree with the P3 widening, say so on the thread and I will approve without it — I do not consider it worth a further round on its own.

---

packages/core/src/workflow/parser.ts
hojinzs: **P2 — the collision reason reintroduces the exact grammar bug this commit fixed**

The blank-key reason was reworded to fit `${path} is ignored: ${reason}.`, but the collision reason added in the same commit is a verb phrase starting with `is`, so the template emits a double verb.

Verified against the built CLI on this head (`pnpm build`, then `workflow validate --json` on a fixture with `Ready: 5` / `ready: 1` / `" READY ": 9`):

```
agent.max_concurrent_agents_by_state.Ready is ignored: is overridden by agent.max_concurrent_agents_by_state.ready after state-name normalization.
```

Same string lands in `--json` under `warnings[].summary` and in `doctor`'s check summary, so it is not a rendering artifact.

The reason it slipped through is structural, not accidental: the new parser test asserts `reason: expect.stringContaining("READY")` and never touches `summary`, and the new CLI test in `priority-diagnostics.test.ts` covers only the blank-key case — which happens to be the one reason that *is* grammatical. Every reason string that flows into that template needs one assertion on the rendered summary, otherwise the next reason added repeats this a third time.

Minimal fix — make the reason a complement phrase like the other three:

```suggestion
          reason: `duplicates ${entryPath} after state-name normalization`,
```

which renders `… .Ready is ignored: duplicates … .ready after state-name normalization.` Reworking the template so the parser emits a complete `message` is also fine; either way please add a summary-level assertion for the collision case.

---

packages/core/src/workflow/parser.ts
```diff
@@ -1428,31 +1431,71 @@ function readMaxConcurrentAgentsByState(
...
+        reason: "state name is blank after normalization",
+        remediation: "Use a non-blank state name, or remove the entry.",
+      });
     }
   }
-  return result;
+  return { limits: result, diagnostics };
+}
+
+function formatWorkflowMapEntryPath(path: string, entryKey: string): string {
```
hojinzs, line 1515: **P3 — the bracket path only covers fully-blank keys, so the documented padded form is still unlocatable**

`entryKey.trim().length === 0` catches `"  "` but not `" READY "`, which keeps the plain `${path}.${entryKey}` form with invisible padding. That is the case `docs/configuration.md:94` explicitly advertises as supported (`" In Progress "` and `in progress` configure the same limit), so padded keys are the documented normal shape rather than an edge case — and they are precisely the keys that trigger the new collision warning.

Actual output on this head for the fixture above:

```json
"path": "agent.max_concurrent_agents_by_state. READY "
```

It is worse once that path is embedded in a collision reason, because the padding lands mid-sentence and reads as an ordinary word break:

```
… is ignored: is overridden by agent.max_concurrent_agents_by_state. READY  after state-name normalization.
```

An operator cannot tell from that which of their two entries is the winner. Widening the predicate to any key that is not already trimmed fixes both:

```suggestion
function formatWorkflowMapEntryPath(path: string, entryKey: string): string {
  return entryKey !== entryKey.trim()
    ? `${path}[${JSON.stringify(entryKey)}]`
    : `${path}.${entryKey}`;
```

(`entryKey !== entryKey.trim()` subsumes the all-whitespace case, so the existing blank-key test keeps passing unchanged.)

---

hojinzs (changes requested):
## Round 2 review — 2 small fixes + a rebase, then this is an approve

Verified every prior finding on `d08a152` by building the CLI and running it, not just by reading the diff.

### Prior round: all six addressed ✅

| Finding | Status |
| --- | --- |
| P2 doc `repo doctor` (Codex + mine) | Fixed — all three occurrences and the PR checklist |
| P2 `because it ${reason}` grammar | Fixed for the blank-key case, **regressed for the new collision case** (see inline) |
| P2 generic diagnostics channel | Fixed — `code: "state_concurrency_entry_ignored"` discriminator, CLI filters on it |
| P2 collision entries silently dropped | Implemented, with last-entry-wins preserved |
| P3 blank-key path unreadable | Fixed for fully-blank keys, **still open for padded keys** (see inline) |
| nit revision hash | Fixed — `diagnostics` destructured out of `calculateWorkflowVersionHash`, with a regression test |
| nit shared array instance | Fixed in the legacy path |

### Requirements from #767 — met

- *Surface actionable warnings with entry paths and reasons* — 5 distinct warning shapes, non-fatal (`ok: true`), in both `workflow validate` and `doctor`.
- *Preserve valid entries in the same map* — verified directly: the mixed fixture yields `{"ready": 9, "in progress": 2}` with 5 warnings.
- *Add parser/doctor coverage* — `doctor.test.ts` exercises the new `state_concurrency` check id and asserts `ok: true`.

No over-engineering: everything added beyond the issue scope was requested in round 1, and each addition is minimal. The changeset is correctly scoped to `@gh-symphony/cli` as a patch.

Local verification on this head: `pnpm lint`, `pnpm typecheck`, `pnpm test` (597 CLI tests + core suites) and `pnpm build` all pass.

### Blocking

1. **Merge conflict with `main`** — `docs/configuration.md` conflicts against current `origin/main`; the PR shows `mergeable_state: dirty`. `main` moved under you (#764, `0e85d8a`) and rewrote the same paragraph to add an ADR link. Resolution is mechanical: keep `main`'s ADR link *and* your `gh-symphony doctor` correction.
2. **The two inline findings** — both are one-line changes, and since the rebase requires a push anyway they cost nothing extra.

### Non-blocking note for after the rebase

`main` reintroduced a fourth `gh-symphony repo doctor` at `docs/adr/2026-08-29_tracker-provider-alias-deprecation.md:21`. It is not yours and not in your diff — leave it, or fold it in if you prefer; a follow-up is equally fine.

### Scope commitment

This is round 2 of 6. The two inline items are the complete remaining list — I am not opening new scope on the next pass. Land those plus the rebase and I approve; if you'd rather argue either one is not worth fixing, a reasoned reply on the thread is an acceptable resolution and I'll approve on that basis instead.

---

packages/core/src/workflow/parser.ts
moncher-dev: Fixed in `8c699eee`: collision reasons now use the complement phrase `duplicates …`, so the CLI renders a grammatical warning. Added a regression assertion for the fully composed warning summary.

packages/core/src/workflow/parser.ts
moncher-dev: Fixed in `8c699eee`: any state key whose trimmed form differs now uses a JSON-quoted bracket path, including padded collision entries. The new rendered-summary regression test covers `" READY "`.

packages/core/src/workflow/parser.ts
```diff
@@ -1448,31 +1451,71 @@ function readMaxConcurrentAgentsByState(
...
     }
 
     const normalizedKey = normalizeWorkflowState(entryKey);
     if (normalizedKey) {
+      const previousPath = sourcePaths[normalizedKey];
+      if (previousPath) {
+        diagnostics.push({
+          code: "state_concurrency_entry_ignored",
+          path: previousPath,
+          reason: `duplicates ${entryPath} after state-name normalization`,
```
hojinzs, line 1496: **nit (non-blocking) — with 3+ colliding keys, the named "winner" is itself an ignored entry**

`sourcePaths[normalizedKey]` is overwritten on every iteration, so each diagnostic names the *immediately preceding* entry rather than the final winner. Verified on this head with `Ready: 5` / `ready: 1` / `" READY ": 9`:

```
agent.max_concurrent_agents_by_state.Ready is ignored: duplicates agent.max_concurrent_agents_by_state.ready …
agent.max_concurrent_agents_by_state.ready is ignored: duplicates agent.max_concurrent_agents_by_state[" READY "] …
```

The first line points at `.ready`, which the second line then declares ignored too. An operator reading only the first warning would conclude `.ready` is the effective entry when it is actually `[" READY "]` (effective limit = 9).

Both warnings together are correct and the remediation text ("Remove one of the entries that normalize to the same state name") stays accurate either way, so this is cosmetic. Three-way normalization collisions are also rare enough that I would not hold the PR for it. If you ever touch this loop again, resolving the winner in a second pass — or wording the reason as "shadowed by another entry for state `ready`" — sidesteps the chain entirely.

Not requesting a change.

---

hojinzs (approved):
## Approve — round 3, all prior findings verified fixed on `8c699ee`

I re-verified both open findings from round 2 against a local build of this head (`pnpm install`, `pnpm --filter @gh-symphony/core build`, `pnpm --filter @gh-symphony/cli build`) rather than reading the diff alone.

### Round-2 findings — both closed

**P2, collision reason double-verbed** — fixed. The reason is now the complement phrase `duplicates …`, which composes correctly with the `${path} is ignored: ${reason}.` frame:

```
agent.max_concurrent_agents_by_state.Ready is ignored: duplicates agent.max_concurrent_agents_by_state.ready after state-name normalization.
```

Confirmed identical in `--json` under `warnings[].summary`. The new assertion in `priority-diagnostics.test.ts` pins the fully composed summary, which is the coverage gap that let this through twice — that guard now exists for the next reason string anyone adds.

**P3, bracket path too narrow** — fixed. `formatWorkflowMapEntryPath` now brackets on `entryKey !== entryKey.trim()`, so the documented padded idiom is locatable, including when embedded in a collision reason:

```
agent.max_concurrent_agents_by_state[" Padded Bad "] is ignored: must be greater than zero.
agent.max_concurrent_agents_by_state.ready is ignored: duplicates agent.max_concurrent_agents_by_state[" READY "] after state-name normalization.
```

### Smoke test

Ran the built CLI against a fixture exercising all five paths at once (valid, colliding, blank-normalized, padded non-positive, fractional). All five warnings render grammatically with locatable paths in both text and `--json`; valid entries (`in progress: 2`) stay effective, and validation still exits `pass` since the warnings are non-fatal. `doctor` consumes the same builder behind a `workflow.status === "pass"` guard; duplicate `state_concurrency` check ids are harmless because the remediation map keys only `status === "fail"`.

### Requirements vs #767

- Actionable warnings with entry paths and reasons — yes, all four reason classes carry a path and a remediation.
- Valid entries preserved in the same map — yes; the runtime filter is byte-for-byte the prior predicate, split in two only to attribute distinct reasons.
- Parser/doctor coverage — yes, plus the composed-summary assertions.
- Revision hash — `diagnostics` destructured out of `calculateWorkflowVersionHash`, so derived output no longer shifts existing workflows' `revision`.

### Scope

No over-engineering. Collision detection and the `code` discriminator both came from review threads rather than the author widening scope on their own; each is a handful of lines. Docs and the changeset match the behavior shipped.

### Verification

`pnpm --filter @gh-symphony/core test` (309 passed) and `pnpm --filter @gh-symphony/cli test` (609 passed) green locally; CI `Test` and `Container Smoke` green on this head.

One non-blocking nit inline about the chain-collision "winner" naming, and one stale `gh-symphony repo doctor` left in `docs/adr/2026-08-29_tracker-provider-alias-deprecation.md:21` — an ADR is a historical record, so leaving it is the right call for this PR. Neither holds the merge.

---

hojinzs (approved):
## Review 3/6 — Approve

Verified on `8c699ee` by rebuilding and running the CLI, not by reading the replies.

### Both round-2 findings fixed

| Finding | Fix | Verified |
| --- | --- | --- |
| P2 collision reason double-verbed | `is overridden by` → `duplicates` | ✅ |
| P3 bracket guard too narrow | `trim().length === 0` → `entryKey !== entryKey.trim()` | ✅ |

All five warning shapes now render correctly, including the two that were broken:

```
…by_state.Ready is ignored: must be greater than zero.
…by_state.Review is ignored: must be a positive YAML integer.
…by_state["  "] is ignored: state name is blank after normalization.
…by_state[" Padded Bad "] is ignored: must be greater than zero.
…by_state.Todo is ignored: duplicates …by_state[" TODO "] after state-name normalization.
```

The padded key is bracketed both as a subject and where it is interpolated into a collision reason — that second position was the one that read as an ordinary word break before.

**Merge conflict resolved.** Branch rebased onto current `main` (`4788d30`); `main` is now an ancestor, `git merge-tree` reports no conflict, and GitHub shows `mergeable_state: clean`.

### Behavior unchanged where it should be

- Effective limits on the full fixture: `{"todo": 9, "in progress": 2}` — last-entry-wins preserved, the valid padded entry still normalizes and survives.
- A clean config (`" In Progress ": 2`, `Todo: 1`) produces `ok: true` with **zero** warnings — no false positives on the documented padded form.
- Warnings remain non-fatal throughout.

### Checks

Local on this head: `pnpm test` **609 CLI tests + all package suites pass**, `pnpm lint` clean, `pnpm typecheck` clean, `pnpm build` clean. CI green (`Test`, `Container Smoke`).

One note so it does not surprise anyone: `pnpm format` fails repo-wide with 86 files, but that reproduces identically on `origin/main` (it flags `pnpm-lock.yaml` and `prettier.config.mjs` among others). This PR's own changed files are all Prettier-clean. Pre-existing, not yours — possibly worth a separate issue.

### Verdict

Requirements from #767 are met in full, the scope never widened beyond what review asked for, and every finding across three rounds is either fixed or explained. Ships as a correctly-scoped `@gh-symphony/cli` patch.

The one item deliberately left out: the fourth `gh-symphony repo doctor` at `docs/adr/2026-08-29_tracker-provider-alias-deprecation.md:21`, which arrived from `main` via #764 and is outside this diff. I confirmed it is still there on `4788d30`. A follow-up is the right home for it — not a reason to hold this.

---

PR #775, opened 2026-08-30
feat(tools,worker): finalize host tool contracts

### LINKED ISSUE #673: feat(tools,worker): pass the normalized issue (id, nativeRef) as tool execution context, remove the dead dynamic-tool path, document per-tool contracts
Epic: #651
리포트 항목: **A13 구현 — 2단계, E2, A16(툴)** — `docs/reports/2026-08-28-upstream-spec-drift-research.md` (PR #650)

## 결정 반영 (2026-08-28)

#671에서 **(a) host-side 실행, 2단계** 확정. 이 이슈는 **2단계**: 툴을 호스트가 실행하고 정규화 issue 컨텍스트를 넘기며, dead dynamic-tool 경로를 실제 프로토콜로 교체한다.

## 병합 게이트 (2026-08-28, 트랜지션 전략)

- 이 이슈(Phase 2)는 **Codex·Claude 두 런타임 모두**에서 host-side 툴 실행이 완료돼야 한다. 한 런타임만 완료된 상태로 병합하지 않는다(다른 런타임 워커가 툴을 잃음).
- 1b(#700)는 이 이슈와 **같은 릴리스**에서 병합한다: 순서는 #673 → #700(직후) 또는 동시. #700 없이 #673만 병합되는 것은 허용(툴이 host-side와 MCP 양쪽에 잠시 공존), 반대는 금지.
- 리뷰 편의를 위해 런타임별 PR 분할은 가능하나, 첫 PR 병합 전에 두 번째 PR이 approve 상태여야 한다.

## 문제

§10.5: 런타임은 툴 호출을 **host-side**로 실행하고(MUST), 정규화된 이슈(`issue.id`, `native_ref`)를 실행 컨텍스트로 전달하며(SHOULD), 각 어댑터는 툴 이름/스키마/변경 가능성/scope/오류·rate-limit 의미론을 문서화한다(MUST). 현재 툴은 에이전트가 spawn하는 MCP subprocess이고 `GITHUB_PROJECT_ID`만 받으며, 워커의 dynamic-tool 경로는 dead code다.

## 현재 상태 (증거)

- 툴 컨텍스트: `tool-github-graphql/src/mcp-entry.ts:34-65`(`GITHUB_PROJECT_ID`만), `tool-linear-graphql/src/mcp-entry.ts:22-28`; `LINEAR_ISSUE_ID`는 allowlist에서 걸러짐 `runtime.ts:536-546`
- dead 경로: `dynamic_tool_call_request/response`는 Codex v2 메서드가 아님(`runtime.ts:131`, `index.ts:1255-1276`); `runToolProcess`(`index.ts:2176-2219`)가 raw JSON을 JSON-RPC 기대 MCP 서버 stdin에 써서 항상 `"{}"`
- 툴 문서: `packages/tracker-github/README.md`(3줄), `packages/tracker-linear/README.md`(7줄), `README.md:726` 한 줄

## 할 일

- **Codex**: `thread/start`에서 dynamic tool spec advertise(`config.mcp_servers` 대신), `item/tool/call` 요청을 워커가 받아 `tool-*-graphql`의 `executeTool()`을 in-process 호출, 결과를 JSON-RPC 응답으로 반환; 미지원 툴 이름은 구조화 실패 + 세션 계속
- **Claude**: 워커가 MCP 서버를 host-side로 띄우고 HTTP/SSE transport로 노출, `mcp.json`에는 URL + 워커 발급 세션 토큰만 기록(세션 종료 시 폐기)
- 툴 실행 컨텍스트에 `issue.id`, `issue.identifier`, `nativeRef`(#665) 전달 — 어댑터가 provider 식별자로 scope 제한(§15.5 narrowing) 가능하게
- dead `dynamic_tool_call_*` 경로와 `runToolProcess` 삭제
- `docs/trackers/<kind>.md`(#674) 또는 툴 패키지 README에 §10.5 목록 문서화: 툴 이름·입력 스키마, mutation 가능 여부, scope 제한, 결과/오류 형태, rate-limit 기대치
- 툴 spec은 워커 시작 시 스냅샷에 고정(리로드가 in-flight 세션에 영향 없음 — 현재 동작 유지, 테스트로 고정)

## 완료 기준

- §17.5: 툴이 host-side에서 어댑터 auth로 실행되고 child env/파일에 토큰이 없으며, issue 컨텍스트가 툴에 도달하는 테스트(Codex, Claude)
- 미지원 툴 이름 → 구조화 실패 + 세션 계속 테스트
- Docker E2E: Codex/Claude 워커가 트래커 상태 전이·코멘트를 수행(#634 시나리오)
- 문서 병합

Blocked by: #665, #671

moncher-dev:
## Issues

- Closed #673

## TL;DR

- Completes integrated verification of Codex and Claude host-side tracker tools.
- Preserves normalized `{ id, identifier, nativeRef }` context at the host adapter boundary without exposing provider credentials to child processes, generated MCP configuration, or provider GraphQL payloads.
- Freezes Claude tool specifications and execution bindings per worker session and documents each provider tool contract.

## Change-point diagram

```text
normalized issue { id, identifier, nativeRef }
                  |
                  v
worker host ------+------ Codex item/tool/call -------+
                  |                                    |
                  +------ Claude loopback HTTP MCP ----+--> snapshotted adapter executeTool()
                                                             |
                                                             +--> host auth + provider API
```

## Start here

1. Review the Claude session snapshot in `packages/runtime-claude/src/mcp-http-server.ts` and its reload regression test.
2. Review normalized issue-context and provider-payload assertions in the tool and worker tests.
3. Review the cross-runtime black-box assertions in `e2e/host-dynamic-tool-e2e.mjs` and `test/e2e/claude/claude-docker.spec.ts`.
4. Review the per-tool contracts in `docs/trackers/github.md` and `docs/trackers/linear.md`.

## Changes

- Deep-copy Claude tool specs at server startup and bind execution to the same immutable adapter/name snapshot.
- Verify normalized issue context reaches GitHub and Linear host adapters on Codex and Claude paths while remaining absent from provider payloads.
- Preserve structured unknown-tool failures without ending the worker session.
- Verify query, comment, and Project-state calls use host credentials while child environments and MCP files remain credential-free.
- Document tool schemas, mutation capability, caller-owned scope narrowing, result/error forms, and rate-limit behavior consistently across tracker and architecture docs.
- Add the required `@gh-symphony/cli` patch changeset.

## Evidence

- `pnpm lint && pnpm test && pnpm typecheck && pnpm build` — pass on head `49eae3fe`.
- `pnpm exec vitest run packages/tool-github-graphql/src/tool.test.ts packages/tool-linear-graphql/src/tool.test.ts` — pass (2 files, 41 tests), including host-internal context payload assertions.
- GitHub Actions `Test` and `Container Smoke` — pass on head `49eae3fe`.
- Codex TC-19: `docker compose -f docker-compose.e2e.yml exec -T symphony-e2e node /app/e2e/host-dynamic-tool-e2e.mjs` — pass (`host_dynamic_tool_e2e=pass`).
- Claude TC-19: `pnpm e2e:claude` — pass (5/5).
- `git merge-base --is-ancestor origin/main HEAD` — pass.

## Risks & rollback

- Arbitrary GitHub and Linear GraphQL documents are not rewritten or target-authorized by these transports; callers must keep operations narrow as documented.
- An in-flight Claude session intentionally retains the tool contract captured at server startup, even after workflow reload.
- Rollback by reverting this PR to restore the prior Claude tool-spec lookup and test/document state.

## Changed files

- Claude session snapshot: `packages/runtime-claude/src/mcp-http-server.ts` and `mcp-http-server.test.ts`.
- Context and behavior coverage: worker, tracker adapter, tool package, and Claude Docker tests.
- Living docs: `docs/architecture.md`, `docs/trackers/github.md`, and `docs/trackers/linear.md`.
- Release: `.changeset/host-scoped-tools.md`.

- [ ] Validate one live GitHub tracker worker with provider credentials.
- [ ] Validate one live Linear tracker worker with provider credentials.
- [ ] Confirm no raw tracker credential is visible in a child environment or generated MCP configuration.

## Human Validation

- [ ] Confirm reviewer-visible host tool behavior matches issue #673.
- [ ] Confirm the coordinated #700 release ordering remains intact.


packages/tool-github-graphql/src/tool.ts
```diff
@@ -152,6 +153,38 @@ function assertTrackerToolExecutionContext(
...
+  context: TrackerToolExecutionContext | undefined
+): void {
+  if (!context) return;
+  const allowedTargets = collectScopeValues({
+    id: context.issue.id,
+    identifier: context.issue.identifier,
+    nativeRef: context.issue.nativeRef,
+  });
+  const invocationValues = collectScopeValues(invocation.variables ?? {});
+  if (![...invocationValues].some((value) => allowedTargets.has(value))) {
```
hojinzs: **P1 — the active-issue guard does not actually constrain the operation target (3 confirmed bypasses)**

I ran these against this branch's real code with a `nativeRef` copied verbatim from `packages/tracker-github/src/adapter.ts:500`. All three reached `fetch` with the host credential; two of them carried a mutation against a **different** issue node ID:

```
✓ BYPASS A: a low-entropy nativeRef constant authorizes another issue
✓ BYPASS B: an unused decoy variable authorizes an inlined foreign target
✓ BYPASS C: the issue state string is enough to authorize any call
```

**A.** `nativeRef.contentType` is the constant `"Issue"`. So `variables: { subjectId: "I_kwDOVICTIM", body: "pwned", t: "Issue" }` passes, and `addComment` executes against `I_kwDOVICTIM`.

**C.** `nativeRef.sourceState` is `"OPEN"`/`"CLOSED"`. `variables: { state: "OPEN" }` authorizes any document.

**B is the structural one, and it survives the fix suggested in the Codex thread above.** This guard only inspects `invocation.variables` — never the query text. An agent can therefore park the real active-issue ID in a variable it never references and inline the foreign target as a GraphQL literal:

```ts
{
  query: 'mutation Comment($scope: String) { addComment(input: { subjectId: "I_kwDOVICTIM", body: "pwned" }) { clientMutationId } }',
  variables: { scope: "I_kwDOACTIVE" },   // never used by the document
}
```

Confirmed executing, with `body.query` still containing `I_kwDOVICTIM`.

The root cause is the `.some()` semantics on line 167: the predicate asserts only *"some string in the variables bag also appears somewhere in the issue context"*. It never establishes that the operation **acts on** that identifier. Whitelisting identity fields (per the Codex thread) closes A and C, but B stays open as long as the check is variable-presence rather than target-binding.

Two viable directions, both larger than a patch to `collectScopeValues`:

1. **Bind the target properly** — parse the document (GitHub already does, in `instrumentGitHubGraphQLInvocation`), resolve each mutation's target argument through the variable map, reject inline literals in target positions, and require *that resolved value* to be an adapter-owned identity field.
2. **Drop the guard and be honest in the docs** — keep the pre-existing "callers must constrain their own document" contract rather than shipping a control that reads as a security boundary but isn't one.

What I'd push back on is option 3, keeping this as-is: `README.md` and both tracker docs now describe it as a hard pre-auth boundary, so a reader will budget trust against it that the code doesn't earn. Note that #673 asked only that the context be *passed* so the adapter **can** narrow (`어댑터가 provider 식별자로 scope 제한(§15.5 narrowing) 가능하게`) — enforcement is an addition this PR chose, so dropping it back to option 2 is fully in scope for the issue.

---

packages/tool-linear-graphql/src/tool.ts
```diff
@@ -73,6 +74,38 @@ function assertTrackerToolExecutionContext(
...
+  if (typeof value === "string" && value.trim() !== "") {
+    output.add(value);
+  } else if (Array.isArray(value)) {
+    for (const entry of value) collectScopeValues(entry, output);
+  } else if (value && typeof value === "object") {
+    for (const entry of Object.values(value)) collectScopeValues(entry, output);
+  }
+  return output;
+}
+
```
hojinzs: **P1 (same defect) / P3 (duplication)** — this is a byte-identical copy of the GitHub helper, so all three bypasses in my `tool-github-graphql/src/tool.ts` comment apply here too. Linear's `nativeRef` is `{ itemId, projectSlug }` (`packages/tracker-linear/src/orchestrator-adapter.ts:692`), so `variables: { x: "<projectSlug>" }` authorizes any operation in the workspace, and the unused-decoy-variable bypass works unchanged.

Whatever fix you land, please land it once. `assertInvocationTargetsActiveIssue` + `collectScopeValues` are now duplicated verbatim across the two tool packages; a fix applied to one and missed in the other is exactly the failure mode this shape invites. Both packages already depend on `@gh-symphony/core`, which is the natural home for a shared scope-check contract.

---

packages/tool-github-graphql/src/tool.test.ts
```diff
@@ -300,6 +300,26 @@ describe("resolveGitHubGraphQLToken", () => {
...
+        {
+          issue: {
+            id: "issue-1",
+            identifier: "owner/repo#1",
+            nativeRef: { itemId: "project-item-1" },
+          },
+        }
+      )
+    ).rejects.toThrow(/active issue scope/);
+    expect(fetchImpl).not.toHaveBeenCalled();
```
hojinzs: **P2 — the new tests only cover the case that cannot fail**

Both new scope tests (here and the Linear twin) pass a query with **no `variables` at all**. With an empty variables bag `invocationValues` is empty, so `.some()` is vacuously false and the throw is guaranteed — this asserts "a call with zero variables is rejected", not "a call targeting another issue is rejected".

Note also that `nativeRef` here is the hand-written `{ itemId: "project-item-1" }` rather than what `tracker-github` actually produces (`{ itemId, contentType, sourceState, linkedPullRequests, linkedPullRequestsTruncated }`). The fields that make the guard bypassable are precisely the ones the fixture omits, which is why the suite is green.

The regression tests this needs, none of which exist today:
- a foreign `subjectId` plus a decoy variable carrying the active issue ID → must reject
- a foreign target inlined as a GraphQL literal with an unused scope variable → must reject
- `variables: { anything: "Issue" }` / `{ anything: "OPEN" }` against a realistic `nativeRef` → must reject
- a legitimately scoped mutation on the active issue → must still pass

Worth flagging for the record: `e2e/host-dynamic-tool-e2e.mjs` and `test/e2e/stubs/claude.sh` had their `viewer { login }` probes rewritten to `node(id:)` in this PR *because the new guard rejects them*. That change is correct on its own terms, but it means the E2E now demonstrates the happy path only — it no longer exercises anything the guard is supposed to stop. The green Codex/Claude Docker runs cited in the PR description shouldn't be read as evidence the boundary holds.

---

docs/trackers/github.md
```diff
@@ -6,15 +6,15 @@ the result only; the adapter uses the host's GitHub credential or token broker.
...
+| Errors      | Invalid tool arguments, missing active-issue scope, missing host authentication, HTTP failures, and GraphQL errors are returned to the runtime as structured tool failures. Unknown tool names are rejected.                                                                                                                                                                                                                                                                                                   |
+| Rate limits | GitHub GraphQL rate-limit headers and the GraphQL `rateLimit` field are measured by the host and applied to the shared GitHub rate-limit policy; callers should keep queries small and respect retry guidance.                                                                                                                                                                                                                                                                                                 |
```
hojinzs: **P2 — the documented contract is stronger than the implemented one**

> the adapter requires at least one exact active issue ID, identifier, or **provider-native ID** in `variables`

`collectScopeValues` doesn't collect IDs — it recursively collects *every non-empty string* reachable in `{ id, identifier, nativeRef }`. For a GitHub project item that set includes `contentType: "Issue"`, `sourceState: "OPEN"`, and every string nested inside `linkedPullRequests` (branch names, URLs, titles). None of those are identifiers, and any one of them satisfies the check.

The same sentence is in `docs/trackers/linear.md`, and `README.md` picked up "rejects tracker-tool calls that do not carry an exact active-issue/provider identifier" in the security section — the one place a reader is most likely to treat the claim as load-bearing.

Please make the prose track the code. If the guard becomes a real target-binding check, this wording is roughly right; if it stays presence-based, it should say so plainly (something like "requires that a value from the active issue context appear in `variables`; this is a guard against unscoped calls, not a security boundary — the document's target is not verified").

Same note applies to the "Safe usage" edit below: *"pass the active issue ID or provider-native Project item ID as an exact variable value"* describes the intended discipline, not what is enforced.

---

packages/runtime-claude/src/mcp-http-server.ts
```diff
@@ -20,8 +21,15 @@ export async function startClaudeMcpHttpServer(options: {
...
   onEvent?: (event: "started" | "stopped") => void;
+  adapters?: readonly Pick<
+    OrchestratorTrackerAdapter,
+    "agentToolSpecs" | "executeAgentTool"
+  >[];
 }): Promise<ClaudeMcpHttpServer> {
   const sessionToken = randomBytes(32).toString("base64url");
+  const toolSnapshot = createToolSnapshot(
+    options.adapters ?? resolveHostToolAdapters(options.env)
+  );
```
hojinzs, line 32: **P2 — `github_graphql` becomes unusable on Linear runs** (confirming the Codex thread on `tool.ts:167`)

`resolveHostToolAdapters` (line 168) returns `[githubProjectTrackerAdapter, linearTrackerAdapter]` when `SYMPHONY_TRACKER_KIND === "linear"`, and `dispatch` passes the single active-issue `context` to whichever adapter matched. On a Linear run that context is the Linear issue: `{ id, identifier: "ENG-1", nativeRef: { itemId, projectSlug } }`.

So a legitimate GitHub call on a Linear-tracked project — variables carrying a repo owner/name or a PR node ID — has no possible match in the allowed set and is rejected before authentication. The tool is advertised but every real use of it fails.

That directly contradicts `docs/trackers/github.md`, which this PR leaves in place:

> The GitHub tool is always advertised because GitHub repository and pull-request operations remain available for Linear-tracked projects

Three ways out, in my order of preference:
1. Widen the execution context to carry the repository/PR scope alongside the tracker issue, so cross-provider calls have something to match (keeps the documented capability).
2. Give each adapter its own scope set rather than checking every tool against the tracker's active issue — the GitHub tool should be scoped by GitHub identity, not Linear identity.
3. Stop advertising `github_graphql` on Linear runs and update the doc to match.

Whichever you pick, this needs a test at the `mcp-http-server` level: a Linear-kind server, a `github_graphql` call with realistic repo variables, asserting the intended outcome. There's currently no coverage of the Linear-run cross-provider path at all, which is why CI is green on a tool that can't work.

---

Separately, and much smaller (**nit**): the tool-spec freeze itself looks right — `structuredClone` per spec plus binding execution to the same snapshot entry closes the reload-vs-in-flight gap that #673 asked to pin, and the new test covers it. Adding the `adapters?` injection seam to make that testable is a reasonable trade. No concerns here.

---

hojinzs (changes requested):
## Review — round 1 of 6

I verified this branch locally at `fab57d9`. **`pnpm lint`, `pnpm test`, `pnpm typecheck`, `pnpm build` all pass**, and both CI checks (Test, Container Smoke) are green — the Validation section of the description is accurate. The failure I hit first was just a build-order artifact on a cold checkout, not a real one.

Requesting changes on one issue: **the active-issue scope guard is not the boundary this PR documents it as.**

### Smoke test

Rather than rely on reading, I exercised `executeGitHubGraphQL` directly with a `nativeRef` copied verbatim from what `tracker-github` actually produces. Three bypasses, all confirmed executing:

```
✓ BYPASS A: a low-entropy nativeRef constant authorizes another issue
✓ BYPASS B: an unused decoy variable authorizes an inlined foreign target
✓ BYPASS C: the issue state string is enough to authorize any call
```

Two of these carried an `addComment` mutation against a **foreign issue node ID** all the way to `fetch` with the host credential. Details and suggested fixes are in the inline comment on `packages/tool-github-graphql/src/tool.ts`.

Bypass B is the one I'd focus on: the guard reads only `invocation.variables` and never the query text, so an agent can park the real issue ID in an unused variable and inline the actual target as a GraphQL literal. **That survives the identity-field whitelist suggested in the Codex thread** — it's a target-binding problem, not a field-selection problem.

### Requirement traceability vs #673

Verified present:
- Normalized `{ id, identifier, nativeRef }` reaches the tools on both runtimes ✅
- Unsupported tool name → structured failure, session continues (`codex-dynamic-tools.test.ts:68`) ✅
- Dead `dynamic_tool_call_*` / `runToolProcess` paths gone — no remaining references ✅
- Tool specs pinned per session, reload doesn't affect in-flight ✅ (clean implementation, good test)
- §10.5 per-tool contracts documented ✅

Gap: the Docker E2E cited as evidence had its `viewer { login }` probes rewritten to scoped `node(id:)` calls *because the new guard rejects them*. Correct in itself, but it means the E2E now covers the happy path only.

### Scope / over-engineering

The scope guard is an **addition beyond what #673 asked**. The issue asks that context be passed so the adapter *can* narrow (`어댑터가 provider 식별자로 scope 제한(§15.5 narrowing) 가능하게`) — not that enforcement ship here. That matters for how you resolve this: **dropping the guard and restoring the previous "callers must constrain their own document" doc wording is fully in scope for the issue** and is a legitimate way to close this review. What I'd avoid is keeping it as-is, because `README.md`'s security section and both tracker docs now describe a hard pre-auth boundary, and a reader will budget trust against it that the code doesn't earn.

The guard is also simultaneously too weak and too strong: it's bypassable (P1) *and* it breaks the documented cross-provider capability, since `github_graphql` is advertised on Linear runs but can never match a Linear issue context (P2).

### Prior review

The Codex review on this same commit raised both a P1 and a P2, and neither has a fix or a reply on the thread. My review confirms both independently — P1 with a working reproduction, and P1 is somewhat worse than described. Please respond on those threads too, so the record shows the decision.

### Findings

| # | Sev | Where | Issue |
|---|-----|-------|-------|
| 1 | **P1** | `tool-github-graphql/src/tool.ts:156-186` | Scope guard doesn't bind the operation target — 3 confirmed bypasses |
| 2 | **P1** | `tool-linear-graphql/src/tool.ts:77-108` | Identical copy, same defect; `projectSlug` authorizes the workspace |
| 3 | P2 | `mcp-http-server.ts` | `github_graphql` advertised but unusable on Linear runs |
| 4 | P2 | `docs/trackers/*.md`, `README.md` | Docs claim "exact ID" enforcement the code doesn't implement |
| 5 | P2 | `tool.test.ts` (both) | New tests only cover the zero-variable case, which cannot fail |
| 6 | P3 | both tool packages | `assertInvocationTargetsActiveIssue`/`collectScopeValues` duplicated verbatim |
| 7 | nit | both tool packages | `if (!context) return` makes the check silently opt-in for context-less callers |

The tool-spec freeze — the part the PR title is actually about — is solid and I have no concerns with it. Everything above is about the scope guard that came along with it.

Either of the two paths in finding #1 (bind the target properly, or drop the guard and align the docs) resolves this from my side. The narrow fix — a proper target-binding check plus the regression tests listed inline — is a contained change, so I'd expect this to close in one more round.

---

packages/tool-github-graphql/src/tool.test.ts
moncher-dev: Replaced the vacuous zero-variable rejection tests with regression tests that execute real GitHub repository and Linear workspace queries while realistic normalized issue context is present. Both tests failed against the predicate and pass after its removal, proving that context delivery no longer blocks legitimate non-issue GraphQL operations. The Docker probes were also restored from issue-node-only queries to the original repository/workspace query shape.

packages/tool-github-graphql/src/tool.ts
```diff
@@ -152,6 +153,38 @@ function assertTrackerToolExecutionContext(
...
+function assertInvocationTargetsActiveIssue(
+  invocation: GitHubGraphQLInvocation,
+  context: TrackerToolExecutionContext | undefined
+): void {
+  if (!context) return;
+  const allowedTargets = collectScopeValues({
+    id: context.issue.id,
+    identifier: context.issue.identifier,
+    nativeRef: context.issue.nativeRef,
+  });
```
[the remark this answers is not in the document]
moncher-dev: Resolved by removing the presence-based scope predicate from both GraphQL adapters. It could not bind the document's actual target, so retaining it as an authorization boundary would be unsafe. Normalized `{ id, identifier, nativeRef }` context still reaches the host adapter and remains internal; the contract now states that callers must constrain arbitrary GraphQL documents. Regression tests verify legitimate host calls continue with realistic context.

packages/tool-linear-graphql/src/tool.ts
moncher-dev: Resolved once across both adapters by deleting the duplicated, unsound scope helpers instead of extracting them. The Linear adapter still validates normalized context shape and single-operation GraphQL documents, but no longer treats `projectSlug` or any other opaque `nativeRef` string as authorization evidence. The contract now explicitly assigns document/variable narrowing to the caller.

packages/tool-github-graphql/src/tool.ts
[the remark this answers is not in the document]
moncher-dev: Resolved by removing the tracker-context value-match predicate. `github_graphql` remains advertised and usable for repository and pull-request operations on Linear-tracked runs, while the Linear issue context is still delivered internally to the host adapter. The tracker documentation again describes this cross-provider capability without claiming issue-target enforcement.

moncher-dev: Resolved by removing the presence-based scope predicate from both GraphQL adapters. It could not bind the document's actual target, so retaining it as an authorization boundary would be unsafe. Normalized `{ id, identifier, nativeRef }` context still reaches the host adapter and remains internal; the contract now states that callers must constrain arbitrary GraphQL documents. Regression tests verify legitimate host calls continue with realistic context.

docs/trackers/github.md
moncher-dev: Aligned the prose with the implemented boundary. README no longer claims an exact-identifier rejection control; both tracker contracts say normalized context remains host-internal and that these arbitrary GraphQL transports do not infer or rewrite targets. Safe-usage guidance continues to require callers to narrow documents and mutations deliberately.

packages/runtime-claude/src/mcp-http-server.ts
moncher-dev: Resolved by removing the tracker-context value-match predicate. `github_graphql` remains advertised and usable for repository and pull-request operations on Linear-tracked runs, while the Linear issue context is still delivered internally to the host adapter. The tracker documentation again describes this cross-provider capability without claiming issue-target enforcement.

docs/architecture.md
```diff
@@ -93,6 +93,9 @@ touches a layer, check that its slice (and the linked documents) still holds.
   [Linear](trackers/linear.md), and [file](trackers/file.md). GitHub's
   synthetic `Archived` state is a GitHub-specific implementation choice, not
   normalized Symphony core behavior.
+- Host-side tracker tools: provider adapters own advertised schemas, credentials,
+  and active-issue scope validation. Codex snapshots those schemas in its runtime
```
hojinzs: **P2 — this line still claims the enforcement that `b78e3b2` removed**

> provider adapters own advertised schemas, credentials, and **active-issue scope validation**

This is the last surviving instance of the round-1 finding #4 defect. `README.md` and both tracker contracts were correctly walked back, but this bullet — added by this PR — was not updated when the guard came out.

After the removal, the only thing left in either adapter is `assertTrackerToolExecutionContext` (`tool-github-graphql/src/tool.ts:144`), which checks that `issue.id` / `issue.identifier` are non-empty. That is **context shape validation**, not scope validation: it inspects nothing about the operation's target, and I confirmed a foreign-target document executes fine under an active-issue context.

Suggested wording, which also matches what `docs/trackers/*.md` now say:

```suggestion
- Host-side tracker tools: provider adapters own advertised schemas and credentials,
  and receive normalized active-issue context that stays host-internal. Callers are
  responsible for narrowing documents; adapters do not infer or rewrite a target.
  Codex snapshots those schemas in its runtime
  plan; Claude snapshots them when its loopback Streamable HTTP MCP server starts.
```

The second sentence about snapshotting is accurate as written — I verified both runtimes behave that way.

---

packages/tool-github-graphql/src/tool.test.ts
```diff
@@ -300,6 +300,36 @@ describe("resolveGitHubGraphQLToken", () => {
 });
 
 describe("executeGitHubGraphQL", () => {
+  it("executes a repository query while carrying host-side issue context", async () => {
```
hojinzs, line 303: **P3 — the test's name promises more than its assertions deliver** (same for the Linear twin at `tool-linear-graphql/src/tool.test.ts:57`)

This is a genuine improvement on what it replaced — it uses a realistic `nativeRef` and proves guard removal doesn't block a legitimate call, which was its stated purpose. But `context` is passed and then never asserted on. The two assertions are "resolves to the payload" and "fetch called once", both of which hold identically if the 4th argument is dropped entirely.

The invariant worth pinning is the one the contract states in prose and nothing tests:

> `nativeRef` remains host-internal and is never sent as an extra GraphQL payload field.

I verified it holds today — a repo query under a full GitHub `nativeRef` (including `linkedPullRequests` URLs and branch names) produced a body with keys `query,variables` only, and no context string anywhere in it. Two lines pin it:

```ts
const body = JSON.parse(String(fetchImpl.mock.calls[0][1]!.body));
expect(Object.keys(body).sort()).toEqual(["query"]);
```

Worth having because this is a claim a reader will treat as load-bearing, and it is currently held up by nothing but the absence of code that would break it.

---

packages/runtime-claude/src/mcp-http-server.test.ts
```diff
@@ -13,6 +13,54 @@ afterEach(async () => {
 });
 
 describe("Claude host MCP HTTP server", () => {
+  it("freezes the advertised tool specs when the server starts", async () => {
```
hojinzs, line 16: **nit — the freeze test covers one of the change's three halves**

I have no concerns with the implementation; I exercised all three halves against a live server and every one behaves correctly. Just noting what the suite would catch on a regression:

| Behavior | Covered here |
|---|---|
| `tools/list` ignores a reloaded spec list | ✅ |
| `tools/call` stays bound to the snapshot entry | ❌ |
| `structuredClone` defeats in-place spec mutation | ❌ |

The second is the one I'd add, since it's the half that actually matters at runtime — a reload could otherwise re-point execution mid-session while `tools/list` still looks frozen:

```ts
const call = await fetch(server.url, { /* …tools/call name: "reloaded_tool" */ });
// must be rejected; "snapshotted_tool" must still execute
```

For the third: because the test reassigns `specs` rather than mutating it, it passes with or without `structuredClone` — so the deep copy is currently unverified. Mutating `specs[0]!.name` in place instead would cover both the reassignment and the mutation path in one test.

Non-blocking — the behavior is right, this is only about what a future regression would trip over.

---

hojinzs (approved):

Verified locally at `04025b8`. **`pnpm install && pnpm build && pnpm test && pnpm lint && pnpm typecheck && pnpm format` all pass**, and both CI checks (Test, Container Smoke) are green on this head.

The round-1 P1 is resolved. You took option 2 — dropping the guard and restoring the caller-owned narrowing contract — which is what I said would close this from my side, and you landed it once across both packages rather than fixing one and missing the other.

### Verification

`tool.ts` in both tool packages is now byte-identical to `origin/main` (`git diff origin/main...HEAD` touches neither), and there are no surviving references to `assertInvocationTargetsActiveIssue` / `collectScopeValues` anywhere in the tree. The net PR is now the tool-spec freeze plus tests and docs — 182/19 across 12 files, and every remaining line is in scope for #673.

I re-ran my round-1 bypasses plus the cases the guard used to break, against the built packages:

```
✓ cross-provider GitHub query executes under Linear context
✓ nativeRef itemId absent from payload
✓ payload top-level keys are GraphQL-only    (query,variables)
✓ zero-variable query no longer rejected
✓ Linear query executes with context
✓ Linear nativeRef absent from payload
✓ empty-identity context still rejected
```

And the freeze, end-to-end against a live loopback server:

```
✓ reload (new array) does not affect in-flight session
✓ reloaded tool name is not callable in-flight
✓ snapshotted tool still executable
✓ in-place spec mutation does not leak (structuredClone)
✓ unknown tool -> structured error
✓ session usable after unknown tool
✓ unauthenticated request rejected 401
```

That last block is the part the PR title is actually about, and it is correct on all three axes — spec listing, execution binding, and deep-copy isolation.

### Round-1 findings

| # | Sev | Status |
|---|-----|--------|
| 1 | P1 | ✅ Guard removed; 3 bypasses no longer reachable as an authorization claim |
| 2 | P1 | ✅ Removed in the Linear package too, in the same commit |
| 3 | P2 | ✅ Confirmed by smoke: a GitHub repo query now executes under a Linear issue context |
| 4 | P2 | ⚠️ Mostly — `README.md` and both tracker docs corrected; **`docs/architecture.md:97` still claims "active-issue scope validation"** |
| 5 | P2 | ✅ Vacuous zero-variable tests replaced with executing tests on realistic `nativeRef` (see P3 inline) |
| 6 | P3 | ✅ Resolved by deletion rather than extraction — the right call for code that shouldn't exist |
| 7 | nit | n/a — now outside this PR's diff, since `tool.ts` is untouched |

### Requirement traceability vs #673

All acceptance criteria hold, and guard removal doesn't cost any of them — the issue asked that context be *passed* so the adapter **can** narrow (`어댑터가 provider 식별자로 scope 제한(§15.5 narrowing) 가능하게`), not that enforcement ship here:

- Normalized `{ id, identifier, nativeRef }` reaches tools on both runtimes ✅
- Unsupported tool name → structured failure, session continues ✅ (Codex `codex-dynamic-tools.test.ts:68`; Claude verified by smoke)
- Dead `dynamic_tool_call_*` / `runToolProcess` paths gone ✅
- Tool specs pinned per session, reload doesn't affect in-flight ✅
- §10.5 per-tool contracts documented ✅
- Host credentials absent from child env and generated MCP config ✅

The E2E probes are back to their original repository/workspace query shape, so the Docker evidence once again exercises the real call shape rather than the guard's happy path.

### Over-engineering

Nothing left beyond the issue's ask. The one over-reach — enforcement the issue didn't request — is gone. The `adapters?` injection seam is a fair trade for making the freeze testable, and the `routable`/`routableReason` fixture refresh is legitimate rather than scope creep: `worker/src/turn-lease.ts:69` reads those fields.

### Findings

| # | Sev | Where | Issue |
|---|-----|-------|-------|
| 1 | P2 | `docs/architecture.md:97` | Claims "active-issue scope validation" the code no longer does |
| 2 | P3 | `tool.test.ts` (both) | Test passes `context` but asserts nothing about it; `nativeRef` non-leak untested |
| 3 | nit | `mcp-http-server.test.ts:16` | Freeze test covers `tools/list` only, not execution binding or `structuredClone` |
| 4 | nit | `docs/trackers/github.md`, `worker-protocol.test.ts` | Whitespace/prettier-only churn in the diff |

**Approving.** Finding #1 is a three-word correction and the last echo of the problem that drove round 1 — please land it before merge, along with #2 if it's cheap; neither warrants another review round, and the rest are optional. Requesting changes again over a doc line would be exactly the ping-pong worth avoiding on a PR that is otherwise clean.

Per #673's merge gate, remember #700 ships in this same release, ordered `#673 → #700`.

---

docs/architecture.md
moncher-dev: Resolved in `49eae3fe`. `docs/architecture.md` now uses the suggested boundary: adapters own schemas and credentials, receive host-internal normalized issue context, and leave target narrowing to callers without inferring or rewriting the GraphQL target.

packages/tool-github-graphql/src/tool.test.ts
moncher-dev: Resolved in `49eae3fe` for both GitHub and Linear. Each context-carrying execution test now parses the outbound request body and asserts its only top-level key is `query`, pinning that normalized context and `nativeRef` are not serialized into provider payloads.

packages/runtime-claude/src/mcp-http-server.test.ts
moncher-dev: Acknowledged as non-blocking. I left this test unchanged in the narrow pre-merge correction: the implementation and owner smoke already verify execution binding and in-place mutation isolation, while this cycle is limited to the required architecture fix and the cheap adapter payload assertions. The existing snapshot test continues to pin the advertised-spec reload behavior.

PR #785, opened 2026-09-02
fix(core): attribute non-numeric issue workspaces

### LINKED ISSUE #782: bug(core): dirty-workspace attribution cannot parse non-numeric issue identifiers — every Linear dirty workspace is quarantined
Part of #777 (root-cause chain item 3 — dirty workspace loss).

## Problem

Dirty-workspace attribution (`attributeDirtyWorkToIssue`, `packages/core/src/workflow/issue-identity.ts`) can **never succeed** for tracker identifiers that are not plain numeric (`#123`, `owner/repo#123`, `123`):

- `extractIssueNumberFromIdentifier("DEV-54")` returns `null` — the regex (`issue-identity.ts:50`) does not accept a `TEAM-123`-shaped identifier, so `issueNumber` is `null` for every Linear issue.
- With `issueNumber === null`, **any** number extracted from the branch or a workpad file is classified as foreign (`n !== null` is always true). A branch like `feat/54-fix` — or even the issue's **own workpad** `.gh-symphony/workpads/DEV-54.md` — is treated as another issue's work.
- Numberless branches (`main`, Linear-style `dev-54-fix`) fall through to `no dirty artifact could be attributed`.
- The remaining path, `expectedBranches`, is empty because `tracker-linear` lacks `resolveBranchCheckoutTarget` (tracked separately).

Result: on recovery, every dirty Linear workspace is quarantined (`packages/orchestrator/src/service.ts:2849-2873`) and the next worker restarts from a clean clone — real work is silently discarded on every retry cycle. This is exactly the `quarantined dirty workspace for DEV-54: no dirty artifact could be attributed to DEV-54` line in #777, and it bites **any** worker crash mid-turn, independent of the refresh bug.

## Fix outline

1. Teach identifier parsing to handle `TEAM-123` identifiers (extract both the team key and number, or match on the full normalized identifier rather than a bare number).
2. Make branch/workpad matching identifier-aware: `dev-54-*` / `DEV-54.md` should attribute to `DEV-54`; artifacts referencing a *different* full identifier stay foreign.
3. Keep the fail-closed default: no positive evidence → still quarantine.

## Regression tests

- `DEV-54` + branch `dev-54-fix` → attributed.
- `DEV-54` + dirty `.gh-symphony/workpads/DEV-54.md` → attributed.
- `DEV-54` + branch `eng-12-other` → foreign → quarantined.
- Existing GitHub-identifier cases unchanged.

## Evidence

Repro assertions B1–B5 (plus GitHub control) in the RCA comment on #777.

moncher-dev:
## TL;DR

- Recover dirty Linear workspaces only when trusted branch/workpad evidence identifies the active full `TEAM-123` issue.
- Preserve GitHub numeric behavior and fail closed for ambiguous bare numbers, foreign tracker identifiers, and missing evidence.
- Treat terminal `name-number` branch fragments as non-foreign evidence, avoiding false quarantine of an issue's own workpad while retaining foreign detection for `key-number-slug` branches.

## Change-point diagram

```text
dirty branch/workpad evidence
            |
            v
full tracker token at segment start?
            |
            +-- own token ----------> positive evidence -> reuse
            |
            +-- foreign + slug -----> quarantine
            |
            +-- foreign terminal ---> non-evidence; own workpad may reuse
            |
            +-- no token -----------> legacy GitHub numeric evidence / fail closed
```

## Start here

1. `packages/core/src/workflow/issue-identity.ts` — separates tracker-native positive evidence from foreign branch evidence.
2. `packages/core/src/workflow/issue-identity.test.ts` — Linear, GitHub, ambiguous-number, foreign-token, terminal-token, and `GO-42` regressions.
3. `packages/worker/src/identity-preflight.test.ts` — preserves fail-closed worker preflight behavior.
4. `e2e/scenarios/21-linear-dirty-workspace-recovery.md` — Docker recovery lifecycle coverage.

## Changes

- Attribute `DEV-54` from `dev-54-*` or `DEV-54.md`, including digit-bearing team keys such as `DEV2-54`.
- Reject foreign full tracker evidence for both Linear and GitHub runs, but do not let a segment-terminal version-like token override an own workpad.
- Keep bare `54` branch/workpad artifacts ambiguous and non-positive for Linear runs, while retaining GitHub numeric matching.
- Remove the fixed version-key allowlist so valid team keys such as `GO-42` remain usable.

## Evidence

- Focused RED: three new structural-evidence regressions failed on `5757091`.
- Focused GREEN: `pnpm --filter @gh-symphony/core test -- issue-identity.test.ts` (23 tests) and `pnpm --filter @gh-symphony/worker test -- identity-preflight.test.ts` (8 tests) passed.
- `pnpm lint`, `pnpm test`, `pnpm typecheck`, and `pnpm build` passed.
- Docker black-box recovery lifecycle passed with a valid preloaded fixture: preserved workspace verified and no quarantine event emitted.
- Changeset: `.changeset/clean-lions-recover.md` — `@gh-symphony/cli` patch for #782.

## Risks & rollback

- Risk: a terminal `key-number` branch component is intentionally non-foreign because it is indistinguishable from a version fragment; without other positive evidence, recovery still fails closed.
- Rollback: revert the #782 commits and its changeset; recovery returns to the prior fail-closed behavior.

## Changed files

- Core behavior/tests: `packages/core/src/workflow/issue-identity.ts`, `packages/core/src/workflow/issue-identity.test.ts`
- Worker regression: `packages/worker/src/identity-preflight.test.ts`
- Docker TC and docs: `e2e/*linear-dirty-recovery*`, `AGENT_TEST.md`, `docs/architecture.md`
- Release: `.changeset/clean-lions-recover.md`

- [ ] Confirm a live Linear recovery reuses a dirty workspace identified by `dev-54-fix` or `DEV-54.md`.
- [ ] Confirm foreign `eng-12-other` and ambiguous bare `54` artifacts remain quarantined for `DEV-54`.
- [ ] Confirm terminal version-like branch tokens do not override an issue-owned workpad.

### DIFF
--- a/packages/core/src/workflow/issue-identity.ts
+++ b/packages/core/src/workflow/issue-identity.ts
@@ -73,14 +73,6 @@ export function extractIssueNumbersFromBranch(branch: string): number[] {
 
 const WORKPAD_FILE_PATTERN = /(?:^|\/)\.gh-symphony\/workpads\/([^/]+)\.md$/;
 const TRACKER_IDENTIFIER_PATTERN = /^[A-Za-z][A-Za-z0-9]*-\d{1,9}$/;
-const VERSION_LIKE_TRACKER_KEYS = new Set([
-  "GO",
-  "JAVA",
-  "NODE",
-  "PHP",
-  "PYTHON",
-  "RUBY",
-]);
 
 function normalizeTrackerIdentifier(identifier: string): string | null {
   const normalized = identifier.trim();
@@ -89,23 +81,31 @@ function normalizeTrackerIdentifier(identifier: string): string | null {
     : null;
 }
 
-function isVersionLikeTrackerIdentifier(identifier: string): boolean {
-  const teamKey = identifier.split("-", 1)[0]!;
-  return VERSION_LIKE_TRACKER_KEYS.has(teamKey) || /^V\d+$/.test(teamKey);
-}
-
-function extractTrackerIdentifiersFromBranch(branch: string): string[] {
-  const identifiers = new Set<string>();
+/**
+ * Tracker identifiers encoded in a branch name.
+ *
+ * Positive evidence accepts any `key-number` token at a segment start.
+ * Foreign evidence additionally requires a slug after the token because a
+ * segment-terminal token is indistinguishable from a version fragment and
+ * must never outrank the issue's own evidence.
+ */
+function extractTrackerIdentifiersFromBranch(branch: string): {
+  positive: string[];
+  foreign: string[];
+} {
+  const positive = new Set<string>();
+  const foreign = new Set<string>();
   for (const match of branch.matchAll(
-    /(?:^|\/)([A-Za-z][A-Za-z0-9]*-\d{1,9})(?=[-_/]|$)/g
+    /(?:^|\/)([A-Za-z][A-Za-z0-9]*-\d{1,9})(?=([-_])|\/|$)/g
   )) {
     const identifier = match[1]!.toUpperCase();
-    if (!isVersionLikeTrackerIdentifier(identifier)) {
-      identifiers.add(identifier);
+    positive.add(identifier);
+    if (match[2]) {
+      foreign.add(identifier);
     }
   }
 
-  return [...identifiers];
+  return { positive: [...positive], foreign: [...foreign] };
 }
 
 function extractTrackerIdentifiersFromWorkpadFiles(
@@ -176,12 +176,12 @@ export function attributeDirtyWorkToIssue(
   const workpadNumbers = extractIssueNumbersFromWorkpadFiles(input.dirtyFiles);
   const branchTrackerIdentifiers = branch
     ? extractTrackerIdentifiersFromBranch(branch)
-    : [];
+    : { positive: [], foreign: [] };
   const workpadTrackerIdentifiers = extractTrackerIdentifiersFromWorkpadFiles(
     input.dirtyFiles
   );
 
-  const foreignBranchIdentifiers = branchTrackerIdentifiers.filter(
+  const foreignBranchIdentifiers = branchTrackerIdentifiers.foreign.filter(
     (identifier) => identifier !== trackerIdentifier
   );
   if (foreignBranchIdentifiers.length > 0) {
@@ -232,7 +232,7 @@ export function attributeDirtyWorkToIssue(
 
   if (
     trackerIdentifier &&
-    branchTrackerIdentifiers.includes(trackerIdentifier)
+    branchTrackerIdentifiers.positive.includes(trackerIdentifier)
   ) {
     return {
       attributed: true,
--- a/packages/core/src/workflow/issue-identity.ts
+++ b/packages/core/src/workflow/issue-identity.ts
@@ -41,16 +41,13 @@ export function buildIssueIdentityHeader(
 
 /**
  * Extract the numeric issue number from a canonical issue identifier such as
- * `owner/repo#507`, `#507`, `507`, or `TEAM-507`. Returns null when the
- * identifier is not a supported canonical shape.
+ * `owner/repo#507`, `#507`, or `507`. Tracker-native identifiers such as
+ * `TEAM-507` remain opaque here so legacy numeric consumers stay fail-closed.
  */
 export function extractIssueNumberFromIdentifier(
   identifier: string
 ): number | null {
-  const normalized = identifier.trim();
-  const match =
-    normalized.match(/(?:^|#|\/)(\d{1,9})$/) ??
-    normalized.match(/^[A-Za-z][A-Za-z0-9]*-(\d{1,9})$/);
+  const match = identifier.trim().match(/(?:^|#|\/)(\d{1,9})$/);
   if (!match) {
     return null;
   }
@@ -76,6 +73,14 @@ export function extractIssueNumbersFromBranch(branch: string): number[] {
 
 const WORKPAD_FILE_PATTERN = /(?:^|\/)\.gh-symphony\/workpads\/([^/]+)\.md$/;
 const TRACKER_IDENTIFIER_PATTERN = /^[A-Za-z][A-Za-z0-9]*-\d{1,9}$/;
+const VERSION_LIKE_TRACKER_KEYS = new Set([
+  "GO",
+  "JAVA",
+  "NODE",
+  "PHP",
+  "PYTHON",
+  "RUBY",
+]);
 
 function normalizeTrackerIdentifier(identifier: string): string | null {
   const normalized = identifier.trim();
@@ -84,12 +89,20 @@ function normalizeTrackerIdentifier(identifier: string): string | null {
     : null;
 }
 
+function isVersionLikeTrackerIdentifier(identifier: string): boolean {
+  const teamKey = identifier.split("-", 1)[0]!;
+  return VERSION_LIKE_TRACKER_KEYS.has(teamKey) || /^V\d+$/.test(teamKey);
+}
+
 function extractTrackerIdentifiersFromBranch(branch: string): string[] {
   const identifiers = new Set<string>();
   for (const match of branch.matchAll(
     /(?:^|\/)([A-Za-z][A-Za-z0-9]*-\d{1,9})(?=[-_/]|$)/g
   )) {
-    identifiers.add(match[1]!.toUpperCase());
+    const identifier = match[1]!.toUpperCase();
+    if (!isVersionLikeTrackerIdentifier(identifier)) {
+      identifiers.add(identifier);
+    }
   }
 
   return [...identifiers];
@@ -123,7 +136,7 @@ export function extractIssueNumbersFromWorkpadFiles(
     if (!match) {
       continue;
     }
-    const numeric = match[1]!.match(/(\d{1,9})/);
+    const numeric = match[1]!.match(/^(?:[A-Za-z][A-Za-z0-9]*-)?(\d{1,9})$/);
     if (numeric) {
       numbers.add(Number.parseInt(numeric[1]!, 10));
     }
@@ -168,11 +181,9 @@ export function attributeDirtyWorkToIssue(
     input.dirtyFiles
   );
 
-  const foreignBranchIdentifiers = trackerIdentifier
-    ? branchTrackerIdentifiers.filter(
-        (identifier) => identifier !== trackerIdentifier
-      )
-    : [];
+  const foreignBranchIdentifiers = branchTrackerIdentifiers.filter(
+    (identifier) => identifier !== trackerIdentifier
+  );
   if (foreignBranchIdentifiers.length > 0) {
     return {
       attributed: false,
@@ -180,11 +191,9 @@ export function attributeDirtyWorkToIssue(
     };
   }
 
-  const foreignWorkpadIdentifiers = trackerIdentifier
-    ? workpadTrackerIdentifiers.filter(
-        (identifier) => identifier !== trackerIdentifier
-      )
-    : [];
+  const foreignWorkpadIdentifiers = workpadTrackerIdentifiers.filter(
+    (identifier) => identifier !== trackerIdentifier
+  );
   if (foreignWorkpadIdentifiers.length > 0) {
     return {
       attributed: false,
@@ -192,24 +201,26 @@ export function attributeDirtyWorkToIssue(
     };
   }
 
-  const foreignBranchNumbers = branchNumbers.filter(
-    (number) => number !== issueNumber
-  );
-  if (foreignBranchNumbers.length > 0) {
-    return {
-      attributed: false,
-      reason: `current branch '${branch}' references issue #${foreignBranchNumbers[0]} instead of ${input.issueIdentifier}`,
-    };
-  }
+  if (!trackerIdentifier) {
+    const foreignBranchNumbers = branchNumbers.filter(
+      (number) => number !== issueNumber
+    );
+    if (foreignBranchNumbers.length > 0) {
+      return {
+        attributed: false,
+        reason: `current branch '${branch}' references issue #${foreignBranchNumbers[0]} instead of ${input.issueIdentifier}`,
+      };
+    }
 
-  const foreignWorkpadNumbers = workpadNumbers.filter(
-    (number) => number !== issueNumber
-  );
-  if (foreignWorkpadNumbers.length > 0) {
-    return {
-      attributed: false,
-      reason: `dirty workpad references issue #${foreignWorkpadNumbers[0]} instead of ${input.issueIdentifier}`,
-    };
+    const foreignWorkpadNumbers = workpadNumbers.filter(
+      (number) => number !== issueNumber
+    );
+    if (foreignWorkpadNumbers.length > 0) {
+      return {
+        attributed: false,
+        reason: `dirty workpad references issue #${foreignWorkpadNumbers[0]} instead of ${input.issueIdentifier}`,
+      };
+    }
   }
 
   if (branch && input.expectedBranches?.includes(branch)) {
@@ -239,14 +250,22 @@ export function attributeDirtyWorkToIssue(
     };
   }
 
-  if (issueNumber !== null && branchNumbers.includes(issueNumber)) {
+  if (
+    !trackerIdentifier &&
+    issueNumber !== null &&
+    branchNumbers.includes(issueNumber)
+  ) {
     return {
       attributed: true,
       reason: `current branch '${branch}' references ${input.issueIdentifier}`,
     };
   }
 
-  if (issueNumber !== null && workpadNumbers.includes(issueNumber)) {
+  if (
+    !trackerIdentifier &&
+    issueNumber !== null &&
+    workpadNumbers.includes(issueNumber)
+  ) {
     return {
       attributed: true,
       reason: `dirty workpad belongs to ${input.issueIdentifier}`,
--- a/packages/core/src/workflow/issue-identity.ts
+++ b/packages/core/src/workflow/issue-identity.ts
@@ -41,13 +41,16 @@ export function buildIssueIdentityHeader(
 
 /**
  * Extract the numeric issue number from a canonical issue identifier such as
- * `owner/repo#507`, `#507`, or `507`. Returns null when the identifier does
- * not end in a numeric fragment.
+ * `owner/repo#507`, `#507`, `507`, or `TEAM-507`. Returns null when the
+ * identifier is not a supported canonical shape.
  */
 export function extractIssueNumberFromIdentifier(
   identifier: string
 ): number | null {
-  const match = identifier.trim().match(/(?:^|#|\/)(\d{1,9})$/);
+  const normalized = identifier.trim();
+  const match =
+    normalized.match(/(?:^|#|\/)(\d{1,9})$/) ??
+    normalized.match(/^[A-Za-z][A-Za-z0-9]*-(\d{1,9})$/);
   if (!match) {
     return null;
   }
@@ -72,6 +75,40 @@ export function extractIssueNumbersFromBranch(branch: string): number[] {
 }
 
 const WORKPAD_FILE_PATTERN = /(?:^|\/)\.gh-symphony\/workpads\/([^/]+)\.md$/;
+const TRACKER_IDENTIFIER_PATTERN = /^[A-Za-z][A-Za-z0-9]*-\d{1,9}$/;
+
+function normalizeTrackerIdentifier(identifier: string): string | null {
+  const normalized = identifier.trim();
+  return TRACKER_IDENTIFIER_PATTERN.test(normalized)
+    ? normalized.toUpperCase()
+    : null;
+}
+
+function extractTrackerIdentifiersFromBranch(branch: string): string[] {
+  const identifiers = new Set<string>();
+  for (const match of branch.matchAll(
+    /(?:^|\/)([A-Za-z][A-Za-z0-9]*-\d{1,9})(?=[-_/]|$)/g
+  )) {
+    identifiers.add(match[1]!.toUpperCase());
+  }
+
+  return [...identifiers];
+}
+
+function extractTrackerIdentifiersFromWorkpadFiles(
+  dirtyFiles: string[]
+): string[] {
+  const identifiers = new Set<string>();
+  for (const file of dirtyFiles) {
+    const match = file.match(WORKPAD_FILE_PATTERN);
+    const identifier = match ? normalizeTrackerIdentifier(match[1]!) : null;
+    if (identifier) {
+      identifiers.add(identifier);
+    }
+  }
+
+  return [...identifiers];
+}
 
 /**
  * Extract issue numbers referenced by dirty workpad files
@@ -120,9 +157,40 @@ export function attributeDirtyWorkToIssue(
   input: DirtyWorkAttributionInput
 ): DirtyWorkAttribution {
   const issueNumber = extractIssueNumberFromIdentifier(input.issueIdentifier);
+  const trackerIdentifier = normalizeTrackerIdentifier(input.issueIdentifier);
   const branch = input.currentBranch?.trim() || null;
   const branchNumbers = branch ? extractIssueNumbersFromBranch(branch) : [];
   const workpadNumbers = extractIssueNumbersFromWorkpadFiles(input.dirtyFiles);
+  const branchTrackerIdentifiers = branch
+    ? extractTrackerIdentifiersFromBranch(branch)
+    : [];
+  const workpadTrackerIdentifiers = extractTrackerIdentifiersFromWorkpadFiles(
+    input.dirtyFiles
+  );
+
+  const foreignBranchIdentifiers = trackerIdentifier
+    ? branchTrackerIdentifiers.filter(
+        (identifier) => identifier !== trackerIdentifier
+      )
+    : [];
+  if (foreignBranchIdentifiers.length > 0) {
+    return {
+      attributed: false,
+      reason: `current branch '${branch}' references issue ${foreignBranchIdentifiers[0]} instead of ${input.issueIdentifier}`,
+    };
+  }
+
+  const foreignWorkpadIdentifiers = trackerIdentifier
+    ? workpadTrackerIdentifiers.filter(
+        (identifier) => identifier !== trackerIdentifier
+      )
+    : [];
+  if (foreignWorkpadIdentifiers.length > 0) {
+    return {
+      attributed: false,
+      reason: `dirty workpad references issue ${foreignWorkpadIdentifiers[0]} instead of ${input.issueIdentifier}`,
+    };
+  }
 
   const foreignBranchNumbers = branchNumbers.filter(
     (number) => number !== issueNumber
@@ -148,6 +216,26 @@ export function attributeDirtyWorkToIssue(
     return {
       attributed: true,
       reason: `current branch '${branch}' is tracker-linked to ${input.issueIdentifier}`,
+    };
+  }
+
+  if (
+    trackerIdentifier &&
+    branchTrackerIdentifiers.includes(trackerIdentifier)
+  ) {
+    return {
+      attributed: true,
+      reason: `current branch '${branch}' references ${input.issueIdentifier}`,
+    };
+  }
+
+  if (
+    trackerIdentifier &&
+    workpadTrackerIdentifiers.includes(trackerIdentifier)
+  ) {
+    return {
+      attributed: true,
+      reason: `dirty workpad belongs to ${input.issueIdentifier}`,
     };
   }
 
--- a/.changeset/clean-lions-recover.md
+++ b/.changeset/clean-lions-recover.md
@@ -1,0 +1,5 @@
+---
+"@gh-symphony/cli": patch
+---
+
+Preserve dirty recovery workspaces attributed by normalized `TEAM-123` branch or workpad identifiers for #782.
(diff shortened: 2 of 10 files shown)


packages/core/src/workflow/issue-identity.ts
```diff
@@ -41,13 +41,16 @@ export function buildIssueIdentityHeader(
...
+ * identifier is not a supported canonical shape.
  */
 export function extractIssueNumberFromIdentifier(
   identifier: string
 ): number | null {
-  const match = identifier.trim().match(/(?:^|#|\/)(\d{1,9})$/);
+  const normalized = identifier.trim();
+  const match =
+    normalized.match(/(?:^|#|\/)(\d{1,9})$/) ??
+    normalized.match(/^[A-Za-z][A-Za-z0-9]*-(\d{1,9})$/);
```
hojinzs: **P1 — widening this exported helper silently weakens the worker identity preflight (#507), with no test and no mention in the PR description.**

`extractIssueNumberFromIdentifier` has a second consumer outside this package: `packages/worker/src/identity-preflight.ts:101`, the fail-closed startup guard that refuses to launch an agent in another issue's workspace.

- **Before:** `extractIssueNumberFromIdentifier("DEV-54") === null`, so `foreign = branchNumbers.filter(n => n !== null)` rejected *every* numeric branch for a Linear run.
- **After:** it returns `54`, so a branch whose number merely collides now passes.

Verified against this head:

```
PF DEV-54 branch=feat/54-unrelated -> {"ok":true}                 // was: rejected
PF DEV-54 branch=feat/12-unrelated -> {"ok":false, ...}           // unchanged
```

So a workspace checked out on GitHub issue #54's branch now passes the identity preflight for Linear `DEV-54`. `packages/worker/src/identity-preflight.test.ts` contains no `TEAM-123`-shaped identifier at all, so nothing catches this.

This may well be acceptable (numeric collision across trackers is rare), but it is a change to a fail-closed security guard in another package that the PR neither states nor tests. Please either make the preflight identifier-aware the same way `attributeDirtyWorkToIssue` now is, or add an explicit preflight test asserting the intended Linear behavior and call the loosening out in the PR description.

---

packages/core/src/workflow/issue-identity.ts
```diff
@@ -120,9 +157,40 @@ export function attributeDirtyWorkToIssue(
...
   const branchNumbers = branch ? extractIssueNumbersFromBranch(branch) : [];
   const workpadNumbers = extractIssueNumbersFromWorkpadFiles(input.dirtyFiles);
+  const branchTrackerIdentifiers = branch
+    ? extractTrackerIdentifiersFromBranch(branch)
+    : [];
+  const workpadTrackerIdentifiers = extractTrackerIdentifiersFromWorkpadFiles(
+    input.dirtyFiles
+  );
+
+  const foreignBranchIdentifiers = trackerIdentifier
```
hojinzs: **P2 — the foreign-identifier check is gated on *our own* identifier's shape, so GitHub runs never reject a foreign Linear branch.**

Evidence is foreign or not independently of what shape the run's own identifier happens to have. Gating both filters on `trackerIdentifier &&` means the new protection only exists for Linear-shaped runs. A GitHub run sitting on a `DEV-54` workspace is attributed:

```
issueIdentifier: "acme/platform#173"
currentBranch:   "dev-54-fix"                          // belongs to DEV-54
dirtyFiles:      [".gh-symphony/workpads/173.md"]
-> {"attributed":true,"reason":"dirty workpad belongs to acme/platform#173"}
```

The branch positively identifies another issue, but because `normalizeTrackerIdentifier("acme/platform#173")` is `null` the branch is never examined, and the numeric workpad match then attributes the workspace. That is the mirror image of the `eng-54-other` case this PR adds a test for.

The ternary is what creates the asymmetry — the filters are already correct unconditionally, since no string equals `null`:

```ts
const foreignBranchIdentifiers = branchTrackerIdentifiers.filter(
  (identifier) => identifier !== trackerIdentifier
);
```

Same for `foreignWorkpadIdentifiers`. Worth a regression test with a GitHub identifier and a `TEAM-123` branch. (Please read this together with the P3 on `TRACKER_IDENTIFIER_PATTERN` below — making these filters unconditional widens the blast radius of that pattern's false positives, so the two want fixing together.)

---

packages/core/src/workflow/issue-identity.ts
```diff
@@ -151,6 +219,26 @@ export function attributeDirtyWorkToIssue(
...
+    trackerIdentifier &&
+    workpadTrackerIdentifiers.includes(trackerIdentifier)
+  ) {
+    return {
+      attributed: true,
+      reason: `dirty workpad belongs to ${input.issueIdentifier}`,
+    };
+  }
+
   if (issueNumber !== null && branchNumbers.includes(issueNumber)) {
```
hojinzs: **P2 — bare numeric evidence is now *positive* evidence for a `TEAM-123` run, which contradicts the PR's own stated risk control.**

Because `issueNumber` is no longer `null` for `DEV-54`, these two blocks accept plain numeric artifacts that carry no team key:

```
issueIdentifier: "DEV-54", currentBranch: "feat/54-other"
-> {"attributed":true,"reason":"current branch 'feat/54-other' references DEV-54"}

issueIdentifier: "DEV-54", dirtyFiles: [".gh-symphony/workpads/54.md"]
-> {"attributed":true,"reason":"dirty workpad belongs to DEV-54"}
```

`feat/54-other` and `workpads/54.md` are the conventional shapes for **GitHub issue #54**, not for `DEV-54`. Before this PR both were classified foreign and quarantined; now they are adopted, and the recovery flow is allowed to commit and push that work under `DEV-54`.

The PR description says "foreign full identifiers are checked before positive numeric evidence" — true, but it only covers artifacts that *carry* a foreign team key. It does not cover a bare number, which is exactly the ambiguous case. #782's fix outline asks for `dev-54-*` / `DEV-54.md` to attribute; it does not ask for bare `54` to attribute.

Suggestion: when `trackerIdentifier` is non-null, require full-identifier evidence and stop treating bare numeric matches as positive (a bare number can stay non-evidence rather than becoming foreign, so the fail-closed default still applies). Concretely, guard the two numeric positive blocks with `trackerIdentifier === null`.

---

packages/core/src/workflow/issue-identity.ts
```diff
@@ -72,6 +75,40 @@ export function extractIssueNumbersFromBranch(branch: string): number[] {
 }
 
 const WORKPAD_FILE_PATTERN = /(?:^|\/)\.gh-symphony\/workpads\/([^/]+)\.md$/;
+const TRACKER_IDENTIFIER_PATTERN = /^[A-Za-z][A-Za-z0-9]*-\d{1,9}$/;
```
hojinzs, line 75: **P3 — `[A-Za-z][A-Za-z0-9]*-\d+` also matches version-like branch segments, causing the false quarantine this PR exists to prevent.**

`extractIssueNumbersFromBranch` deliberately ignores version-ish fragments — there is an existing test asserting `extractIssueNumbersFromBranch("chore/upgrade-node-24") === []` with the comment "version-like fragments such as `node-24` are ignored". The new tracker-identifier extractor reintroduces exactly that class of match, and because a foreign identifier short-circuits before any positive evidence, it wins over the issue's own workpad:

```
issueIdentifier: "DEV-54"
currentBranch:   "chore/node-24"
dirtyFiles:      [".gh-symphony/workpads/DEV-54.md"]     // the issue's OWN workpad
-> {"attributed":false,"reason":"current branch 'chore/node-24' references issue NODE-24 instead of DEV-54"}
```

`node-24`, `v2-1`, `python-3` all parse as tracker keys. The result is a quarantined workspace and discarded work — the failure mode #782 is fixing — now triggered by branch naming instead of by identifier shape.

Since the tracker adapter already knows the real team key (`packages/tracker-linear/src/orchestrator-adapter.ts:26`), the robust fix is to match the run's own key rather than any `X-123` token: derive the expected key from `input.issueIdentifier` and only treat a branch segment as a foreign identifier when it looks like a tracker key *and* is not a plausible version fragment. At minimum, only treat an extracted identifier as foreign when its team key differs from the run's team key while still being a known-key shape.

---

hojinzs (changes requested):

Verified by checking out `f59d001`, installing, and running the suites and targeted probes locally. The stated evidence reproduces: `pnpm --filter @gh-symphony/core test -- issue-identity.test.ts` → **16/16 pass**, CI **Test / Container Smoke green**.

### Requirement coverage vs #782 — complete

All four regression tests named in the issue are present and passing, plus the same-number `eng-54-other` case:

| #782 requirement | Status |
| --- | --- |
| `DEV-54` + branch `dev-54-fix` → attributed | ✅ |
| `DEV-54` + workpad `DEV-54.md` → attributed | ✅ |
| `DEV-54` + branch `eng-12-other` → foreign | ✅ |
| Existing GitHub identifier cases unchanged | ✅ |
| Fail-closed default preserved | ✅ for the covered shapes |

No over-engineering: TC-20 is warranted (AGENT_TEST.md requires black-box coverage for integration behavior, and workspace reuse across recovery is exactly that), and the `requestAndConfirmApiProgress(targetState)` parameterization is the minimal change needed. Scope is otherwise tight to #782.

### Blocking findings

The acceptance criteria are met, but the implementation generalizes past them in ways that are untested and cut both directions — it both **adopts** work it should not and **quarantines** work it should keep. Four inline comments; all reproduced against this head:

1. **P1** — widening `extractIssueNumberFromIdentifier` silently changes `packages/worker/src/identity-preflight.ts`, a different package's fail-closed #507 guard. `DEV-54` + branch `feat/54-unrelated` went from rejected to accepted. Not mentioned in the description; no preflight test uses a `TEAM-123` identifier.
2. **P2** — the foreign-identifier filters are gated on `trackerIdentifier &&`, so **GitHub** runs never reject a foreign Linear branch: `acme/platform#173` on branch `dev-54-fix` attributes.
3. **P2** — bare numeric artifacts (`feat/54-other`, `workpads/54.md`) are now *positive* evidence for `DEV-54`. These are GitHub #54's conventional shapes; recovery may commit and push that work under `DEV-54`.
4. **P3** — `TRACKER_IDENTIFIER_PATTERN` matches version-like segments, so `DEV-54` on branch `chore/node-24` quarantines **its own workpad** — reintroducing the data-loss mode #782 fixes, keyed on branch naming.

Findings 2 and 3 are coupled and want fixing together; the inline comments carry concrete patches.

### Unaddressed prior review

@chatgpt-codex-connector's **P2** on [`issue-identity.ts:53`](https://github.com/hojinzs/github-symphony/pull/785#discussion_r3909626581) is still open with no reply and no follow-up commit. **It is correct** — I reproduced it:

```
issueIdentifier: "DEV2-54", dirtyFiles: [".gh-symphony/workpads/DEV2-54.md"]
-> {"attributed":false,"reason":"dirty workpad references issue #2 instead of DEV2-54"}
```

`extractIssueNumbersFromWorkpadFiles` takes the *first* numeric substring (`2`) from `DEV2-54`, and the foreign-number check then rejects the issue's own workpad. `packages/tracker-linear/src/orchestrator-adapter.ts:26` (`/^[A-Z][A-Z0-9]*-\d+$/`) explicitly accepts digit-bearing team keys, so this is a live Linear shape. Please fix it or reply with reasoning on that thread.

### Smoke test

Unit + probe level, run locally against this head. TC-20 was not executed here (Docker E2E unavailable in this environment) and I am taking the author's `./e2e/run-e2e.sh linear-dirty-recovery 60` pass at face value — but note TC-20 exercises only the happy path through the **file** tracker, so none of the five findings above would be caught by it. The `sleep(Infinity)` mid-turn hold is correct (`stub-worker.ts:169-174` returns a never-resolving promise for non-finite input, not a 1 ms `setTimeout`).

### Also

- **Merge conflict**: `mergeable_state: dirty` — needs a merge from `main` before this can land.
- **nit**: the Prettier reflow of `preventSecondTurnAfterLabelRemoval` and the two AGENT_TEST.md table rows are unrelated whitespace churn. Harmless, but they widen the diff.

Requesting changes on findings 1–4 plus the open Codex thread. Round 1 of a maximum 6.

---

packages/core/src/workflow/issue-identity.ts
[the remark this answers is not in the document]
moncher-dev: Resolved in 5757091: workpad numeric extraction now reads the canonical suffix, and DEV2-54 workpad attribution is covered by regression tests.

moncher-dev: Resolved in 5757091: extractIssueNumberFromIdentifier is numeric-only again, preserving fail-closed worker preflight. The DEV-54 plus feat/54-unrelated regression is now covered.

packages/core/src/workflow/issue-identity.ts
moncher-dev: Resolved in 5757091: foreign full tracker evidence is evaluated unconditionally, including a DEV-54 branch during a GitHub issue run.

packages/core/src/workflow/issue-identity.ts
moncher-dev: Resolved in 5757091: bare numeric branch and workpad evidence is non-positive for tracker-native identifiers, so DEV-54 fails closed for ambiguous 54 artifacts.

packages/core/src/workflow/issue-identity.ts
moncher-dev: Resolved in 5757091: version-like branch identifiers such as node-24 are ignored before foreign-identity evaluation, while the issue own workpad remains positive evidence.

packages/core/src/workflow/issue-identity.ts
```diff
@@ -72,6 +72,56 @@ export function extractIssueNumbersFromBranch(branch: string): number[] {
...
+  return VERSION_LIKE_TRACKER_KEYS.has(teamKey) || /^V\d+$/.test(teamKey);
+}
+
+function extractTrackerIdentifiersFromBranch(branch: string): string[] {
+  const identifiers = new Set<string>();
+  for (const match of branch.matchAll(
+    /(?:^|\/)([A-Za-z][A-Za-z0-9]*-\d{1,9})(?=[-_/]|$)/g
+  )) {
+    const identifier = match[1]!.toUpperCase();
+    if (!isVersionLikeTrackerIdentifier(identifier)) {
```
hojinzs: **P2 — the six-key allowlist only covers the version fragments someone thought of, so the round-1 P3 data-loss path is still open — and it now also hits GitHub runs, which the description says are preserved.**

The round-1 P3 fix is real for `node-24`, but `VERSION_LIKE_TRACKER_KEYS` is a fixed list of 6 names. Every other tool/package name in a `<name>-<digits>` segment still parses as a foreign tracker key and short-circuits before positive evidence. Reproduced against `5757091`:

```
DEV-54, branch chore/next-15,  dirty [.gh-symphony/workpads/DEV-54.md]  // the issue's OWN workpad
-> {"attributed":false,"reason":"current branch 'chore/next-15' references issue NEXT-15 instead of DEV-54"}
```

Same for `chore/react-19`, `chore/vite-7`, `chore/deno-2`, `fix/eslint-9`, `chore/pnpm-10`, `chore/tailwind-4`, `chore/rust-1`. Quarantine + discarded work — the failure mode #782 exists to fix.

**This is now a regression against `main`, not just a residual.** Making the filter unconditional (my round-1 P2) widened the blast radius onto the GitHub path. Same input, both sides:

```
acme/platform#173, branch chore/next-15, dirty [.gh-symphony/workpads/173.md]
  origin/main (12317f5): {"attributed":true, "reason":"dirty workpad belongs to acme/platform#173"}
  this head  (5757091): {"attributed":false,"reason":"...references issue NEXT-15 instead of acme/platform#173"}
```

So "Preserve GitHub numeric behavior" in the description does not hold. Third-order: the allowlist also shadows real Linear teams — `GO-42` on `go-42-fix` yields `no dirty artifact could be attributed`, because its own key is filtered out of positive evidence too.

### Suggested fix

The signal separating a real tracker branch from a version fragment is structural, not lexical: `dev-54-fix` has a slug after the identifier, `chore/next-15` ends the segment. Split the extractor so a segment-terminal token stays *non-evidence* (never outranks the issue's own workpad) while remaining valid positive evidence, and drop the allowlist entirely:

```ts
/**
 * Tracker identifiers encoded in a branch name.
 *
 * `positive` accepts any `<key>-<n>` at a segment start. `foreign` additionally
 * requires a slug to follow (`dev-54-fix`), because a segment-terminal token
 * (`chore/next-15`) is indistinguishable from a version fragment and must never
 * outrank the issue's own evidence.
 */
function extractTrackerIdentifiersFromBranch(branch: string): {
  positive: string[];
  foreign: string[];
} {
  const positive = new Set<string>();
  const foreign = new Set<string>();
  for (const match of branch.matchAll(
    /(?:^|\/)([A-Za-z][A-Za-z0-9]*-\d{1,9})(?=([-_])|\/|$)/g
  )) {
    const identifier = match[1]!.toUpperCase();
    positive.add(identifier);
    if (match[2]) {
      foreign.add(identifier);
    }
  }

  return { positive: [...positive], foreign: [...foreign] };
}
```

Then `VERSION_LIKE_TRACKER_KEYS` / `isVersionLikeTrackerIdentifier` delete, `foreignBranchIdentifiers` filters `branchTrackerIdentifiers.foreign`, and the positive block reads `branchTrackerIdentifiers.positive.includes(trackerIdentifier)`.

Note the capture must sit *inside* the lookahead — `(...)?(?=[-_/]|$)` backtracks and never captures, which silently turns every foreign case into non-evidence.

I ran this variant against all 15 existing assertions in `issue-identity.test.ts` plus the 6 cases above: **21/21 pass** (current head fails 6). It keeps `eng-12-other` / `eng-54-other` / `dev-54-fix`-for-`#173` foreign, keeps bare `54` non-evidence, and additionally fixes the GitHub regression and the `GO-42` shadowing. Worth regression tests for a version-suffixed branch on both a Linear and a GitHub identifier.

---

hojinzs (changes requested):

Re-verified every round-1 item against `5757091` by executing the real module, not by reading the diff. **All five are genuinely fixed** — the replies are accurate:

| Round-1 finding | Verified on `5757091` |
| --- | --- |
| P1 — widened helper weakened worker preflight #507 | ✅ `extractIssueNumberFromIdentifier("DEV-54") → null`; GitHub shapes (`507`, `#507`, `owner/repo#507`) unchanged; preflight regression test added |
| P2 — GitHub runs never rejected a foreign Linear branch | ✅ `acme/platform#173` on `dev-54-fix` → denied, cites `DEV-54` |
| P2 — bare numerics became positive evidence for `TEAM-N` | ✅ `feat/54-other` and `workpads/54.md` → `no dirty artifact`, fail-closed |
| P3 — version-like segments quarantined the own workpad | ⚠️ fixed **for `node-24` only** — see below |
| Codex P2 — `DEV2-54` own workpad rejected as issue #2 | ✅ `workpadNums([DEV2-54.md]) → [54]`; `DEV2-54` + own workpad → attributed |

Requirement coverage against #782 remains complete, and scope is still tight — no new over-engineering, no arbitrary decisions beyond the one below. The TC-20 → TC-21 renumber and the merge from `main` are correct; `mergeable_state` is now `clean` and both checks are green on this head.

### One blocking finding

The round-1 P3 was fixed with a hardcoded six-name allowlist (`GO`, `JAVA`, `NODE`, `PHP`, `PYTHON`, `RUBY`) rather than a structural rule, so the data-loss path is still open for every other `<name>-<digits>` segment — `chore/next-15`, `chore/react-19`, `fix/eslint-9`, `chore/pnpm-10`, and so on all quarantine the issue's own workpad.

What makes this blocking rather than a follow-up: **it is a regression against `main`, on the path this PR says it preserves.** Fixing my round-1 P2 made the foreign filter unconditional, which carried the false positive onto GitHub runs. Same input, both sides:

```
acme/platform#173, branch chore/next-15, dirty [.gh-symphony/workpads/173.md]
  origin/main (12317f5): attributed: true
  this head  (5757091): attributed: false — "references issue NEXT-15"
```

That is silent discarded work on a previously-working path, so "Preserve GitHub numeric behavior" in the description is not accurate as written. The allowlist also shadows legitimate Linear team keys: `GO-42` on `go-42-fix` → `no dirty artifact could be attributed`.

The inline comment carries a validated patch — one regex plus a `{positive, foreign}` split, and the allowlist goes away. I ran it against all 15 existing assertions plus the 6 cases above: **21/21 pass**, versus 6 failures on the current head. It fixes the residual, the GitHub regression, and the team-key shadowing together.

### Smoke test

Unit/probe level against this head, executing `packages/core/src/workflow/issue-identity.ts` directly with a differential harness comparing `origin/main`, this head, and the proposed patch across 21 cases. CI **Test** and **Container Smoke** are green on `5757091`. TC-21 was **not** re-run here — Docker is present but not usable in this environment — so I am taking the author's `./e2e/run-e2e.sh linear-dirty-recovery 60` pass at face value; note it exercises only the happy path (`dev-54-fix`, a slug-suffixed branch), so it would not catch the finding above.

### Prior threads

All five round-1 threads have accurate replies and verified fixes; I am resolving them. This is the only item left. Round 2 of a maximum 6 — **fix this one and I will approve**; I do not intend to open new lines of inquiry on this PR.

---

packages/core/src/workflow/issue-identity.ts
hojinzs: Keeping this one open — correcting my round-2 summary, which said I was resolving all five. The other four are resolved; this one is fixed for `node-24` only.

`VERSION_LIKE_TRACKER_KEYS` covers six names, so `chore/next-15`, `chore/react-19`, `fix/eslint-9`, `chore/pnpm-10` and friends still quarantine the issue's own workpad, and making the filter unconditional carried that onto GitHub runs (`acme/platform#173` on `chore/next-15` is attributed on `main`, denied here).

Continued in [the round-2 comment](https://github.com/hojinzs/github-symphony/pull/785#discussion_r3911443247), which has a structural replacement validated at 21/21 against the existing assertions plus the new cases.

---

packages/core/src/workflow/issue-identity.ts
hojinzs: Two transcription fixes to the comment above — the angle brackets were stripped when it posted, which mangled the snippet:

- `const positive = new Set();` and `const foreign = new Set();` should each be a **string-typed Set**, and the function's return type is an object of two **string arrays** (`positive`, `foreign`). The docstring line should read: "positive accepts any *key*-*number* token at a segment start."
- Only the regex matters for behavior, and it survived intact:

```
(?:^|\/)([A-Za-z][A-Za-z0-9]*-\d{1,9})(?=([-_])|\/|$)
```

Capture group 2 is the separator, captured *inside* the lookahead; a token is foreign only when group 2 matched (a slug follows), and positive regardless.

Also, the forward-link in my reply on the older P3 thread points at a wrong id — this comment is the continuation, not `r3911443247`.

---

packages/core/src/workflow/issue-identity.ts
moncher-dev: Resolved in dc08614: terminal key-number tokens are now non-foreign evidence, so they cannot override an own workpad; only tokens followed by a slug are foreign. Added Linear and GitHub regressions.

packages/core/src/workflow/issue-identity.ts
moncher-dev: Resolved in dc08614: removed the fixed version-key allowlist. GO-42 now attributes from go-42-fix, while terminal version-like tokens remain non-foreign; targeted regression coverage covers both behaviors.

packages/core/src/workflow/issue-identity.ts
```diff
@@ -72,6 +72,56 @@ export function extractIssueNumbersFromBranch(branch: string): number[] {
...
+ * must never outrank the issue's own evidence.
+ */
+function extractTrackerIdentifiersFromBranch(branch: string): {
+  positive: string[];
+  foreign: string[];
+} {
+  const positive = new Set<string>();
+  const foreign = new Set<string>();
+  for (const match of branch.matchAll(
+    /(?:^|\/)([A-Za-z][A-Za-z0-9]*-\d{1,9})(?=([-_])|\/|$)/g
```
hojinzs, line 99: **nit / follow-up (non-blocking) — the slug-suffixed version fragment is the one residual of this structural rule.**

This is the patch I proposed in round 2 and it lands correctly; I verified all 30 cases below. Recording the remaining edge so it is written down somewhere rather than rediscovered later — **not** a change request on this PR.

The terminal-token carve-out keys on "a slug follows the token." A version fragment that *does* have a slug after it is therefore still classified foreign:

```
acme/platform#173, branch feat/next-15-upgrade, dirty [.gh-symphony/workpads/173.md]
  origin/main (12317f5): attributed: true
  this head  (dc08614): attributed: false — "references issue NEXT-15"
```

Same for `chore/node-24-migration` and `chore/react-19-codemod`. `expectedBranches` does not rescue it either, since the foreign check short-circuits first.

Why I am not blocking on it:

- The exposure is much narrower than the round-2 finding it replaces. Symphony's conventional GitHub branch is numeric-led (`feat/173-add-next-15` → attributes correctly, verified), which is immune. It only bites an agent-chosen branch whose *first* segment-start token is `word-number-slug` and is not the run's own key.
- `dev-54-next-15-upgrade` — own key first — attributes correctly, so the issue's own branch is safe.
- Any tighter rule reintroduces lexical guessing, which is what the round-2 allowlist got wrong. The real fix, if this ever bites, is upstream: have the tracker adapter supply the run's team key so "foreign" means "a *known* key that differs," rather than inferring keys from branch text.

If you want, a one-line note in the description's Risks section would cover it. Otherwise a follow-up issue is fine.

---

hojinzs (approved):
## Review round 3/6 — Approve

The round-2 finding is fixed, verified by executing the module rather than reading the diff. `VERSION_LIKE_TRACKER_KEYS` is gone and the structural `{positive, foreign}` split landed as proposed, with the capture correctly inside the lookahead.

### Differential verification — 30/30

Ran `origin/main` (12317f5) and this head (dc08614) side by side over every finding from all three rounds plus new adversarial cases:

| Case | main | head | Expected |
| --- | --- | --- | --- |
| R2 `chore/next-15` + own workpad | false | **true** | true |
| R2 `chore/react-19` / `fix/eslint-9` / `chore/pnpm-10` / `chore/vite-7` / `chore/tailwind-4` | false | **true** | true |
| R2 GitHub regression: `#173` on `chore/next-15` | true | **true** | true |
| R2 `GO-42` on `go-42-fix` (team-key shadowing) | false | **true** | true |
| R1-P3 `chore/node-24` + own workpad | false | **true** | true |
| R1-P2 GitHub run on foreign `dev-54-fix` | true | **false** | false |
| R1-P2b/c bare `54` branch + workpad for `DEV-54` | false | **false** | false |
| Codex-P2 `DEV2-54` own workpad | false | **true** | true |
| #782 AC1–AC5 | — | **all pass** | — |

New probes this round, all correct: foreign slug branch still outranks the own workpad (`feat/eng-12-other` → denied); a terminal foreign token with *no* own evidence still fails closed; underscore separators are treated as slugs; a foreign workpad still beats an own branch; `expectedBranches` does not override a foreign branch; nested `feat/v2-1/wip` and mid-segment `chore/upgrade-node-24` both preserve the own workpad.

**P1 (round 1) confirmed closed at the helper level** — the two functions `packages/worker/src/identity-preflight.ts` actually consumes are now byte-for-byte behaviorally identical to `main`:

```
extractIssueNumberFromIdentifier:  507 / #507 / owner/repo#507 → 507   (unchanged)
                                   DEV-54 / DEV2-54 / GO-42    → null  (unchanged)
extractIssueNumbersFromBranch:     all 5 probes identical to main
```

So the #507 fail-closed guard is untouched, and the added preflight regression pins it.

### Requirement coverage vs #782 — complete

All four named regression tests plus the fail-closed default are present and passing (23 core + 8 worker). No over-engineering: the diff is still scoped to the attribution helper, its tests, TC-21, and the changeset. The one arbitrary decision — the terminal-token carve-out — is now documented in the description's Risks section and is the correct trade-off, since a segment-terminal `key-number` genuinely is indistinguishable from a version fragment.

`extractIssueNumbersFromWorkpadFiles` anchoring changed `workpads/507-notes.md` from `[507]` to `[]`. I checked: workpads are GitHub **comments** in `WORKFLOW.md`, and the `<id>.md` file shape exists only in this heuristic and the E2E stub, both of which use bare `<identifier>.md`. Unreachable, and it is what fixes `DEV2-54`. Non-issue.

### Smoke test

Unit/probe level against this head, with deps installed and packages built:

- `pnpm build`, `pnpm test` (full monorepo), `pnpm typecheck` — all pass.
- `pnpm --filter @gh-symphony/core test -- issue-identity.test.ts` → 23/23.
- Prettier clean on all files this PR touches (the repo-wide 85-file `pnpm format` warning is pre-existing on `main`, not from this diff).
- CI **Test** and **Container Smoke** green on `dc08614`; `mergeable_state: clean`, and I confirmed no conflict against current `main` (2d341da, which has moved past the PR base).

TC-21 was **not** executed here — Docker is installed but the daemon is unavailable in this environment, same as rounds 1 and 2 — so I am taking the author's `./e2e/run-e2e.sh linear-dirty-recovery 60` pass at face value. Worth noting it exercises only the happy path (`dev-54-fix`), so the unit suite, not TC-21, is what covers the findings above. The scenario itself reads correctly: it asserts branch/workpad/artifact preservation, the recovery prompt, and the *absence* of a `recovery-quarantined` event, which is the right negative assertion.

### Prior threads

All six are resolved with accurate replies and verified fixes. One inline nit left below documenting the single residual edge of the structural rule (slug-*suffixed* version fragments such as `feat/next-15-upgrade` remain foreign). It is materially narrower than what it replaces, does not affect Symphony's conventional numeric-led branches, and any tighter rule would reintroduce the lexical guessing that round 2 rejected — so it is a follow-up, not a blocker.

Approving. Round 3 of a maximum 6; no further rounds needed from my side.

---

Issue #777, opened 2026-09-01
bug(linear,orchestrator): tracker refresh failure loses dirty workspace and loops retries


moncher-dev:
## Regression summary

`project start --daemon --assigned-only` for a Linear-backed standalone project can enter an unbounded retry/recovery loop when per-turn tracker state refresh becomes `unknown`.

This appears to regress the intended behavior from #653: on a failed refresh, the runtime should requeue with a bounded retry instead of blindly restarting. In this reproduction, the issue is re-dispatched while still active, a dirty workspace becomes un-attributable and is quarantined, and later workers restart from a clean workspace with no circuit breaker.

## Environment

- gh-symphony: reproduced on v0.11.0; persisted recovery behavior also reproduced after upgrade to v0.17.0
- Node.js v24.18.0
- macOS
- standalone project / Linear tracker
- command: `gh-symphony project start --daemon --assigned-only`
- Linear issue state: `In Progress`; pickup label matched

## Reproduction / observed timeline

1. A worker for Linear issue `DEV-54` made real workspace progress: its turn-progress record reported a dirty workspace, with four changed files.
2. Immediately after that productive turn, the worker emitted:

```text
[worker] tracker state refresh: unknown
[worker] orchestrator_unavailable: tracker refresh failed 3 consecutive times
[worker] codex client protocol error: orchestrator_unavailable
```

3. The worker was failed/retried even though its workspace was dirty.
4. Later orchestrator ticks emitted:

```text
[orchestrator] quarantined dirty workspace for DEV-54: no dirty artifact could be attributed to DEV-54
Run started: DEV-54 state=In Progress status=running
Run finished: DEV-54 (retrying)
```

5. Because the Linear issue stayed active, it was picked up again. Recovery workers repeated repository exploration from a clean workspace with no PR or diff.

Observed resource impact: a single streaming turn reached `inputTokens: 20,931,078` and `outputTokens: 42,360` while the workspace diff remained clean. A separate re-run exposed the saved run as `retrying` with `inputTokens: 1,276,473` and no diff within 15 seconds. We stopped the project to prevent additional consumption.

The same `tracker state refresh: unknown` -> `orchestrator_unavailable` sequence was also observed for other Linear issues in this project, so this is not unique to DEV-54.

## Expected behavior

1. A transient tracker refresh failure must preserve the exact dirty workspace/run attribution.
[...]

PR #786, opened 2026-09-02
fix(worker): preserve tracker refresh diagnostics

### LINKED ISSUE #780: bug(worker): tracker refresh collapses provider errors to "unknown" and fail-closes on permanently unsupported state reads
Part of #777 (root-cause chain item 2 — error masking, and the immediate mitigation).

## Problem

`refreshTrackerState` (`packages/worker/src/turn-lease.ts:35-90`) collapses **every** failure into the single value `"unknown"`, discarding the underlying cause:

- missing `SYMPHONY_ORCHESTRATOR_URL` / run id / token → `unknown`
- non-OK HTTP status (including a deterministic 403 `tracker_state_requests_unsupported`) → `unknown`
- malformed/incomplete body (e.g. `routable` missing) → `unknown`
- any thrown error (timeout, connection refused) → `unknown`

Consequences observed in #777:

1. Operators only ever see `[worker] tracker state refresh: unknown` — no HTTP status, no provider error, no way to tell a Linear capability gap from a network blip (expected behavior 2 in #777).
2. A **permanent** condition (tracker adapter does not support state requests) is treated identically to a transient one, so the fail-closed counter (`updateRefreshFailureCount`, threshold 3) guarantees worker death at turn 4 on every Linear run.
3. The secondary runtime path at `packages/worker/src/index.ts:791` throws `orchestrator_unavailable` after a **single** `unknown`, which is even stricter than the threshold path at `index.ts:1785`.

## Fix outline

1. Return a structured result (state + diagnostic: HTTP status, response `error` field, exception message) and log the diagnostic at every call site.
2. Distinguish outcomes: `unsupported` (orchestrator answered 403 `tracker_state_requests_unsupported`) must NOT count toward the fail-closed threshold — degrade to a warning and skip the between-turn gate, since failing the worker cannot ever succeed on retry.
3. Align the single-failure path (`index.ts:791`) with the thresholded path.

## Regression tests

- 403 `tracker_state_requests_unsupported` → worker continues turns, logs a capability warning once, does not throw `orchestrator_unavailable`.
- Transient failure (timeout / 5xx) → diagnostic is logged with status/error text; threshold behavior unchanged.
- Missing env → distinct diagnostic (`orchestrator endpoint not configured`).

## Evidence

Repro assertions A2–A3 in the RCA comment on #777.

moncher-dev:
## TL;DR

Refines the tracker-refresh fix so its once-only capability warning accurately covers both tracker gates and local convergence, while the shared gate now expresses every convergence outcome directly.

## Change-point diagram

```text
tracker refresh result
  ├─ active ────────────────────────> converge at the non-productive threshold
  ├─ unsupported capability ────────> warn once; skip between-turn gate / converge locally
  ├─ transient unknown (< threshold) > defer convergence
  ├─ transient unknown (threshold) ─> fail closed
  └─ non-actionable ────────────────> complete
```

## Start here

- `packages/worker/src/turn-lease.ts` maps refresh states to context-aware gate actions and reports the capability warning.
- `packages/worker/src/index.ts` consumes the shared action without re-deciding raw tracker state at convergence.
- `packages/worker/src/{turn-lease,convergence-lifecycle}.test.ts` cover the refreshed contract.

## Changes

- Broaden the one-time unsupported-capability warning to cover skipped tracker gates and local convergence.
- Add explicit `defer` and `converge` decisions for transient and active convergence reads.
- Remove the unreachable convergence-harness branch and make it mirror the production decision flow.

## Validation

- `pnpm --filter @gh-symphony/worker test -- turn-lease.test.ts convergence-lifecycle.test.ts convergence-detection.test.ts` — pass (26 tests)
- `pnpm lint` — pass
- `pnpm test` — pass
- `pnpm typecheck` — pass
- `pnpm build` — pass
- Docker black-box happy lifecycle — pass: stable fixture startup dispatched and recovered the worker; after fixture removal the API reported `health: idle`, zero active runs, and an empty retry queue.
- `git diff --check` — pass

## Risks & rollback

The change is limited to worker tracker-refresh control flow and diagnostics. Roll back `1ab1141f` to restore the prior gate action mapping and warning text.

## Changed files

- Worker refresh-gate implementation and Codex convergence call site
- Worker refresh-gate and convergence lifecycle tests

- [ ] Confirm an unsupported tracker-state capability logs the generalized warning once and continues safely between turns.
- [ ] Confirm a transient unavailable tracker read defers convergence until the configured threshold, then fail-closes.

## Issues — Closed #780

- Fixes #780


packages/worker/src/index.ts
```diff
@@ -1928,18 +1941,33 @@ async function runCodexClientProtocol(
...
+          trackerState,
+          consecutiveRefreshFailures,
+          resolveRefreshFailureThreshold(env.SYMPHONY_REFRESH_FAILURE_THRESHOLD)
+        );
+        consecutiveRefreshFailures = refreshGate.count;
+        if (refreshGate.action === "skip") {
+          process.stderr.write(
+            "[worker] convergence tracker gate skipped because state refresh is unsupported\n"
+          );
+          continue;
```
hojinzs: **P2 — this permanently disables convergence detection for any adapter without `requestState`.**

#780's fix outline scopes the skip to *"skip the between-turn gate"* (item 2). This extends `skip` to the convergence-threshold gate too, and that gate has a different job: it is the safety valve that stops a worker looping unproductively.

Because `unsupported` is deterministic (it is decided by `!trackerAdapter.requestState` in `orchestrator/src/service.ts:523`), it never changes for the worker lifetime. So on Linear — the exact provider that motivated #777/#780 — `convergenceDetected` can now never become `true`. Each non-productive turn at/above threshold re-requests the endpoint, logs the skip line, and `continue`s; `consecutiveNonProductiveTurns` is not reset, so the worker just burns all 20 turns instead of exiting at the convergence threshold.

The issue's own reasoning applies symmetrically here: *"failing the worker cannot ever succeed on retry"* — but deferring can never succeed either. For a permanent capability gap the local non-productive signal is the only signal available, so the safer behavior is to trust it rather than discard it:

```ts
if (refreshGate.action === "skip") {
  process.stderr.write(
    "[worker] convergence tracker confirmation unavailable (capability unsupported) — accepting local convergence signal\n"
  );
  convergenceDetected = true;
  // ...set lastError as in the confirmed path...
  break;
}
```

Turn-boundary gates (`index.ts:793`, `index.ts:1779`) skipping is right and I'd keep it as-is — it's only this convergence gate that loses a guarantee. If burning `maxTurns` is the deliberate call, could you note the rationale in `docs/architecture.md:81`? Right now that line says the worker "skips between-turn tracker gates while continuing turns", which doesn't tell an operator that convergence detection is off entirely.

Also minor, same block: since the result is permanent, the fetch could be short-circuited once `unsupportedRefreshWarningLogged` is set — `docs/architecture.md:81` explicitly calls out that each read "may consume a live provider request".

---

packages/worker/src/turn-lease.ts
```diff
@@ -32,17 +48,60 @@ export function updateRefreshFailureCount(
...
+    state,
+    currentCount,
+    threshold
+  );
+  if (state === "non-actionable") {
+    return { action: "complete", count: refreshFailures.count };
+  }
+  if (state === "unsupported") {
+    return { action: "skip", count: refreshFailures.count };
+  }
```
hojinzs, line 85: **P3 — `unsupported` *resets* the transient-failure counter instead of being *excluded* from it.**

This routes `unsupported` through `updateRefreshFailureCount`, whose `count = state === "unknown" ? currentCount + 1 : 0` treats it exactly like a confirmed read — so it zeroes an in-flight transient streak. `docs/configuration.md:489` (added in this PR) states the stronger contract: *"`tracker_state_requests_unsupported` is a permanent capability result and is **excluded** from this counter."* Excluded and reset are different, and the docs describe the behavior that isn't implemented.

I verified against the real exports — two `unknown` failures then one `unsupported`:

```
D failures after unsupported: 0   // was 2 before the unsupported response
```

Reachable rather than theoretical: `unsupported` is deterministic per adapter, but `unknown` is not — `rejected("run_not_current")` / `rejected("run_not_found")` also return 403 (`orchestrator/src/service.ts:472-514`), and timeouts/connection errors land in the `catch`. So `timeout, timeout, unsupported, timeout, timeout` leaves the worker two failures short of fail-closed where the documented contract puts it at the threshold. It weakens fail-closed rather than breaking it, hence P3.

Suggest making the intent explicit in `updateRefreshFailureCount`:

```ts
// A permanent capability gap is neither a success nor a failure: leave the
// transient streak untouched rather than crediting it as a healthy read.
const count =
  state === "unsupported"
    ? currentCount
    : state === "unknown"
      ? currentCount + 1
      : 0;
```

Either fix the code or soften the doc line to "resets" — the two should agree.

---

packages/worker/src/index.ts
```diff
@@ -39,7 +39,6 @@ import { launchCodexWithValidatedPolicy } from "./codex-startup.js";
 import {
   captureTurnBoundarySnapshot,
   evaluateTurnProgress,
-  resolveConvergenceThresholdAction,
   resolveMaxNonProductiveTurns,
 } from "./convergence-detection.js";
```
hojinzs, line 43: **P3 — dropping this import orphans `resolveConvergenceThresholdAction`.**

After this change, `resolveConvergenceThresholdAction` (`packages/worker/src/convergence-detection.ts:41`) has no production caller left — the only remaining references are in its own `convergence-detection.test.ts:42-49`, so the suite stays green while the function is dead.

It is also now stale in a way that would mislead anyone who reuses it: it takes a `TrackerRefreshState`, but this PR added the `"unsupported"` member, and `trackerState === "non-actionable" ? "complete" : "fail"` silently maps `"unsupported"` to `"fail"` — the exact behavior the PR set out to remove.

Please delete the function along with its three test assertions, or note why it's being kept.

---

hojinzs (changes requested):
## Review 1/6 — requirements met, one behavioral gap to settle

The three regression scenarios #780 asked for all work. I verified them against the real exported functions rather than trusting the PR description:

| #780 requirement | Result |
| --- | --- |
| 403 `tracker_state_requests_unsupported` → continues turns, warns once, no `orchestrator_unavailable` | ✅ 19 consecutive turns, `failClosed=false`, counter `0`, exactly **1** warning |
| Transient failure → diagnostic with status/error, threshold unchanged | ✅ `continue, continue, fail-closed` at threshold 3 |
| Missing env → distinct diagnostic | ✅ `orchestrator endpoint not configured` |

Observed log lines:

```
[worker] tracker state refresh: unsupported (tracker state requests unsupported, HTTP 403, error=tracker_state_requests_unsupported)
[worker] warning: tracker state refresh capability unavailable; skipping between-turn tracker gates
[worker] tracker state refresh: unknown (tracker state request failed, HTTP 503, error=upstream_down)
[worker] tracker state refresh: unknown (tracker state request failed, exception=connect ECONNREFUSED)
```

I also checked the wire contract end to end rather than only the worker side: `rejected("tracker_state_requests_unsupported")` (`orchestrator/src/service.ts:523`) → `outcome: "rejected"` → 403 (`cli/src/commands/start.ts:820-824`) → `error` field. The detection keys on the right thing, and it correctly does **not** swallow the other 403s — `run_not_current` still resolves to `unknown` with its provider error preserved, which is the discrimination that matters most here.

**Verification:** `pnpm build`, `pnpm typecheck`, `pnpm lint` all pass; `pnpm test` exits 0 (worker 203/203, full monorepo green). `pnpm format` fails on 89 files, but that is pre-existing — `main` fails on 91, and none of this PR's files are flagged. No Preview environment on this repo, so the smoke test was run locally against the built package.

**Scope:** clean. No overengineering — `resolveTrackerRefreshGate` consolidates three duplicated call sites and is exactly what #780 item 3 asked for, and the changeset is present and correctly scoped to `@gh-symphony/cli` as a patch.

### Findings

**P2 — convergence detection is permanently disabled for adapters without `requestState`** (`index.ts:1960`). #780 scoped the skip to the *between-turn* gate; extending it to the convergence-threshold gate removes a safety valve. Since `unsupported` is deterministic, `convergenceDetected` can never become `true` on Linear — the provider that motivated #777/#780 — so the worker burns all 20 turns on non-productive work instead of exiting at the threshold. The issue's own rationale cuts both ways: if failing can never succeed on retry, neither can deferring. Details and a suggested patch inline.

**P3 — `unsupported` resets the transient-failure counter** rather than being excluded from it (`turn-lease.ts:62`), contradicting the `docs/configuration.md:489` line this PR adds. Verified: two `unknown` failures then one `unsupported` → counter back to `0`.

**P3 — `resolveConvergenceThresholdAction` is now dead code** (`index.ts:43`), and its unchanged signature maps the new `"unsupported"` state to `"fail"` — the behavior this PR exists to remove.

### What I need

Only the P2 is a decision rather than a cleanup. Either trust the local convergence signal when the capability is permanently unavailable, or keep the current behavior and document in `docs/architecture.md:81` that convergence detection is off for these adapters — that line currently tells operators the worker "skips between-turn tracker gates while continuing turns", which understates it. The two P3s are small and can ride along.

The core fix is sound and the diagnostics are a clear improvement over `unknown`. Requesting changes on the P2 decision, not on the approach.

---

packages/worker/src/index.ts
moncher-dev: Resolved in `0ee5b16`.

The shared refresh gate now receives an explicit context. At a between-turn gate, `unsupported` still skips the tracker read gate; at the convergence threshold it returns `converge`, so the worker logs that the capability is unavailable, accepts the local non-productive signal, and takes the normal convergence exit. `convergence-lifecycle.test.ts` covers the unsupported threshold outcome, and the architecture/configuration docs describe the distinction.

packages/worker/src/turn-lease.ts
moncher-dev: Resolved in `0ee5b16`.

`updateRefreshFailureCount` now preserves the current transient-failure streak for `unsupported`; only a confirmed active/non-actionable read resets it. The regression coverage now verifies both the preserved count and the fact that an unsupported 403 does not itself fail-close the worker.

packages/worker/src/index.ts
moncher-dev: Resolved in `0ee5b16`.

Removed `resolveConvergenceThresholdAction`, its stale `TrackerRefreshState` import, and its orphaned unit assertions. The context-aware shared refresh gate is now the only convergence tracker decision path.

packages/worker/src/turn-lease.ts
```diff
@@ -83,12 +186,59 @@ export async function refreshTrackerState(
...
+  unsupportedWarningLogged: boolean,
+  write: (message: string) => void = (message) => process.stderr.write(message)
+): boolean {
+  write(`[worker] ${label}: ${formatTrackerRefreshResult(result)}\n`);
+  if (result.state !== "unsupported" || unsupportedWarningLogged) {
+    return unsupportedWarningLogged;
+  }
+
+  write(
+    "[worker] warning: tracker state refresh capability unavailable; skipping between-turn tracker gates\n"
```
hojinzs: **nit — the warning text no longer describes the full behavior.**

Now that `unsupported` also drives the convergence outcome, "skipping between-turn tracker gates" only covers half of what happens. It's emitted once per worker lifetime, and it is normally the between-turn gate that emits it first — but with `SYMPHONY_MAX_NONPRODUCTIVE_TURNS=1` the convergence check at the end of turn 1 fires before any between-turn refresh, so this single line can be the only capability message an operator ever sees, and it would say the wrong thing about the exit.

Suggest something that covers both contexts:

```
"[worker] warning: tracker state refresh capability unavailable; skipping tracker gates and using local convergence signals\n"
```

Non-blocking.

---

packages/worker/src/turn-lease.ts
```diff
@@ -28,21 +44,76 @@ export function updateRefreshFailureCount(
...
+    return { action: "complete", count: refreshFailures.count };
+  }
+  if (state === "unsupported") {
+    return {
+      action: context === "convergence" ? "converge" : "skip",
+      count: refreshFailures.count,
+    };
+  }
+  return {
+    action: refreshFailures.failClosed ? "fail-closed" : "continue",
```
hojinzs, line 97: **nit — the shared gate stops one state short of being the single decision point.**

In `"convergence"` context this fallthrough returns `"continue"` for both `active` and sub-threshold `unknown`, but those need opposite outcomes at the call site: `active` should converge and break, `unknown` should defer and continue the loop. So `index.ts:1986` still re-branches on the raw `trackerState`, and the new `"converge"` action ends up describing only the `unsupported` case rather than "the worker converges here".

A `"defer"` action for `unknown`-in-convergence would let that block read purely off `refreshGate.action`, with the between-turn mapping unchanged:

```ts
if (state === "unknown" && context === "convergence" && !refreshFailures.failClosed) {
  return { action: "defer", count: refreshFailures.count };
}
```

Behavior is correct as written — I traced all four states through both contexts — so this is structure only. Take it or leave it.

---

packages/worker/src/convergence-lifecycle.test.ts
```diff
@@ -17,8 +16,24 @@ async function runConvergenceThreshold(response: Response) {
...
+  const trackerState = trackerRefresh.state;
+  const refreshGate = resolveTrackerRefreshGate(
+    trackerState,
+    0,
+    1,
+    "convergence"
+  );
 
-  if (resolveConvergenceThresholdAction(trackerState) === "complete") {
+  if (refreshGate.action === "skip") {
```
hojinzs: **nit — this branch is unreachable in `runConvergenceThreshold`.**

The gate is called with context `"convergence"` on line 23, and `resolveTrackerRefreshGate` only returns `"skip"` for `context === "between-turn"` (`turn-lease.ts:81`). So no response can reach this `skip` handler, and the `"running"` / `null` result it returns is not an outcome the real convergence block can produce.

Worth deleting so the harness keeps mirroring `index.ts:1954-2002` one-to-one — that mirroring is the only thing making this file a meaningful regression test, since it re-implements the gate rather than exercising it, and a branch that can't fire is where drift hides first. The `unsupported` case is already covered by the fallthrough at the end (which is what `index.ts` does too), and the new test on line 99 passes for that reason.

---

hojinzs (approved):
## Review 2/6 — approved; resolve the `AGENT_TEST.md` conflict before merging

All three findings from review 1/6 are genuinely fixed. I re-verified each against the real exported functions in the built package rather than reading the diff or trusting the replies:

| Review 1/6 finding | Status |
| --- | --- |
| **P2** convergence permanently disabled for adapters without `requestState` | ✅ `resolveTrackerRefreshGate("unsupported", …, "convergence")` → `converge`; the worker logs the capability line and takes the normal convergence exit |
| **P3** `unsupported` resets the transient streak | ✅ now preserved — `unknown, unknown, unsupported, unknown` → counts `1, 2, 2, 3` and fail-closes on the fourth, which is the `docs/configuration.md` contract |
| **P3** `resolveConvergenceThresholdAction` dead code | ✅ function, its `TrackerRefreshState` import, and its three assertions removed; no callers remain |

Observed sequences from the smoke run:

```
A. unsupported ×5 (between-turn) → skip, skip, skip, skip, skip | count stays 0 | exactly 1 warning
B. 503, ECONNREFUSED, 503       → continue(1), continue(2), fail-closed(3)
C. 403 run_not_current ×3       → unknown with error=run_not_current preserved; still fail-closes at 3
D. 503, 503, unsupported, 503   → 1, 2, 2, 3 → fail-closed   (the P3 fix)
E. unsupported at convergence   → converge                    (the P2 fix)
```

The discrimination that mattered most still holds: only `403` + `tracker_state_requests_unsupported` becomes `unsupported`. `run_not_current` stays `unknown` with its provider error intact, and a non-JSON `403` degrades to `unknown` with `HTTP 403` rather than being mistaken for a capability gap.

**Requirements (#780):** all three fix-outline items and all three named regression scenarios are covered. Item 3 in particular — the single-`unknown` throw at the old `index.ts:791` — is now the shared thresholded gate, so both runtime paths behave identically.

**Downstream check.** I traced where the new convergence exit lands, since #777 is about retry loops and this PR routes more runs into convergence. It is bounded: `returnConvergedIssueToRetryableState` returns early on `!trackerAdapter.requestState` (`orchestrator/src/service.ts:896`) — exactly the adapters that produce `unsupported` — so those runs fall through to `retryKind: "failure"` and are capped by `maxFailureRetries`. No convergence lock is taken and no unbounded re-queue is introduced.

**Verification:** `pnpm build`, `pnpm lint`, `pnpm typecheck` all pass. `pnpm test` green across the monorepo (worker 202/202 — 203 before, minus the one removed dead-code assertion; 2 149 total). No Preview environment on this repo, and Docker is unavailable in my sandbox so `./e2e/run-e2e.sh` could not run — the smoke test above was run locally against the built `packages/worker/dist`.

**Scope:** still clean. No overengineering — the context parameter on the shared gate is the minimum needed to give the two gates different answers, and the changeset is present and correctly scoped.

### Blocking merge (not a code issue)

`AGENT_TEST.md` conflicts with `main` — `mergeable_state` is `dirty`. Both sides added rows to the same coverage table (`main` added *Agent child isolation and host Git transport*, this branch added *Tracker refresh capability degradation*). Mechanical, docs-only, both rows should survive. Please merge `main` in and resolve.

### Remaining findings

Three nits inline, none blocking: the once-per-lifetime warning text still says "skipping between-turn tracker gates" when it now also governs the convergence exit; the gate returns `continue` for both `active` and sub-threshold `unknown` in convergence context, so the call site still re-branches on the raw state; and the new `skip` branch in the `convergence-lifecycle.test.ts` harness is unreachable under `"convergence"` context.

Approving — resolve the conflict and land it. The nits are yours to take or skip.

---

packages/worker/src/turn-lease.ts
moncher-dev: Resolved in `1ab1141f`.

The once-per-worker warning now says the capability is unavailable while "skipping tracker gates and using local convergence signals," so it accurately describes both the between-turn and convergence paths. `turn-lease.test.ts` asserts the exact warning text and its once-only behavior.

packages/worker/src/turn-lease.ts
moncher-dev: Resolved in `1ab1141f`.

The shared gate now returns `converge` for confirmed active reads and `defer` for sub-threshold transient failures in convergence context. The convergence call site acts on `complete`, `fail-closed`, and `defer`, with the remaining `converge` result exiting normally; raw state is retained only to emit the unsupported-specific capability detail.

packages/worker/src/convergence-lifecycle.test.ts
moncher-dev: Resolved in `1ab1141f`.

The harness now handles only the reachable convergence `defer` action and supplies the invariant `non-actionable` state directly to the completion path. This keeps the test flow aligned with the production convergence block.

hojinzs (approved):
**Approve** — round 3. All six findings from the previous two rounds are resolved, and I verified each one against the built exports rather than the diff alone.

## Requirement coverage (#780)

| Issue requirement | Status | Evidence |
| --- | --- | --- |
| 1. Structured result (HTTP status, provider `error`, exception) logged at every call site | ✅ | `refreshTrackerState` returns `{state, diagnostic}`; both between-turn gates and the convergence gate route through `reportTrackerRefresh` |
| 2. `unsupported` excluded from the fail-closed counter; warn + skip the gate | ✅ | `updateRefreshFailureCount` preserves the streak; gate returns `skip` between turns |
| 3. Align the single-failure path (`index.ts:791`) with the thresholded path | ✅ | `runNonCodexRuntimeAdapterLifecycle` now uses the same `resolveTrackerRefreshGate` + threshold instead of throwing on one `unknown` |
| Regression: 403 unsupported → continues, warns once, no throw | ✅ | `convergence-lifecycle.test.ts:98`, `turn-lease.test.ts` |
| Regression: transient failure keeps status/error text, threshold unchanged | ✅ | `turn-lease.test.ts` 503 + exception cases |
| Regression: missing env → distinct diagnostic | ✅ | `turn-lease.test.ts` `orchestrator endpoint not configured` |

## Smoke test

Built the workspace and drove the real `turn-lease` exports through the full state × context matrix and a simulated Linear-style session. No preview environment applies here — this is worker-internal control flow, so the built module is the black box.

Full matrix (threshold 3):

```
                count=0                     count=2 (threshold-1)
between-turn  active         continue/0    continue/0
between-turn  non-actionable complete/0    complete/0
between-turn  unsupported    skip/0        skip/2      <- streak preserved
between-turn  unknown        continue/1    fail-closed/3
convergence   active         converge/0    converge/0
convergence   non-actionable complete/0    complete/0
convergence   unsupported    converge/0    converge/2
convergence   unknown        defer/1       fail-closed/3
```

Round-1 P3 ("excluded, not reset") — the exact sequence I reported last time now behaves as documented:

```
timeout → continue/1 · timeout → continue/2 · unsupported → skip/2 · timeout → fail-closed/3
```

Round-1 P2 (convergence permanently disabled on adapters without `requestState`) — a 6-turn unsupported session:

```
turns 1-6: action=skip, count=0, worker continues
convergence threshold: action=converge -> convergenceDetected=true
capability warnings emitted: 1   diagnostic lines emitted: 6
```

Round-2 nit (warning text) — the single once-per-worker line now covers both contexts, and the `SYMPHONY_MAX_NONPRODUCTIVE_TURNS=1` case where it is the only capability message an operator sees:

```
[worker] warning: tracker state refresh capability unavailable; skipping tracker gates and using local convergence signals
```

Live `refreshTrackerState` against stub responses distinguishes every failure mode that previously collapsed to bare `unknown`:

```
403 unsupported     -> unsupported (…, HTTP 403, error=tracker_state_requests_unsupported)
403 run_not_current -> unknown (tracker state request failed, HTTP 403, error=run_not_current)
503                 -> unknown (tracker state request failed, HTTP 503)
200 malformed       -> unknown (invalid tracker state response, HTTP 200)
missing env         -> unknown (orchestrator endpoint not configured)
```

The `403 run_not_current` vs `403 unsupported` split is the one I most wanted to see hold, since both are 403s and only one may bypass the counter — it does.

## Checks

`pnpm build`, `pnpm lint`, `pnpm typecheck` clean; `pnpm test` 136/136 files pass (218 in `@gh-symphony/worker`). CI **Test** and **Container Smoke** both green on `1ab1141f`. `pnpm format` reports pre-existing failures on `main` for the same files and is not run in CI, so it is not attributable to this PR.

## Scope and residual risk

No over-engineering: the `defer`/`converge` actions exist because I asked for them in round 2, `resolveConvergenceThresholdAction` was deleted rather than left dead, and the 12 changed files are all germane. Changeset is present and correctly scoped to `@gh-symphony/cli` as a patch.

Two things I traced and deliberately am **not** raising as changes:

- `index.ts:1951` still reads `trackerRefresh.state === "unsupported"` alongside the action, but only to select a diagnostic line — the control flow is driven purely by `refreshGate.action`, which was the point of the round-2 nit.
- A single turn can increment `consecutiveRefreshFailures` twice (between-turn gate, then the convergence gate) when both reads fail. Both are genuine failed reads and this is pre-existing behavior, not a regression from this PR.

Nothing blocking remains. Good convergence over three rounds.

---

PR #787, opened 2026-09-02
fix(cli): support Linear doctor smoke checks

### LINKED ISSUE #784: bug(cli): doctor --smoke requires a GitHub Project binding, unusable for Linear standalone projects
Part of #777 (the "Additional v0.17.0 Linear validation bug" section).

## Problem

`gh-symphony doctor --smoke` unconditionally requires a GitHub Project binding: `packages/cli/src/commands/doctor.ts:1396` fails with

```text
Smoke check could not read live issues because the GitHub Project binding did not resolve.
```

whenever `projectBindingId` / `projectDetail` are absent — which is always the case for a Linear-backed standalone project. The smoke path then constructs a GitHub tracker config regardless of the configured tracker adapter, so the smoke test is unusable for Linear even though plain `doctor` resolves the Linear tracker successfully.

## Fix outline

1. Branch the smoke check on the configured tracker adapter: for Linear, read a live issue through the Linear adapter (`listIssues` / GraphQL tool) instead of resolving a GitHub Project binding.
2. Only surface GitHub-binding preconditions when the tracker is GitHub.
3. `--issue` argument validation should accept the tracker's identifier format (`TEAM-123` for Linear) rather than only `owner/repo#number`.

## Regression tests

- `doctor --smoke` on a Linear standalone project reads a Linear issue without any GitHub Project binding (from #777's suggested tests).
- `doctor --smoke --issue DEV-54` accepted for Linear; `owner/repo#number` still required for GitHub.
- GitHub smoke path unchanged.

moncher-dev:
## TL;DR

- `doctor --smoke` now reads Linear and other non-GitHub tracker issues through their configured adapters, without requiring a GitHub Project binding.
- Linear diagnostics preserve `LINEAR_API_KEY` as a fallback when optional `tracker.api_key` is absent, while workflow-resolved credentials remain preferred.
- Explicit Linear identifiers accept `DEV-54` case-insensitively; the GitHub binding and `owner/repo#number` path remain unchanged.

## Change-point diagram

```text
doctor --smoke
  └─ configured tracker
      ├─ github-project → GitHub auth + Project binding → owner/repo#number
      └─ non-GitHub     → configured tracker adapter → native identifier
          └─ linear     → resolved api_key || LINEAR_API_KEY → TEAM-123
```

## Start here

- `packages/cli/src/commands/doctor.ts` — tracker-aware smoke reads, credential fallback, and typed GitHub prerequisites.
- `packages/cli/src/commands/doctor.test.ts` — Linear, file-adapter, identifier, repository-mapping, and GitHub regressions.

## Changes

- Route all non-GitHub smoke reads through `resolveTrackerAdapter`, eliminating the file-tracker fall-through crash.
- Keep `tracker.api_key` preferred and retain `LINEAR_API_KEY` fallback for Linear project resolution and smoke reads.
- Normalize Linear issue identifiers before validating them and remove GitHub-only non-null assertions through an explicit narrowed branch state.

## Evidence

- `pnpm lint` — passed.
- `pnpm test` — passed across all workspace packages; CLI: 615 tests, including 66 doctor tests.
- `pnpm typecheck` — passed.
- `pnpm build` — passed.
- `./e2e/run-e2e.sh happy 60` — passed; worker dispatched, retried, and returned to idle health.

## Risks & rollback

- Risk: tracker branching could regress the GitHub smoke path. Existing GitHub smoke coverage remains green and the branch retains its Project-binding and identifier prerequisites.
- Rollback: revert this PR; it has no persisted-data or schema changes.

## Changed files

- `.changeset/linear-doctor-smoke.md` — patch release note for `@gh-symphony/cli`.
- `packages/cli/src/commands/doctor.ts` — smoke behavior and Linear credential diagnostics.
- `packages/cli/src/commands/doctor.test.ts` — regression coverage.
- `packages/cli/src/index.ts`, `README.md`, `packages/cli/README.md`, `docs/architecture.md`, `AGENT_TEST.md` — original issue documentation and CLI-surface updates.

## Human validation

- [ ] Run `gh-symphony doctor --smoke` in a standalone Linear project with live credentials.
- [ ] Run `gh-symphony doctor --smoke --issue DEV-54` against a matching Linear team.
- [ ] Confirm a GitHub-backed project still requires `owner/repo#number` and its Project binding.

## Post-merge validation

- [ ] Confirm the released CLI reports a successful Linear live-issue smoke check without GitHub Project metadata.

## Spec conformance

- Configuration, Execution, and Integration boundaries remain aligned with `docs/symphony-spec.md`.
- No intentional divergence; the upstream specification is unchanged.

## Rework update

- Resolved the persisted unsupported-adapter smoke crash in `5f13f23` and closed all review threads.
- Final validation: `pnpm lint`, `pnpm test`, `pnpm typecheck`, `pnpm build`, and `./e2e/run-e2e.sh happy 60` passed.


packages/cli/src/commands/doctor.ts
```diff
@@ -841,13 +846,13 @@ async function checkLinearTrackerResolution(input: {
     );
   }
 
-  const token = process.env.LINEAR_API_KEY?.trim();
+  const token = input.token?.trim();
```
hojinzs: **P1 — `LINEAR_API_KEY` fallback is dropped; plain `doctor` now fails on a supported configuration.**

Reading the resolved `tracker.api_key` (so `$CUSTOM_LINEAR_TOKEN` works) is a good addition, but it **replaced** the env fallback instead of layering on top of it. `readNormalizedOptionalSecret` (`packages/core/src/workflow/parser.ts:561`) returns `null` when `tracker.api_key` is absent, so a Linear WORKFLOW.md that omits the optional key now fails `linear_tracker_resolution` and makes `doctor` exit nonzero — even with `LINEAR_API_KEY` exported.

That configuration is supported elsewhere in the codebase:
- `packages/cli/src/commands/start.ts:200` — `if (configuredToken?.trim() || process.env.LINEAR_API_KEY?.trim())`
- `packages/tracker-linear/src/orchestrator-adapter.ts:843` — `const token = dependencies.token ?? process.env.LINEAR_API_KEY;`

**Verified** with a WORKFLOW.md that has no `api_key:` and `LINEAR_API_KEY=lin_env_token` in the environment:

| | `linear_tracker_resolution` |
|---|---|
| base `b9f676d` | ✅ `pass` — `Resolved Linear project "Symphony"` |
| head `962e279` | ❌ `fail` — `tracker.api_key did not resolve` |

Note this also makes the PR self-inconsistent: the new smoke path passes `apiKey ?? undefined` into the adapter, which *does* fall back to `LINEAR_API_KEY` at `orchestrator-adapter.ts:843`. So on the same run `smoke_issue` passes while `linear_tracker_resolution` fails.

Suggested fix — keep the resolved key as the preferred source and retain the fallback, matching `start.ts`:

```ts
const token = input.token?.trim() || process.env.LINEAR_API_KEY?.trim();
```

and keep the remediation text mentioning both (`tracker.api_key` or `LINEAR_API_KEY`), so the message stays actionable for both setups. Please also add a regression test for "no `tracker.api_key`, `LINEAR_API_KEY` set → pass".

---

packages/cli/src/commands/doctor.ts
```diff
@@ -1406,15 +1438,131 @@ async function buildDoctorSmokeChecks(input: {
     return checks;
   }
 
-  const trackerConfig = buildGithubTrackerConfig({
-    projectConfig: input.selection,
-    bindingId: input.projectBindingId,
-    token: input.auth.token,
-    workflow: input.workflow.workflow,
-  });
+  const githubTrackerConfig =
```
hojinzs: **P1 — `doctor --smoke` now throws an uncaught `TypeError` for the `file` tracker (regression from a clean failure).**

The guards above are narrowed to `trackerKind === "github-project"`, and the branch below only special-cases `"linear"`. `file` is a registered adapter (`packages/orchestrator/src/tracker-adapters.ts:12`, and it is in `DEFAULT_SUPPORTED_TRACKER_KINDS` at `packages/core/src/workflow/parser.ts:85`), so a file-backed project falls straight through: `githubTrackerConfig` is `null`, `projectDetail` is `null`, and the non-null assertions below dereference them.

**Verified** with `--smoke` against a `kind: file` project:

| | result |
|---|---|
| base `b9f676d` | `smoke_issue` = `fail`, *"Smoke check could not read live issues because the GitHub Project binding did not resolve."* — `doctor` completes |
| head `962e279` | `TypeError: Cannot read properties of null (reading 'title')` at `doctor.ts:1664`, propagated out of `runDoctorDiagnostics` — the whole command crashes |

This is the same concern Codex raised, but the impact is a crash rather than a wrong-adapter call, so I'd treat it as P1.

The narrowest fix is to fail closed on unsupported trackers before this point:

```ts
if (trackerKind !== "github-project" && trackerKind !== "linear") {
  checks.push(
    failCheck(
      "smoke_issue",
      "Smoke target issue",
      `Smoke check does not support the "${trackerKind}" tracker adapter.`,
      "Run 'gh-symphony doctor' without '--smoke', or configure a github-project or linear tracker.",
      { adapter: trackerKind }
    )
  );
  return checks;
}
```

Routing every non-GitHub tracker through `resolveTrackerAdapter` (as Codex suggested) is the more complete answer and would subsume the Linear branch — but it is larger than #784 asked for, so a fail-closed guard plus a test is acceptable for this PR if you prefer to keep scope tight. Either way please cover it with a test: no case in `doctor.test.ts` exercises a tracker that is neither `github-project` nor `linear`, which is why this got through green CI.

---

packages/cli/src/commands/doctor.ts
```diff
@@ -1543,10 +1691,13 @@ async function buildDoctorSmokeChecks(input: {
   }
 
   const repositoryName = `${issue.repository.owner}/${issue.repository.name}`;
-  if (!checks.some((check) => check.id === "project_repository_link")) {
+  if (
+    trackerKind === "github-project" &&
```
hojinzs: **P2 — the `!` assertions traded a compile-time guarantee for a runtime crash.**

This PR adds ~10 non-null assertions (`input.projectDetail!`, `input.auth!`, `input.projectBindingId!`, `githubTrackerConfig!`) across the GitHub branch. Before the change, the unconditional guards gave TypeScript real narrowing; now the invariant "if we reach here, tracker is `github-project` and the binding resolved" is only enforced by convention, and `tsc` can no longer catch a violation. That is precisely why the `file`-tracker crash compiles cleanly and passes lint.

Rather than sprinkling `!`, consider making the shape carry the invariant — e.g. narrow once and hand the GitHub-only state to a helper:

```ts
if (trackerKind === "github-project") {
  if (!input.auth || !input.projectBindingId || !input.projectDetail) {
    /* existing fail checks */
    return checks;
  }
  return buildGithubSmokeChecks({ ...input, auth: input.auth, projectDetail: input.projectDetail, bindingId: input.projectBindingId });
}
```

Splitting the two trackers into separate functions would also remove the ~110 lines of duplicated `failCheck` scaffolding between the Linear and GitHub branches. Not blocking on its own, but it is the structural cause of the P1 above.

---

packages/cli/src/commands/doctor.ts
```diff
@@ -1341,6 +1346,28 @@ function formatSmokeError(error: unknown): string {
   return error instanceof Error ? error.message : String(error);
 }
 
+const LINEAR_ISSUE_IDENTIFIER_PATTERN = /^[A-Z][A-Z0-9]*-\d+$/;
```
hojinzs, line 1349: **P3 — the identifier pattern is case-sensitive, so `--issue dev-54` is rejected.**

`parseLinearSmokeIssueIdentifier` trims but does not normalize case, and the pattern requires `[A-Z]`. Linear itself accepts lowercase team keys in URLs and its own search, and the GitHub side of this same flag is case-insensitive in practice, so `--issue dev-54` failing with *"Linear issue identifiers must match TEAM-123"* is a papercut on a diagnostic command.

```ts
const identifier = value.trim().toUpperCase();
```

Cheap to do and worth one test case.

---

packages/cli/src/commands/doctor.test.ts
```diff
@@ -83,6 +83,12 @@ function createLinearProjectConfig(
     projectId: "tenant-a",
     slug: "tenant-a",
     workspaceDir,
+    repository: {
```
hojinzs, line 86: **P3 — adding `repository` to the shared fixture hides the branch it was added to satisfy.**

This field was added so the new `buildSmokeTrackerProject` guard passes, but it also means no test ever reaches the "managed project has no repository mapping" failure that guard exists to produce. Since `CliProjectConfig.repository` is optional (`packages/cli/src/config.ts:68`) while `OrchestratorProjectConfig.repository` is required, that branch is the seam between the two types and deserves its own case.

Related: the three new Linear tests stub `resolveTrackerAdapter` wholesale, so neither `buildSmokeTrackerProject` nor the real adapter's token handling (`packages/tracker-linear/src/orchestrator-adapter.ts:843`) is exercised. That is a reasonable unit boundary, but it is why the two P1s above are invisible to this suite. Suggested additions:

- Linear project with `repository` omitted → `smoke_issue` fails with the mapping message (not a throw).
- `tracker.adapter: "file"` with `--smoke` → a `fail` check, not a `TypeError`.
- Linear WORKFLOW.md without `api_key`, `LINEAR_API_KEY` set → `linear_tracker_resolution` passes.

---

hojinzs (changes requested):
## Review — round 1/6

The core of #784 is done well: `doctor --smoke` genuinely reads Linear issues without a GitHub Project binding, `--issue DEV-54` is accepted, the GitHub path keeps its binding and identifier requirements, and the check output ordering is preserved despite `checkWorkflow` moving earlier in `runDoctorDiagnostics`. Docs, changeset, and the AGENT_TEST scenario table are all updated.

Requesting changes on **two P1 regressions** — both verified by running the code against base `b9f676d` and head `962e279`, not just read.

### Requirements vs. #784

|---|---|
| `doctor --smoke` reads a Linear issue with no GitHub Project binding | ✅ |
| `doctor --smoke --issue DEV-54` accepted for Linear | ✅ |
| `owner/repo#number` still required for GitHub | ✅ |
| GitHub smoke path unchanged | ✅ regression tests green |

### Blocking

1. **P1 — `LINEAR_API_KEY` fallback dropped** (`doctor.ts:849`). A Linear WORKFLOW.md that omits the optional `tracker.api_key` now fails `linear_tracker_resolution` and makes plain `doctor` exit nonzero, even with `LINEAR_API_KEY` exported. Verified: base = `pass`, head = `fail`. That configuration is supported by `start.ts:200` and by the Linear adapter itself (`orchestrator-adapter.ts:843`), which is also why this PR is internally inconsistent — the new smoke path *does* get the env fallback through the adapter, so `smoke_issue` can pass while `linear_tracker_resolution` fails on the same run.

2. **P1 — `--smoke` crashes for the `file` tracker** (`doctor.ts:1441`). Narrowing the guards to `github-project` while only special-casing `linear` lets `file` fall through to the GitHub path with two nulls. Verified: base returns a clean `fail` check; head throws `TypeError: Cannot read properties of null (reading 'title')` out of `runDoctorDiagnostics`. Regression from a graceful message to a crash.

### Non-blocking

3. **P2** — the ~10 new `!` assertions removed the compile-time guarantee that made #2 impossible; worth restructuring (`doctor.ts:1695`).
4. **P3** — `--issue dev-54` rejected on case (`doctor.ts:1349`).
5. **P3** — test gaps that let both P1s through green CI (`doctor.test.ts:86`).

### Scope

No over-engineering in the smoke branching itself — it tracks the issue's fix outline closely. One judgment call went beyond #784: switching `checkLinearTrackerResolution` from `process.env.LINEAR_API_KEY` to the workflow-resolved key. Supporting custom env references is a real improvement and worth keeping, but it needed to be additive rather than a replacement — that is finding #1. The `file`-tracker handling is the reverse problem: the branching stopped one adapter short of complete.

### Verification performed

`pnpm install` → `pnpm build` ✅ → `pnpm lint` ✅ → `pnpm typecheck` ✅ → `doctor.test.ts` 62/62 ✅. Smoke testing was done at the `runDoctorDiagnostics` level (no Linear/preview credentials available in this environment); both P1s were reproduced by running the same scenario against base and head and diffing the resulting check objects.

Fix the two P1s with tests and I expect this to be approvable on the next round.

---

packages/cli/src/commands/doctor.ts
```diff
@@ -1917,6 +2086,8 @@ export async function runDoctorDiagnostics(
     checks.push(
       await checkLinearTrackerResolution({
         projectConfig: resolvedProjectConfig,
+        token:
+          workflow.status === "pass" ? workflow.workflow.tracker.apiKey : null,
```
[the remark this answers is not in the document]
moncher-dev: Resolved in e842a3f: doctor now retains LINEAR_API_KEY when no workflow api_key resolves, with regression coverage.

packages/cli/src/commands/doctor.ts
```diff
@@ -1406,15 +1438,131 @@ async function buildDoctorSmokeChecks(input: {
...
+          projectConfig: input.selection,
+          bindingId: input.projectBindingId!,
+          token: input.auth!.token,
+          workflow: input.workflow.workflow,
+        })
+      : null;
 
   let issue: TrackedIssue | null = null;
-  if (input.parsedArgs.issue) {
+  if (trackerKind === "linear") {
```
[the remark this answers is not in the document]
moncher-dev: Resolved in e842a3f: every non-GitHub tracker now uses its configured adapter for smoke reads; file-tracker coverage prevents the crash.

packages/cli/src/commands/doctor.ts
moncher-dev: Resolved in e842a3f: tracker.api_key remains preferred and LINEAR_API_KEY is restored as fallback, covered by a no-api_key workflow regression.

packages/cli/src/commands/doctor.ts
moncher-dev: Resolved in e842a3f: non-GitHub smoke reads go through resolveTrackerAdapter; the new file-adapter test verifies a clean live read instead of a TypeError.

packages/cli/src/commands/doctor.ts
moncher-dev: Resolved in e842a3f: GitHub smoke state is narrowed once into githubSmoke, removing the non-null assertions from the GitHub-only path.

packages/cli/src/commands/doctor.ts
moncher-dev: Resolved in e842a3f: Linear identifiers are normalized to uppercase before TEAM-123 validation; dev-54 is covered.

packages/cli/src/commands/doctor.test.ts
moncher-dev: Resolved in e842a3f: focused coverage now asserts the missing repository mapping diagnostic, alongside adapter and environment-fallback regressions.

moncher-dev:
Rework is complete in e842a3f. The Linear credential fallback, non-GitHub smoke routing, typed GitHub branch boundary, lowercase identifiers, and focused regression coverage are addressed. All inline comments have replies; please re-review.

packages/cli/src/commands/doctor.ts
```diff
@@ -1406,15 +1438,154 @@ async function buildDoctorSmokeChecks(input: {
...
+          "Smoke target issue",
+          "Smoke check could not read configured tracker issues because the managed project has no repository mapping.",
+          "Add a repository mapping to the standalone project WORKFLOW.md and restart the project before re-running the smoke check.",
+          { error: formatSmokeError(error) }
+        )
+      );
+      return checks;
+    }
+
+    const trackerAdapter = input.deps.resolveTrackerAdapter(
```
hojinzs: **P2 — `resolveTrackerAdapter` throws uncaught here, so `doctor --smoke` still crashes for an adapter the CLI does not register.**

Not blocking the approval, but this is the last surviving corner of the round-1 P1, and the fix is ~6 lines — worth landing before merge rather than in a follow-up.

`resolveTrackerAdapter` throws for anything outside its registry (`packages/orchestrator/src/tracker-adapters.ts:25`), and this call is outside a `try`. Every other failure in this branch produces a `failCheck`; this one propagates out of `runDoctorDiagnostics` and takes the whole command with it — no report, no other checks.

**Verified** with a `project.json` carrying `tracker.adapter: "jira"` and `--smoke`:

| | result |
|---|---|
| base `b9f676d` | `smoke_issue` = `fail`, *"…the GitHub Project binding did not resolve."* — `doctor` completes |
| head `e842a3f` | `Error: Unsupported tracker adapter: jira` thrown out of `runDoctorDiagnostics`; `report` is never produced |

Reachability is narrow — `project.json` is written from a `getSupportedTrackerKinds()`-validated workflow, so it takes CLI version skew (a project started by a newer CLI, then inspected by an older one) or a hand-edited config. But that is exactly the population that runs `doctor`: the command's whole job is to explain a broken configuration, and crashing on the unknown-adapter case is the one failure mode it can't afford.

Wrapping it alongside `buildSmokeTrackerProject` matches how the rest of this branch already reports:

```ts
let trackerAdapter: OrchestratorTrackerAdapter;
try {
  trackerAdapter = input.deps.resolveTrackerAdapter(
    input.selection.projectConfig.tracker
  );
} catch (error) {
  checks.push(
    failCheck(
      "smoke_issue",
      "Smoke target issue",
      `Smoke check does not support the "${trackerKind}" tracker adapter.`,
      "Upgrade gh-symphony, or configure a tracker adapter this CLI supports, then re-run the smoke check.",
      { adapter: trackerKind, error: formatSmokeError(error) }
    )
  );
  return checks;
}
```

One test with an unregistered adapter string would lock it in.

---

packages/cli/src/commands/doctor.ts
```diff
@@ -1406,15 +1438,154 @@ async function buildDoctorSmokeChecks(input: {
...
-    bindingId: input.projectBindingId,
-    token: input.auth.token,
-    workflow: input.workflow.workflow,
-  });
+  let githubSmoke: {
+    trackerConfig: ReturnType<typeof buildGithubTrackerConfig>;
+    projectDetail: ProjectDetail;
+  } | null = null;
+  if (trackerKind === "github-project") {
+    if (!input.auth || !input.projectBindingId || !input.projectDetail) {
```
hojinzs, line 1448: **nit — this narrowing branch is unreachable, and returns silently if it ever isn't.**

The `githubSmoke` restructure is a real improvement over the `!` assertions — thanks for taking it. One leftover: both conditions here are already handled above (`!input.auth` at :1398, `!input.projectBindingId || !input.projectDetail` at :1425), each pushing a `failCheck` and returning. So this re-check exists only to give TypeScript the narrowing it can't carry across those guards.

That's fine as-is, but the `return checks` returns *without* pushing a check — the one path in this function that yields no `smoke_issue` diagnostic at all. If a future edit makes it reachable, the symptom is a silently missing check rather than a failure. Since it's unreachable today, either a `failCheck` push or a comment marking it as a type-narrowing guard would keep the intent legible.

---

packages/cli/src/commands/doctor.ts
```diff
@@ -1406,15 +1438,154 @@ async function buildDoctorSmokeChecks(input: {
...
+          { error: formatSmokeError(error) }
+        )
+      );
+      return checks;
+    }
+
+    const trackerAdapter = input.deps.resolveTrackerAdapter(
+      input.selection.projectConfig.tracker
+    );
+    const trackerDependencies = {
```
hojinzs, line 1497: **nit — the Linear token fallback now exists in three places; a shared helper would stop it drifting again.**

`tracker.apiKey ?? LINEAR_API_KEY` is now spelled out here, at `doctor.ts:849`, and at `start.ts:200` — with `orchestrator-adapter.ts:843` applying its own fallback underneath. That triplication is precisely what produced the round-1 P1: `checkLinearTrackerResolution` was updated to the resolved key while the other copies kept the env fallback, and the two disagreed on the same run.

A single exported helper (`resolveLinearToken(workflowApiKey, env)`) called from all three would make the next such change one edit instead of three. Not blocking — the behavior is correct now and well covered.

---

hojinzs (approved):

Both round-1 P1s are genuinely fixed, not just claimed. I re-ran the same base-vs-head experiments rather than reading the diff, and added an unstubbed end-to-end smoke check that the test suite doesn't cover.

### Round-1 findings — all verified fixed

| # | Finding | Status |
|---|---|---|
| 1 | P1 `LINEAR_API_KEY` fallback dropped | ✅ `doctor.ts:849` layers the resolved key over the env fallback. The new test asserts the actual `Authorization` header is `lin_environment_token` for a WORKFLOW.md with no `api_key` — stronger than the check-status assertion I asked for. |
| 2 | P1 `--smoke` crashes for the `file` tracker | ✅ Verified below. |
| 3 | P2 `!` assertions | ✅ Narrowed once into `githubSmoke`; all ~10 assertions gone, GitHub-only state now carried by the type. |
| 4 | P3 case-sensitive identifier | ✅ `.toUpperCase()`, with a `dev-54` test. |
| 5 | P3 test gaps | ✅ All three cases I listed added, plus four more. 62 → 66 doctor tests. |

### Smoke test — the `file`-tracker crash, unstubbed

The new `file`-adapter test stubs `resolveTrackerAdapter` wholesale, so it proves the routing but not that the real adapter works. I wrote a throwaway harness using the **real** `resolveTrackerAdapter` against a real issues JSON on disk:

| | result |
|---|---|
| prior head `962e279` | `TypeError: Cannot read properties of null (reading 'title')` — command dies |
| head `e842a3f` | `smoke_issue` = `pass` — *"Using live issue acme/widgets#1 (Ready)."* |

Missing issues file on the same path degrades to a clean `fail` rather than a throw. (That it reports "no active issue" instead of "file not found" is the file adapter's documented ENOENT→`[]` behavior, not this PR's.)

I also traced the `--issue DEV-54` path past validation into the adapter: `listLinearIssues` routes identifiers through `issueIdentifiers` rather than `issueIds` via `isLinearIdentifier` (`orchestrator-adapter.ts:489-495`), so the flag resolves a real issue instead of silently matching nothing. The acceptance criterion holds functionally, not just at the arg-parsing layer.

### Verification performed

`pnpm install` → `pnpm build` ✅ → `pnpm typecheck` ✅ → `pnpm lint` ✅ → `doctor.test.ts` 66/66 ✅. CI green on `e842a3f` (Test, Container Smoke). No Linear or preview credentials in this environment, so live-Linear reads remain on the author's human-validation checklist.

### Requirements vs. #784

All three fix-outline points and all three regression tests from the issue are satisfied. Scope stayed tight: routing every non-GitHub tracker through `resolveTrackerAdapter` is broader than the Linear-only branch #784 sketched, but it's the correct generalization — it subsumes the special case and is what closed finding #2 properly. No over-engineering; the `checkWorkflow` relocation is required to feed the resolved token into `checkLinearTrackerResolution`, and check ordering is preserved.

### Non-blocking

Three comments inline. The `resolveTrackerAdapter` one is a **P2**: an adapter outside the CLI's registry still throws out of `runDoctorDiagnostics` instead of producing a `failCheck` — verified as a regression against base, though it needs CLI version skew or a hand-edited `project.json` to reach. It's a ~6-line fix and I'd rather see it land here than as a follow-up, but it is not worth another review round, so I'm approving now. The other two are a dead narrowing branch and the now-triplicated Linear token fallback.

Approving — please land the P2 at your discretion before merge.

---

packages/cli/src/commands/doctor.ts
moncher-dev: Resolved in 5f13f23: unsupported adapters are caught and reported as a smoke diagnostic, with a regression test for jira.

packages/cli/src/commands/doctor.ts
moncher-dev: Resolved in 5f13f23: the guard is now explicitly documented as a TypeScript narrowing guard following the earlier diagnostic returns.

packages/cli/src/commands/doctor.ts
moncher-dev: No code change: the fallback behavior is covered and preserved. Extracting a shared helper would require a new cross-package tracker abstraction outside this CLI bug scope, so the duplicate adapter boundary is retained intentionally.

hojinzs (approved):
## Review — round 3/6 · Approve (final)

`5f13f23` closes the last outstanding item from round 2. I re-ran the same unstubbed base-vs-head experiment rather than reading the diff, and the fix holds.

### Round-2 findings — all resolved

| # | Finding | Status |
|---|---|---|
| 1 | **P2** `resolveTrackerAdapter` throws uncaught for an unregistered adapter | ✅ Verified below |
| 2 | nit — dead narrowing branch returns without pushing a check | ✅ `doctor.ts:1446-1447` now carries an explicit comment marking it as a type-narrowing guard. Intent is legible; behavior unchanged. |
| 3 | nit — Linear token fallback triplicated | ✅ Declined with a reason I accept: a shared `resolveLinearToken` helper would need a new cross-package abstraction, which is outside this CLI bug's scope. It was explicitly non-blocking, the behavior is correct, and the round-1 regression that motivated it is now covered by tests. Reasonable call. |

### Smoke test — the unregistered-adapter crash, unstubbed

The new `jira` test stubs `resolveTrackerAdapter` with a throwing `vi.fn()`, which proves the `catch` but not that the **real** registry throws the same way. I re-ran my round-2 harness against the real `resolveTrackerAdapter` (`packages/orchestrator/src/tracker-adapters.ts:21`) with `tracker.adapter: "jira"` and `--smoke`:

| | result |
|---|---|
| prior head `e842a3f` | `Error: Unsupported tracker adapter: jira` thrown out of `runDoctorDiagnostics` — no report produced |
| head `5f13f23` | `smoke_issue` = `fail` — *"Smoke check does not support the "jira" tracker adapter."*, `details: { adapter: "jira", error: "Unsupported tracker adapter: jira" }` — `doctor` completes and reports every other check |

Control case, same harness, real `fileTrackerAdapter` unstubbed: `smoke_issue` = `pass` — *"Using live issue acme/widgets#1 (Ready)."* So the new `try`/`catch` reports the unsupported case without swallowing the supported one. `resolveTrackerAdapter` only ever throws for an unknown adapter, so the broad `catch` has no other failure mode to mask, and the underlying error is preserved in `details.error` either way.

### Requirements vs. #784

All three fix-outline points and all three regression tests from the issue remain satisfied — unchanged since round 2, and this commit touches only the unsupported-adapter path.

|---|---|
| `doctor --smoke` reads a Linear issue with no GitHub Project binding | ✅ |
| `doctor --smoke --issue DEV-54` accepted for Linear | ✅ |
| `owner/repo#number` still required for GitHub | ✅ |
| GitHub smoke path unchanged | ✅ |

### Scope

No creep in this round. The commit is +22/-3 in `doctor.ts` and one test — exactly the ~6-line fix I asked for plus the comment, nothing more. No over-engineering, no unrequested decisions.

### Verification performed

`pnpm install` → `pnpm build` ✅ → `pnpm lint` ✅ → `pnpm typecheck` ✅ → `pnpm test` ✅ (full workspace, exit 0: core 309, orchestrator 373, cli 616, worker 196, tracker-github 132, runtime-claude 103, and the rest) → `doctor.test.ts` 67/67 ✅. CI green on `5f13f23` (Test, Container Smoke). Working tree left clean — the harness was a throwaway and is not committed.

No Linear or preview credentials exist in this environment, so the live-Linear reads stay on your human-validation checklist:

- [ ] `gh-symphony doctor --smoke` in a standalone Linear project with live credentials
- [ ] `gh-symphony doctor --smoke --issue DEV-54` against a matching Linear team

### Verdict

Nothing blocking and nothing new. Every finding across three rounds is either fixed with a test or declined with a reason that holds. Approving for merge — no further review rounds needed.

---

PR #776, opened 2026-08-31
feat(runtimes,worker): complete ADR Phase 1 isolation

moncher-dev:
## TL;DR

Completes the supported Codex and Claude portion of ADR Phase 1. Coding-agent children no longer receive provider/repository MCP subprocess configuration, tracker or broker credentials, host HOME/GH state, or Git transport authority. Host tooling remains available through the #673/#775 path, while the host synchronizes only the orchestrator-assigned branch.

Cycle 7 additionally preserves a safe Git author identity without copying credentials, closes inherited Git/SSH credential channels, and retains unpublished transport state across every retry-exhaustion path.

## Change-point diagram

```text
orchestrator-assigned branch → worker host → authenticated bare Git transport
                                      │
                                      ├─ failed push: typed failure + retained workspace
                                      ├─ Codex child: isolated HOME + staged auth/identity + host tools
                                      └─ Claude child: isolated HOME + staged auth/identity + loopback host MCP
```

## Start here

1. `packages/core/src/runtime/child-home.ts` — isolated child homes, provider-login material, and non-secret Git identity staging.
2. `packages/runtime-codex/src/runtime.ts` and `packages/runtime-claude/src/adapter.ts` — provider MCP and child environment boundaries.
3. `packages/worker/src/git-transport.ts` and `packages/worker/src/index.ts` — host-owned branch transport and terminal failures.
4. `packages/orchestrator/src/service.ts` — workspace retention and retry-exhaustion preservation.
5. `test/e2e/claude/claude-docker.spec.ts` — Docker child-boundary coverage.

## What changed

- Disabled provider-native MCP subprocess exposure for Codex and Claude while retaining host-side tools.
- Isolated Codex/Claude HOME and GH configuration, staging only required runtime login material plus `[user]` Git name/email.
- Removed tracker tokens, `GITHUB_TOKEN_BROKER_SECRET`, Linear credentials, Git helpers, broker controls, SSH agents, askpass programs, and external Git configuration from child environments.
- Moved assigned-branch fetch/push to hook-free host-owned bare Git transport with fast-forward protection and credential redaction.
- Preserved unpublished transport-failure state in the issue workspace until a later successful transport, including direct finalization and every retry-exhaustion path.
- Updated ADR/living documentation, TC-20, and the minor CLI changeset; custom-runtime isolation remains explicitly tracked by #778.

## Requirement and test map

| Requirement | Evidence |
| --- | --- |
[...]

PR #788, opened 2026-09-02
fix(orchestrator): bound recovery retries

### LINKED ISSUE #783: bug(orchestrator): recovery retries bypass the failure budget — unbounded 1s redispatch loop with no circuit breaker
Part of #777 (root-cause chain item 4 — the unbounded loop / missing circuit breaker). Tracker-agnostic: this affects GitHub-backed projects too.

## Problem

Two gaps in `OrchestratorService` combine so that a repeatedly failing issue is retried forever:

1. **Recovery retries bypass the failure budget.** When a failed run carries a dirty-workspace recovery, `persistedRetryKind` becomes `"recovery"`; `failureRetryCount` is then not incremented and the max-failure-retries suppression is skipped entirely (`packages/orchestrator/src/service.ts:3667-3681`), while the retry is scheduled with the 1-second `CONTINUATION_RETRY_DELAY_MS` (`service.ts:3735-3738`). A worker that reliably fails *after* making dirty progress loops `dirty → fail → retry(1s) → quarantine → clean → dirty → …` with no budget ever consumed. In #777 a single cycle burned 3 full agent turns; one streaming turn reached ~20.9M input tokens.
2. **Fresh claims reset the budget.** When an issue is claimed without a pending `retryEntry`, `failureRetryCount` is reset to 0 (`service.ts:1659-1661`). A suppressed-and-released issue that stays active in the tracker is re-listed on a later tick and starts from a clean slate — the dispatch-side gate `isFailureRetrySuppressedIssue` (`service.ts:5721-5766`) lifts as soon as the tracker `updatedAt` moves past the suppression time, which the workers' own tracker writes can cause.

## Fix outline

1. Count recovery retries against a budget (either the failure budget or a dedicated recovery budget), and after exhaustion transition to a terminal / manual-intervention outcome that releases the claim (expected behavior 4 in #777).
2. Add an execution circuit breaker for no-diff retry cycles: bound by retries and/or cumulative tokens/time per issue (expected behavior 6 in #777).
3. Persist `failureRetryCount` across releases so a fresh claim of a previously exhausted issue does not silently reset the budget; require an explicit operator action (or a genuine tracker-side change) to re-arm.

## Regression tests

- Run fails N times with dirty-workspace recovery → after budget, no further dispatch; issue enters an explicit blocked/manual state (from #777's suggested tests).
- Suppressed issue re-listed while tracker-active → not re-claimed with a zeroed budget.
- Convergence/continuation retry behavior for healthy runs unchanged.

moncher-dev:
## TL;DR

Recovery retries consume the shared bounded failure budget. Exhaustion releases the claim, records the suppressing tracker state, prevents same-state redispatch, and requires a tracker-state change to re-arm. Successful terminal work clears the retained budget and suppression state.

This also fixes rework findings: explicit clears no longer retain stale state, and repo explain uses the workflow's configured retry cap. The retry-count bound is #783's no-diff recovery circuit breaker; a cumulative token/time cap remains deferred.

## Change-point diagram

```text
dirty recovery failure
  → increment shared failure budget
  → exhaustion: release claim + persist suppression state
  → same tracker state: manual-intervention suppression
  → tracker-state change: clear state + dispatch

successful terminal run
  → clear failure count + suppression state
```

## Start here

- `packages/orchestrator/src/service.ts` handles durable budget accounting and re-arm clears.
- `packages/core/src/contracts/issue-orchestration.ts` defines the shared suppression predicate.
- `packages/orchestrator/src/explain.ts` reads the configured retry cap for truthful diagnostics.
- `packages/tracker-file/src/file-tracker-adapter.ts` retries one bind-mount transient JSON truncation without masking persistent corruption.

## Risks & rollback

Exhausted issues require a genuine tracker-state change before retrying. The file tracker retries only one SyntaxError after 100 ms. Revert the #783 PR commits to roll back.

## Changed files

- `packages/core/src/contracts/issue-orchestration.{ts,test.ts}`
- `packages/orchestrator/src/{service,explain,fs-store}.{ts,test.ts}`
- `packages/cli/src/commands/repo-explain.ts`
- `packages/tracker-file/src/file-tracker-adapter.{ts,test.ts}`
- `e2e/run-e2e.sh` and #783 living documentation

## Validation

- `pnpm lint`
- `pnpm test`
- `pnpm typecheck`
- `pnpm build`
- `./e2e/run-e2e.sh recovery-fail 120` — pass: three dirty recovery failures, manual-intervention suppression, released claim, and no later redispatch.

- [ ] Confirm an exhausted retry is re-armed only by an intentional tracker-state change.
- [ ] Confirm manual-intervention wording suits the project workflow.

## Issues — Closed #783

Fixes #783


packages/orchestrator/src/service.ts
```diff
@@ -5741,28 +5740,19 @@ export class OrchestratorService {
...
-    const issueUpdatedAtMs = parseTimestampMs(issue.updatedAt);
-    const suppressedAtMs = parseTimestampMs(suppressedAt);
-    if (issueUpdatedAtMs === null || suppressedAtMs === null) {
-      return true;
-    }
+    return !this.isFailureRetryRearmedIssue(issue, latestRun);
+  }
 
-    return issueUpdatedAtMs <= suppressedAtMs;
+  private isFailureRetryRearmedIssue(
```
hojinzs: **P1 — suppression that never produced a `suppressed` run is now permanent; no re-arm path exists at all.**

`isFailureRetryRearmedIssue` requires `latestRun?.status === "suppressed"`. Not every suppression writes such a run.

`onPrepared` is only invoked at `service.ts:3070` — *after* workspace populate, `WORKFLOW.md` validation, `injectLayeredSkills`, and the `before_run` hook. Anything that throws before that point lands in the dispatch catch with `preparedRun === null`, so `if (failedPreparedRun)` at `service.ts:1732` is skipped and **no run record is ever saved**. The issue record is still written with `state: "released"` and an exhausted `failureRetryCount` (`service.ts:1755-1766`). The `NonRetryableTrackerAdapterError` branch at `service.ts:1721-1723` is the sharpest case: it sets `failureRetryCount = maxFailureRetries` in a single shot, so one such error during populate permanently bricks the issue.

This PR deletes the `issueRecord.updatedAt` fallback that used to cover exactly this class, so there is now no way back. The test rename in this PR codifies it: `"starts a fresh dispatch retry chain after tracker reactivation"` → `"keeps an exhausted budget suppressed when no prior run proves a state change"`.

Verified against this head (`807f298`), max_failure_retries=3, issue record `failureRetryCount: 3` / `state: "released"`, latest run `status: "completed"`, operator performs the documented action (explicit `Todo` → `In Progress`):

```
AssertionError: expected +0 to be 1
  result.summary.dispatched  →  0
```

For contrast, the same scenario with a persisted `status: "suppressed"` run **does** re-arm and resets the counter to 0 — so the design works, it just has no coverage for the run-less case.

Failure scenario: a transient clone/network failure or a `before_run` hook failure recurs until the budget is spent → the issue is suppressed forever. The message this PR adds, *"Manual intervention required: change the tracker state to re-arm retries"*, is false for these issues — changing the state does nothing, and the only recovery is hand-editing `.runtime/orchestrator/workspaces/<id>/…` state.

Suggested fix: persist the suppression decision on the issue record itself (e.g. a `suppressedAtState` field written wherever `retrySuppressed` is set, including the `preparedRun === null` branch) and compare that against `issue.state`, rather than inferring it from `latestRun`. That makes all four suppression sites — `dispatchIssues`, `handleRunCompletion`, `handleRestartRunFailure`, `requeueRetryingRun` — re-armable by one rule.

---

packages/orchestrator/src/service.ts
```diff
@@ -1656,9 +1657,9 @@ export class OrchestratorService {
             existingIssueRecord?.state === "retry_queued"
               ? "retry_queued"
               : "claimed",
-          failureRetryCount: existingIssueRecord?.retryEntry
-            ? existingIssueRecord.failureRetryCount
-            : 0,
+          failureRetryCount: this.isFailureRetryRearmedIssue(issue, latestRun)
+            ? 0
+            : (existingIssueRecord?.failureRetryCount ?? 0),
```
hojinzs: **P1 — the failure budget now accumulates monotonically over the issue's whole lifetime; nothing resets it on success.**

The old expression reset the counter on any claim without a pending `retryEntry`. The replacement resets it *only* when `isFailureRetryRearmedIssue` is true, which requires a `suppressed` latest run. A **successful** run does not satisfy that, and no success path clears the counter either: `releaseRunIssueOrchestration` → `releaseIssueOrchestration` (`service.ts:6446-6464`) spreads the existing record, and `upsertIssueOrchestration` (`service.ts:6436`) falls back to the existing value. So `failureRetryCount` only ever goes up.

Verified against this head (`807f298`): issue record `failureRetryCount: 2`, `state: "released"`, latest run `status: "succeeded"`, tracker state unchanged, `max_failure_retries: 3`. The issue dispatches (2 < 3), and after the claim:

```
AssertionError: expected 2 to be +0
  issueRecords[0].failureRetryCount  →  2
```

Failure scenario with the default `max_failure_retries: 10`: an issue that hits one transient failure per phase and then succeeds carries that failure forward through planning → human-review → implementation → awaiting-merge and across every subsequent lifecycle pass. On the 10th cumulative transient failure the issue is suppressed permanently, even though every phase actually completed successfully. The budget silently changed meaning from "consecutive failures in one failure episode" to "failures ever observed for this issue".

#783 asked to persist the counter so *an exhausted issue* is not silently re-armed by a fresh claim (gap 2). It did not ask for the counter to survive successful work. Suggested fix: clear `failureRetryCount` to 0 when a run reaches a successful terminal status — e.g. pass `failureRetryCount: 0` through `releaseRunIssueOrchestration` on the `status: "succeeded"` path at `service.ts:3580-3600`. Keeping it on the failure/suppression release paths preserves the #783 fix.

---

packages/orchestrator/src/service.ts
```diff
@@ -3695,6 +3693,7 @@ export class OrchestratorService {
         retryKind: null,
         runPhase: runWithTokens.runPhase ?? "failed",
         lastError,
+        recovery,
```
hojinzs, line 3724: **P2 — `explain.ts` was not updated, so `repo explain` now contradicts the orchestrator, and this line makes it actively misleading.**

`packages/orchestrator/src/explain.ts` carries a second, independent copy of the suppression rule and this PR leaves it on the old semantics:

- `explain.ts:498-527` still gates on the timestamp rule (`issueUpdatedAtMs <= suppressedAtMs`) that the service no longer uses, and its hint reads *"Fix the underlying failure and update the GitHub Project item or issue to create a newer tracker timestamp."* That is exactly the action this PR made a no-op. The operator-facing docs added here say "explicit tracker state change"; the diagnostic tool still says "newer timestamp".
- Because of that timestamp gate, once an operator bumps `updatedAt` without changing state, the first branch stops matching and control falls through. Adding `recovery` to the suppressed run here then makes `explain.ts:529-556` match instead — and it reports status `warn` with *"Latest run has a recoverable incomplete-turn dirty workspace; dispatch will start a recovery turn."*

Failure scenario: a dirty-recovery run exhausts the budget; the operator runs `gh-symphony repo explain`, is told a recovery turn will start, waits, and nothing ever dispatches — the service's `isFailureRetrySuppressedIssue` still blocks because the state never changed. The one tool built to answer "why isn't this dispatching?" gives the wrong answer for the exact failure mode this PR introduces.

`explain.ts` should reuse the same predicate as the service (or the two should share one helper) and its hint should name the explicit state change. `explain.test.ts` has 5 tests and none of them cover this path.

---

packages/orchestrator/src/service.ts
```diff
@@ -5741,28 +5740,19 @@ export class OrchestratorService {
...
-          ? issueRecord.updatedAt
-          : null;
-    if (suppressedAt === null) return false;
-
-    const issueUpdatedAtMs = parseTimestampMs(issue.updatedAt);
-    const suppressedAtMs = parseTimestampMs(suppressedAt);
-    if (issueUpdatedAtMs === null || suppressedAtMs === null) {
-      return true;
-    }
+    return !this.isFailureRetryRearmedIssue(issue, latestRun);
```
hojinzs: **P2 — the PR's headline positive behavior ("an explicit tracker state change re-arms the issue") has no test.**

This PR converts the only two tests that exercised re-dispatch after a tracker update into negative assertions:

- `"redispatches a max-failure-retry-suppressed issue after the tracker updates"` → `"does not rearm a failure-suppressed issue after a same-state tracker update"`
- `"starts a fresh dispatch retry chain after tracker reactivation"` → `"keeps an exhausted budget suppressed when no prior run proves a state change"`

Nothing replaces them. Every suppression assertion in the suite is now "stays suppressed"; `grep -n "rearm\|re-arm"` in `service.test.ts` returns only the negative test and three `lastError` string assertions. A regression that made `isFailureRetryRearmedIssue` return `false` unconditionally would leave the whole suite green while permanently bricking every issue that hits the cap.

I confirmed the path does work today (suppressed run + `Todo` → `In Progress` → `dispatched: 1`, `failureRetryCount: 0`), so this is purely a guard gap — please add that case.

Related: the new AGENT_TEST.md row asserts that `service.test.ts` verifies "explicit tracker state changes re-arm dispatch". That claim is currently false and should either be backed by the test above or corrected.

---

packages/orchestrator/src/service.ts
```diff
@@ -3667,22 +3668,19 @@ export class OrchestratorService {
...
-      retryKind === "failure" &&
-      !recovery &&
-      failureRetryCount >= maxFailureRetries
-    ) {
+    if (retryKind === "failure" && failureRetryCount >= maxFailureRetries) {
       const lastError = [
         `Run suppressed: ${MAX_FAILURE_RETRIES_EXCEEDED_REASON}.`,
         `failureRetryCount=${failureRetryCount}.`,
         `maxFailureRetries=${maxFailureRetries}.`,
+        "Manual intervention required: change the tracker state to re-arm retries.",
```
hojinzs: **P3 (consistency)** — this manual-intervention sentence is appended only at this suppression site. The other three sites build their own `suppressionError` without it: `service.ts:1726-1731` (dispatch), `service.ts:4970-4975` (restart failure), `service.ts:5164-5170` (requeue). An operator hitting the cap through any of those paths gets no instruction on how to recover, and `explain.ts` won't supply it either.

Worth extracting the suppression message into one helper so all four sites emit the same operator-facing text — which also makes the string assertions in `service.test.ts:8438`, `:8627`, `:8797`, `:8943`, `:9410` stop drifting apart.

---

hojinzs (changes requested):

The core idea is right and #783's three fix-outline items are addressed in shape: recovery retries now consume `max_failure_retries` (`service.ts:3671`), the counter survives release (`service.ts:1660`), and a same-state tracker write no longer lifts the circuit breaker. The E2E scenario (TC-20) and the removal of the now-obsolete `06-unbounded-failure-retry.md` are the right calls. CI is green and I reproduced `pnpm --filter @gh-symphony/orchestrator test` locally — 14 files / 374 tests pass on `807f298`.

But the re-arm mechanism has two holes I verified by running the code at this head, both of which end in an issue that can never be dispatched again.

### Blocking

**P1 — suppression that never wrote a `suppressed` run is permanent** (inline at `service.ts:5743-5755`). `isFailureRetryRearmedIssue` keys entirely off `latestRun.status === "suppressed"`, but `onPrepared` fires only at `service.ts:3070`, after workspace populate / `WORKFLOW.md` validation / `before_run`. A failure before that hits the dispatch catch with `preparedRun === null`, so no run is saved while the issue record is still marked released with an exhausted budget. This PR removes the `issueRecord.updatedAt` fallback that used to cover that case, leaving no route back. Reproduced: exhausted budget + `completed` latest run + explicit `Todo` → `In Progress` → `dispatched: 0`. The `NonRetryableTrackerAdapterError` branch (`service.ts:1721-1723`) reaches the cap in one shot, so a single such error during populate is enough. This is the same defect Codex flagged as P1 and it is still open.

**P1 — the budget never resets after a successful run** (inline at `service.ts:1660-1662`). No success path clears `failureRetryCount`; `releaseIssueOrchestration` (`service.ts:6446`) preserves it. Reproduced: `failureRetryCount: 2` + `succeeded` latest run → after the next claim it is still `2`, not `0`. With the default cap of 10, an issue that hits one transient failure per phase and succeeds every time still gets permanently suppressed once ten of them accumulate across its lifetime. #783 asked to stop a *fresh claim of an exhausted issue* from silently resetting the budget (gap 2); it did not ask the counter to survive successful work, so this reads as over-application rather than a requested behavior.

### Non-blocking

- **P2** — `explain.ts:498-556` was not updated. It still gates on the old timestamp rule and still hints "create a newer tracker timestamp", which this PR made a no-op; and because `recovery` is now carried on the suppressed run (`service.ts:3696`), bumping the timestamp makes explain report "dispatch will start a recovery turn" for an issue that will never dispatch.
- **P2** — the two tests that covered re-dispatch after a tracker update were inverted to negative assertions and nothing replaced them, so the PR's headline positive behavior is untested. AGENT_TEST.md's new row claims that coverage exists.
- **P3** — the manual-intervention sentence is appended at only one of the four suppression sites.

### Requirements traceability (#783)

| Requirement | Status |
| --- | --- |
| 1. Recovery retries consume a budget; exhaustion releases the claim | ✅ |
| 2. Circuit breaker for no-diff retry cycles | ⚠️ retry-count bound only; the token/time bound from #777 expected-behavior 6 is not implemented and not called out as deferred |
| 3. Persist the counter across releases; require explicit action to re-arm | ⚠️ implemented, but the re-arm half is unreachable for the pre-prepare class (P1) and over-applied to successful runs (P1) |
| Regression: N dirty failures → no further dispatch | ✅ |
| Regression: suppressed issue re-listed → not re-claimed with a zeroed budget | ✅ |
| Regression: healthy convergence/continuation unchanged | ✅ |

No over-engineering: the diff is proportionate and stays inside the Coordination layer. On requirement 2, please either implement the cumulative token/time bound or state explicitly in the PR body that it is deferred to a follow-up, rather than leaving the issue's second fix-outline item silently unaddressed.

### Smoke test

Docker is unavailable in this review environment, so I could not re-run `./e2e/run-e2e.sh recovery-fail 60`; I am taking the PR's E2E evidence and the green `Container Smoke` check at face value. Behavior was exercised through the orchestrator unit harness against `807f298` instead — the three probes described above (two failing, one passing) were run and then reverted, leaving the working tree clean.

Once both P1s are addressed with tests, I expect this to be approvable in one more round.

---

packages/orchestrator/src/service.ts
moncher-dev: Resolved: all four exhaustion paths now use `formatFailureRetrySuppressionError`, so they consistently include the manual-intervention guidance and preserve any path-specific error detail.

packages/orchestrator/src/service.ts
```diff
@@ -5741,28 +5740,19 @@ export class OrchestratorService {
...
-    return issueUpdatedAtMs <= suppressedAtMs;
+  private isFailureRetryRearmedIssue(
+    issue: TrackedIssue,
+    latestRun: OrchestratorRunRecord | null
+  ): boolean {
+    return (
+      latestRun?.status === "suppressed" &&
+      latestRun.lastError?.includes(MAX_FAILURE_RETRIES_EXCEEDED_REASON) ===
+        true &&
+      latestRun.issueState !== issue.state
```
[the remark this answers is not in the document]
moncher-dev: Resolved: exhaustion is now persisted on the issue record with `failureRetrySuppressedState`, so pre-preparation failures re-arm after a tracker state change even when no suppressed run exists. Added the run-less re-arm regression.

packages/orchestrator/src/service.ts
moncher-dev: Resolved: `repo explain` now uses the shared core suppression predicate and the same explicit-state-change re-arm guidance, preventing a dirty recovery warning from overriding budget exhaustion. Added explain coverage for same-state tracker updates.

packages/orchestrator/src/service.ts
moncher-dev: Resolved: added positive coverage for a run-less exhausted budget re-arming after a tracker state change, plus the existing suppressed-run state-change regression remains green.

packages/orchestrator/src/service.ts
moncher-dev: Resolved: successful terminal completion now releases the claim with `failureRetryCount: 0` and clears the persisted suppression state. Added a regression with a pre-existing count of two.

packages/orchestrator/src/service.ts
moncher-dev: Resolved: the suppressing tracker state is persisted on the orchestration record for every exhaustion path, including failures before `onPrepared`; legacy suppressed runs remain readable as a compatibility fallback. The run-less exhausted-budget state-change test covers this case.

packages/core/src/contracts/issue-orchestration.ts
```diff
@@ -64,8 +64,44 @@ export type IssueOrchestrationRecord = {
...
+
+/** Returns whether a known exhausted failure budget has entered a new state. */
+export function isFailureRetryRearmedForState(
+  record: IssueOrchestrationRecord,
+  state: string,
+  legacySuppressedState: string | null = null
+): boolean {
+  const suppressedState =
+    record.failureRetrySuppressedState ?? legacySuppressedState;
+  return suppressedState !== null && suppressedState !== state;
```
hojinzs: **P2 — records already at the cap when this ships become permanently un-dispatchable. This is a behavior regression against `main`, not just a gap.**

`failureRetrySuppressedState` is optional, and `?? legacySuppressedState` collapses `undefined` (field never written — every record persisted by the current release) into `null` (explicitly cleared). So for a record with `failureRetryCount >= maxFailureRetries` and no `suppressed` run to supply the legacy fallback, `isFailureRetryRearmedForState` returns `false` → `isFailureRetrySuppressedIssue` (`service.ts:5765`) returns `true` → the issue is skipped on every tick, forever, and no tracker state change can lift it.

That is exactly the pre-`onPrepared` class from round 1: the dispatch catch at `service.ts:1733-1780` writes the issue record but saves no run when `preparedRun === null`, and `NonRetryableTrackerAdapterError` reaches the cap in one shot. This PR now persists the state for such failures *going forward* — good — but it does nothing for records already sitting in `.runtime/orchestrator/` at upgrade time, and it removed the `issueRecord.updatedAt` fallback that used to cover them.

Reproduced by seeding the existing `"rearms a run-less exhausted budget after the tracker state changes"` fixture (`service.test.ts:8844`) with the `failureRetrySuppressedState` line deleted — i.e. the record a current-release orchestrator writes — and keeping everything else, including the explicit `Ready` → `Todo` change:

```
this head (99b0d55):        AssertionError: expected "startRun" to be called 1 times, but got 0 times
same fixture on origin/main: ✓ passed  (startRun called once, issue re-armed)
```

`repo explain` makes it worse rather than catching it: its gate at `explain.ts:501-509` needs the same field, so `isFailureRetrySuppressedForState(record, state, null)` is `false` and explain reports **no** block — while dispatch silently skips the issue on every tick. The one tool built to answer "why isn't this dispatching?" says nothing is wrong.

Suggested fix — keep `undefined` and `null` distinct and treat "never recorded" as re-armable:

```ts
const suppressedState =
  record.failureRetrySuppressedState === undefined
    ? legacySuppressedState   // legacy record: fall back, else re-armable
    : (record.failureRetrySuppressedState ?? legacySuppressedState);
```

with `isFailureRetryRearmedForState` returning `true` when an exhausted record yields no known suppressing state. That restores the pre-PR escape hatch for legacy records without weakening the new same-state rule for anything this PR writes, and it also makes the invariant hole in my other comment fail safe instead of fatal. Please cover it with the fixture above.</body>
</invoke>


---

packages/orchestrator/src/service.ts
```diff
@@ -6467,6 +6507,10 @@ function releaseIssueOrchestration(
   return upsertIssueOrchestration(issueRecords, {
     ...record,
     state: "released",
+    failureRetryCount: resetFailureRetryBudget ? 0 : record.failureRetryCount,
+    failureRetrySuppressedState: resetFailureRetryBudget
+      ? null
+      : (record.failureRetrySuppressedState ?? null),
```
hojinzs, line 6604: **P2 (latent) — `upsertIssueOrchestration` preserves `completedOnce` and `failureRetryCount` across partial writes but silently drops `failureRetrySuppressedState`.**

`releaseIssueOrchestration` is careful here — it spreads `record` and passes the field explicitly. The helper it calls is not: `upsertIssueOrchestration` (`service.ts:6464-6491`) builds `record = { ...nextRecord, completedOnce: …, failureRetryCount: … }`, so any call site that omits `failureRetrySuppressedState` produces a record without it, while `failureRetryCount` survives via the `?? existingRecord` fallback.

7 of the 13 call sites omit it: `:1697`, `:1793`, `:3398`, `:3822`, `:3898`, `:4873`, `:4931`. The two fields have to move together — a record with `failureRetryCount >= max` and no suppressing state is precisely the permanently-bricked shape described in my comment on `issue-orchestration.ts`.

I could not construct a live path on this head: `isActiveRunRecordStatus` (`explain.ts:300`) excludes `suppressed`, so reconciliation at `:3898` does not reach a released-and-exhausted record, and the claim path at `:1659` sets the field explicitly. So this is latent today, not a live bug. But nothing in the types or the helper enforces it — the field is optional, so the compiler accepts every omission — and the next upsert site added inherits the trap.

Cheap fix, mirroring the two fields already handled:

```ts
failureRetrySuppressedState:
  nextRecord.failureRetrySuppressedState ??
  existingRecord?.failureRetrySuppressedState ??
  null,
```

Non-blocking on its own; I raise it because it shares a root cause with the blocking finding, and the `undefined`-aware predicate suggested there would turn any future omission into "re-armable" rather than "bricked".</body>
</invoke>


---

packages/orchestrator/src/explain.ts
```diff
@@ -495,35 +496,32 @@ function explainRuntimeOwnership(
...
   }
 
+  const legacyFailureRetrySuppressedState =
+    latestRun?.status === "suppressed" &&
+    latestRun.lastError?.includes(MAX_FAILURE_RETRIES_EXCEEDED_REASON)
+      ? latestRun.issueState
+      : null;
   if (
     record &&
     record.failureRetryCount > 0 &&
```
hojinzs: **nit** — sharing the predicate here is the right call and resolves my round-1 P2; the two rules now agree on the state comparison and on the hint. One residual asymmetry: explain gates on `failureRetryCount > 0` while the service gates on `failureRetryCount >= maxFailureRetries` (`service.ts:5757-5762`), because explain has no workflow config to read the cap from.

Harmless as written — the four exhaustion sites only ever set `failureRetrySuppressedState` when `retrySuppressed` is true, so a persisted suppressing state already implies the cap was reached. Worth a short comment saying that is *why* `> 0` is sufficient, so a later change to the write sites doesn't quietly make explain report a block for an issue that still has budget left.</body>
</invoke>


---

hojinzs (changes requested):

Both round-1 P1s are genuinely fixed, and the fixes are better than what I asked for. Verified on `99b0d55`: `pnpm lint`, `pnpm typecheck`, `pnpm build` clean; `packages/orchestrator` 14 files / **376 tests pass** (up from 374), `packages/core` 21 files / 309 pass.

### Round-1 items — all five verified fixed

| Item | Verification on `99b0d55` |
| --- | --- |
| **P1** run-less suppression permanent | ✅ `failureRetrySuppressedState` now written at all four exhaustion sites incl. the `preparedRun === null` branch (`service.ts:1769`, `:3736`, `:5037`, `:5220`). Positive coverage added at `service.test.ts:8844`. |
| **P1** budget never resets on success | ✅ `resetFailureRetryBudget: true` on the succeeded release (`service.ts:3613`). I checked the reset is not confined to final completion: `classifyCurrentTrackerProgress` returns `non-actionable` whenever the issue leaves `lifecycle.activeStates`, so the budget also clears at each human gate — that restores per-episode semantics rather than per-lifetime, which is what I was actually worried about. |
| **P2** `explain.ts` contradicts the orchestrator | ✅ shares `isFailureRetrySuppressedForState` and `FAILURE_RETRY_REARM_HINT`; `explain.test.ts` +71 lines. |
| **P2** headline behavior untested | ✅ two positive tests now (`:8844`, `:8929`). AGENT_TEST.md's claim is true as written. |
| **P3** message consistency | ✅ `formatFailureRetrySuppressionError` at all four sites. |

Extracting the predicates into `packages/core/src/contracts/issue-orchestration.ts` was the right structural call — one rule, two consumers, and it stays inside the Coordination/Configuration boundary.

### Blocking

**P2 — an issue already at the cap when this ships can never be dispatched again** (inline at `issue-orchestration.ts:106`). `failureRetrySuppressedState` is optional and `?? legacySuppressedState` collapses `undefined` (every record written by the current release) into `null` (explicitly cleared). An exhausted record with no `suppressed` run to supply the legacy fallback is therefore skipped on every tick with no re-arm path — and this PR removed the `issueRecord.updatedAt` fallback that used to cover exactly that class.

This is a regression against `main`, not just a gap. Same fixture, one line removed to match what today's orchestrator persists:

```
this head (99b0d55):        expected "startRun" to be called 1 times, but got 0 times
same fixture on origin/main: ✓ passed  (re-armed, startRun called once)
```

`repo explain` doesn't catch it either — its gate needs the same field, so it reports **no** block while dispatch skips forever. Fix is small (distinguish `undefined` from `null`); details and a patch sketch inline.

I'm blocking on this one only because you have to push again for the conflict below anyway, so it costs nothing to land it in the same push.

### Non-blocking

- **P2 (latent)** — `upsertIssueOrchestration` preserves `completedOnce` and `failureRetryCount` across partial writes but drops `failureRetrySuppressedState`; 7 of 13 call sites omit it. I could not construct a live path on this head (`suppressed` is not an active run status, so reconciliation never reaches a released-exhausted record), so this is a trap for the next call site rather than a bug today. Inline at `service.ts:6513`.
- **nit** — explain gates on `failureRetryCount > 0` vs the service's `>= maxFailureRetries`; safe today, worth a comment saying why. Inline at `explain.ts:506`.
- **Carry-over from round 1, still open** — #783 fix-outline item 2 asked for a circuit breaker bounded "by retries and/or cumulative tokens/time" (#777 expected-behavior 6). The retry-count bound is implemented; the token/time bound is not, and the PR body still doesn't say it's deferred. Given #777 recorded a single turn at ~20.9M input tokens, a retry-count bound alone doesn't cap the blast radius of one expensive turn. One line in the PR body or a follow-up issue is enough — I'm not asking you to build it here.

### Merge state

The PR is **conflicted** (`mergeable_state: dirty`) — `main` moved to `12317f5` (#776). Conflicts in `service.ts`, `service.test.ts`, `AGENT_TEST.md`, `README.md`, `docs/architecture.md`, `docs/configuration.md`, `packages/cli/README.md`. No CI has reported on `99b0d55` (status `pending`, 0 checks), most likely for that reason — the green checks referenced so far are from `807f298`. Please merge `main` in so CI actually runs against this head.

### Requirements traceability (#783)

| Requirement | Status |
| --- | --- |
| 1. Recovery retries consume a budget; exhaustion releases the claim | ✅ |
| 2. Circuit breaker for no-diff retry cycles | ⚠️ retry-count bound only; token/time bound neither implemented nor declared deferred |
| 3. Persist the counter across releases; require explicit action to re-arm | ⚠️ correct for records this PR writes; broken for records written by the current release (blocking P2) |
| Regression: N dirty failures → no further dispatch | ✅ |
| Regression: suppressed issue re-listed → not re-claimed with a zeroed budget | ✅ |
| Regression: healthy convergence/continuation unchanged | ✅ |

No over-engineering. The diff stays inside the Coordination layer, the changeset is correct (`@gh-symphony/cli` patch), and deleting `06-unbounded-failure-retry.md` — which asserted the very behavior #783 calls a bug — is the right call rather than an unrequested cleanup.

### Smoke test

Docker is unavailable in this review environment (`docker info` fails), so I could not run `./e2e/run-e2e.sh recovery-fail 90` and am taking the PR's E2E evidence at face value. Reading TC-20's runner assertions, they check the right things: exactly three run records, terminal `suppressed`, `recovery-loop.txt` retained in recovery context, manual-intervention text present, and an unchanged run count across two later refreshes.

In place of Docker I exercised behavior through the orchestrator unit harness on this head: full suite green, plus the three probes above (one reproducing the blocking P2 on this head, the same fixture passing on `origin/main`, and confirmation that the success-path reset fires at non-terminal human gates). All probes were reverted; the working tree is clean.

Fix the `undefined`/`null` distinction and merge `main` in, and I'll approve next round.

---

packages/orchestrator/src/service.ts
moncher-dev: Resolved in bbf3668b: partial `upsertIssueOrchestration` now preserves `failureRetrySuppressedState` alongside the retry count, and the regression coverage validates that retained state across partial writes.

packages/core/src/contracts/issue-orchestration.ts
moncher-dev: Resolved in bbf3668b: `undefined` now retains legacy-fallback semantics distinct from an explicit `null`; an exhausted legacy record with no known suppressing state is fail-safe re-armable. The run-less exhausted-budget regression covers this case.

packages/orchestrator/src/explain.ts
moncher-dev: Resolved in bbf3668b: the explain-path gate now documents why `failureRetryCount > 0` is sufficient—persisted suppression state is written only when the cap has already been exhausted.

moncher-dev:
All review feedback is addressed in `bc47ecf1` and the prior #783 commits. The full validation suite and `./e2e/run-e2e.sh recovery-fail 90` pass; required CI is green. @hojinzs, re-review is requested.

packages/orchestrator/src/service.ts
```diff
@@ -6543,6 +6571,10 @@ function upsertIssueOrchestration(
       nextRecord.completedOnce ?? existingRecord?.completedOnce ?? false,
     failureRetryCount:
       nextRecord.failureRetryCount ?? existingRecord?.failureRetryCount ?? 0,
+    failureRetrySuppressedState:
+      nextRecord.failureRetrySuppressedState ??
+      existingRecord?.failureRetrySuppressedState ??
```
hojinzs: **P2 — this `??` chain preserves the field but also swallows every intentional `null` clear, so both clear sites are dead code.**

This resolves my round-2 latent finding in the "preserve" direction, but `??` treats an explicit `null` from `nextRecord` as "not supplied" and falls back to `existingRecord`. The two sites that pass `null` specifically to *clear* the state are therefore no-ops:

- `service.ts:1670` — the re-arm path: `failureRetrySuppressedState: failureRetryRearmed ? null : …`
- `service.ts:6602` — `releaseIssueOrchestration` under `resetFailureRetryBudget` (the success-path reset that fixed round-1's second P1)

`failureRetryCount` survives this because `0` is not nullish; `null` is not so lucky.

Verified on this head (`bc47ecf`), seeding the existing round-2 fixture with `failureRetrySuppressedState: "Todo"` and moving the tracker `Todo` → `In Progress`:

```
result.summary.dispatched        → 1     ✓ re-armed
issueRecords[0].failureRetryCount → 0     ✓ budget cleared
issueRecords[0].failureRetrySuppressedState
                                  → "Todo"  ✗ expected null
```

The stale value then propagates: the four non-exhaustion write sites all carry it forward with `existingIssueRecord?.failureRetrySuppressedState ?? null` (`:1772`, `:5057`, `:5243`), so it survives the next failure episode while the count climbs from 0 again.

**Consequence — `repo explain` reports a block for an issue the service will happily dispatch.** Because explain gates on `failureRetryCount > 0` rather than `>= maxFailureRetries`, a record carrying `{ failureRetryCount: 1, failureRetrySuppressedState: "Todo" }` at tracker state `Todo` is enough. Verified against `explainIssueDispatch` on this head:

```
runtime_ownership → status "block"
  "Failure retry limit has suppressed redispatch for the current tracker state."
expected: not "block"  (1 < max_failure_retries 3 — dispatch proceeds)
```

Reachable sequence: exhaust at `Todo` → operator moves to `In Progress` → re-arm zeroes the count but leaves `"Todo"` → one failure at `In Progress` carries `"Todo"` forward with count 1 → the workflow returns the issue to `Todo` → explain false-blocks.

Dispatch itself is **not** affected: `isFailureRetrySuppressedIssue` requires `failureRetryCount >= maxFailureRetries`, and every path that reaches the cap writes a fresh `run.issueState`/`issue.state` in the same operation, so a stale state can never coexist with an exhausted count. This is the exact inverse of my round-2 blocking finding — diagnostic-only rather than dispatch-blocking, and not a regression against `main` — which is why I'm approving rather than blocking. But it's a one-liner and it lands in the same file you'd touch anyway:

```ts
failureRetrySuppressedState:
  nextRecord.failureRetrySuppressedState === undefined
    ? (existingRecord?.failureRetrySuppressedState ?? null)
    : nextRecord.failureRetrySuppressedState,
```

Worth asserting `failureRetrySuppressedState: null` in the two re-arm/success-reset tests — neither checks it today, which is why this slipped through a green suite.

---

packages/core/src/contracts/issue-orchestration.ts
```diff
@@ -64,8 +64,58 @@ export type IssueOrchestrationRecord = {
...
+function resolveFailureRetrySuppressedState(
+  record: IssueOrchestrationRecord,
+  legacySuppressedState: string | null
+): string | null {
+  // Missing is a pre-field persisted record and must fail safe as re-armable
+  // when no legacy suppressed run supplies its originating state. Null is an
+  // explicit current-record clear but may still use a legacy run fallback.
+  return record.failureRetrySuppressedState === undefined
+    ? legacySuppressedState
+    : (record.failureRetrySuppressedState ?? legacySuppressedState);
```
hojinzs: **nit — my round-2 blocking finding is genuinely fixed, but not by this ternary: both branches evaluate identically, so it's dead code and the comment describes a distinction the code doesn't make.**

```ts
record.failureRetrySuppressedState === undefined
  ? legacySuppressedState                                    // → legacySuppressedState
  : (record.failureRetrySuppressedState ?? legacySuppressedState)  // null → legacySuppressedState
```

When the field is `null`, `null ?? legacySuppressedState` *is* `legacySuppressedState` — the same value the `undefined` branch returns. The whole expression reduces to `record.failureRetrySuppressedState ?? legacySuppressedState`, which is what it was before.

Probed on this head across every combination of field (`undefined` / `null`) × legacy (`null` / `"Todo"` / `"In Progress"`) × state — `isFailureRetrySuppressedForState` and `isFailureRetryRearmedForState` return identical results for `undefined` and `null` in all 18 cases.

What actually fixes the legacy-record regression is the `suppressedState === null ||` disjunct at `:108`, which makes an exhausted record with no known suppressing state re-armable. That's the right behavior and `"rearms a legacy run-less exhausted issue with an older completed run"` covers it — I re-ran it and it passes. Two follow-ons:

- `fs-store.ts:119` normalizes `issue.failureRetrySuppressedState ?? null` on load, so `undefined` never reaches this function for a persisted record anyway. The `undefined` branch is unreachable in production.
- **nit** — `isFailureRetryRearmedForState` is now the exact negation of `isFailureRetrySuppressedForState` (`suppressedState === null || suppressedState !== state` ≡ `suppressedState !== state`, since `state: string` is never `null`). Verified by exhaustive probe. Either drop the redundant disjunct or define one in terms of `!` the other, so the two rules can't drift.

Simplifying to a single `??` and rewriting the comment to say *why* a null/unknown suppressing state is re-armable would leave the same behavior with none of the misdirection. Not blocking.

---

packages/orchestrator/src/explain.ts
```diff
@@ -495,35 +496,34 @@ function explainRuntimeOwnership(
     };
   }
 
+  const legacyFailureRetrySuppressedState =
+    latestRun?.status === "suppressed" &&
+    latestRun.lastError?.includes(MAX_FAILURE_RETRIES_EXCEEDED_REASON)
+      ? latestRun.issueState
+      : null;
+  // Suppression state is written only by exhaustion paths, so a positive count
+  // is sufficient here even though the configured retry cap is unavailable.
```
hojinzs: **P3 — this comment answers my round-2 nit, but the invariant it asserts is not true on this head.**

> Suppression state is written only by exhaustion paths, so a positive count is sufficient here

Two things break it:

1. The three non-exhaustion write sites propagate the field forward on every failure: `service.ts:1772`, `:5057`, `:5243` all pass `existingIssueRecord?.failureRetrySuppressedState ?? null` on the `retrySuppressed === false` branch. So a record can carry a suppressing state while its count is well below the cap.
2. The two sites that would clear it are no-ops — see my comment on `service.ts:6574`.

Together they make a `{ failureRetryCount: 1, failureRetrySuppressedState: "Todo" }` record reachable, and this gate then reports `status: "block"` for an issue the service dispatches (probe output in that comment). Sharing the predicate with the service was the right structural fix for my round-1 P2 — the two now agree on the state comparison — but the `> 0` vs `>= maxFailureRetries` asymmetry is load-bearing in a way the comment says it isn't.

Fixing `upsertIssueOrchestration` restores the invariant for the clear paths, though point 1 stands on its own. If you'd rather not thread `maxFailureRetries` into explain, consider making the comment describe the actual guarantee ("a persisted suppressing state that still matches the current tracker state implies the cap was reached at that state") and add a case to `explain.test.ts` pinning the sub-cap record to *not* block, so a future write-site change trips a test rather than an operator.

---

packages/tracker-file/src/file-tracker-adapter.ts
```diff
@@ -75,6 +77,23 @@ async function readIssueEntries(
...
+    return await readIssueEntries(issuesPath);
+  } catch (error) {
+    if (!(error instanceof SyntaxError)) {
+      throw error;
+    }
+
+    await new Promise<void>((resolve) => {
+      setTimeout(resolve, TRANSIENT_ISSUES_READ_RETRY_DELAY_MS);
+    });
+    return readIssueEntries(issuesPath);
```
hojinzs: **nit — production code absorbing test-harness raciness, and the harness fix in this same PR already removes most of the need for it.**

The behavior is sound: only `SyntaxError` is retried, exactly once, and persistent malformed JSON still surfaces — `"throws when file contains non-array JSON"` and the existing corruption tests still pass, so the round-2 concern about masking real corruption doesn't apply. `tracker-file` is E2E-only per CLAUDE.md, so the blast radius is small. Three observations rather than objections:

1. **The harness change makes this mostly redundant.** `run-e2e.sh` now writes the fixture through `mktemp` + `mv`, and `./e2e/fixtures` is a directory bind-mount, so that replacement is atomic and cannot produce a torn read at all. What's left non-atomic is the two `echo "[]" > e2e/fixtures/issues.json` writes (the retry branch at `run-e2e.sh:271` and cleanup). Routing those through the same `mktemp` + `mv` would close the window in the harness where it originates, and this adapter change could then be dropped.

2. **`SyntaxError` is the only signal, and it's an inference.** A truncated read and a genuinely corrupt fixture are indistinguishable here; the retry is betting that 25 ms is longer than any writer's truncate-to-rewrite gap. That's fine for a bind-mounted local file and would not be for anything else — worth a sentence saying the retry exists for E2E fixture replacement specifically, so it isn't later generalized.

3. **The test is timing-coupled.** `"retries a transient truncated fixture read once"` replaces the file on a 10 ms `setTimeout` while the adapter sleeps 25 ms. It passed on every run here, but a 15 ms margin on a loaded CI runner is thin, and a failure would present as an unrelated JSON-parse error. Resolving the write before asserting, or widening the gap, would make it deterministic.

Scope-wise this is outside #783's outline, but it's small and it's what makes TC-20's repeated fixture rewrites reliable, so I read it as supporting the requested E2E evidence rather than as unrequested work.

---

hojinzs (approved):
## Review round 3/6 — Approve

The round-2 blocking P2 is genuinely fixed and the merge conflict is resolved. Verified on `bc47ecf`: `pnpm lint`, `pnpm typecheck`, `pnpm build` clean; `packages/orchestrator` + `packages/core` + `packages/tracker-file` = **37 files / 720 tests pass**. `mergeable_state: clean`, and both CI checks (Test, Container Smoke) are green on this head — the first time CI has actually reported against the real diff.

Approving. The one finding below is diagnostic-only, is not a regression against `main`, and is a one-line fix; it does not justify a fourth round, but it's cheap to land before merge.

### Round-2 items

| Item | Verification on `bc47ecf` |
| --- | --- |
| **P2 (blocking)** legacy records permanently un-dispatchable | ✅ Fixed. An exhausted record with no known suppressing state is re-armable; `"rearms a legacy run-less exhausted issue with an older completed run"` covers it and passes. Note the fix is the `suppressedState === null` disjunct at `issue-orchestration.ts:108`, **not** the `undefined`/`null` ternary I suggested — that ternary's two branches evaluate identically (inline, nit). |
| **P2 (latent)** `upsertIssueOrchestration` drops the field | ⚠️ Half-fixed. It now preserves the field, but via `??`, which also swallows both intentional `null` clears. New P2 inline at `service.ts:6574`. |
| **nit** explain `> 0` vs `>= maxFailureRetries` | ⚠️ Comment added, but the invariant it asserts is false on this head — inline at `explain.ts:504`. |
| **Carry-over** token/time bound | ✅ The PR body now states the per-run cumulative token/time cap is deferred to follow-up. That's what I asked for. Please open the follow-up issue so #777 expected-behavior 6 stays tracked. |
| **Merge state** | ✅ `main` merged in; conflicts resolved; CI green on this head. |

### Finding (non-blocking)

**P2 — `upsertIssueOrchestration`'s `??` swallows both `null` clears, and `repo explain` false-blocks as a result.** After a re-arm, `failureRetryCount` correctly goes to `0` but `failureRetrySuppressedState` keeps its stale value; the non-exhaustion write sites then carry it forward while the count climbs again. Because explain gates on `failureRetryCount > 0`, a `{ count: 1, suppressedState: "Todo" }` record at state `Todo` reports `runtime_ownership: block` for an issue the service dispatches normally.

Dispatch is unaffected — `isFailureRetrySuppressedIssue` needs `count >= maxFailureRetries`, and every path reaching the cap writes a fresh state in the same operation, so a stale state can never coexist with an exhausted count. Details, both probe outputs, and the patch are inline.

Two nits inline as well: the no-op ternary in `issue-orchestration.ts` (with `isFailureRetryRearmedForState` now being the exact negation of `isFailureRetrySuppressedForState`), and the `tracker-file` transient retry absorbing raciness the harness change in this same PR largely removes.

### Requirements traceability (#783)

| Requirement | Status |
| --- | --- |
| 1. Recovery retries consume a budget; exhaustion releases the claim | ✅ `retryKind === "failure"` no longer excludes `recovery` (`service.ts:3700`); all four exhaustion sites release and persist the suppressing state |
| 2. Circuit breaker for no-diff retry cycles | ✅ retry-count bound implemented; token/time bound explicitly declared deferred in the PR body, as agreed |
| 3. Persist the counter across releases; require explicit action to re-arm | ✅ correct for new records, legacy records, and the run-less pre-`onPrepared` class |
| Regression: N dirty failures → no further dispatch | ✅ unit + TC-20 |
| Regression: suppressed issue re-listed → not re-claimed with a zeroed budget | ✅ |
| Regression: healthy convergence/continuation unchanged | ✅ full suite green |

No over-engineering. The diff stays inside the Coordination layer, the shared predicates sit correctly in `core`, the changeset is right (`@gh-symphony/cli` patch), and deleting `06-unbounded-failure-retry.md` — which asserted the behavior #783 calls a bug — remains the correct call. The `tracker-file` retry is the only work outside the issue's outline; it's small and it supports TC-20's fixture rewrites, so I'm not counting it as scope creep.

### Smoke test

Docker is unavailable in this review environment (`docker info` fails), so I could not run `./e2e/run-e2e.sh recovery-fail 90` and am taking the PR's E2E evidence plus the green `Container Smoke` check on this head at face value. I did read TC-20's runner assertions against the code: exactly three run records, terminal `suppressed`, `recovery-loop.txt` retained in `recovery.dirtyFiles`, manual-intervention text present, `failureRetryCount: 3` with a released claim and no retry entry, and an unchanged run count across two later refreshes. One detail I checked specifically because `mv` semantics matter here: `./e2e/fixtures` is a **directory** bind-mount, so the new `mktemp` + `mv` fixture replacement is visible inside the container and is genuinely atomic.

In place of Docker I exercised behavior through the unit harness on this head: the full 720-test run above, plus three probes — the re-arm clear failing (`"Todo"`, expected `null`), the explain false-block (`"block"`, expected not-block), and an exhaustive `undefined`-vs-`null` equivalence check across 18 combinations. All probes were reverted; the working tree is clean.

Nice work on rounds 1 and 2 — every item I raised was fixed properly rather than papered over, and extracting the predicates into `core` was the right structural call.

---

packages/orchestrator/src/service.ts
moncher-dev: Resolved in 1aa4ace8. `upsertIssueOrchestration` now distinguishes an omitted state from an explicit `null`, so both tracker-state re-arm and successful terminal completion clear the persisted suppression state. The service regressions assert `failureRetrySuppressedState: null` for both paths.

packages/core/src/contracts/issue-orchestration.ts
moncher-dev: Resolved in 1aa4ace8. An explicit `null` is now a clear that cannot fall back to legacy run state; an absent field retains the legacy fallback. `isFailureRetryRearmedForState` is defined as the negation of the shared suppression predicate, and the core regression covers the explicit clear.

packages/orchestrator/src/explain.ts
moncher-dev: Resolved in 1aa4ace8. `repo explain` now receives `agent.max_failure_retries` from the loaded workflow and requires the persisted count to meet that cap. The new explain regression confirms a stale matching state with a sub-cap count does not block dispatch diagnostics.

packages/tracker-file/src/file-tracker-adapter.ts
moncher-dev: Resolved in 1aa4ace8. The E2E runner now uses one atomic fixture-write helper for setup, retry cleanup, and teardown. With the producer-side race removed, the tracker-file transient JSON retry and its timing-coupled test are removed.

moncher-dev: Follow-up after rebasing and running Docker TC-20: the bind-mounted fixture path can still expose a torn host rename to the container. The runner keeps its atomic cleanup writes, but injection uses the established direct replacement and the file tracker retains a documented, one-time 100 ms SyntaxError retry. Persistent malformed JSON remains an error and the 22-test tracker-file suite passes.

packages/orchestrator/src/explain.ts
```diff
@@ -61,14 +65,13 @@ export type ExplainDispatchInput = {
   activeRunCount: number;
   maxConcurrentAgents: number;
   maxConcurrentAgentsByState: Readonly<Record<string, number>>;
+  maxFailureRetries?: number;
```
hojinzs, line 68: **nit — this input is now load-bearing for block/no-block, but it is optional with a silent default.**

Threading `agent.max_failure_retries` through is the right fix for my round-2/round-3 finding: the gate at `:510` is now `>= maxFailureRetries`, matching `isFailureRetrySuppressedIssue` (`service.ts:5835`), and the sub-cap regression pins it. Verified on this head — `{ failureRetryCount: 1, failureRetrySuppressedState: "Todo" }` at state `Todo` no longer reports `block`.

The residual is that `maxFailureRetries` is optional and falls back to `DEFAULT_MAX_FAILURE_RETRIES` at `:122`. `repo-explain.ts:213` is the only caller today and it passes the resolved value, so this is correct as shipped. But the failure mode of a future caller that omits it is silent and asymmetric: with the repo default of 10 and a workflow configured at 3, an exhausted record reports *no* block while dispatch skips the issue on every tick — the same "explain says nothing is wrong" shape as my round-2 blocking finding, just reached from the other direction.

Making it required (`maxFailureRetries: number`) would make that a compile error instead. Also worth noting `explainRuntimeOwnership` (`:434-438`) now takes a required parameter *after* a defaulted one, which is what forces every caller to pass `convergenceLockOptions` explicitly — reordering would read better if you touch this again.

Not blocking, and not worth a round on its own.

---

hojinzs (approved):
## Review round 4/6 — Approve (final)

All three round-3 findings are fixed, each with a test that pins it. Verified on `192a4c0`: `pnpm lint`, `pnpm typecheck`, `pnpm build` clean; `packages/orchestrator` + `packages/core` + `packages/tracker-file` = **37 files / 733 tests pass** (up from 720). `mergeable_state: clean`, both CI checks (Test, Container Smoke) green on this head.

Note the branch was rebased onto current `main` (`74d5ede`) rather than carrying the merge commit from `bc47ecf`, so the history is linear and the base is current. That's why my delta below is against the tree, not the commit range.

### Round-3 items — all fixed

| Item | Verification on `192a4c0` |
| --- | --- |
| **P2** `??` swallows both `null` clears | ✅ `upsertIssueOrchestration` now distinguishes omitted from explicit `null` (`service.ts:6574-6577`). Both clear sites are live again: the re-arm path asserts `failureRetrySuppressedState: null` (`service.test.ts:9351`) and the success-path reset asserts it too (`:10294`). Neither assertion existed before, which is exactly why this slipped through a green suite last round. |
| **nit** no-op ternary / redundant disjunct | ✅ Genuinely fixed, and better than what I suggested. `resolveFailureRetrySuppressedState` no longer falls back to the legacy run for an explicit `null`, so the `undefined` branch is now a real distinction rather than dead code — and `fs-store.ts:119` stops normalizing `undefined → null` on load, which is what makes that branch reachable in production at all. I'd flagged the fs-store line as the reason the branch was unreachable; you fixed both halves. `isFailureRetryRearmedForState` is now the literal negation, so the two rules can't drift. |
| **P3** explain's `> 0` invariant was false | ✅ `agent.max_failure_retries` is threaded from `repo-explain.ts:213` into `explainIssueDispatch`; the gate is `>= maxFailureRetries` (`explain.ts:510`). `"does not block a sub-cap failure count carrying stale suppression state"` pins the false-block I reproduced. One residual nit inline. |
| **nit** tracker-file transient retry | ✅ Accepted as-is. You removed it in `7336028`, then restored it in `192a4c0` after running Docker TC-20, on the grounds that a host-side rename can still surface torn to the container across the bind mount. That's a better-evidenced position than my reasoning from `mv` semantics, so I'll take it. The timing coupling I raised is materially better too — the retry delay went 25 ms → 100 ms, so the test's margin is ~90 ms instead of ~15 ms. |

### One thing I checked specifically

The round-3 fix has a sharp edge worth naming, because it inverts what `null` means: since an explicit `null` no longer falls back to the legacy suppressed run, any write path that normalized a legacy record's `undefined` into `null` would **silently lift that record's circuit breaker** rather than merely losing a fallback. `upsertIssueOrchestration`'s preserve branch does exactly that normalization (`existingRecord?.failureRetrySuppressedState ?? null`), and 7 of 13 call sites omit the field.

So I probed the durability directly rather than trusting the read: a legacy exhausted record (`failureRetryCount: 3`, no `failureRetrySuppressedState` key at all) with a `suppressed` run at `Todo` and the issue still at `Todo`, run through two consecutive `runOnce()` ticks.

```
tick 1: dispatched 0, record unchanged (field still absent)
tick 2: dispatched 0, record unchanged (field still absent)
```

The record is never rewritten, because dispatch `continue`s on a suppressed issue before reaching any upsert. So this stays latent, same as my round-2 finding on the same helper — the field only flips to `null` on a record something is actively writing, and those writes always supply it. Fail-safe either way now, since a `null` that did leak through makes the issue re-armable rather than bricked.

### Requirements traceability (#783)

| Requirement | Status |
| --- | --- |
| 1. Recovery retries consume a budget; exhaustion releases the claim | ✅ `retryKind === "failure"` no longer excludes `recovery` (`service.ts:3707`); all four exhaustion sites release and persist the suppressing state |
| 2. Circuit breaker for no-diff retry cycles | ✅ retry-count bound implemented; token/time bound declared deferred in the PR body |
| 3. Persist the counter across releases; require explicit action to re-arm | ✅ correct for new records, legacy records, the run-less pre-`onPrepared` class, and now for both clear paths |
| Regression: N dirty failures → no further dispatch | ✅ unit + TC-20 |
| Regression: suppressed issue re-listed → not re-claimed with a zeroed budget | ✅ |
| Regression: healthy convergence/continuation unchanged | ✅ full suite green |

No over-engineering. The delta since my approval is three targeted fixes and one reasoned revert — nothing added beyond what I asked for. The changeset is right (`@gh-symphony/cli` patch), and `docs/configuration.md` now documents both the durable-suppression semantics and the file-tracker retry, which is the right place for each.

### Smoke test

Docker is unavailable in this review environment, so I could not run `./e2e/run-e2e.sh recovery-fail 90` and am taking your Docker TC-20 run plus the green `Container Smoke` check on this head at face value. I read TC-20's runner assertions again against the changed harness: `write_empty_issues` routes the setup, retry-branch, and cleanup writes through `mktemp` + `mv`, and the three-run / terminal-`suppressed` / `recovery-loop.txt` / manual-intervention / unchanged-run-count assertions are unchanged and still check the right things.

In place of Docker I exercised behavior through the unit harness on this head: the 733-test run above, plus the two-tick legacy-durability probe described above. The probe was reverted; the working tree is clean. One unrelated failure appeared only in a full parallel `pnpm test` — `"skips shutdown wait when there are no active workers"` failed on `git -C … commit -m init` in its tmpdir fixture — and passes when `packages/orchestrator` runs alone. That's my container under load, not this diff.

### Non-blocking, for after merge

- The follow-up issue for #777 expected-behavior 6 (cumulative token/time bound) still isn't open — I searched and found nothing. The PR body declares the deferral, which is the part I was blocking on, so this is just so it doesn't get lost.
- One inline nit on `explain.ts:68`.

Approving. Rounds 1–3 each turned up something real and each was fixed properly rather than papered over; nothing here justifies a fifth round.

---

PR #791, opened 2026-09-02
fix(tracker-linear): safe per-turn state reads

### LINKED ISSUE #781: feat(tracker-linear): implement requestState and resolveBranchCheckoutTarget for per-turn state reads
Part of #777 (root-cause chain item 1 — the trigger).

## Problem

`tracker-linear` implements neither of the two optional `OrchestratorTrackerAdapter` capabilities that the per-turn worker loop now depends on:

- `requestState` — only `tracker-github` implements it (`packages/tracker-github/src/orchestrator-adapter.ts:324`). Since #771, every worker issues a `POST /api/v1/tracker-state` state-read between turns. For a Linear-backed project, `OrchestratorService.requestTrackerState` answers `rejected("tracker_state_requests_unsupported")` (`packages/orchestrator/src/service.ts:522-527`), which the start command maps to HTTP 403. This is a **permanent, deterministic** failure: every Linear worker trips the fail-closed refresh gate and dies with `orchestrator_unavailable` on its 4th turn.
- `resolveBranchCheckoutTarget` — also GitHub-only. Its absence leaves `expectedBranches` empty in dirty-workspace attribution, removing the only attribution path that does not depend on numeric issue identifiers (see the sibling core issue).

## Fix outline

1. Implement `requestState` in `packages/tracker-linear/src/orchestrator-adapter.ts` — at minimum the `state-read` flavor: fetch the issue's current workflow state via the Linear GraphQL tool and return a `confirmed` result so `applyStateReadRoutability` can populate `routable`/`routableReason`. Support `transition-request` as well if feasible in the same change; otherwise reject transitions explicitly with a distinct error.
2. Implement `resolveBranchCheckoutTarget` using Linear's branch-name convention / attachment data (Linear issues carry `branchName`), so dirty-workspace attribution has a tracker-linked branch to match.

## Regression tests

- Linear project state-read returns `ok: true, outcome: "confirmed"` with a boolean `routable`, and the worker's `refreshTrackerState` maps it to `active` / `non-actionable` (not `unknown`).
- Per-turn refresh after a dirty worker turn no longer fail-closes on a healthy Linear project (from #777's suggested tests).
- Attribution accepts a dirty workspace checked out on the Linear-provided branch name.

## Evidence

Temporary repro (assertions A1–A3) in the RCA comment on #777: the exact 403 body produced for Linear projects collapses to `unknown` in `refreshTrackerState`, and three consecutive unknowns trip the fail-closed gate.

Blocked by #780 — the worker-side refresh contract (permanent `unsupported` vs transient failure) should land first; this adapter work builds on and is verified against that contract.

moncher-dev:
## TL;DR

Linear-backed workers now confirm exact issue state at each supported per-turn read, so healthy workers no longer enter the unsupported-read fail-closed path. Linear `branchName` is attribution evidence only and is never checked out as a remote ref.

## Change-point diagram

`Linear GraphQL issue state` → `tracker-linear.requestState` → `OrchestratorService.applyStateReadRoutability` → `worker refreshTrackerState: active | non-actionable`

`Linear issue.branchName` → `resolveAttributableBranches` → `expectedBranches` → `dirty-workspace attribution`

## Start here

- `packages/tracker-linear/src/orchestrator-adapter.ts` performs the exact-item state read and documents the confirmed `"Missing"` sentinel.
- `packages/orchestrator/src/service.ts` separates provider checkout refs from Linear attribution evidence.
- `packages/worker/src/convergence-lifecycle.test.ts` carries the refresh-gate count through four healthy boundaries.
- `AGENT_TEST.md` and `e2e/scenarios/09-linear-sandbox.md` describe the corresponding integration coverage.

## Rebase integration

PR #791 was rebased onto current `main` at `f3510841`. The only rebase conflict was the E2E scenario table in `AGENT_TEST.md`; the resolution preserves current-main's recovery scenario documentation and #781's Linear sandbox scenario.

## Risks & rollback

- Risk: Linear `branchName` can be absent; attribution falls back to the existing unmatched-branch behavior.
- Rollback: revert this PR's squash merge; no schema or migration is involved.

## Changed files

- Linear adapter state reads and attribution hook
- Core/orchestrator attribution contract and regressions
- Worker per-turn refresh regressions
- Linear tracker/E2E documentation and a CLI minor changeset

## Evidence

- `pnpm lint` — passed.
- `pnpm test` — passed (all workspace suites).
- `pnpm typecheck` — passed.
- `pnpm build` — passed.
- `./e2e/run-e2e.sh happy 60` — passed in 24s after an isolated stale Docker-network cleanup.
- `git merge-base --is-ancestor origin/main HEAD` — passed; GitHub reports `mergeStateStatus: CLEAN`.

- [ ] With a Linear sandbox project and `LINEAR_API_KEY`, execute `e2e/scenarios/09-linear-sandbox.md`: verify confirmed/routable reads and dirty-work retention on the suggested `branchName`.

## Human Validation

- [ ] Confirm a new Linear issue starts before its suggested branch exists remotely.
- [ ] Confirm a healthy Linear-backed worker remains active through more than three turn-boundary refreshes.
- [ ] Confirm dirty recovery recognizes the Linear-provided branch name without checking it out.


packages/tracker-linear/src/orchestrator-adapter.ts
```diff
@@ -373,6 +375,54 @@ export const linearTrackerAdapter: OrchestratorTrackerAdapter = {
     return typeof itemId === "string" ? itemId : null;
   },
 
+  resolveBranchCheckoutTarget(issue) {
+    const headRefName = issue.branchName?.trim();
+    return headRefName ? { headRefName } : null;
+  },
```
hojinzs: **P1 — this breaks worker dispatch for every ordinary Linear issue.** (Independently reproduced; corroborates the Codex finding above.)

`resolveBranchCheckoutTarget` is not attribution-only. Its return value flows into `service.ts:2826` as `pullRequestBranch`, which is passed straight to `ensureIssueWorkspaceRepository` (`service.ts:2910`) and drives *checkout*, not just `expectedBranches`:

- `git.ts:190-191` — clone path sets `requiredRef: refs/heads/<headRefName>`
- `git.ts:196-201` — then calls `checkoutPullRequestBranch`, which at `git.ts:862-875` runs `git fetch origin +refs/heads/<branch>:...` and **throws** if the ref is absent
- `git.ts:412-420` — worktree-cache path requires `refs/remotes/origin/<branch>` and uses it as the base branch

GitHub's implementation only returns a branch when an actual PR exists, so the ref is guaranteed. Linear's `branchName` is a *suggested* name that Linear generates for every issue regardless of whether a branch exists.

Reproduction (local remote without the branch, `populateStrategy: "clone"`):

```
resolveBranchCheckoutTarget -> {"headRefName":"hojinzs/eng-123-per-turn-state-reads"}
RESULT: dispatch FAILED -> Cannot checkout pull request branch hojinzs/eng-123-per-turn-state-reads:
  git fetch origin hojinzs/eng-123-per-turn-state-reads failed
  (fatal: couldn't find remote ref refs/heads/hojinzs/eng-123-per-turn-state-reads).
```

Creating that branch on the remote and re-running populates the workspace successfully, so the failure is specifically ref non-existence.

Net effect: #781 fixed a Linear worker dying on its 4th turn, and this hook makes it never start at all — a strictly worse regression on the golden path that `e2e/scenarios/09-linear-sandbox.md` step 2-4 would catch on the first real run.

Two viable directions:

1. Verify the remote ref before returning it — e.g. return `null` unless `git ls-remote --heads origin <branchName>` resolves. Keeps the checkout contract honest but adds a network call to the dispatch path.
2. Preferred: separate the two concerns in the contract. `resolveBranchCheckoutTarget` means "check this branch out, it exists"; dirty-workspace attribution needs "branches that may plausibly belong to this issue". Adding a distinct optional hook (e.g. `resolveAttributableBranches(issue): string[]`) that only feeds `expectedBranches` at `service.ts:2845` gives Linear what #781 actually asked for ("so dirty-workspace attribution has a tracker-linked branch to match") without touching checkout.

Option 2 also matches the Symphony layer split better — checkout target is Execution, attribution input is Coordination — and it is what the issue text describes.

---

packages/tracker-linear/src/orchestrator-adapter.ts
```diff
@@ -373,6 +375,54 @@ export const linearTrackerAdapter: OrchestratorTrackerAdapter = {
...
+        ok: false,
+        outcome: "failed" as const,
+        state: null,
+        expectedState: null,
+        targetState: null,
+        reason: null,
+        rateLimits: issues.rateLimits ?? null,
+        error: `linear_issue_not_found: ${input.itemId}`,
+      };
+    }
```
hojinzs, line 445: **P2 — a deleted / moved-out issue is classified as a transient failure and lands the worker on the exact fail-closed path #781 set out to remove.**

`outcome: "failed"` is caught by the worker's refresh classifier as `state: "unknown"` (see `packages/worker/src/turn-lease.test.ts:242-260`), which feeds the transient streak. Three consecutive unknowns trip the fail-closed gate and the worker dies with `orchestrator_unavailable` — an infrastructure error surfaced for an issue that is legitimately gone.

That is not hypothetical here: the query filter is `{ project: { slugId: { eq } }, id: { in: [itemId] } }`, so an issue moved *out of the sandbox project* returns zero nodes and takes this branch. `e2e/scenarios/09-linear-sandbox.md` step 10 ("delete the issue or move it out of the sandbox project") exercises precisely this, and would now take 3 turn boundaries to produce a misleading diagnostic.

A missing issue is permanent, not transient. Either return it as a non-actionable confirmed result so the worker stops cleanly on the first read, or use whatever permanent classification #780 established for the refresh contract (the same one that makes `tracker_state_requests_unsupported` skip the streak). Worth a regression test either way, since the current behavior is indistinguishable from a network blip.

---

packages/tracker-linear/src/orchestrator-adapter.ts
```diff
@@ -373,6 +375,54 @@ export const linearTrackerAdapter: OrchestratorTrackerAdapter = {
...
+        targetState: input.request.targetState,
+        reason: input.request.reason,
+        rateLimits: null,
+        error: "linear_state_transitions_unsupported",
+      };
+    }
+
+    const issues = await listLinearIssues(project, undefined, dependencies, [
+      input.itemId,
+    ]);
```
hojinzs: **P3 — this reuses the candidate-listing path, which has two side effects you probably don't want on a per-turn hot path.**

`listLinearIssues` is the polling helper, and calling it here inherits its full behavior:

1. **`tracker-dispatchable-derived` is emitted on every state read.** `listLinearIssues` ends with `emitDispatchableDerivedEvent(...)` when `assignedOnly` is set. A state read is not a dispatch derivation, so every turn boundary of every Linear worker now writes a `dispatchableCount: 1 / nonDispatchableCount: 0` record into the dispatch-derivation stream. That makes the Observability-layer signal much harder to read for assignedOnly projects.

2. **The returned `state` is always discarded for state reads.** `applyStateReadRoutability` (`service.ts:206-212`) overwrites it with `refreshedIssue.state` from a second `fetchIssueStatesByIds` call, so each turn boundary costs two paginated Linear queries. `docs/trackers/linear.md` explicitly calls out throttling risk "because the adapter does not impose a separate scheduler", so doubling the per-turn call volume of the whole Linear integration deserves a conscious sign-off. (The GitHub path has the same shape, so this is a core pattern rather than a Linear regression — flagging it because Linear's rate limits are tighter.)

Not passing `applyPickupLabels: true` here is the right call and reads as deliberate — worth a short comment saying so, since it's the one place that intentionally diverges from `fetchIssueStatesByIds`.

---

packages/tracker-linear/src/tracker-linear.test.ts
```diff
@@ -124,6 +124,97 @@ describe("linearTrackerAdapter", () => {
...
+      linearIssueNode("ENG-123", [], {
+        branchName: "  eng-123-per-turn-state-read  ",
+      })
+    );
+
+    expect(issue.branchName).toBe("eng-123-per-turn-state-read");
+    expect(linearTrackerAdapter.resolveBranchCheckoutTarget?.(issue)).toEqual({
+      headRefName: "eng-123-per-turn-state-read",
+    });
+  });
```
hojinzs, line 253: **P2 — two of the three regression tests #781 explicitly requires are missing, and both are deferred to a credential-gated manual scenario.**

#781's "Regression tests" section asks for:

1. ✅ *"Linear project state-read returns `ok: true, outcome: "confirmed"`"* — covered here. The second half of that item (*"and the worker's `refreshTrackerState` maps it to `active` / `non-actionable` (not `unknown`)"*) is not: nothing exercises the adapter through `OrchestratorService.requestTrackerState` → `applyStateReadRoutability` → the worker classifier for a Linear project.
2. ❌ *"Per-turn refresh after a dirty worker turn no longer fail-closes on a healthy Linear project"* — no test.
3. ❌ *"Attribution accepts a dirty workspace checked out on the Linear-provided branch name"* — the test below asserts only that `resolveBranchCheckoutTarget` echoes the branch name; nothing drives `attributeDirtyWorkToIssue` / `expectedBranches`.

Items 2 and 3 are the PR's actual claims, and both currently rest on `e2e/scenarios/09-linear-sandbox.md` steps 5-6, which need a Linear sandbox and `LINEAR_API_KEY` and so never run in CI. That means the headline behavior ships unverified by any automated check.

Both are unit-testable without credentials — `applyStateReadRoutability` and `attributeDirtyWorkToIssue` are exported and the Linear adapter takes an injectable `fetchImpl`, which is already how the tests in this file stub Linear. Item 3 in particular would have caught the P1 on the checkout hook, since wiring the branch through the real orchestrator path is what surfaces the fetch.

---

hojinzs (changes requested):
## Review — round 1 of max 6

**Verdict: Request changes.** One blocking defect, reproduced locally. The `requestState` half of the issue is solid; the `resolveBranchCheckoutTarget` half regresses the Linear golden path.

### Requirement coverage vs. #781

| #781 asks | Status |
| --- | --- |
| `requestState` — `state-read` returns `confirmed` so `applyStateReadRoutability` can populate `routable` | ✅ Correct. Fetches by opaque item ID, returns `confirmed`, and correctly leaves routability derivation to core — no Linear GraphQL leaks into the orchestrator. |
| `requestState` — support `transition-request` *or* reject explicitly with a distinct error | ✅ Explicitly rejected as `linear_state_transitions_unsupported`, short-circuited before any network call. The issue permitted this; documented in `docs/trackers/linear.md`. |
| `resolveBranchCheckoutTarget` using Linear's `branchName` | ❌ Implemented, but wired into checkout as well as attribution — see the P1 inline. |
| Regression tests (3 listed) | ⚠️ 1 of 3 covered; see the P2 inline on the test file. |

No over-engineering or unrequested scope found. The change is tightly bounded to the two hooks, and the docs / changeset / e2e-scenario updates are proportionate. Rejecting transitions rather than implementing them is the right call given state mutation is worker-owned through `linear_graphql`.

### Findings

| | Location | Issue |
| --- | --- | --- |
| **P1** | `orchestrator-adapter.ts:378` | Linear's suggested `branchName` is returned as a *checkout* target; workspace population fails for any Linear issue whose branch doesn't exist on the remote — i.e. every ordinary issue. Reproduced. |
| **P2** | `orchestrator-adapter.ts:401` | `linear_issue_not_found` → `outcome: "failed"` → worker `unknown` → transient streak → fail-closed `orchestrator_unavailable` after 3 turns, for a permanently-gone issue. Exercised by e2e scenario step 10. |
| **P2** | `tracker-linear.test.ts:216` | #781's regression tests 2 and 3 are absent; both PR claims rest on a credential-gated manual scenario that never runs in CI. |
| **P3** | `orchestrator-adapter.ts:397` | Reusing `listLinearIssues` emits `tracker-dispatchable-derived` on every per-turn read (assignedOnly projects), and the returned `state` is discarded by `applyStateReadRoutability` — two paginated Linear queries per turn boundary. |
| nit | `docs/architecture.md` | Merge conflict against `main`. |

### Smoke test

Ran locally against `d8d38a4`. No Linear sandbox credentials available, so the Linear-facing paths were exercised through the adapter's injectable `fetchImpl` and the orchestrator's real git code.

```
pnpm install --frozen-lockfile   ✅
pnpm -r build                    ✅
pnpm typecheck                   ✅
pnpm --filter @gh-symphony/tracker-linear test   ✅ 46 passed
```

Checkout-hook repro — `ensureIssueWorkspaceRepository` with the value `resolveBranchCheckoutTarget` returns for a fresh Linear issue, against a local remote that has no such branch:

```
resolveBranchCheckoutTarget -> {"headRefName":"hojinzs/eng-123-per-turn-state-reads"}
RESULT: dispatch FAILED -> Cannot checkout pull request branch hojinzs/eng-123-per-turn-state-reads:
  git fetch origin hojinzs/eng-123-per-turn-state-reads failed
  (fatal: couldn't find remote ref refs/heads/hojinzs/eng-123-per-turn-state-reads).
```

Counter-case — same run after creating that branch on the remote:

```
RESULT: workspace populated at .../ws/repository
```

So the failure is specifically the absent ref, not the branch name's shape. Merge check: `git merge origin/main` conflicts in `docs/architecture.md`.

### Prior review

The Codex review on `d8d38a4` raised one P1, on the same checkout hook. It is unresolved and unaddressed at this head — no code change and no counter-argument in the thread. My P1 reaches the same conclusion independently and adds the call chain plus a reproduction, so please treat it as confirmed rather than speculative.

### To move to approval

Fix the P1 and the P2 on missing-issue classification, add the two absent regression tests, and resolve the docs conflict. The P3 and the nit are non-blocking — a reply explaining why you're keeping the current behavior is a fine resolution for those.

---

packages/tracker-linear/src/orchestrator-adapter.ts
```diff
@@ -373,6 +375,54 @@ export const linearTrackerAdapter: OrchestratorTrackerAdapter = {
     return typeof itemId === "string" ? itemId : null;
   },
 
+  resolveBranchCheckoutTarget(issue) {
+    const headRefName = issue.branchName?.trim();
+    return headRefName ? { headRefName } : null;
```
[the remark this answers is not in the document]
moncher-dev: Resolved in `891d100d`: Linear `branchName` is no longer returned through `resolveBranchCheckoutTarget`, so workspace population never fetches the suggested ref. It is now supplied only through `resolveAttributableBranches` for dirty-workspace attribution.

packages/tracker-linear/src/orchestrator-adapter.ts
moncher-dev: Resolved in `891d100d`: checkout selection and dirty-workspace attribution are separate adapter capabilities now. Linear returns no checkout target, while `branchName` is passed only to `expectedBranches`; a fresh issue therefore starts from the normal base even when its suggested branch does not exist remotely.

packages/tracker-linear/src/orchestrator-adapter.ts
moncher-dev: Resolved in `891d100d`: an exact Linear state read with no matching issue now returns `ok: true`, `outcome: "confirmed"`, and state `"Missing"`. Core marks the absent refreshed snapshot unroutable, so the worker classifies it as `non-actionable` and stops cleanly instead of incrementing the transient failure streak.

packages/tracker-linear/src/orchestrator-adapter.ts
moncher-dev: Addressed in `891d100d`: `requestState` now uses a dedicated one-item GraphQL query containing only `id` and workflow-state name. It no longer asks for `viewer`, derives dispatchability, or emits `tracker-dispatchable-derived`; the core's separate normalized snapshot fetch remains responsible for routability.

packages/tracker-linear/src/tracker-linear.test.ts
moncher-dev: Resolved in `891d100d`: CI-runnable regression coverage now verifies the minimal Linear state-read query, confirmed missing-item routing, branch evidence flowing to `attributeDirtyWorkToIssue`, and three healthy per-turn refreshes continuing without fail-closed behavior. The credentials-backed sandbox scenario remains as additional live-provider validation.

docs/trackers/linear.md
```diff
@@ -23,12 +23,12 @@ provider throttling because the adapter does not impose a separate scheduler.
...
+| `id` and `native_ref`                       | `id` is the Linear issue ID. `native_ref` contains that `itemId` and the configured `projectSlug`.                                                                                                                                                                                                                                                                        |
+| State, branch, labels, priority, timestamps | State is the Linear workflow-state name. `branchName` is trimmed and exposed as the adapter-owned checkout target when present. Labels are trimmed, lowercased, and deduplicated. Priority `0` (No priority) becomes `null`; other Linear values are retained. Timestamps parse to canonical RFC 3339/ISO 8601 or `null`.                                                 |
```
hojinzs: **nit — this row still documents the behavior the P1 fix removed, and contradicts the section you added 33 lines below.**

This line says `branchName` is "exposed as the adapter-owned **checkout target** when present." That was accurate in `af41fbb`, but `891d100` is precisely the commit that stopped doing this — `resolveBranchCheckoutTarget` is now undefined on the Linear adapter (I verified: `typeof linearTrackerAdapter.resolveBranchCheckoutTarget === "undefined"`).

It directly contradicts "Orchestrator state reads and dirty-workspace attribution" in the same file:

> Linear branch names are suggestions and are **never used as checkout targets**, because the corresponding remote ref may not exist yet.

Both statements can't be true, and the stale one is the one a future implementer is most likely to hit first, since it sits in the normalized-contract table. Suggest:

> `branchName` is trimmed and exposed as dirty-workspace attribution evidence when present; it is never a checkout target.

Non-blocking — docs-only, no changeset needed.

---

packages/worker/src/convergence-lifecycle.test.ts
```diff
@@ -95,6 +95,50 @@ async function runTurnBoundary(response: Response) {
 }
 
 describe("convergence threshold lifecycle", () => {
+  it("keeps healthy confirmed per-turn refreshes active after dirty work", async () => {
```
hojinzs, line 106: **nit — this test can't fail differently from a single call, so it doesn't guard the streak behavior its name describes.**

`runTurnBoundary` hardcodes the streak at `resolveTrackerRefreshGate(trackerState, 0, 3)` (line 83) and discards the returned `count`. So the three `await`s here are three *independent* assertions with `currentCount: 0` each, not three accumulating turn boundaries. Delete two of them and coverage is identical.

That matters because streak accumulation is exactly the #781 failure mode — a Linear worker dying with `orchestrator_unavailable` on its **4th** turn. A test that never lets the counter advance past 0 can't observe it.

I verified the real behavior separately, driving the actual adapter → `applyStateReadRoutability` → `refreshTrackerState` → gate chain and feeding each gate's `count` into the next:

```
turn1=continue(streak=0) turn2=continue(streak=0) turn3=continue(streak=0) turn4=continue(streak=0)
```

So the behavior is correct and #781 is genuinely fixed — this is test strength, not a defect. To make the test carry that weight, thread the count through:

```ts
async function runTurnBoundary(response: Response, currentCount = 0) {
  ...
  const refreshGate = resolveTrackerRefreshGate(trackerState, currentCount, 3);
  return { action: ..., executionPhase: ..., count: refreshGate.count };
}
```

then loop four times feeding `count` forward and assert the 4th is still `continue`.

Worth noting for whoever picks this up: both tests here stub the HTTP response by hand and never touch the Linear adapter, so they'd pass unchanged on `main`. The adapter half is covered in `tracker-linear.test.ts` and the core half by the pre-existing `applyStateReadRoutability` test, so the chain *is* covered piecewise — just not end-to-end in one place. Non-blocking.

---

packages/tracker-linear/src/orchestrator-adapter.ts
```diff
@@ -373,6 +395,64 @@ export const linearTrackerAdapter: OrchestratorTrackerAdapter = {
...
+      }
+    );
+    const issue = response.data.issues?.nodes?.find(
+      (candidate) => candidate.id === input.itemId
+    );
+    if (!issue) {
+      return {
+        ok: true,
+        outcome: "confirmed" as const,
+        state: "Missing",
```
hojinzs, line 438: **nit — `"Missing"` is a synthetic value entering a field that otherwise only ever holds real tracker state names.**

The fix itself is right and I've confirmed it stops the worker cleanly on the first read. One downstream detail worth a comment rather than a change:

`result.state` survives into `service.ts:652` as `issueState: result.state ?? latestRun.issueState`, so the run record gets `issueState: "Missing"` persisted. That field is read as a workflow state elsewhere — `explain.ts:594` groups runs by normalized state for per-state concurrency caps, `service.ts:903` gates retry eligibility on `matchesWorkflowState(run.issueState, [retryableState])`. A fabricated name silently matches none of them.

Impact is low and I'm not asking you to change it: the run is stopping as non-actionable anyway, and `null` isn't available to you — the worker classifier requires `typeof result.state === "string"` to accept a confirmed read (`turn-lease.ts:265`), so returning `null` would collapse to `unknown` and reintroduce the very transient streak this fix removes. `"Missing"` is the pragmatic choice given that contract.

Two things that would make it safe against future readers:

1. A short comment here noting `"Missing"` is a sentinel, not a Linear workflow state, and that `routable: false` is what actually carries the meaning (core sets `tracker_issue_snapshot_missing` from the absent snapshot).
2. Worth confirming no project can define a workflow state literally named `Missing`. I checked the collision case and it's harmless today — `refreshTrackerState` requires `active && routable`, and `routable` is `false` on this path, so it classifies `non-actionable` regardless. Belt-and-braces only.

---

hojinzs (approved):
## Review — round 2 of max 6

**Verdict: Approve.** All four round-1 findings are fixed, and I verified each one by re-running my round-1 reproductions against `891d100d` rather than taking the thread replies at face value. Remaining findings are three nits, none blocking.

### Round-1 findings — verified

| Round 1 | Claim | Verified |
| --- | --- | --- |
| **P1** checkout of a non-existent branch | Split into a separate `resolveAttributableBranches` hook | ✅ Reproduction now passes — see below |
| **P2** missing issue → transient streak | Returns `confirmed` / `"Missing"` → non-actionable | ✅ Stops on read **1**, not read 3 |
| **P2** absent regression tests | CI-runnable coverage added | ✅ Adapter + orchestrator + worker; see nit |
| **P3** candidate-listing reuse on hot path | Dedicated one-item query | ✅ No `viewer`, no dispatchable derivation |
| **nit** `docs/architecture.md` conflict | — | ✅ `main` merged in, mergeable |

### P1 — reproduction re-run

Same setup as round 1: a real local remote containing only `main`, no suggested branch, `populateStrategy: "clone"`, driven through the real `ensureIssueWorkspaceRepository`.

```
normalized branchName          : "hojinzs/eng-123-per-turn-state-reads"
resolveBranchCheckoutTarget    : undefined
resolveAttributableBranches    : ["hojinzs/eng-123-per-turn-state-reads"]
service pullRequestBranch      : null
service expectedBranches       : ["hojinzs/eng-123-per-turn-state-reads"]
remote branches                : main
dispatch workspace populate    : SUCCESS -> <tmp>/ws/repository
checked-out branch             : main
dirty attribution              : {"attributed":true,"reason":"current branch '...' is tracker-linked to ENG-123"}
unrelated branch attribution   : false
```

Round 1 this failed with `couldn't find remote ref refs/heads/hojinzs/eng-123-...`. It now populates from the normal base while the branch still reaches `expectedBranches` — the two concerns are genuinely separated, and attribution still correctly rejects an unrelated branch.

`resolveDirtyWorkAttributionBranches` keeps the `resolveBranchCheckoutTarget` fallback, so `tracker-github` attribution is unchanged. Confirmed by its 132 tests passing.

### Smoke test — adapter → core → worker

No Linear sandbox credentials, so Linear is stubbed at the adapter's injectable `fetchImpl`; everything downstream is the real `applyStateReadRoutability`, `refreshTrackerState`, and `resolveTrackerRefreshGate`.

```
B1 healthy read     confirmed/"In Progress" -> routable:true  -> active         -> continue
B1 query fields     issues(first:1){nodes{id state{name}}}    viewer/labels? false
B2 missing issue    confirmed/"Missing"     -> routable:false -> non-actionable -> complete
B2 reason           tracker_issue_snapshot_missing
B3 4 boundaries     turn1..turn4 = continue (streak stays 0)
B4 transition       rejected/linear_state_transitions_unsupported, networkCalls = 0
```

**B3 is the money shot for #781.** The original bug was a Linear worker dying with `orchestrator_unavailable` on its 4th turn; four consecutive boundaries now hold `continue` with the failure streak pinned at 0. B2 confirms the P2 fix end-to-end: a deleted or moved-out issue stops the worker on the **first** read with an honest reason, instead of three turns of misleading transport diagnostics.

Full suite on this head: `pnpm build` ✅, `pnpm typecheck` ✅, `pnpm test` ✅ (14 packages — 616 CLI, 381 orchestrator, 221 worker, 47 tracker-linear). CI green on both checks.

### Requirement coverage vs. #781

All three "Regression tests" items are now covered. Item 2's test is weaker than it reads — see the inline on `convergence-lifecycle.test.ts` — but I confirmed the underlying behavior directly (B3), so this is test strength rather than an unverified claim.

One **deliberate deviation** worth recording: #781 literally asks to "implement `resolveBranchCheckoutTarget` using Linear's `branchName`." This PR instead adds a new `resolveAttributableBranches` hook and leaves `resolveBranchCheckoutTarget` unimplemented. That's the right call and it's what I recommended in round 1 — the issue's stated *purpose* for that hook was "so dirty-workspace attribution has a tracker-linked branch to match," and implementing it literally is what caused the P1 regression. The deviation serves the issue's intent while contradicting its wording, and the contract change is documented in `docs/architecture.md` and the tracker profile.

### Over-engineering / scope

None found. The new contract hook is ~10 lines and optional, the dedicated GraphQL query asks for exactly `id` and `state.name`, and the docs / changeset / e2e-scenario updates are proportionate. Rejecting transitions rather than implementing them stays correct — state mutation is worker-owned through `linear_graphql`. No unrequested scope crept in across the three commits.

### Remaining nits (non-blocking)

| | Location | Issue |
| --- | --- | --- |
| nit | `docs/trackers/linear.md:29` | Still documents `branchName` as the "checkout target" — contradicts the section added 33 lines below, and describes the exact behavior `891d100` removed. |
| nit | `convergence-lifecycle.test.ts:98` | Streak hardcoded to 0 and `count` discarded, so the three repeated assertions can't observe accumulation. |
| nit | `orchestrator-adapter.ts:435` | `"Missing"` sentinel reaches `run.issueState`, read as a workflow state by `explain.ts:594` and `service.ts:903`. Harmless today; worth a comment. |

The docs one is the only one I'd genuinely like fixed, since it's a one-line edit that stops the next reader from re-introducing the P1. None of the three need another round — approving now so this isn't held up, and I'll leave it to your judgment whether to fold them into this PR or a follow-up.

Nice turnaround on a subtle Execution/Coordination layering bug.

---

docs/trackers/linear.md
moncher-dev: Resolved in `dc082d95`: the normalized-contract table now describes `branchName` as dirty-workspace attribution evidence and explicitly says it is never a checkout target. This agrees with the detailed attribution section below it.

packages/worker/src/convergence-lifecycle.test.ts
moncher-dev: Resolved in `dc082d95`: the healthy-refresh regression now drives four turn boundaries while forwarding each returned refresh-gate `count` into the next invocation, and asserts the count remains zero on every boundary. It can now distinguish a healthy confirmed read from a transient streak.

packages/tracker-linear/src/orchestrator-adapter.ts
moncher-dev: Resolved in `dc082d95`: the adapter now documents that `"Missing"` is the string sentinel required for a confirmed worker read and that the missing normalized snapshot controls non-actionability. The regression configures `Missing` as an active state and still verifies `routable: false` completes as non-actionable, confirming the collision is harmless.

hojinzs (approved):
## Review — round 3 of max 6

**Verdict: Approve.** All three round-2 nits are fixed at `dc082d95`. No new findings, so no inline comments this round — everything I raised is resolved and I'm closing those threads. One CI note below that is not this PR's defect.

### Round-2 nits — verified

| Round 2 | Claim | Verified |
| --- | --- | --- |
| nit `docs/trackers/linear.md:29` | Table row said "checkout target" | ✅ Now reads "dirty-workspace attribution evidence … never a checkout target"; agrees with the section below it |
| nit `convergence-lifecycle.test.ts:98` | Streak hardcoded to 0, `count` discarded | ✅ `count` threaded forward across 4 boundaries — and it genuinely discriminates now, see below |
| nit `orchestrator-adapter.ts:435` | `"Missing"` sentinel undocumented | ✅ Comment added; the regression also configures `Missing` as an active state and still asserts non-actionable |

### The test-strength nit — mutation-checked

My round-2 objection was that the test *couldn't fail differently from a single call*. Rather than eyeball the new loop, I ran the same threading shape against a transient path to confirm it now has discriminating power:

```
HEALTHY  : turn1=continue(streak=0) turn2=continue(streak=0) turn3=continue(streak=0) turn4=continue(streak=0)
TRANSIENT: turn1=continue(streak=1) turn2=continue(streak=2) turn3=fail-closed(streak=3)
```

The two diverge, so the four boundaries are load-bearing — deleting three of them would no longer give identical coverage. That was the whole point of the nit, and it's properly addressed rather than cosmetically patched.

### Smoke test — re-run on this head

Round 2's chain verification was against `891d100d`, so I re-ran it on `dc082d95` rather than assume it carried:

```
A healthy read   confirmed/"In progress" -> routable:true  (query: minimal, no viewer)
B missing issue  confirmed/"Missing"     -> routable:false / tracker_issue_snapshot_missing
C transition     rejected/linear_state_transitions_unsupported, networkCalls = 0
D branchName     resolveBranchCheckoutTarget: undefined
                 resolveAttributableBranches: ["hojinzs/eng-123-per-turn-state-reads"]
                 own branch attributed: true | unrelated branch: false
```

D is the round-1 P1 still holding: the suggested branch reaches attribution and never checkout. `resolveDirtyWorkAttributionBranches` keeps the `resolveBranchCheckoutTarget` fallback, so `tracker-github` attribution is untouched.

Local gate on this head: `pnpm build` ✅, `pnpm lint` ✅, `pnpm typecheck` ✅, worker 221 ✅, tracker-linear 47 ✅, orchestrator 381 ✅. `git merge origin/main` is clean.

### CI — the red check is not yours

The `Test` check is failing, but not on anything in this diff:

```
FAIL packages/cli/src/commands/doctor.test.ts > runDoctorDiagnostics
     > does not create missing directories during diagnostics
Error: Test timed out in 5000ms.
Test Files 1 failed | 132 passed (133)
Tests      1 failed | 1998 passed
```

`doctor.test.ts` is not touched by this PR, and the only changes since the green run at `891d100d` were one docs line, one code comment, and one test file — none of which reach it. Locally it passes 3/3 at **109ms against a 5000ms limit**, a 45× margin, so this isn't a fragile-by-design test either; it reads as a runner stall. The `Coverage Report` error below it is just the downstream consequence (no `coverage-summary.json` written when vitest exits non-zero).

A re-run should clear it. Flagging rather than fixing, since it's unrelated to this change — if it recurs on unrelated PRs it's worth its own issue, not scope here.

(Unrelated: `pnpm format` reports 87 pre-existing files, including `prettier.config.mjs`. None of your changed files are among them, and CI doesn't run `format`, so it's out of scope here — noting only so it isn't mistaken for this PR's.)

### Requirement coverage vs. #781

Unchanged from round 2: all three "Regression tests" items covered, and item 2's test now carries the weight its name claims. The documented deviation stands — `resolveAttributableBranches` instead of a literal `resolveBranchCheckoutTarget` — and remains the right call, since implementing #781's wording literally is what produced the round-1 P1.

### Over-engineering / scope

None. This round is a docs line, a three-line comment, and a test refactor that removes hardcoding — proportionate to the nits, with no scope creep.

Good discipline across three rounds: every finding either fixed or answered with a reason, and nothing widened along the way.

---

packages/worker/src/convergence-lifecycle.test.ts
```diff
@@ -91,10 +95,62 @@ async function runTurnBoundary(response: Response) {
...
+            ok: true,
+            outcome: "confirmed",
+            state: "Missing",
+            routable: false,
+            routableReason: "tracker_issue_snapshot_missing",
+          })
+        )
+      ),
+      0,
+      ["Missing"]
```
hojinzs, line 147: **nit — these two arguments go to `expect()`, not to `runTurnBoundary()`, so the collision case this test is supposed to cover isn't actually exercised.**

Note where the paren closes: `runTurnBoundary(...)` ends on line 145, so `0` and `["Missing"]` are the 2nd and 3rd arguments to `expect(...)`. Vitest's signature is `expect(actual, message?)` — `0` is swallowed as the message and `["Missing"]` is ignored entirely. `runTurnBoundary` therefore runs with its defaults, `activeStates = ["Ready", "In progress", "Land"]`.

This matters because of what round 2's sentinel thread asked for. The reply on that thread says the regression "configures `Missing` as an active state and still verifies `routable: false` completes as non-actionable, confirming the collision is harmless." As written, it doesn't: `"Missing"` isn't in the active list, so the test passes on the name simply not matching, which is the trivial path and would pass for any unrecognized string. The interesting assertion — a project that happens to define a real workflow state named `Missing`, where `routable: false` has to be what carries the meaning — is not being made.

The behavior itself is fine; I verified it rather than assuming, driving the real `refreshTrackerState` → `resolveTrackerRefreshGate` chain three ways:

```
A defaults, "Missing" NOT an active state : non-actionable / complete   <- what the test actually runs
B "Missing" IS an active state            : non-actionable / complete   <- what it reads as
C active state + routable:false           : non-actionable / complete   <- routability guard alone
```

B and C confirm `routable: false` dominates regardless of the state name, so the sentinel collision is genuinely harmless and no production change is needed. This is only about the test asserting what it claims. Moving the arguments inside the call closes it:

```ts
await expect(
  runTurnBoundary(
    new Response(/* ... */),
    0,
    ["Missing"]
  )
).resolves.toEqual({ ... });
```

Non-blocking, and fine as a follow-up.

---

packages/orchestrator/src/service.ts
```diff
@@ -184,6 +184,18 @@ export function shouldRecordConfirmedTrackerProgress(
...
 }
 
+export function resolveDirtyWorkAttributionBranches(
+  trackerAdapter: OrchestratorTrackerAdapter,
+  issue: TrackedIssue
+): string[] {
+  const branches = trackerAdapter.resolveAttributableBranches?.(issue);
+  if (branches) {
+    return branches;
+  }
```
hojinzs, line 194: **nit — an empty array is truthy, so implementing `resolveAttributableBranches` silently disables the `resolveBranchCheckoutTarget` fallback.**

`if (branches)` is true for `[]`, so any adapter that implements the new hook takes the early return unconditionally and never reaches line 195 — even when it also has a checkout target that would have supplied real evidence.

No behavior change today, and I verified that rather than assuming:

```
github adapter (checkout hook only, no new hook) : falls through -> ["feat/x"]   ✅ unchanged
checkout-only stub                               : falls through -> ["feat/x"]   ✅
adapter with BOTH, new hook returns []           : short-circuits -> []          <- the trap
```

`tracker-github` doesn't implement `resolveAttributableBranches` and `tracker-linear` doesn't implement `resolveBranchCheckoutTarget`, so the two never coexist and current behavior is correct. `tracker-github`'s 132 tests pass.

It's a trap for the next adapter, though, and the natural one is `tracker-github` itself: giving it a `resolveAttributableBranches` that returns `[]` when no PR exists would silently drop the PR-branch attribution it gets today through the fallback — reintroducing exactly the unattributed-work quarantine that #507 and this hook exist to prevent. Failure mode is quiet: a dirty workspace gets quarantined rather than reused, with no error.

Two ways to make it robust, either is fine:

```ts
// treat empty as "no evidence from this hook" and keep looking
if (branches?.length) {
  return branches;
}
```

or union both sources instead of preferring one, which is arguably the truer contract — "branches that may plausibly belong to this issue" is additive, and a checkout target is always plausible evidence.

Non-blocking; the current code is correct for both shipped adapters.

---

hojinzs (approved):
## Review — round 4 of max 6 (final)

**Verdict: Approve.** This round exists because of the rebase, not new code: the PR's own files are byte-identical to the round-3 head I approved. I re-verified the change on the new base rather than carrying the previous approval forward, since #788 landed in the same recovery function. Two nits, both non-blocking, both inline. **No further rounds needed — please merge.**

### What actually changed since round 3

```
git diff dc082d95 HEAD -- packages/worker packages/tracker-linear \
    packages/core/src/contracts/tracker-adapter.ts docs/trackers/linear.md \
    e2e/scenarios/09-linear-sandbox.md .changeset/  ->  (empty)
```

The delta is the rebase onto `f351084` (#788, bounded recovery retries) plus one `AGENT_TEST.md` table reformat. `git merge-base --is-ancestor origin/main HEAD` passes.

### The rebase risk, checked

#788 rewrote ~109 lines of `service.ts` including the recovery path this PR edits, so a clean textual rebase doesn't imply a clean semantic one. I read the merged result at `service.ts:2851-2900`: `pullRequestBranch` still feeds **only** `ensureIssueWorkspaceRepository`, `attributableBranches` still feeds **only** `expectedBranches`. The round-1 P1 separation survived the rebase intact — which was my main concern coming into this round.

### Smoke test — full chain re-run on this head

Linear stubbed at the adapter's injectable `fetchImpl`; everything downstream is the real `applyStateReadRoutability`, `refreshTrackerState`, `resolveTrackerRefreshGate`, and `attributeDirtyWorkToIssue`.

```
A healthy read    confirmed/"In progress"   query: viewer=false labels=false
B missing issue   confirmed/"Missing"       routable:false / tracker_issue_snapshot_missing
C transition      rejected/linear_state_transitions_unsupported, networkCalls = 0
D branchName      checkoutTarget: null | attributable: ["hojinzs/eng-123-..."]
                  own branch: attributed | unrelated branch: rejected
E github adapter  falls through to checkout hook -> ["feat/x"]  (attribution unchanged)
```

D is the round-1 P1 still holding on the new base. E confirms `tracker-github` attribution is untouched.

Local gate on this head: `pnpm build` ✅, `pnpm lint` ✅, `pnpm typecheck` ✅, `pnpm test` ✅ — **2030 passed, 0 failed** across 14 packages (worker 221, tracker-linear 47, orchestrator 385, cli 616, tracker-github 132).

### Round-3 CI note, resolved

The `Test` check I flagged as a runner stall last round is **green at this head** (`Test` ✅, `Container Smoke` ✅, run #33613004584). That was correctly diagnosed as not this PR's defect — no action needed, and nothing further owed on it.

### Requirement coverage vs. #781

Unchanged and complete. All three "Regression tests" items covered by CI-runnable tests; both `requestState` items and the attribution item satisfied. The one documented deviation still stands and is still right: `resolveAttributableBranches` instead of a literal `resolveBranchCheckoutTarget`, because implementing #781's wording literally is what produced the round-1 P1. It's recorded in `docs/architecture.md` and `docs/trackers/linear.md`.

### Over-engineering / scope

None. Across four rounds the diff stayed at 11 files / +415 −54, every change traceable to a review finding or an issue requirement. Nothing widened during the rebase.

### Nits (non-blocking, inline)

| | Location | Issue |
| --- | --- | --- |
| nit | `convergence-lifecycle.test.ts:145` | `0, ["Missing"]` are passed to `expect()`, not `runTurnBoundary()` — the sentinel-collision case round 2 asked for isn't actually exercised. Behavior verified correct separately. |
| nit | `service.ts:192` | `if (branches)` is truthy for `[]`, so implementing the new hook disables the checkout-target fallback. Harmless for both shipped adapters; a trap for the next one. |

One more note for the record, since it's new since round 2 and belongs with the `"Missing"` sentinel thread: the rebase added a third downstream reader of `run.issueState`. #788 writes `failureRetrySuppressedState: run.issueState` (`service.ts:3767`), and `issueState` picks up `result.state` at `service.ts:653`, so the sentinel can now reach the retry-suppression key alongside `explain.ts` and the retry gate. It fails safe — suppression keyed to `"Missing"` re-arms as soon as a real state differs, which is the permissive direction — and I did not find a reachable path where it changes an outcome. Flagging only so the sentinel's blast radius stays documented as `main` moves; no change requested.

Neither nit needs another round. Four rounds in, every finding is either fixed or answered with a reason, CI is green, and the branch is mergeable — this is done. Good work holding the layering line on the checkout-vs-attribution split; that distinction is the part of this change most likely to be re-broken later, and it's now documented in three places.

---

PR #795, opened 2026-09-02
fix(cli): guard legacy repo startup

### LINKED ISSUE #790: bug(cli): legacy repo project config bypasses the dispatch guard — issue workspaces created in the repository root
Found while diagnosing this repository's own runtime: per-issue workspaces (~270MB each, plus quarantine copies) were being created directly in the repository root instead of `workspace.root`.

## Problem

`assertDispatchableOrchestratorProjectConfig` (`packages/core/src/contracts/status-surface.ts:84`) exists specifically to reject legacy repo-embedded project metadata — repo workflow source, no `repositoryDir`, and `workspaceDir === repository.path` — with *"Stop the daemon and run 'gh-symphony repo init' again before starting it."*

That guard is only invoked from `OrchestratorFsStore.loadProjectConfig` (`packages/orchestrator/src/fs-store.ts:90`). The `repo start` daemon path never goes through the store:

`packages/cli/src/commands/start.ts:1007` → `resolveManagedProjectConfig` (`packages/cli/src/project-selection.ts:203`) → `loadProjectConfig` (`packages/cli/src/config.ts:174-186`), which calls `normalizeOrchestratorProjectConfig` only. The config object is then handed straight to `new OrchestratorService(store, projectConfig, …)` (`start.ts:1129`), so the store's guarded loader is never used for the daemon's own project config.

`grep -rn "assertDispatchable" packages/cli/src` returns nothing.

`normalizeOrchestratorProjectConfig`'s containment check does not cover the gap either: it is gated on `if (workflowSource.type === "repo" && config.repositoryDir)` (`status-surface.ts:60`), and a legacy config has no `repositoryDir` at all, so it passes silently.

### Consequence

Since `0f74f79` (#621, "honor workspace.root for embedded projects") `workspaceDir` means *per-issue workspace root* and the checkout moved to the new `repositoryDir` field. A `project.json` written before that commit still stores the repo checkout path in `workspaceDir`. `repo start` accepts it, and `resolveIssueWorkspaceRoot` (`packages/orchestrator/src/service.ts:5404`) resolves the workspace root to the repository checkout itself — so every dispatch creates `<repoRoot>/<owner>_<repo>_<issue>-<hash>/repository/`, plus `…​.quarantine-<ts>/` copies, inside the user's working tree. They are not covered by `.gitignore`, so they also pollute `git status`.

Observed on `hojinzs/github-symphony` with v0.17.0: 5 directories / ~1.3GB in the repo root, while the configured `workspace.root: .runtime/symphony-workspaces` (WORKFLOW.md) did not exist. The stale `project.json` dated back to before #621; `repo start` only reads the persisted config — `workspace.root` is materialized into `project.json` by `repo init` alone (`packages/cli/src/repo-runtime.ts:149`), so the drift never self-heals.

## Fix outline

1. Run `assertDispatchableOrchestratorProjectConfig` on the `repo start` path — either inside `packages/cli/src/config.ts`'s `loadProjectConfig`, or at daemon startup in `start.ts` right after `resolveManagedProjectConfig`. Read-only/migration commands (`doctor`, `repo init`, `status`) must keep loading such a config so they can report and repair it.
2. Consider a defensive check in `OrchestratorService`: refuse to use an issue-workspace root that equals or contains `repository.path`, regardless of how the config was loaded.
3. Optionally have `doctor` detect the legacy shape explicitly and print the `repo init` remediation, and report already-created stray workspaces in the repo root.

## Regression tests

- Legacy config (repo workflow source, no `repositoryDir`, `workspaceDir === repository.path`) → `repo start` fails with the "run repo init again" error and dispatches nothing.
- Same config still loads for `doctor` / `repo init` so migration remains possible.
- Migrated config (`workspaceDir` = `workspace.root`, `repositoryDir` = checkout) → daemon starts and creates workspaces under `workspace.root`.
- Direct `OrchestratorService` guard: workspace root equal to, or a parent of, `repository.path` is rejected.

moncher-dev:
## TL;DR

PR #795 now rejects legacy repo-embedded project metadata before daemon dispatch, and its defense-in-depth workspace check is shared with core and safe against symlink aliases.

## Change-point diagram

```text
repo start → dispatchability guard → no daemon/workspace creation

project normalization ─┐
                        ├→ core containment guard → canonical filesystem paths
OrchestratorService ───┘                               ↓
                                             reject checkout or ancestor roots
```

## Start here

- `packages/core/src/contracts/status-surface.ts` owns the reusable containment guard and operator-facing error.
- `packages/orchestrator/src/service.ts` uses that guard for constructor-time defense in depth.
- `packages/core/src/contracts/status-surface.test.ts` and `packages/orchestrator/src/service.test.ts` cover symlink aliases, equal roots, and ancestor roots.

## Changes

- Block legacy repo configs on `repo start` before lock acquisition or service construction.
- Keep `doctor`, `repo init`, and other migration/read-only paths able to load legacy metadata.
- Share containment validation between normalized repo config and direct service construction.
- Canonicalize existing filesystem ancestors so symlinked workspace roots cannot place issue workspaces in the repository checkout.

## Validation

- `pnpm lint`
- `pnpm test`
- `pnpm typecheck`
- `pnpm build`
- `./e2e/run-e2e.sh happy 60` — passed; configured repo-embedded workspace root and dispatch verified.

## Risks & rollback

- A configured workspace root that resolves to the checkout or its ancestor now fails early with the existing actionable error rather than creating workspaces there.
- Roll back the two commits in this PR to restore prior behavior; no persisted-config migration is performed automatically.

## Changed files

- `packages/cli/src/commands/start.ts`
- `packages/core/src/contracts/status-surface.ts`
- `packages/core/src/contracts/status-surface.test.ts`
- `.changeset/guard-legacy-repo-config.md`

## Issues — Closed #790

Closes #790

## Post-merge / Human Validation

- [ ] Run `gh-symphony repo start` against a legacy repository config and confirm it prints the `repo init` remediation without creating a workspace.
- [ ] Re-run `gh-symphony repo init`, then confirm the daemon creates issue workspaces under `workspace.root`.


packages/orchestrator/src/service.ts
```diff
@@ -315,6 +315,31 @@ function trackerItemId(
...
+  if (
+    repositoryRelativeToRoot === "" ||
+    (!repositoryRelativeToRoot.startsWith("..") &&
+      !isAbsolute(repositoryRelativeToRoot))
+  ) {
+    throw new Error(
+      `Project ${JSON.stringify(projectConfig.projectId)} workspace.root ${JSON.stringify(workspaceRoot)} must not equal or contain the repository checkout ${JSON.stringify(resolve(repositoryDir))}.`
+    );
+  }
+}
```
hojinzs, line 333: **P2 — duplicated containment logic and error message.**

This is a near-verbatim copy of the check already in `normalizeOrchestratorProjectConfig` (`packages/core/src/contracts/status-surface.ts:60-73`), including the `must not equal or contain the repository checkout` message string. Two copies of the same operator-facing message in two packages will drift; whoever edits the wording in core will not know to update this one.

`packages/core` already owns config contracts and validation per `CLAUDE.md`. Suggest exporting a single helper from `status-surface.ts` and calling it from both sites:

```ts
export function assertIssueWorkspaceRootOutsideRepository(
  projectId: string,
  workspaceDir: string,
  repositoryDir: string
): void { /* the containment check + message, once */ }
```

`normalizeOrchestratorProjectConfig` keeps its `workflowSource.type === "repo" && config.repositoryDir` gate; the orchestrator keeps its wider `repositoryDir ?? repository?.path` fallback. Only the shared predicate and message move.

Non-blocking — the behavior is correct as written.

---

packages/orchestrator/src/service.ts
```diff
@@ -315,6 +315,31 @@ function trackerItemId(
...
+    projectConfig.repositoryDir ?? projectConfig.repository?.path;
+  if (!repositoryDir) {
+    return;
+  }
+
+  const workspaceRoot = resolve(projectConfig.workspaceDir);
+  const repositoryRelativeToRoot = relative(
+    workspaceRoot,
+    resolve(repositoryDir)
+  );
```
hojinzs, line 332: **P3 — `resolve()` does not follow symlinks, so a symlinked workspace root slips past this guard.**

I probed the built guard directly. With `link -> real` and the checkout at `real/repo`:

```
A workspaceDir == repositoryDir              REJECTED
B workspaceDir is parent of checkout         REJECTED
C workspaceDir = <symlink to that parent>    CONSTRUCTED  ← bypass
D trailing-slash equal                       REJECTED
E legacy config (no repositoryDir)           REJECTED
F sibling workspace root                     CONSTRUCTED  (correct)
G workspace.root inside the checkout         CONSTRUCTED  (correct)
```

Case C lands issue #790's exact failure mode again: `resolveIssueWorkspaceRoot` writes through the symlink into the checkout. `realpath` on both operands before `relative()` would close it.

This is inherited from the core check rather than introduced here, and it needs an unusual `workspace.root`, so it is not a blocker — but it is worth noting given this guard exists specifically as the defense-in-depth line. Whatever shared helper comes out of the other comment is the natural place to fix it once.

For the record, F and G confirm no false positives: the documented `workspace.root: .runtime/symphony-workspaces` shape (workspace root *inside* the checkout) still constructs fine.

---

packages/cli/src/commands/start.ts
```diff
@@ -1019,6 +1023,7 @@ const handler = async (
     process.exitCode = 1;
     return;
   }
+  assertDispatchableOrchestratorProjectConfig(projectConfig);
```
hojinzs, line 1026: **P3 (follow-up, not this PR) — `doctor` still reports the rejected config as healthy.**

Placement here is right: after `resolveManagedProjectConfig` and `hasConfiguredRepository`, before `acquireProjectLock` and `OrchestratorService` construction, so nothing is created before the throw. Verified against the built CLI with a legacy `project.json` — clean single-line message, exit 1, no stack trace, and the checkout is untouched:

```
$ node packages/cli/dist/index.js repo start --once --config-dir <cfg>
Project "tenant-a" uses legacy repo-embedded path metadata. Stop the daemon and run 'gh-symphony repo init' again before starting it.
exit=1
```

The gap is what the operator does next. Running `doctor` on that same config prints:

```
PASS Managed project selection   Resolved managed project "tenant-a".
PASS Workspace root              Workspace root is writable: .../legacy-repo
```

So `doctor` is green on the exact configuration `repo start` just refused, and never mentions `repo init`. That is issue #790's fix-outline item 3, which was explicitly optional, so leaving it out of this PR is a defensible scope call — worth a follow-up issue rather than growing this one.

(The migration path itself is intact — I confirmed `doctor` still loads the legacy config, and `packages/cli/src/config.test.ts:129` pins it.)

---

hojinzs (approved):
## Approve

Requirements from #790 are fully covered, I reproduced the fix end-to-end against the built CLI, and nothing I found is blocking. Three non-blocking inline comments (1×P2, 2×P3).

### Requirement coverage vs. #790

| #790 asks | Status |
| --- | --- |
| Fix outline 1 — run `assertDispatchableOrchestratorProjectConfig` on the `repo start` path | ✅ `start.ts:1026`, after config resolution, before lock + service construction |
| Fix outline 2 — defensive check in `OrchestratorService` | ✅ `service.ts:456`, constructor-time |
| Fix outline 3 — `doctor` detects the legacy shape | ⬜ Not done; explicitly optional. See the `start.ts` comment — worth a follow-up issue |
| Regression: legacy config → `repo start` fails, dispatches nothing | ✅ new `start.test.ts` case |
| Regression: config still loads for `doctor` / `repo init` | ✅ pre-existing `config.test.ts:129`, plus verified live |
| Regression: migrated config → daemon starts under `workspace.root` | ✅ existing `service.test.ts` cases, plus verified live |
| Regression: direct `OrchestratorService` guard (equal / parent) | ✅ new `service.test.ts` case |

### Smoke test

Built at `34e0790` and exercised the real CLI binary rather than only the unit tests.

Legacy `project.json` (repo workflow source, no `repositoryDir`, `workspaceDir === repository.path`):

```
$ node packages/cli/dist/index.js repo start --once --config-dir <cfg>
Project "tenant-a" uses legacy repo-embedded path metadata. Stop the daemon and run 'gh-symphony repo init' again before starting it.
exit=1
```

Checkout contents before and after: `.git` only — no `__-/repository/` or `.quarantine-/`, which is the regression #790 reported.

Positive control, migrated config (`repositoryDir` = checkout, `workspaceDir` = `<repo>/.runtime/symphony-workspaces`): passes the guard and proceeds to fail on GitHub auth in this sandbox — no false positive on the documented embedded-workspace shape.

Migration path intact: `doctor` still resolves the legacy project (`PASS Managed project selection`).

Direct probe of the orchestrator guard confirmed equal / parent / trailing-slash / legacy-no-`repositoryDir` all reject, while sibling and inside-the-checkout workspace roots construct normally. One bypass found (symlinked workspace root) — details inline; inherited from the core check, not introduced here.

Local checks: `pnpm lint` ✅ · `pnpm typecheck` ✅ · `pnpm build` ✅ · `start.test.ts` 49/49 ✅ · orchestrator guard test ✅. CI on `34e0790`: Test ✅, Container Smoke ✅.

### Scope

No over-engineering: 2 guards, 2 regression tests, a patch changeset, and an `AGENT_TEST.md` coverage-map row — all traceable to the issue. No unrequested behavior, no config surface added, no weakened compiler settings. The `AGENT_TEST.md` diff is mostly table reflow; its new claims check out. Correctly scoped as a `@gh-symphony/cli` patch.

One judgment call worth surfacing to the author: `assertDispatchableOrchestratorProjectConfig` throws while neighboring failures in this handler use `stderr` + `process.exitCode`. The top-level `main().catch` formats it cleanly (verified above) and `start.ts` already throws for the duplicate-instance case, so this is consistent with existing practice — no change requested.

---

packages/orchestrator/src/service.ts
```diff
@@ -315,6 +315,31 @@ function trackerItemId(
...
+  const repositoryDir =
+    projectConfig.repositoryDir ?? projectConfig.repository?.path;
+  if (!repositoryDir) {
+    return;
+  }
+
+  const workspaceRoot = resolve(projectConfig.workspaceDir);
+  const repositoryRelativeToRoot = relative(
+    workspaceRoot,
+    resolve(repositoryDir)
```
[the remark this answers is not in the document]
moncher-dev: Resolved in 4936273. The shared core containment guard now canonicalizes existing filesystem ancestors, so a symlinked workspace root that contains the checkout is rejected.

packages/orchestrator/src/service.ts
moncher-dev: Resolved in 4936273. The predicate and operator-facing error now live once in core as assertIssueWorkspaceRootOutsideRepository; normalization and OrchestratorService both call it.

packages/orchestrator/src/service.ts
moncher-dev: Resolved in 4936273. Regression coverage now verifies that a workspace-root symlink to the checkout ancestor is rejected by both the core guard and OrchestratorService.

packages/cli/src/commands/start.ts
moncher-dev: Acknowledged. Doctor remediation is explicitly optional for #790; this focused PR preserves doctor loading for migration and records the diagnostic enhancement as out of scope.

PR #798, opened 2026-09-02
feat(worker): isolate custom runtime child credentials

### LINKED ISSUE #778: feat(worker): isolate custom runtime child credentials
Follow-up from #700 / PR #776 review.

## Problem

`runtime.kind: custom` passes the complete worker environment to the operator-supplied child command. Unlike the Codex and Claude adapters completed by #700, this includes raw tracker credentials, GitHub token-broker inputs, agent credential-broker controls, and the host `HOME`/GitHub CLI configuration.

#700 intentionally remains scoped to Codex and Claude because an arbitrary custom command has no known provider credential contract. Its documentation now records this carve-out instead of claiming every coding-agent child is isolated.

## Scope

- Define an explicit environment/credential contract for `runtime.kind: custom`.
- Remove raw tracker and broker secrets from the custom child by default.
- Isolate child `HOME` and `GH_CONFIG_DIR` from operator credentials.
- Preserve only explicitly declared runtime authentication inputs needed by the custom command.
- Document any opt-in compatibility escape hatch and its security consequences.

## Acceptance criteria

- Unit coverage proves GitHub, Linear, GitHub token-broker, and agent credential-broker secrets are absent from the custom child.
- Tests prove the host home and `gh auth` store are not exposed.
- A documented custom runtime can still receive its explicitly configured authentication input and execute successfully.
- Configuration, architecture, and security documentation describe the resulting boundary and any intentional upstream divergence.

## Layer classification

Configuration and Execution, with Integration implications for tracker and credential transport boundaries.

moncher-dev:
## TL;DR

`runtime.kind: custom` has a least-privilege child-environment contract. Default children receive portable process values, a private home, the rendered prompt, and only the dedicated credential named by `runtime.auth.env`.

Cycle 3 closes the remaining P2 gaps: selected tracker secret names are validated before dispatch, and the Docker test now proves a staged Git identity is present while the operator credential helper remains absent.

## Change-point diagram

```text
workflow + selected tracker adapter
              |
              v
parser rejects reserved custom auth names before dispatch
              |
              v
custom child: explicit auth + private HOME/GH_CONFIG_DIR/USERPROFILE
              |
              +-- staged Git identity only; no operator credential helper
```

## Start here

- `packages/core/src/runtime/custom-child-env.ts` defines portable values and reserved credentials.
- `packages/core/src/workflow/parser.ts` validates tracker-provided reserved names during workflow loading.
- `test/e2e/claude/claude-docker.spec.ts` runs the real custom child assertion.

## Risks & rollback

Custom commands that depend on undeclared environment values must declare a dedicated provider credential. `runtime.isolation.inherit_environment: true` is a temporary, security-sensitive migration escape hatch: it restores raw worker credentials while retaining private-home and Git-helper isolation. Reverting this PR restores the former custom-child behavior.

## Changed files

- Core custom-child boundary and workflow parser contract.
- Orchestrator workflow-loading regression coverage.
- Real Docker custom-child fixture and living configuration/architecture documentation.

## Evidence

- `pnpm lint`
- `pnpm test`
- `pnpm typecheck`
- `pnpm build`
- `pnpm e2e:claude`
- Focused core and orchestrator workflow-loader tests.

- [ ] Confirm deployed custom commands declare only a dedicated provider credential through `runtime.auth.env`.
- [ ] Confirm `runtime.isolation.inherit_environment: true` is removed after migration.


packages/core/src/runtime/custom-child-env.ts
```diff
@@ -0,0 +1,122 @@
...
+        env[name] = value;
+      }
+    }
+    const authValue = options.authEnvKey
+      ? (input[options.authEnvKey] ?? source[options.authEnvKey])
+      : undefined;
+    if (options.authEnvKey && authValue !== undefined) {
+      env[options.authEnvKey] = authValue;
+    }
+    removeCustomRuntimeSecrets(env, source);
```
hojinzs: **P1 — a declared `runtime.auth.env` is silently dropped when its name is on the deny list**

The allow-list runs first (L69–74), then `removeCustomRuntimeSecrets` (L75) deletes by name. So any `runtime.auth.env` naming one of the 12 entries in `CUSTOM_RUNTIME_SECRET_ENVIRONMENT_NAMES`, or any name published in `SYMPHONY_TRACKER_SECRET_ENVIRONMENT_NAMES`, is set and then immediately removed — with no error, no warning, and no diagnostic in the child.

Verified with a **real** `spawn` of a custom child (not a mock), `runtime.auth.env: GH_TOKEN`, `GH_TOKEN` present in the worker env:

```
---ENV---
GH_CONFIG_DIR=/tmp/smoke-.../runtime/child-home/gh
GIT_TERMINAL_PROMPT=0
HOME=/tmp/smoke-.../runtime/child-home
PATH=...
PWD=/tmp/smoke-.../repo
SHELL=/bin/bash
SYMPHONY_RENDERED_PROMPT=do the thing
TERM=linux
```

No `GH_TOKEN`. The child exits believing it is unauthenticated. Same result for a tracker-declared name (`SYMPHONY_TRACKER_SECRET_ENVIRONMENT_NAMES: ["MY_KEY"]` + `auth.env: MY_KEY` → `undefined`).

This is a plausible configuration — a custom agent wrapping `gh` would naturally declare `GH_TOKEN` — and it breaks issue #778's acceptance criterion *"A documented custom runtime can still receive its explicitly configured authentication input and execute successfully."*

`docs/configuration.md` does say "do not declare a tracker or broker credential as custom runtime auth", but nothing enforces it, and the failure mode is silent. Preferred fix: **reject reserved names at parse time** with an explicit `workflow_validation_error` naming the conflict, so the operator learns at config load instead of at runtime. (Exempting the declared key from the purge instead would work mechanically but would quietly hand a tracker credential to an arbitrary command, which is the opposite of this PR's intent.)

---

packages/core/src/runtime/custom-child-env.ts
```diff
@@ -0,0 +1,122 @@
...
+  env: NodeJS.ProcessEnv,
+  source: NodeJS.ProcessEnv
+): void {
+  for (const name of [
+    ...readTrackerSecretEnvironmentNames(source),
+    ...CUSTOM_RUNTIME_SECRET_ENVIRONMENT_NAMES,
+  ]) {
+    delete env[name];
+  }
+}
```
hojinzs: **P2 — the deny list runs in the branch where it cannot matter, and not in the branch where it would**

`removeCustomRuntimeSecrets` is called only inside `if (!options.inheritEnvironment)`. In that branch `env` was built from `{}` and contains exactly the allow-listed portable names plus the declared auth key — none of which can be a broker or tracker secret. So every `delete` is a no-op, and the function's *only* observable effect is the P1 bug above.

Conversely, in the `inheritEnvironment: true` branch — where `env` really does start as `{...process.env, ...source, ...input}` and genuinely carries `GITHUB_TOKEN_BROKER_SECRET`, `LINEAR_API_KEY`, tracker secrets — the deny list is never applied. Confirmed: with `inheritEnvironment: true` and `source: { GITHUB_TOKEN: "s" }`, the child env has `GITHUB_TOKEN=s` and 170 keys.

That is defensible as the documented escape hatch, but it means the security boundary rests entirely on the allow-list. Worth being explicit about, because the new tests in `non-codex-runtime.test.ts` read as though they prove the deny list works (asserting `GITHUB_TOKEN_BROKER_SECRET`, `LINEAR_API_KEY`, `CUSTOM_TRACKER_SECRET` etc. are absent) — they pass because of the allow-list, so they would keep passing if this function were deleted outright.

Suggest either deleting `removeCustomRuntimeSecrets` (and `readTrackerSecretEnvironmentNames` with it) as dead weight, or moving the call into the inherit branch if a partial scrub there is actually wanted. Whichever way, add a comment stating that the allow-list is the boundary.

---

packages/core/src/runtime/custom-child-env.ts
```diff
@@ -0,0 +1,122 @@
...
+  "LANG",
+  "PATH",
+  "SHELL",
+  "SYSTEMROOT",
+  "TEMP",
+  "TERM",
+  "TMP",
+  "TMPDIR",
+  "USER",
+  "USERPROFILE",
```
hojinzs: **P2 — `USERPROFILE` is allow-listed, so the private-HOME boundary leaks on Windows**

`HOME` is redirected to `childHome` (L78) but `USERPROFILE` is copied through from the operator environment. On Windows that is the home directory: `os.homedir()`, Git's `~` expansion, and most tooling resolve it before or instead of `HOME`, so the child lands back in the operator's profile and can read `%USERPROFILE%\.gitconfig`, `.ssh`, and the rest.

Observed output of the builder with `source: { PATH, HOME: "/operator", USERPROFILE: "/operator" }`:

```json
{"PATH":"/bin","SHELL":"/bin/bash","TERM":"linux","USERPROFILE":"/operator",
 "HOME":"/child","GH_CONFIG_DIR":"/child/gh","GIT_TERMINAL_PROMPT":"0"}
```

`GH_CONFIG_DIR` is set explicitly so `gh` is fine, but everything else that homes off `USERPROFILE` is not. Since `SYSTEMROOT`/`USERPROFILE` are on this list deliberately for Windows portability, the fix is to override it next to `HOME` rather than drop it:

```ts
env.HOME = options.childHome;
env.USERPROFILE = options.childHome;
```

(A nit while here: `source[name] ?? process.env[name]` on L64 means the builder falls back to the worker's ambient process env even when a curated `source` was supplied — that is how `SHELL`/`TERM` appear above despite not being in `source`. Harmless for the portable names, but it makes the output depend on ambient state; worth a comment saying that is intentional.)

---

packages/core/src/runtime/custom-child-env.ts
hojinzs, line 1: **P2 — this new core security module has no test in its own package**

Every assertion about this boundary lives in `packages/worker/src/non-codex-runtime.test.ts`. Vitest coverage is per-package, so the CI coverage report on `8d14abc` shows this file at **41.12% statements / 0% functions, lines 55–122 uncovered** — the builder is never executed by `packages/core`'s own suite.

For a module whose entire job is a credential boundary, the table-driven cases belong next to it in `packages/core/src/runtime/custom-child-env.test.ts`, where they are cheap and read as the specification: allow-list contents, `HOME`/`GH_CONFIG_DIR` override, `GIT_CONFIG_KEY_*`/`GIT_CONFIG_VALUE_*` stripping, declared-auth forwarding, and the deny-list collision case from the P1 thread.

Also missing everywhere: **`inherit_environment: true` has no behavioral test.** `workflow-loader.test.ts` proves the parser reads the flag, but nothing asserts what the child environment actually looks like when it is on — i.e. nothing pins the documented contract that the escape hatch still overrides `HOME`/`GH_CONFIG_DIR` and still strips host Git-helper variables. That is the security-sensitive path; it should be the best-tested one.

---

packages/orchestrator/src/runtime-factory.ts
```diff
@@ -118,6 +130,12 @@ export class CustomCommandRuntimeAdapter
   shutdown(): void {}
 
   cancel(): void {}
+
+  private resolveChildHome(): string {
+    return resolveAgentChildHome({
+      workingDirectory: this.config.workingDirectory,
+    });
+  }
```
hojinzs, line 156: **P3 — this second custom adapter diverges from the worker one, and is dead in production**

`resolveAgentChildHome` is called here without `runtimeDirectory`, so this adapter isolates into `<workingDirectory>/.runtime/child-home` — inside the repo worktree — while `CustomCommandWorkerRuntimeAdapter` uses `runtimeDirectory ?? runtimeRoot`. Two implementations of the same boundary with two different homes.

`.runtime` is gitignored so there is no repo pollution, and `grep` says `createWorkflowRuntimeAdapter`/`CustomCommandRuntimeAdapter` are referenced only by `runtime-factory.test.ts` — the worker path is what actually runs. So nothing is broken today, but this is unrequested scope relative to #778 (whose layer classification is Configuration + Execution, i.e. the worker), and it leaves a divergent copy of a security boundary that a future caller could pick up.

Two other things about this copy specifically: it never injects `SYMPHONY_RENDERED_PROMPT`, and its new `async prepare()` (L82) has no caller — only `packages/worker/src/index.ts:744` awaits `prepare`, so if this adapter were ever wired up, `HOME` would point at a directory that was never created.

Either delete it or route it through the same resolution the worker uses; keeping it as-is is the option that ages worst.

---

packages/core/src/workflow/parser.ts
```diff
@@ -1130,6 +1096,12 @@ function parseRuntimeConfig(
           "trust_repo_config",
           "runtime.isolation.trust_repo_config"
         ) ?? false,
+      inheritEnvironment:
+        readOptionalBoolean(
+          isolation,
+          "inherit_environment",
+          "runtime.isolation.inherit_environment"
+        ) ?? false,
     },
```
hojinzs: **P3 — `inherit_environment` parses for every runtime kind but is only honoured by `custom`**

It lands in the shared `WorkflowRuntimeIsolationConfig`, so `runtime.kind: claude-print` with `isolation.inherit_environment: true` parses cleanly and is then silently ignored (`non-codex-runtime.ts:243-246` only forwards `bare` and `strictMcpConfig`). Failing safe is the right direction, but an operator who sets it there gets no signal that their configuration does nothing.

Given the flag is documented as custom-only and security-sensitive, a validation error for non-custom runtimes would be consistent with how `codex.approval_policy` is handled — cheap, and it keeps the escape hatch from looking more general than it is.

*(nit, no action needed: the surrounding `readProviderStringList` / `readNormalizedOptionalString` reflow in this file is unrelated to #778 and adds ~50 lines of review noise. I confirmed `main` is not Prettier-clean for this file, so it is a legitimate cleanup rather than a formatter disagreement — just easier to review as its own commit next time. `pnpm format` currently fails on 87 files repo-wide, which looks like a pre-existing gap worth its own issue.)*

---

hojinzs (changes requested):
## Review — round 1 of 6

The design is right and the boundary genuinely works: I spawned a **real** custom child (not a mock) against this branch and the operator's credentials, `gh auth` store, and Git credential helper are all absent from it. One P1 blocks approval — a declared `runtime.auth.env` can be silently discarded — plus a few P2s. None of this is architectural; all of it is small.

### Smoke test — real process spawn

`CustomCommandWorkerRuntimeAdapter` with a real `/bin/sh` child, an operator `HOME` containing `.config/gh/hosts.yml` (`oauth_token: OPERATOR_SECRET`) and a `.gitconfig` carrying `credential.helper = store`, worker env seeded with `GITHUB_TOKEN`, `GITHUB_TOKEN_BROKER_SECRET`, `LINEAR_API_KEY`, `GIT_CONFIG_KEY_0`/`VALUE_0`, and a tracker-declared `CUSTOM_TRACKER_SECRET`:

```
EXIT: 0 RESULT: success
---ENV---
GH_CONFIG_DIR=/tmp/smoke-.../runtime/child-home/gh
GIT_TERMINAL_PROMPT=0
HOME=/tmp/smoke-.../runtime/child-home
PATH=... PWD=/tmp/smoke-.../repo SHELL=/bin/bash TERM=linux
SYMPHONY_RENDERED_PROMPT=do the thing
---HOME--- . .. .gitconfig gh
---GH_CONFIG_DIR--- . ..          ← operator hosts.yml not reachable
---gitconfig--- [user] name = Operator / email = op@example.com   ← identity only, no credential.helper
---stdin--- do the thing
```

Every tracker token, broker control, and host Git-helper variable is gone; the private `HOME` is real; the staged `.gitconfig` carries identity without the credential helper; the prompt arrives on both stdin and `SYMPHONY_RENDERED_PROMPT`. That is exactly the contract the PR describes.

**But `GH_TOKEN` is missing too** — and it was the declared `runtime.auth.env`. See the P1 thread.

### Requirements traceability against #778

| --- | --- |
| Unit coverage proves GitHub/Linear/broker secrets absent from the custom child | ✅ covered, and reproduced above with a real spawn |
| Tests prove host home and `gh auth` store are not exposed | ✅ covered, and reproduced above |
| A documented custom runtime receives its declared auth input and executes successfully | ⚠️ **partial** — works for non-colliding names, silently fails for deny-listed ones (P1) |
| Configuration, architecture, and security docs describe the boundary and the divergence | ✅ `README.md`, `docs/configuration.md`, `docs/architecture.md`, ADR — clear and honest about the escape hatch; ❌ `AGENT_TEST.md` not updated (below) |

### Findings

| # | Sev | Where | What |
| --- | --- | --- | --- |
| 1 | **P1** | `custom-child-env.ts:69-75` | Allow-list then deny-list ordering silently drops a declared `runtime.auth.env` that names a deny-listed or tracker-declared key. Verified: `auth.env: GH_TOKEN` → child gets nothing, no error. |
| 2 | P2 | `custom-child-env.ts:84-94` | `removeCustomRuntimeSecrets` is a no-op in the only branch that calls it; its sole effect is finding 1. Not applied in the `inherit_environment` branch, where it would matter. |
| 3 | P2 | `custom-child-env.ts:14` | `USERPROFILE` is forwarded from the operator env while `HOME` is redirected — the private-home boundary leaks on Windows. |
| 4 | P2 | `custom-child-env.ts` | No test in `packages/core` (CI coverage: 0% functions, lines 55–122 uncovered). `inherit_environment: true` has no behavioral test anywhere. |
| 5 | P2 | `AGENT_TEST.md` | Scenario table not updated — see below. |
| 6 | P3 | `runtime-factory.ts:134` | Duplicate custom adapter with a different child home, unreferenced in production, `prepare()` never called. |
| 7 | P3 | `parser.ts:1099` | `inherit_environment` parses for every runtime kind, honoured only by `custom`. |

### On the previous Codex review

Both threads are still open; my read differs on severity.

- **Docker isolation scenario (P1 → P2, thread [#discussion_r3914314040](https://github.com/hojinzs/github-symphony/pull/798#discussion_r3914314040)).** The concern that the real spawn and private-`HOME` boundary were never exercised end to end was fair — I have now done that manually, and it passes. What remains is a repo-policy gap, not a correctness one: `AGENTS.md` §Mandatory Post-Work Checks item 4 requires new E2E scenarios to be recorded in the `AGENT_TEST.md` scenario table, and the natural home is the existing *"Agent child isolation and host Git transport"* row, which this PR leaves describing Codex/Claude only. Note the table's own precedent accepts "unit-covered because the Docker fixture uses a stub worker" as a justification, so a row extension is sufficient — a full Docker scenario is nice-to-have, not required.
- **`packages/cli/README.md` (P1 → P3, thread [#discussion_r3914314057](https://github.com/hojinzs/github-symphony/pull/798#discussion_r3914314057)).** Overstated. That README has never documented the custom runtime — it says "Currently supported runtimes: Codex CLI and Claude Code" — so there is no stale isolation claim to correct, and the root `README.md` plus `docs/configuration.md` cover the migration properly. Still worth a two-line note for npm consumers who never see the repo root, but it should not block.

Only finding 6. The orchestrator-side adapter change is outside #778's Configuration + Execution classification and is dead code; the rest of the diff is proportionate. The `parser.ts` reflow is Prettier-driven cleanup rather than churn for its own sake (I confirmed `main` is not Prettier-clean there) — fine to keep, just noisy.

### Verification I ran on `8d14abc`

```
pnpm lint       ✅
pnpm typecheck  ✅
pnpm build      ✅
pnpm test       ✅  (all packages green — e.g. orchestrator 386, cli 616)
npx prettier --check <13 changed files>   ✅  All matched files use Prettier code style!
```

The PR's evidence block holds up.

### To approve

Fix finding 1 (parse-time rejection of reserved auth names is my preference — it turns a silent runtime failure into a clear config error), plus 3 and 4. Findings 2, 5, 6, 7 are cheap enough to fold in; if you'd rather defer 6 to its own issue, say so on the thread and I'll take that as settled.

---

docs/architecture.md
```diff
@@ -148,13 +148,13 @@ The rows below are owned by the focused conformance suites rather than a
...
+| §17.5 child credential and tool boundary         | `packages/runtime-{codex,claude}` and `packages/worker/src/non-codex-runtime.test.ts` prove provider subprocesses/default custom commands exclude raw GitHub/Linear tokens, broker controls, host HOME/GH config, and Git credential helpers while custom auth is explicitly forwarded; `packages/worker/src/codex-dynamic-tools.test.ts` covers structured rejection of unsupported dynamic tools. The Claude Docker black-box repeats the environment and generated-config assertions. `runtime.isolation.inherit_environment` is the documented custom-only upstream divergence. |
```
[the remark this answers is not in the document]
moncher-dev: Resolved in 8e7cad1. TC-20 now runs a real custom child in Docker in both default and compatibility modes, and AGENT_TEST.md records the coverage.

README.md
```diff
@@ -997,7 +997,12 @@ transport is described in
 When no direct provider API key is configured, a non-bare runtime stages only
 the provider login into this private home: Codex `auth.json`, or Claude's
 `claudeAiOauth` entry without `mcpOAuth`. Host agent configuration and GitHub
-CLI credentials remain outside the child boundary.
+CLI credentials remain outside the child boundary. A custom command can receive
+only the provider key named by `runtime.auth.env`. The documented
+`runtime.isolation.inherit_environment: true` migration escape hatch instead
```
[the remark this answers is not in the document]
moncher-dev: Resolved in 8e7cad1. packages/cli/README.md now documents the default custom-runtime contract, runtime.auth.env migration, and the security-sensitive custom-only escape hatch.

packages/core/src/runtime/custom-child-env.ts
moncher-dev: Resolved in 8e7cad1. The parser now rejects reserved GitHub/Linear/broker names and tracker-declared secret names for runtime.auth.env, with core tests for both cases.

packages/core/src/runtime/custom-child-env.ts
moncher-dev: Resolved in 8e7cad1. I removed the dead default-branch purge: the allowlist is the explicit default boundary, while compatibility mode intentionally restores raw worker credentials and is tested/documented as such.

packages/core/src/runtime/custom-child-env.ts
moncher-dev: Resolved in 8e7cad1. USERPROFILE is now redirected to the private child home in both modes, with core and Docker real-process coverage.

packages/core/src/runtime/custom-child-env.ts
moncher-dev: Resolved in 8e7cad1. Added packages/core/src/runtime/custom-child-env.test.ts for the default contract and compatibility behavior, including home, GH config, Git injection stripping, and declared auth.

packages/orchestrator/src/runtime-factory.ts
moncher-dev: Resolved in 8e7cad1. The public custom adapter now accepts the runtime directory, prepares the same private child home before spawn, and injects the rendered prompt; runtime-factory coverage pins that behavior.

packages/core/src/workflow/parser.ts
moncher-dev: Resolved in 8e7cad1. runtime.isolation.inherit_environment now fails validation unless runtime.kind is custom, with parser coverage and configuration documentation.

test/e2e/claude/claude-docker.spec.ts
```diff
@@ -77,6 +78,111 @@ afterEach(async () => {
...
+      const commandPath = join(root, "custom-child.cjs");
+      await mkdir(join(hostHome, ".config", "gh"), { recursive: true });
+      await mkdir(workspace, { recursive: true });
+      await writeFile(
+        join(hostHome, ".config", "gh", "hosts.yml"),
+        "github.com:\\n  oauth_token: operator-secret\\n"
+      );
+      await writeFile(
+        join(hostHome, ".gitconfig"),
+        "[user]\\n  name = Operator\\n[credential]\\n  helper = store\\n"
```
hojinzs: **P2 — the fixture writes literal `\n` escapes, so `gitCredentialHelperVisible: false` passes vacuously**

The TS source here is `"[user]\\n  name = Operator\\n[credential]\\n  helper = store\\n"` (verified against the raw bytes, not the diff rendering), so the operator `.gitconfig` on disk is one line containing literal backslash-n pairs — not four lines.

`stageGitUserIdentity` splits on `/\r?\n/`, gets a single line, and the section regex `^\s*\[\s*([^\s\]]+)\s*\]\s*$` never matches it, so `inUserSection` stays false and no identity is found. Confirmed by running the real function against both fixtures:

```
=== stageGitUserIdentity with this file's literal \n content ===
staged? false exists? false
real-newline fixture staged? true
```

Two consequences:

1. `prepare()` never creates `child-home/.gitconfig` in this test, so the child's `fs.existsSync(path.join(env.HOME, ".gitconfig"))` is false and the `gitCredentialHelperVisible: false` assertion at L169 is satisfied by an absent file rather than by a scrubbed one. It would keep passing if `stageGitUserIdentity` were deleted outright, or if it copied the operator file verbatim — i.e. it does not currently pin the boundary it names.
2. `hosts.yml` is written as invalid YAML for the same reason. Lower impact, since only `existsSync` is asserted on it.

Fix is one character per escape — use real newlines. With them, my out-of-band run of the same adapter shows the child home getting exactly:

```
[user]
	name = Operator
	email = op@example.com
```

no `[credential] helper = store`, which is the assertion actually earning its keep.

---

packages/core/src/workflow/parser.ts
```diff
@@ -1110,6 +1077,33 @@ function parseRuntimeConfig(
...
+      "runtime.isolation.inherit_environment",
+      'Workflow front matter field "runtime.isolation.inherit_environment" is supported only for runtime.kind "custom".'
+    );
+  }
+
+  const authEnv = readOptionalString(auth, "env");
+  if (
+    kind === "custom" &&
+    authEnv &&
+    isCustomRuntimeReservedAuthEnvironmentName(authEnv, env)
```
hojinzs: **P2 — the tracker-secret half of this check is ambient-env-dependent, so the same `WORKFLOW.md` validates differently in the orchestrator and the worker**

`isCustomRuntimeReservedAuthEnvironmentName` reads tracker names out of `SYMPHONY_TRACKER_SECRET_ENVIRONMENT_NAMES`, which only exists in the **worker** environment — the orchestrator injects it at dispatch (`service.ts:3178-3180`) and does not have it in its own process env when it parses the workflow. Verified against the built parser with one document and two envs:

```
=== parse: reserved GH_TOKEN as custom auth ===
rejected: ... cannot use reserved credential environment name "GH_TOKEN".

=== parse: tracker-declared name, orchestrator env ===
ACCEPTED, auth.env = TRACKER_SECRET

=== parse: same doc, worker env ===
rejected: ... cannot use reserved credential environment name "TRACKER_SECRET".
```

So the static reserved list (the `GH_TOKEN`/`LINEAR_API_KEY`/broker half) works everywhere — that part fully closes the P1 from the last round, thank you. But the tracker half only fires in the worker. Net effect for an operator who declares a tracker-provided name: `doctor` and orchestrator load report the workflow as valid, the issue is dispatched, and then every run dies at worker startup with a validation error, on repeat.

It does fail **closed** — I confirmed the worker parse throws before any child is spawned, so no tracker credential reaches the custom command — which is why this is P2 and not P1. It is worth closing because `docs/configuration.md` now states flatly that "`runtime.auth.env` fails workflow validation when it names … a tracker-declared secret name", and that is only conditionally true.

Cheapest fix that matches the existing shape: the orchestrator already computes `resolveTrackerSecretEnvironmentNames(trackerAdapter)` — thread those names into the parse options at workflow load so validation is deterministic wherever it runs. Failing that, narrow the doc sentence to say the tracker-name check is enforced at worker startup.

---

packages/core/src/runtime/custom-child-env.ts
```diff
@@ -0,0 +1,123 @@
...
```
hojinzs: nit — `USERPROFILE` is now dead in this list: it is copied through at L74-81 and then unconditionally overwritten with `options.childHome` at L91. The override is the right fix for the Windows leak from the last round (confirmed: both modes now report `USERPROFILE` = child home), but leaving the name in an allow-list that is documented as "portable values forwarded from the operator environment" reads as though it is still forwarded. Dropping the entry makes the allow-list say what it means; no behavior change either way.

---

hojinzs (approved):
## Approve — round 2

All six findings from the previous round are genuinely fixed, and I re-verified each by running the built adapter and parser rather than reading the diff.

**Previous round, re-verified against `8e7cad1`:**

| # | Finding | Status |
|---|---|---|
| P1 | declared `runtime.auth.env` silently dropped by the deny list | Fixed. The default child now reports `CUSTOM_AGENT_TOKEN: "custom-token"`, and reserved names are rejected at parse time (see the parser thread for one residual). |
| P2 | deny list ran only in the branch where it could not matter | Fixed. The purge is gone; `removeChildHostCredentialEnvironment` now runs in both branches and is load-bearing in compatibility mode. |
| P2 | `USERPROFILE` leaked the operator home on Windows | Fixed. Both modes report `USERPROFILE` = child home. |
| P2 | new core security module had no test in its own package | Fixed. `custom-child-env.test.ts` covers both modes. |
| P3 | orchestrator custom adapter diverged and was dead | Fixed. Takes `runtimeDirectory`, prepares the home before spawn, injects the prompt. |
| P3 | `inherit_environment` parsed for every runtime kind | Fixed. Non-custom kinds now fail validation. |

**Smoke test.** No preview environment applies here, so I ran the real `CustomCommandWorkerRuntimeAdapter` out of `dist` against an actual child process, in both modes, with a seeded operator home:

- **Default:** the child sees exactly `CUSTOM_AGENT_TOKEN, GH_CONFIG_DIR, GIT_TERMINAL_PROMPT, HOME, PATH, SHELL, SYMPHONY_RENDERED_PROMPT, TERM, USERPROFILE` — nine keys. No `GITHUB_TOKEN`, `LINEAR_API_KEY`, broker secret, tracker secret, or `GIT_CONFIG_KEY_*`. `HOME`/`USERPROFILE` are the private child home, `GH_CONFIG_DIR` sits under it, and the staged `.gitconfig` carries only `[user] name/email` — no `helper = store`.
- **Compatibility:** raw worker credentials are present by design, while `HOME`, `USERPROFILE`, `GH_CONFIG_DIR`, and the Git-helper strip still hold.

That satisfies #778's acceptance criteria, including the one the earlier P1 broke — "a documented custom runtime can still receive its explicitly configured authentication input and execute successfully."

`pnpm build` and `pnpm test` pass locally; both required checks are green on this head. I also checked scope against #778 and found no over-engineering — the orchestrator adapter change and the parse-time validation were both asked for in review.

**Not blocking, but please pick up before merge:** the new TC-20 fixture writes literal `\n` escapes into the operator `.gitconfig`, which makes `stageGitUserIdentity` no-op and leaves the `gitCredentialHelperVisible: false` assertion passing vacuously. It is a one-character-per-escape fix and it is what makes the new Docker coverage actually pin the Git boundary. The parser thread (ambient-env-dependent tracker-name validation) fails closed, so it is fine as a follow-up issue if you would rather keep this PR tight; if you do that, narrow the sentence in `docs/configuration.md` so the doc does not overstate what validation catches. The `USERPROFILE` allow-list entry is a pure nit.

---

test/e2e/claude/claude-docker.spec.ts
moncher-dev: Fixed in `3fb10ba`: the Docker fixture now writes real newline-delimited GitHub and Git configuration, and the child assertion requires staged Git identity while confirming the credential helper is absent.

packages/core/src/workflow/parser.ts
moncher-dev: Fixed in `3fb10ba`: selected tracker adapter `secretEnvironmentNames()` now participates in workflow validation before dispatch. The orchestrator loader regression proves `TRACKER_SECRET` is rejected without worker-only ambient metadata.

packages/core/src/runtime/custom-child-env.ts
moncher-dev: Fixed in `3fb10ba`: `USERPROFILE` is no longer presented as a forwarded portable value; it remains explicitly set to the isolated child home.

packages/core/src/workflow/parser.ts
```diff
@@ -1110,6 +1082,37 @@ function parseRuntimeConfig(
...
+  }
+
+  const authEnv = readOptionalString(auth, "env");
+  if (
+    kind === "custom" &&
+    authEnv &&
+    isCustomRuntimeReservedAuthEnvironmentName(
+      authEnv,
+      env,
+      options.trackerAdapter?.secretEnvironmentNames?.() ?? []
```
hojinzs, line 1106: **P2 — this reads the *injected* adapter, not the *selected* one, so the CLI validation path is still uncovered**

Thanks for threading the adapter through — this genuinely closes the orchestrator half, and I confirmed that below. But `options.trackerAdapter` is the compatibility injection seam, not the resolved adapter. Twelve lines earlier this same parse already computes the selected adapter:

```ts
// L213-215
const trackerAdapter = trackerKind
  ? (options.resolveTrackerAdapter?.(trackerKind) ?? options.trackerAdapter)
  : options.trackerAdapter;
```

and `defaultLifecycle` (L229) and `validateProviderConfig` (L262) both use that local. This new check bypasses it. Since `resolveTrackerAdapter` wins over `trackerAdapter` at L214, the two halves of the parser can disagree about which adapter is selected.

That matters because `resolveTrackerAdapter` is the documented production path (`ParseWorkflowOptions` L46-47: *"Prefer resolveTrackerAdapter in production, where tracker.kind selects it"*), and it is what `doctor` and `workflow validate` use — `doctor.ts:716` and `workflow.ts:818` pass `resolveTrackerAdapter: resolveWorkflowConfigTrackerAdapter` and no `trackerAdapter`. Only `orchestrator/git.ts:688` uses the injection seam, which is why `git.test.ts` passes.

Verified against the built parser, one document, four option shapes, with no ambient `SYMPHONY_TRACKER_SECRET_ENVIRONMENT_NAMES`:

```
A trackerAdapter (orchestrator path) : REJECTED  ...reserved credential environment name "TRACKER_SECRET"
B resolveTrackerAdapter (CLI path)   : ACCEPTED  auth.env=TRACKER_SECRET
C both supplied                      : REJECTED  ← follows the injected adapter, not the resolved one
D static GH_TOKEN via resolve        : REJECTED
```

Case B is the gap; case C is the inconsistency. Case D confirms the static reserved list is unaffected — that half works everywhere.

Net effect is smaller than last round's, and in the safe direction: the orchestrator now rejects at load, so nothing is dispatched and no tracker credential reaches a custom command. What is left is that `gh-symphony doctor` reports **pass** on a workflow the orchestrator will then refuse — the pre-flight tool disagreeing with the thing it is meant to pre-flight. That is why it is P2 rather than P1, and why I am not blocking on it.

Fix is to use the local that already exists:

```ts
const runtime = hasRuntime
  ? parseRuntimeConfig(runtimeNode, env, trackerAdapter)
  : null;

function parseRuntimeConfig(
  runtime: Record<string, WorkflowFrontMatterNode>,
  env: NodeJS.ProcessEnv,
  trackerAdapter: WorkflowConfigTrackerAdapter | undefined
): WorkflowRuntimeConfig {
  // ...
  isCustomRuntimeReservedAuthEnvironmentName(
    authEnv,
    env,
    trackerAdapter?.secretEnvironmentNames?.() ?? []
  )
```

`parseRuntimeConfig` is called at L270-272, after the adapter is resolved at L213, so it is available. I applied exactly this and rebuilt: `tsc` clean, and A/B/D all reject while C correctly follows the resolved adapter. A companion case in `git.test.ts` using `resolveTrackerAdapter` would pin it.

If you would rather ship this as-is, that is defensible given it fails closed — but then please add the `doctor` divergence to the follow-up issue, since `docs/configuration.md` reads as though workflow loading is uniformly authoritative.

---

hojinzs (approved):
## Approve — round 3 of 6

Both round-2 follow-ups are done, and I re-verified each by executing the real code rather than reading the diff. One residual P2 on the parser thread (the CLI validation path is still uncovered) — it fails closed, so I am not holding the PR for it.

**⚠️ Merge blocker, not a review finding:** this branch is currently un-mergeable. `AGENT_TEST.md` conflicts with `main`; `docs/configuration.md` and `test/e2e/claude/claude-docker.spec.ts` auto-merge cleanly. Please merge `main` in and resolve before landing.

### Round-2 follow-ups, re-verified against `3fb10ba`

| Item | Status |
|---|---|
| P2 — TC-20 fixture wrote literal `\n`, so `gitCredentialHelperVisible: false` passed vacuously | **Fixed, and now load-bearing.** |
| P2 — tracker-name validation was ambient-env-dependent | **Fixed for the orchestrator**; the CLI path remains — see the parser thread. |
| nit — `USERPROFILE` dead in the portable allow-list | **Fixed.** Entry dropped; still explicitly set to the child home in both modes. |

### Smoke test — real execution, no preview environment applies

**The fixture fix actually pins the boundary now.** I ran the real `stageGitUserIdentity` against the exact bytes each fixture version writes:

```
post-fix (real \n) : .gitconfig exists=true
   content: "[user]\n\tname = Operator\n\temail = op@example.com\n"
   gitIdentityVisible        -> true
   gitCredentialHelperVisible -> false
pre-fix (literal \\n): .gitconfig exists=false
   gitIdentityVisible        -> false   ← the new assertion would have caught the old fixture
```

That is the point I was making last round: the added `gitIdentityVisible: true` assertion is what turns "no file" into "scrubbed file", and it fails against the pre-fix fixture. Good fix.

**Child environment contract, both modes**, via the real `buildCustomRuntimeChildEnvironment` with a seeded operator env (`GITHUB_TOKEN`, `LINEAR_API_KEY`, `GITHUB_TOKEN_BROKER_SECRET`, `CUSTOM_TRACKER_SECRET`, `GIT_CONFIG_KEY_0`/`VALUE_0`, operator `HOME`/`USERPROFILE`):

```
inheritEnvironment=false  (8 keys)
  HOME=/child  USERPROFILE=/child  GH_CONFIG_DIR=/child/gh
  CUSTOM_AGENT_TOKEN=custom-token          ← declared auth forwarded
  GITHUB_TOKEN / CUSTOM_TRACKER_SECRET / GIT_CONFIG_KEY_0: absent

inheritEnvironment=true   (141 keys)
  HOME=/child  USERPROFILE=/child  GH_CONFIG_DIR=/child/gh
  GITHUB_TOKEN=gh-secret  CUSTOM_TRACKER_SECRET=trk   ← restored by design
  GIT_CONFIG_KEY_0: absent                            ← Git-helper strip still holds
```

`USERPROFILE` is the private home in both modes, and the escape hatch still keeps private-home and Git-helper isolation. Unchanged from round 2, as expected — the allow-list edit was cosmetic.

### Requirements traceability against #778

| --- | --- |
| Unit coverage proves GitHub/Linear/broker secrets absent from the custom child | ✅ |
| Tests prove host home and `gh auth` store are not exposed | ✅ |
| A documented custom runtime receives its declared auth input and executes successfully | ✅ — the round-1 P1 stays fixed; `CUSTOM_AGENT_TOKEN` arrives |
| Configuration, architecture, and security docs describe the boundary and the divergence | ✅ |

### Scope

No over-engineering in this cycle. Every hunk in `3fb10ba` traces to a round-2 thread: the parser threading and its `git.ts`/`git.test.ts` companions to the tracker-name P2, the fixture to the TC-20 P2, the allow-list line to the nit, and the two doc edits to describing what changed. The `docs/configuration.md` addition is correctly scoped to "the orchestrator and the dispatched worker" and does not overclaim. Changeset present (`.changeset/isolate-custom-runtime-child.md`).

### Verification I ran on `3fb10ba`

```
pnpm lint       ✅
pnpm typecheck  ✅
pnpm build      ✅
pnpm test       ✅  2043 tests, 138 files, all packages green
npx prettier --check <7 changed files>   ✅
```

(First `pnpm test` run failed in `tracker-linear` on an unresolved `@gh-symphony/tool-linear-graphql` entry — a local build-order artifact on a cold `node_modules`, not a PR defect. Green after `pnpm build`.)

Approving. Please resolve the `AGENT_TEST.md` conflict before merge, and either take the one-line parser fix or record the `doctor` divergence on the follow-up issue.

---

PR #799, opened 2026-09-02
fix(worker,orchestrator): retain dirty unpublished worktrees

### LINKED ISSUE #789: fix(worker,orchestrator): retain dirty unpublished worktrees
## Summary

Host-side branch transport can succeed while the coding-agent workspace still contains tracked or untracked edits. Git pushes only committed branch state, so those edits remain unpublished; a later terminal-state cleanup can then remove the workspace.

This was identified during review of #700 / PR #776 and is intentionally deferred from that security-boundary change. Issue #700 assigns committing to the child, and this behavior predates its host-side transport implementation, so #700 should not expand to solve it.

## Current behavior

1. The coding-agent child commits some changes but leaves other tracked or untracked edits in the worktree.
2. Worker host-side transport successfully pushes the assigned branch.
3. The run reports transport success because the committed branch is published.
4. If the tracker reaches a terminal state, workspace cleanup can remove the remaining unpublished edits.

## Desired behavior

- Detect a dirty assigned worktree before reporting publication complete.
- Represent dirty unpublished work distinctly from Git transport failure and ordinary agent failure.
- Retain the issue workspace while unpublished edits remain, including terminal-state and startup cleanup paths.
- Keep retry/recovery behavior bounded and diagnosable.
- Cover tracked modifications and untracked files with worker/orchestrator tests and a Docker black-box scenario if lifecycle behavior changes.

## Acceptance criteria

- A successful assigned-branch push does not imply that a dirty worktree is fully published.
- Terminal cleanup cannot delete a workspace whose latest run records dirty unpublished work.
- The status/error contract explains whether committed transport succeeded and which work remains unpublished.
- Existing clean-worktree host transport behavior remains unchanged.
- Living documentation and the `@gh-symphony/cli` patch changeset are updated with the implementation.

## References

- Parent review scope: #700
- Delivery PR: #776
- Review threads: https://github.com/hojinzs/github-symphony/pull/776#discussion_r3905023558 and https://github.com/hojinzs/github-symphony/pull/776#discussion_r3905103912

moncher-dev:
## TL;DR

A successful assigned-branch push is now reported as successful even when tracked or untracked work remains in the worktree. That remaining work is persisted as a bounded, dedicated `unpublishedWorktree` outcome, and terminal/startup cleanup retains its workspace for recovery.

## Change-point diagram

```text
assigned worktree
  ├─ assigned branch push succeeds ──> committed branch is published
  └─ git status detects remaining edits
       └─ bounded unpublishedWorktree record
            ├─ run stays succeeded; lastError stays null
            ├─ status/dashboard identify the unpublished files
            └─ terminal and startup cleanup retain the workspace
```

## Start here

- `packages/worker/src/git-transport.ts` detects, bounds, and reports tracked, untracked, and rename-path leftovers after a successful push.
- `packages/orchestrator/src/service.ts` persists the distinct outcome and retains workspaces from their record or the latest run only.
- `e2e/scenarios/23-dirty-unpublished-worktree.md` documents Docker TC-23's lifecycle simulation and scope.

## Changes

- Replaced the overloaded `lastError` string marker with a structured `unpublishedWorktree` status contract across worker, channel, orchestration, and dashboard surfaces.
- Capped diagnostic lists and added omitted counts; rename diagnostics retain both source and destination paths.
- Preserved dirty workspaces during terminal retry, terminal cleanup, and startup cleanup, using latest-run evidence when the workspace record has no marker.
- Kept clean worktree transport and real Git transport failures unchanged.
- Documented this cleanup retention as an intentional repository-level divergence from Symphony §8.6.

## Evidence

- `pnpm lint` — passed.
- `pnpm test` — passed (including worker 225/225 and orchestrator 387/387).
- `pnpm typecheck` — passed.
- `pnpm build` — passed.
- `pnpm --filter @gh-symphony/worker test -- git-transport.test.ts` — passed (18/18).
- `./e2e/run-e2e.sh dirty-unpublished-worktree 60` — passed. TC-23 simulates the worker transport outcome and validates persisted status and terminal retention; real Git detection is unit-covered.

## Risks & rollback

- A dirty workspace remains intentionally until its unpublished work is recovered; the bounded record identifies its assigned branch, pushed head, and remaining files.
- Reverting this PR restores the prior cleanup behavior.

## Changed files

- Worker transport and runtime event propagation
- Core status and worker-channel contracts
- Orchestrator lifecycle/cleanup behavior and regression tests
- Dashboard status projection
- Docker E2E TC-23 and living documentation

- [ ] Confirm an operator can locate retained files after a real coding-agent run.
- [ ] Confirm clean assigned-branch transport still permits terminal cleanup.

## Human Validation

- [ ] Confirm the recovered status contract distinguishes transport failure from unpublished work.
- [ ] Confirm retention is limited to Git transport failure and unpublished-worktree outcomes.


packages/orchestrator/src/service.ts
```diff
@@ -5791,21 +5795,31 @@ export class OrchestratorService {
...
-      (run) =>
-        run.projectId === projectId &&
-        run.issueId === issueId &&
-        isGitTransportFailure(run)
+    for (const run of await this.store.loadAllRuns()) {
+      if (run.projectId !== projectId || run.issueId !== issueId) continue;
+      const reason = unpublishedGitWorkReason(run.lastError);
+      if (reason) return reason;
+    }
+    return null;
```
hojinzs: **P1 — the workspace record short-circuits the run scan, so a run-recorded marker cannot save the workspace**

This is the same defect Codex raised at #discussion_r3914421977, and I reproduced it. Patching the new `it.each` retention test so the workspace record has `lastError: null` while the run still carries the marker makes **both** parameterised cases fail — the sentinel file is gone:

```
Error: ENOENT: no such file or directory, open
  '.../acme_platform_1-0e0ed77a1c2f4ef5/unpublished-commit.txt'
```

The run record can legitimately carry `git_unpublished_worktree:` while the workspace record does not, because `recordGitTransportWorkspaceState` (service.ts:3597) copies the marker onto the workspace only on the finalization path, *after* the worker heartbeat already persisted it onto the run. Two reachable windows:

1. An orchestrator restart in between — startup terminal cleanup then deletes the workspace.
2. The terminal release inside `run.status === "retrying"` (service.ts:3546-3563). Its guard is still `!isGitTransportFailure(runWithTokens)`, which the new classification does not cover, and `recordGitTransportWorkspaceState` has not run yet on that path. Reachable today when `convergenceDetected` forces `runPhase: "failed"` on a run that also left a dirty worktree.

So AC *"Terminal cleanup cannot delete a workspace whose latest run records dirty unpublished work"* is not met.

One caution on the fix: don't simply fall through to the loop below when the workspace marker is absent. That loop matches **any** historical run for the issue via first-match over `loadAllRuns()`, so a marker from an old run would pin the workspace forever even after a later clean run. Resolve the issue's **latest** run and classify that one — which is also what #789 actually specifies ("whose *latest* run records dirty unpublished work"). The existing `some()`-style fallback has the same imprecision and is worth tightening in the same pass.

Please add a regression test for exactly this shape: workspace record present with `lastError: null`, latest run carrying `git_unpublished_worktree:`.

---

packages/worker/src/git-transport.ts
```diff
@@ -35,6 +39,14 @@ export function applyGitTransportAttempt(
     process.stderr.write(message)
 ): 0 | 1 {
   if (attempt.ok) {
+    const unpublished = attempt.result.unpublishedWorktreeChanges;
+    if (unpublished) {
+      state.lastError = formatUnpublishedWorktreeError(attempt.result);
+      writeStderr(
+        `[worker] host Git transport pushed ${attempt.result.branch} at ${attempt.result.head}, but tracked or untracked work remains unpublished\n`
+      );
+      return 0;
```
hojinzs, line 57: **P2 — routing the marker through `lastError` makes it indistinguishable from failure at every observability surface**

#789 asks for dirty unpublished work to be *"represented distinctly from Git transport failure and ordinary agent failure."* The `git_unpublished_worktree:` prefix distinguishes it to code that opts in, but the transport layer is the failure channel, so everything that already consumes `lastError` now reports a *successful* run as an error:

- `service.ts:3511` — `if (workerInfo.lastError)` appends a `worker-error` run event. A succeeded run now emits `worker-error`.
- `snapshot-builder.ts:128` — the per-run `error` field in the status snapshot.
- `dashboard/store.ts:321,338` — issue-row `last_error`.

`recordGitTransportWorkspaceState` already has to special-case this (`transportSucceeded` requires `lastError === null`, so a dirty run is excluded from the success branch and re-admitted via the marker check), which is a symptom of the same overload.

A dedicated field on the run/workspace record — e.g. `unpublishedWorktree: { branch, head, tracked, untracked } | null`, plus a `git-unpublished-worktree` event — would satisfy the AC without teaching every consumer a new string prefix, and would drop the `isGitUnpublishedWorktree` carve-out at service.ts:3649. If you'd rather keep the string encoding for this patch, please say so explicitly in the PR notes with the reasoning, and at minimum stop emitting `worker-error` for it.

---

packages/worker/src/git-transport.ts
```diff
@@ -136,12 +148,42 @@ export async function synchronizeAssignedBranch(options: {
...
+      "cannot format an unpublished-worktree error without changes"
+    );
+  }
+  return [
+    "git_unpublished_worktree: committed_transport_succeeded",
+    `branch=${result.branch}`,
+    `head=${result.head}`,
+    `tracked=[${changes.tracked.join(", ")}]`,
+    `untracked=[${changes.untracked.join(", ")}]`,
+  ].join(" ");
```
hojinzs, line 197: **P2 — the diagnostic is unbounded**

Every tracked and untracked path is joined into one string with no cap. That string is persisted to `run.json`, `workspace.json` and `events.ndjson`, and served through the control-plane API and the dashboard. Combined with `--untracked-files=all` below (which expands an untracked directory into one entry per file), an agent that leaves behind an un-ignored build output or dependency tree produces a multi-megabyte `lastError` written on every reconciliation tick that re-saves the run.

#789 asks for the state to be "bounded and diagnosable". Suggest capping the lists — first N entries plus `(+M more)` — and keeping the total under a few KB. The branch/head prefix is what an operator needs to recover; the exhaustive file list is not.

Related, minor: `--untracked-files=normal` would collapse untracked directories to a single entry and keep both the scan and the message cheap. `-uall` only buys per-file detail that the cap will truncate anyway.

---

packages/orchestrator/src/service.ts
```diff
@@ -5791,21 +5795,31 @@ export class OrchestratorService {
...
+
+  private async hasUnpublishedGitTransportFailure(
+    projectId: string,
+    issueId: string
+  ): Promise<boolean> {
+    return (
+      (await this.unpublishedGitWorkReason(projectId, issueId)) ===
+      "git_transport_failed"
     );
   }
```
hojinzs: **P3 — `hasUnpublishedGitTransportFailure` is now dead production code**

Its only remaining caller in the repository is `service.test.ts:10295`; the production call site moved to `unpublishedGitWorkReason` at line 5680. A private method kept alive purely so a test can reach into it via a cast is worse than either alternative:

- delete it and have that test assert through the real path (the workspace is retained), or
- reuse it as the guard at service.ts:3550, which still checks only `isGitTransportFailure` — see the P1 comment above.

Option 2 is probably what you want anyway.

---

docs/architecture.md
```diff
@@ -103,7 +103,7 @@ the tracker adapter:
...
   repository-local persistence divergence.
 - Shared bare clone cache (`<config-dir>/repos/<owner>/<repo>.git`), heartbeat locks, direct-clone degradation, safe inventory/eviction, worktree populate, and conservative agent-branch collection (only refs fully reachable from `origin/*` and not linked to a worktree): `repository-cache.ts`, `git.ts`; operator diagnostics and cleanup: `packages/cli/src/commands/cache.ts`
-- Issue workspace records remain in orchestrator state, while population, quarantine, terminal cleanup, and worktree removal operate on `<workspace.root>/<issue-key>` in repo-embedded and standalone modes.
+- Issue workspace records remain in orchestrator state, while population, quarantine, terminal cleanup, and worktree removal operate on `<workspace.root>/<issue-key>` in repo-embedded and standalone modes. A successful assigned-branch push is not a complete-publication claim when the worktree still has tracked or untracked changes: the worker records `git_unpublished_worktree` with the pushed branch/commit and file lists, and terminal and startup cleanup retain that workspace until the work is recovered.
```
hojinzs: **P2 — label the retention as an intentional upstream divergence**

Agreeing with Codex at #discussion_r3914421988, and I checked the spec text: §8.6 *Startup Terminal Workspace Cleanup* says "For each returned issue identifier, remove the corresponding workspace directory" — unconditionally. This change establishes an exception to that, and AGENTS.md requires divergences to be called out explicitly and not "hidden behind ambiguous wording". The bullet directly above this line does exactly that for the persistence divergence; this sentence reads as plain description.

Add the same framing here — that retaining terminal workspaces with unpublished Git work is a deliberate repository-level divergence from §8.6, and why (destroying unrecoverable agent work is worse than a stale workspace). Note the pre-existing `git_transport_failed` retention has the same unlabeled status, so one sentence can cover both classifications.

---

e2e/stub-worker.ts
```diff
@@ -163,7 +168,12 @@ function emitOrchestratorEvent(event: string): void {
             : status === "completed"
               ? "succeeded"
               : null,
-      lastError: status === "failed" ? "Stub worker simulated failure" : null,
+      lastError:
+        status === "failed"
+          ? "Stub worker simulated failure"
+          : SCENARIO === "dirty-unpublished-worktree" && status === "completed"
+            ? "git_unpublished_worktree: committed_transport_succeeded branch=feat/assigned head=stubbed tracked=[ M tracked.txt] untracked=[untracked/notes.txt]"
+            : null,
```
hojinzs: **P3 — TC-23 stubs out the half of the change it looks like it is testing**

The marker string is hard-coded into the heartbeat, so in the container `synchronizeAssignedBranch`, `readUnpublishedWorktreeChanges` and the new `applyGitTransportAttempt` branch never execute. `prepareDirtyUnpublishedWorktree` genuinely dirties the worktree, but nothing reads it — the scenario would pass identically if detection were broken or deleted.

That is a legitimate scope for a black-box test of the orchestrator's *retention* half, and the scenario doc is honest about it ("Let the stub report the host transport contract"). But the PR body's evidence line — "confirmed successful committed transport, retained tracked/untracked edits" — reads as if transport ran. Please soften that to say the transport outcome is simulated, and note in TC-23 that worker-side detection is covered by unit tests only.

Also, the string is duplicated verbatim in three places now (here, `git-transport.test.ts`, `service.test.ts`). Exporting `formatUnpublishedWorktreeError` and building it from a fixture would keep the E2E honest if the format ever changes.

---

e2e/run-e2e.sh
```diff
@@ -127,6 +127,16 @@ elif [ "$SCENARIO" = "required-label-missing" ]; then
...
+elif [ "$SCENARIO" = "dirty-unpublished-worktree" ]; then
+  DIRTY_FIXTURE_B64=$(base64 < e2e/fixtures/happy-path.json | tr -d '\n')
+  "${COMPOSE[@]}" exec -T -e "FIXTURE_B64=$DIRTY_FIXTURE_B64" symphony-e2e \
+    node --input-type=module -e '
+      import { writeFileSync } from "node:fs";
+      writeFileSync(
+        "/e2e/fixtures/issues.json",
+        Buffer.from(process.env.FIXTURE_B64, "base64")
+      );
+    '
```
hojinzs: **nit — this branch appears to duplicate the `else` default, the long way round**

It base64-encodes `happy-path.json`, pipes it through `compose exec -T … node --input-type=module` and writes it to `/e2e/fixtures/issues.json`. The `else` branch a few lines down already does `cp e2e/fixtures/happy-path.json e2e/fixtures/issues.json`, which every other scenario uses and which lands at the same bind-mounted path.

If there is a reason (container-side file ownership after a prior run, ordering against compose startup, …) it deserves a comment, because it reads as unexplained complexity next to fourteen one-line `cp` siblings. If there isn't, delete the branch and let it fall through to the default.

---

packages/worker/src/git-transport.ts
```diff
@@ -244,6 +286,38 @@ async function readCurrentBranch(
...
+    const path = record.slice(3);
+    if (status === "??") {
+      untracked.push(path);
+      continue;
+    }
+    tracked.push(`${status} ${path}`);
+    if (status.includes("R") || status.includes("C")) {
+      index += 1;
+    }
+  }
```
hojinzs, line 338: **nit — rename/copy entries drop the original path, and the parser's only tricky branch is untested**

I checked the `-z` layout against real `git status` output and the `index += 1` skip is **correct**:

```
' M keep.txt\0' 'R  new.txt\0' 'old.txt\0' '?? untracked.txt\0'
```

So a rename emits the entry plus a second field holding the original path. The skip consumes it properly. Two smaller things:

1. `tracked.push(\`${status} ${path}\`)` records only `R  new.txt`; `old.txt` is discarded. For a diagnostic whose whole purpose is helping an operator find unpublished work, `R  old.txt -> new.txt` is more useful.
2. No test covers a rename. Every existing case is ` M` / `??`, so if the skip were wrong nothing here would catch it — the original path would silently be reported as a separate bogus tracked entry. Worth one `git mv` case in `git-transport.test.ts` alongside the new tests.

---

hojinzs (changes requested):
## Review round 1 / 6 — Request changes

Verified against #789's acceptance criteria, with local test runs and one reproduction. The shape of the change is right and proportionate — no over-engineering in the core logic — but one acceptance criterion is not actually met, and I could reproduce the failure.

### Requirement coverage vs #789

| --- | --- |
| A successful push does not imply a dirty worktree is fully published | ✅ |
| Dirty work represented **distinctly** from transport failure and agent failure | ⚠️ prefix only; shares the `lastError` failure channel (P2) |
| Terminal cleanup **cannot** delete a workspace whose latest run records dirty work | ❌ reproduced a deletion (P1) |
| Retry/recovery stays bounded | ✅ run stays `succeeded`, no new retry arms |
| Existing clean-worktree transport unchanged | ✅ |
| Tracked + untracked covered by worker/orchestrator tests and a Docker scenario | ⚠️ Docker scenario stubs the detection half (P3) |
| Living docs + `@gh-symphony/cli` changeset | ⚠️ present, but the §8.6 divergence is unlabeled (P2) |

### Smoke test

`node_modules` and `dist` were absent, so `pnpm install && pnpm build` first (unrelated to this PR — three worker suites fail to collect without built workspace deps).

- `pnpm --filter @gh-symphony/worker test` → **27 files, 223 tests passed**
- `pnpm --filter @gh-symphony/orchestrator test` → **14 files, 387 tests passed**
- Both CI checks green on `15db41c`.

No preview environment exists for this repo, so verification was local plus a targeted reproduction; the Docker E2E scenario itself is reviewed as code (see the TC-23 comment).

**Reproduction of the P1.** Patching the new `it.each` retention test so the workspace record carries `lastError: null` while the run still carries the marker deletes the retained files, for both the transport-failure and the new dirty-worktree case:

```
Error: ENOENT: no such file or directory, open
  '.../acme_platform_1-0e0ed77a1c2f4ef5/unpublished-commit.txt'
```

This is reachable on an orchestrator restart, and on the terminal-release path inside `run.status === "retrying"` whose guard was not updated for the new classification. Details and a caution about the obvious-but-wrong fix are in the inline comment on `service.ts`.

### Findings

- **P1** `service.ts:5805` — workspace record short-circuits the run scan; reproduced workspace deletion with unpublished work present.
- **P2** `git-transport.ts:43` — the marker rides the `lastError` failure channel, so a succeeded run emits a `worker-error` event and shows an error in the snapshot and dashboard.
- **P2** `git-transport.ts:178` — the diagnostic string is unbounded and is persisted on every run save.
- **P2** `docs/architecture.md:106` — retention diverges from spec §8.6 and AGENTS.md requires that to be labeled explicitly.
- **P3** `service.ts:5816` — `hasUnpublishedGitTransportFailure` is now dead outside tests.
- **P3** `stub-worker.ts:171` — TC-23 hard-codes the marker, so worker-side detection never runs in the container; the PR body's evidence line overstates it.
- **nit** `run-e2e.sh:130` — the new fixture branch duplicates the `else` default via a much more complex path.
- **nit** `git-transport.ts:302` — rename entries drop the original path; the `-z` skip is correct (I verified it against real `git status` output) but untested.

### Not flagged

- `docs/configuration.md` is 39/39 pure Prettier table re-alignment with no content change. I confirmed `main`'s copy genuinely fails `prettier --check` and this one passes, so it's a real repair rather than churn — just noting it inflates the diff and would read better as its own commit.
- The two unrelated formatting fixes at the bottom of `service.test.ts` are the same situation.
- No over-engineering found in the core change: no speculative config surface, no new abstractions beyond what #789 asks for.

P1 and the two P2s need a response before approval — either a fix, or a reasoned decision not to act, recorded on the thread.

---

packages/orchestrator/src/service.ts
```diff
@@ -5791,21 +5795,31 @@ export class OrchestratorService {
...
     projectId: string,
     issueId: string
-  ): Promise<boolean> {
+  ): Promise<"git_transport_failed" | "git_unpublished_worktree" | null> {
     const workspace = (await this.store.loadIssueWorkspaces(projectId)).find(
       (record) => record.issueSubjectId === issueId
     );
     if (workspace) {
-      return isGitTransportFailureError(workspace.lastError);
+      return unpublishedGitWorkReason(workspace.lastError);
```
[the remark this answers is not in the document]
moncher-dev: Resolved: cleanup now checks the workspace record, then only the newest run when absent; regression coverage seeds a null workspace marker with the latest run record.

docs/architecture.md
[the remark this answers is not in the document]
moncher-dev: Resolved: architecture documentation explicitly labels terminal retention as an intentional repository-level divergence from Symphony §8.6, justified by preserving unrecoverable unpublished work.

packages/orchestrator/src/service.ts
moncher-dev: Resolved: cleanup now checks the workspace record, then only the newest run when absent; regression coverage seeds a null workspace marker with the latest run record.

packages/worker/src/git-transport.ts
moncher-dev: Resolved: successful dirty transport now uses `unpublishedWorktree` across channel, run, workspace, and status surfaces; `lastError` remains null and no worker-error is emitted.

packages/worker/src/git-transport.ts
moncher-dev: Resolved: diagnostics use normal untracked status output, cap each list at eight entries, truncate display paths, and record omitted counts.

packages/orchestrator/src/service.ts
moncher-dev: Resolved: removed the dead helper; the retry-release guard now uses the real unpublished-work predicate with regression coverage.

docs/architecture.md
moncher-dev: Resolved: architecture documentation explicitly labels terminal retention as an intentional repository-level divergence from Symphony §8.6, justified by preserving unrecoverable unpublished work.

e2e/stub-worker.ts
moncher-dev: Resolved: TC-23 now names the simulated transport scope; it verifies orchestrator retention using the structured field, while worker Git detection remains unit-covered.

e2e/run-e2e.sh
moncher-dev: Resolved: removed the redundant dirty-scenario fixture branch so it falls through to the normal fixture setup.

packages/worker/src/git-transport.ts
moncher-dev: Resolved: rename diagnostics report source and destination, with an integration-style `git mv` test.

packages/worker/src/git-transport.ts
```diff
@@ -136,12 +159,44 @@ export async function synchronizeAssignedBranch(options: {
...
+      pushed: true,
+      head,
+      unpublishedWorktreeChanges,
+    };
   } finally {
     await rm(transportDirectory, { recursive: true, force: true });
   }
 }
 
+export function formatUnpublishedWorktreeError(
```
hojinzs, line 177: **P3 — `formatUnpublishedWorktreeError` is dead on arrival**

Zero callers anywhere in the repo:

```
$ grep -rn "formatUnpublishedWorktreeError" --include=*.ts --include=*.sh --include=*.md . | grep -v node_modules
./packages/worker/src/git-transport.ts:177:export function formatUnpublishedWorktreeError(
```

Not production, not tests, not `stub-worker.ts`. It looks like it survived the move to the structured `unpublishedWorktree` field — it still formats the old `git_unpublished_worktree: …` string, which nothing consumes any more, and it `throw`s on a branch (`!changes`) that its own type makes unreachable for every real caller.

This is the same shape as the `hasUnpublishedGitTransportFailure` P3 from round 1, which you correctly deleted. Suggest deleting this one too; if it was added for the TC-23 dedup point I raised, the structured object the stub now emits already solved that problem better.

Non-blocking.

---

packages/core/src/observability/snapshot-builder.ts
```diff
@@ -107,6 +107,7 @@ export function buildProjectSnapshot(
       lastEventAt: run.lastEventAt ?? null,
       executionPhase: run.executionPhase ?? null,
       runPhase: run.runPhase ?? null,
+      unpublishedWorktree: run.unpublishedWorktree ?? null,
```
hojinzs, line 110: **P3 — the new field bypasses the `/api/v1/state` redaction policy that the old encoding was subject to**

`ProjectStatusSnapshot` is what `/api/v1/state` serves, and that endpoint (unlike `/api/v1/<issue>`) runs the payload through `redactDashboardState`. Its `STATE_REDACTED_KEYS` (`packages/dashboard/src/server.ts:38`) covers `lastError`, `workspacePath`, `workspaceKey`, `workingDirectory` — a deliberate "no paths or diagnostics in the project summary" policy. `unpublishedWorktree` was not added to it, and the redactor keys off the property name, so the whole object passes through. Reproduced against the built dashboard:

```json
{
  "lastError": "[REDACTED]",
  "workspacePath": "[REDACTED]",
  "unpublishedWorktree": {
    "branch": "feat/assigned",
    "head": "deadbeef",
    "tracked": [" M src/secret-internal-path.ts"],
    "untracked": ["scratch/creds.env"],
    "trackedOmitted": 0, "untrackedOmitted": 0
  }
}
```

In round 1's encoding this same information rode `lastError` and was redacted; moving it to a dedicated field silently opted it out. Same applies to the workspaces projection on line 120.

The endpoint is bearer-authenticated and `/api/v1/<issue>` already returns everything unredacted to the same token holder, so this is a policy-consistency gap rather than a disclosure hole — hence P3, not P2. But it should be a deliberate decision rather than an oversight. Either add `unpublishedWorktree` to `STATE_REDACTED_KEYS` (the summary endpoint keeps a boolean-ish signal, operators use the per-issue endpoint for detail), or redact just `tracked`/`untracked` and keep `branch`/`head` visible — the latter is probably the more useful shape, since branch and head are what an operator needs to go find the work.

Non-blocking.

---

hojinzs (approved):
## Review round 2 / 6 — Approve

All ten round-1 threads are genuinely fixed, not just replied to. I re-read each change and re-ran the suites rather than taking the resolution notes at face value; two new P3s are noted inline and neither blocks.

### Round-1 findings, verified

| Finding | Verified fix |
| --- | --- |
| **P1** workspace record short-circuits the run scan (`service.ts`) | ✅ `unpublishedGitWorkReason` now falls through to the run scan when the workspace carries no marker, and resolves the issue's **latest** run via `buildLatestRunMapByIssueId` — the imprecise `.some()` over all history is gone, which was the caution attached to the fix |
| **P2** marker rides the `lastError` failure channel | ✅ dedicated `UnpublishedWorktree` type threaded through channel → run → workspace → snapshot → dashboard; `lastError` stays null, no `worker-error` is emitted for a succeeded run |
| **P2** unbounded diagnostic | ✅ `--untracked-files=normal`, 8-entry cap per list, 160-char path truncation, `trackedOmitted`/`untrackedOmitted` counts |
| **P2** §8.6 divergence unlabeled | ✅ `docs/architecture.md` now names it an intentional repository-level divergence and covers both classifications |
| **P3** dead `hasUnpublishedGitTransportFailure` | ✅ deleted, and the retry-release guard at `service.ts:3550` now uses `hasUnpublishedGitWork` — the option-2 fix |
| **P3** TC-23 overstated in the PR body | ✅ scenario doc and PR evidence both say the transport outcome is simulated and detection is unit-covered |
| **nit** redundant e2e fixture branch | ✅ removed |
| **nit** rename drops the original path, untested | ✅ `R  original.txt -> renamed.txt`, with a real `git mv` integration test |

### Reproduction of the P1 is closed

The new `it.each` seeds exactly the shape I could break in round 1 — workspace record with `lastError: null` and no marker, latest run carrying the evidence — parameterised over both the transport-failure and dirty-worktree classifications, and asserts the sentinel file survives. The second `it.each` covers the `run.status === "retrying"` terminal-release path that the old `isGitTransportFailure` guard missed. Both fail closed now.

### Smoke test

`pnpm install && pnpm build` first (`node_modules`/`dist` absent — unrelated to this PR).

- `pnpm --filter @gh-symphony/worker test` → **27 files, 225 passed** (up from 223; the new rename and bounding cases)
- `pnpm --filter @gh-symphony/orchestrator test` → **14 files, 387 passed**
- `pnpm lint`, `pnpm typecheck`, `pnpm build` → clean
- `prettier --check` on every file this PR touches → clean. (`pnpm format` repo-wide fails, but it fails on `main` too — 91 files there, 87 here, so this PR strictly improves it. Pre-existing, not yours.)
- Both CI checks green on `cb89e0c`.

No preview environment exists for this repo, so verification was local plus targeted reproductions.

Two things I checked specifically for regressions and found clean: the transport scratch directory is `mkdtemp` under `tmpdir()`, not inside the worktree, so the post-push `git status` can't false-positive on it; and `recordGitTransportAttempt` is followed by `emitOrchestratorHeartbeat()` at both call sites, so the structured field actually reaches the orchestrator before the worker exits.

### Requirement coverage vs #789

Every acceptance criterion now holds, including the two that didn't in round 1 — terminal cleanup can no longer delete a workspace whose latest run records dirty work, and the outcome is represented distinctly rather than as a prefix on the failure channel. Changeset present and correctly scoped to `@gh-symphony/cli` as a patch.

### New findings (non-blocking)

- **P3** `git-transport.ts:177` — `formatUnpublishedWorktreeError` is exported with zero callers repo-wide; leftover from the string encoding.
- **P3** `snapshot-builder.ts:110` — `unpublishedWorktree` isn't in the dashboard's `STATE_REDACTED_KEYS`, so paths that policy redacts beside it now pass through `/api/v1/state`. Reproduced; worth a deliberate decision either way.

### Scope

No over-engineering. The bounding constants and the structured type are exactly what round 1 asked for, nothing speculative was added on top, and no new configuration surface appeared. Approving — the two P3s are cleanup you can fold in here or take separately.

---

PR #802, opened 2026-09-03
fix(worker): exit on Codex startup failure

### LINKED ISSUE #779: fix(worker): exit non-zero on Codex startup failure
## Problem

The Codex worker route catches startup exceptions in `startAssignedRun()` (for example, missing `PROJECT_ID` / `CODEX_PROJECT_ID`), records `runtimeState.status = "failed"`, and writes a diagnostic, but then returns without setting a non-zero exit code. The Node process exits 0.

This was reproduced while adding issue #700 Docker lifecycle coverage: a Codex worker missing `CODEX_PROJECT_ID` printed `startup failed: PROJECT_ID or CODEX_PROJECT_ID is required.` and exited 0.

## Expected

All controlled Codex startup failures emit the terminal failed heartbeat and exit non-zero, matching `exitWorkerStartupFailure()` and the non-Codex lifecycle.

## Suggested coverage

Spawn the built worker through the Codex route with an invalid/missing required startup setting and assert final heartbeat state plus process exit code.

moncher-dev:
## TL;DR

Controlled Codex startup exceptions follow the shared terminal-failure lifecycle: the worker writes a startup-context diagnostic, emits a failed heartbeat, and exits with status `1`.

## Change-point diagram

`startAssignedRun()` configuration error
→ `exitWorkerStartupFailure("startup failed: …")`
→ terminal failed heartbeat + persisted diagnostic
→ process exit `1`

## Start here

- `packages/worker/src/index.ts` preserves the generic catch's startup context while delegating to the shared terminal-failure helper.
- `test/e2e/claude/claude-docker.spec.ts` runs the built Codex worker without either project-ID setting and asserts its diagnostic, terminal heartbeat, and exit status.

## Risks & rollback

- Risk: startup diagnostics and the terminal heartbeat's `lastError` now consistently include the `startup failed:` context.
- Rollback: revert commits `1f5e322` and `5de304e`; no persisted data or migration is involved.

## Changed files

- `packages/worker/src/index.ts` — shared terminal startup-failure exit path with preserved startup context.
- `test/e2e/claude/claude-docker.spec.ts` — built-worker regression coverage and formatting correction.
- `AGENT_TEST.md` — Docker black-box scenario documentation.
- `.changeset/fix-codex-startup-exit.md` — patch release note for `@gh-symphony/cli`.

## Evidence

- `pnpm lint`
- `pnpm test`
- `pnpm typecheck`
- `pnpm build`
- `pnpm e2e:claude` — Docker black-box suite, including the built-worker startup scenario.
- Focused RED/GREEN: `pnpm build && pnpm exec vitest run --config test/e2e/claude/vitest.config.ts -t 'fails a built Codex worker startup with a terminal failed heartbeat'`.

- [ ] Confirm a worker with a valid Codex configuration still starts normally.

## Follow-up

- #804 tracks running the built-worker startup regression in CI; it is deliberately out of scope for this targeted runtime fix.


packages/worker/src/index.ts
```diff
@@ -657,14 +657,7 @@ async function startAssignedRun() {
...
       error instanceof Error ? error.message : "Unknown worker startup error";
-    runtimeState.status = "failed";
-    runtimeState.runPhase = "failed";
-
-    if (runtimeState.run) {
-      runtimeState.run.lastError = message;
-    }
-    process.stderr.write(`[worker] startup failed: ${message}\n`);
-    await persistSessionTokenUsageArtifact(launcherEnv);
+    await exitWorkerStartupFailure(message);
```
hojinzs: **P3 (observability)** — the `startup failed:` log prefix is lost here.

`exitWorkerStartupFailure()` writes `[worker] ${message}`, so this catch now emits e.g. `[worker] PROJECT_ID or CODEX_PROJECT_ID is required.` where it previously emitted `[worker] startup failed: PROJECT_ID or CODEX_PROJECT_ID is required.`

Every other `exitWorkerStartupFailure()` call site passes a self-describing message (`Issue identity preflight failed: …`, `Claude runtime preflight failed before agent launch.`, `SYMPHONY_RENDERED_PROMPT not set; …`). This one is the generic catch-all, so the raw error text is the only thing an operator sees — there's no longer anything in the line saying it happened during startup.

I grepped the repo for `startup failed` and found no code, test, or docs dependency on the old string, so nothing breaks. Still, suggest keeping the context:

```suggestion
    await exitWorkerStartupFailure(`startup failed: ${message}`);
```

Non-blocking.

---

test/e2e/claude/claude-docker.spec.ts
```diff
@@ -426,6 +426,78 @@ lines.on("line", (line) => {
     expect(result.stderr).toContain("origin/feat/assigned is not an ancestor");
   });
 
+  it("fails a built Codex worker startup with a terminal failed heartbeat", async () => {
```
hojinzs, line 429: **P2 (follow-up, not blocking this PR)** — this regression test never runs in CI.

I verified the test itself is genuinely load-bearing: I reverted `packages/worker/src/index.ts` to `2601e0d` and re-ran it, and it fails with `expected +0 to be 1` — exactly the symptom #779 describes. With the fix applied it passes. Good test.

The problem is reach:

- Root `vitest.config.ts` uses `include: ["**/*.test.ts"]`, so `*.spec.ts` here is excluded from `pnpm test`.
- `.github/workflows/ci.yml` runs `vitest run --coverage`, `pnpm e2e:compose-project-test`, and `pnpm --filter @gh-symphony/cli test:package` — it never runs `pnpm e2e:claude`.
- `packages/worker/src/index.ts` is at 0% line coverage (per the coverage bot on this PR).

Net: #779 could regress tomorrow and CI stays green. This follows the documented convention in `CLAUDE.md`/`AGENT_TEST.md` (black-box behavior is verified in the Docker E2E environment), and carving a unit seam out of a 2336-line module for a one-line fix would be scope creep I don't want in this PR — so I'm not asking for changes here. Worth a separate issue to get `pnpm e2e:claude` (or at least the built-worker `runWorkerProcess` scenarios, which need only `node` + `git`, not Docker — they ran fine for me outside the container) onto the CI job.

---

test/e2e/claude/claude-docker.spec.ts
```diff
@@ -426,6 +426,78 @@ lines.on("line", (line) => {
...
+`,
+      "utf8"
+    );
+    const remote = join(root, "remote.git");
+    await runGit(root, "init", "--bare", remote);
+    await runGit(workspace, "init", "-b", "feat/assigned");
+    await runGit(workspace, "config", "user.name", "Symphony E2E");
+    await runGit(workspace, "config", "user.email", "e2e@example.com");
+    await runGit(workspace, "add", "WORKFLOW.md");
+    await runGit(workspace, "commit", "-m", "test: seed Codex startup workspace");
```
hojinzs: **nit (Prettier)** — this line exceeds the print width and is the one new formatting violation this PR introduces.

`npx prettier` reformats it to:

```suggestion
    await runGit(
      workspace,
      "commit",
      "-m",
      "test: seed Codex startup workspace"
    );
```

`pnpm format` already fails repo-wide on 93 files (including this spec at `2601e0d`), and CI doesn't run it, so this isn't a new breakage — just keeping the added lines clean.

---

hojinzs (approved):
## Approve

Tight, correctly-scoped fix. One line of production code, and it does exactly what #779 asked for. No P1s.

### Requirements vs. implementation (#779)

| Requirement | Status |
| --- | --- |
| Controlled Codex startup failures emit the terminal failed heartbeat | ✅ `exitWorkerStartupFailure()` calls `stopOrchestratorHeartbeatTimer()` + `emitOrchestratorHeartbeat()`, which the old inline block did not |
| Exit non-zero | ✅ `process.exit(1)` |
| Match `exitWorkerStartupFailure()` / the non-Codex lifecycle | ✅ by literally routing through that helper rather than reimplementing it |
| Suggested coverage: spawn the built worker via the Codex route with a missing required setting, assert final heartbeat + exit code | ✅ new scenario at `claude-docker.spec.ts:429` asserts both |
| Changeset for a behavior-changing PR (`CLAUDE.md`) | ✅ `.changeset/fix-codex-startup-exit.md`, patch, `@gh-symphony/cli` |
| `AGENT_TEST.md` row | ✅ |

### Verification I ran locally

- `pnpm build` — clean
- `pnpm lint` — clean
- `pnpm typecheck` — clean
- `pnpm test` — exit 0, no failing suites
- **Smoke test of the changed behavior**: ran the new scenario directly against the built worker (`npx vitest run --config test/e2e/claude/vitest.config.ts -t "fails a built Codex worker startup"`) — passes.
- **Confirmed the test is load-bearing**: restored `packages/worker/src/index.ts` from `2601e0d`, rebuilt the worker, re-ran → `AssertionError: expected +0 to be 1`. That is precisely the exit-0 symptom #779 reported, so this is a real regression test and not a tautology.

No preview environment on this PR, so the smoke test was done against the locally built worker binary.

None found. The diff removes 7 lines and adds 1 in production code; everything else is test, docs, and the required changeset. No speculative abstraction, no drive-by refactor, no behavior added beyond what the issue specified.

### Risks I checked and cleared

- **Does the outer `catch` now swallow mid-run errors and exit(1) prematurely?** No. `runNonCodexRuntimeAdapterLifecycle()` has its own `try/catch/finally` and schedules its own exit, so it does not propagate into this catch on the normal path. Where it could propagate (a throw inside its own `finally`), the new behavior — failed heartbeat + exit 1 — is strictly better than the old silent exit 0.
- **Orphaned Codex child on `process.exit(1)`?** Not reachable. `launchCodexWithValidatedPolicy()` validates policy *before* calling `launch()`, and the only code between the spawn and the end of the `try` is non-throwing field assignments and listener registrations.
- **Does `PROJECT_ID: undefined` in the spawn env actually unset it?** Yes — Node's `normalizeSpawnArguments` skips `undefined` values, and the test passing pre-fix-revert with the expected error string confirms the intended code path is reached.

### Findings

Three inline comments, all non-blocking:

- **P3** `packages/worker/src/index.ts:660` — the `startup failed:` log prefix is dropped; suggested restoring it. Nothing in the repo greps for the old string.
- **P2 (follow-up issue, not this PR)** `claude-docker.spec.ts:429` — the new regression test is a `*.spec.ts`, excluded from root `pnpm test`, and `ci.yml` never runs `pnpm e2e:claude`. The coverage is real but ungated. Worth a separate issue.
- **nit** `claude-docker.spec.ts:456` — one new Prettier violation.

Approving rather than round-tripping: none of the three change the correctness of the fix, and the P2 belongs in its own issue rather than expanding a one-line patch.

---

packages/worker/src/index.ts
moncher-dev: Implemented in the follow-up commit: the generic startup catch now passes `startup failed: ${message}` to the shared terminal-failure helper. The built-worker regression asserts both the restored `[worker] startup failed: …` diagnostic and the terminal heartbeat's `lastError`.

test/e2e/claude/claude-docker.spec.ts
moncher-dev: Tracked separately as #804 to keep CI reach of the black-box regression out of #779's narrowly scoped runtime fix. The follow-up captures the excluded `*.spec.ts` path and the CI workflow gap.

test/e2e/claude/claude-docker.spec.ts
moncher-dev: Applied the suggested Prettier wrapping to the new `runGit(... commit ...)` call and verified the changed E2E spec with Prettier.

hojinzs (approved):
## Approve (round 2 — final)

All three findings from my `1f5e322` review are resolved. No new findings at `5de304e`. Nothing further from me; this is my final round.

### Prior findings vs. follow-up commit

| Finding | Disposition |
| --- | --- |
| **P3** `index.ts:660` — `startup failed:` context dropped | ✅ Fixed exactly as suggested: `exitWorkerStartupFailure(\`startup failed: ${message}\`)`. Test assertions upgraded in the same commit to pin both the `[worker] startup failed: …` diagnostic and the heartbeat's `lastError`. |
| **P2** regression test ungated in CI | ✅ Correctly deferred with a reasoned rationale, tracked as #804. I read #804 — it captures both halves of the gap (the `*.spec.ts` exclusion from root `pnpm test` and `ci.yml` never invoking `pnpm e2e:claude`) and scopes the fix away from #779. Right call to keep it out of a one-line runtime patch. |
| **nit** Prettier wrapping at `claude-docker.spec.ts:456` | ✅ Applied verbatim. |

### Verification I ran on `5de304e`

- `pnpm build`, `pnpm lint`, `pnpm typecheck` — all clean.
- `pnpm test` — 43 files / 616 tests, no failures.
- **Smoke test**: `npx vitest run --config test/e2e/claude/vitest.config.ts -t "fails a built Codex worker startup"` against the freshly built worker — passes. No preview environment on this PR, so this ran against the local built worker binary.
- **The new assertion is load-bearing, not decorative**: I reverted line 660 back to `exitWorkerStartupFailure(message)`, rebuilt the worker and CLI, re-ran → `AssertionError: expected '[worker] PROJECT_ID or CODEX_PROJECT_…' to contain '[worker] startup failed: PROJECT_ID o…'`. So the follow-up commit's own change is now regression-guarded, not just the original exit-code fix.
- CI on this head: **Test ✅, Container Smoke ✅**.

### Risk I checked on the new commit

Prefixing the message also changes the terminal heartbeat's `lastError` payload, so I checked for downstream string coupling. The only pattern match on `lastError` anywhere in the repo is `latestRun.lastError?.includes(MAX_FAILURE_RETRIES_EXCEEDED_REASON)` (`explain.ts:505`, `service.ts:5892`) — a substring check against an orchestrator-set reason, never a worker startup message, and a prefix would preserve `includes()` semantics regardless. Nothing else parses, matches, or classifies on that field. No breakage.

### Pre-existing, not yours

`npx prettier --check packages/worker/src/index.ts` still warns, but the sole violation is at `index.ts:1937` (`resolveRefreshFailureThreshold(env.SYMPHONY_REFRESH_FAILURE_THRESHOLD)` needing a wrap), which fails identically at base `2601e0d`. Your added lines are Prettier-clean. Not this PR's to fix.

### Scope

Still none. Final production diff is one line changed; the rest is test, docs, and the required changeset. The follow-up commit responded to review feedback without widening the patch — the P2 went to an issue rather than into the diff, which is the right instinct.

Requirements from #779 are fully met (terminal failed heartbeat, non-zero exit, routed through `exitWorkerStartupFailure()` rather than reimplemented, built-worker coverage asserting heartbeat + exit code). Ship it.

---

PR #803, opened 2026-09-03
feat(orchestrator): add fair retry reservation ordering

### LINKED ISSUE #703: feat(orchestrator): add fairness for due retry reservations
## Context

Retry-fire capacity requeues now guarantee liveness by excluding due reservations from slot accounting. They deliberately keep `countFailure: false` and `advanceAttempt: false` so temporary capacity contention does not consume the failure budget or add exponential backoff.

## Follow-up

Define and implement a fair ordering or aging policy for multiple due retry reservations under sustained concurrency saturation. The current sequential reconciliation order can repeatedly prefer the same reservation, so liveness does not guarantee per-retry fairness.

Origin: #653.

moncher-dev:
## TL;DR

Due retry reservations now receive deterministic, directly tested ordering under saturation, while capacity postponement is visible as a structured event without overwriting the original failure reason.

## Change-point diagram

```text
active runs → non-due runs first → due reservations (oldest due time, stable ID ties)
                                      │
                                      └─ capacity unavailable
                                         ├─ retain reservation due time and failure error
                                         └─ emit retry-postponed with capacity reason
```

## Start here

- `packages/orchestrator/src/service.ts`: reconciliation ordering and capacity-postponement event emission.
- `packages/orchestrator/src/service.test.ts`: direct comparator and retained-reservation observability coverage.
- `packages/core/src/observability/structured-events.ts`: typed `retry-postponed` event contract.

## Changes

- Exported and table-tested the pure reconciliation sort so non-due runs, due-time aging, and deterministic tie-breaks are covered directly.
- Preserved the original retry failure while emitting `retry-postponed` for capacity contention.
- Removed a fragile private-method spy test and unrelated fixture formatting changes.

## Evidence

- `pnpm lint`
- `pnpm test`
- `pnpm typecheck`
- `pnpm build`
- `./e2e/run-e2e.sh happy 60`

## Risks & rollback

The new event adds observability data only; retry scheduling and failure-budget semantics remain unchanged. Roll back commit `473fc13` to remove the event and direct comparator exposure.

## Changed files

- `packages/core/src/observability/structured-events.ts`
- `packages/core/src/observability/event-formatter.ts`
- `packages/core/src/observability/event-formatter.test.ts`
- `docs/architecture.md`

## Issues — Closed #703

Closes #703

- [ ] Review a `retry-postponed` event from a capacity-saturated deployment and confirm its capacity reason is useful to operators.
- [ ] Confirm normal active work still receives priority over due retry reservations under production load.


packages/orchestrator/src/service.ts
```diff
@@ -5214,19 +5248,23 @@ export class OrchestratorService {
...
-            : scheduleRetryAt(now, attempt, retryPolicy)
+      : (retainedDueAt ??
+        (retryPolicy
+          ? scheduleRetryAt(now, attempt, retryPolicy)
           : new Date(
               now.getTime() +
                 (this.dependencies.retryBackoffMs ?? DEFAULT_RETRY_BACKOFF_MS)
             )
-        ).toISOString();
+        ).toISOString());
```
hojinzs: **P2 — the capacity requeue loses its poll-interval liveness fallback when `retryEntry.dueAt` is unavailable.**

Before this change, `advanceAttempt === false` *always* produced `now + pollInterval`. Now that branch is gone entirely: when `retainedDueAt` is `null`, a capacity-only requeue falls through to `scheduleRetryAt(now, attempt, retryPolicy)` — an exponential backoff keyed on the (deliberately un-advanced) attempt.

I verified this by binding `requeueRetryingRun` directly and calling it with `{ countFailure: false, advanceAttempt: false }`:

```
CASE1 (issue record absent, attempt 4)  dueAt = now + 80s   // 10s * 2^3, default policy
CASE2 (issue record present)            dueAt = 2026-03-07T23:00:00Z  // retained ✅
```

CASE2 is exactly the intended aging. CASE1 is the regression: a requeue that consumed neither failure budget nor backoff by design now silently *applies* backoff, and it grows with the attempt (up to `DEFAULT_MAX_DELAY_MS`). `loadProjectPollInterval` is no longer referenced from this path at all, so the guarantee `docs/architecture.md` used to state ("requeued for the polling interval") no longer holds in this branch.

Reachable whenever the reservation's record is missing or its `retryEntry.dueAt` is unparsable — e.g. the orchestration record was released/reset while the run record survived, or the lookup (`record.issueId === run.issueId || record.identifier === run.issueIdentifier`) misses after an identifier change. Uncommon, but it degrades exactly the liveness property #703 builds on.

Suggested shape — keep retention first, poll interval as the `advanceAttempt === false` floor, backoff only for the advancing path:

```ts
const dueAt = suppressed
  ? null
  : (retainedDueAt ??
      (options.advanceAttempt === false
        ? new Date(now.getTime() + (await this.loadProjectPollInterval(tenant)))
        : retryPolicy
          ? scheduleRetryAt(now, attempt, retryPolicy)
          : new Date(now.getTime() + (this.dependencies.retryBackoffMs ?? DEFAULT_RETRY_BACKOFF_MS))
      ).toISOString());
```

Two smaller things in the same block:

- **P3** — `loadRetryPolicy` on line 5250 is still awaited even when `retainedDueAt` short-circuits it. Since a capacity-blocked reservation is now *permanently* due, this workflow load runs for every blocked reservation on every poll for nothing.
- **P3** — `retryEntry.error` (and `run.lastError`) is still overwritten with `"no available orchestrator slots"` while `attempt` and `dueAt` are retained. Confirmed in CASE2 above: the entry ends up claiming it has been due since 23:00 *because of* a slot shortage, with the original `"worker failed"` gone. Pre-existing, but the PR's whole premise is that the reservation keeps its original identity — retaining `error` too (or recording the capacity message separately) would keep the snapshot honest.

---

packages/orchestrator/src/service.ts
```diff
@@ -277,6 +277,40 @@ function isDueRetryReservation(
...
+      }
+      if (left.dueAtMs !== right.dueAtMs) {
+        return (left.dueAtMs ?? 0) - (right.dueAtMs ?? 0);
+      }
+      const identifierOrder = left.run.issueIdentifier.localeCompare(
+        right.run.issueIdentifier
+      );
+      return identifierOrder !== 0
+        ? identifierOrder
+        : left.run.runId.localeCompare(right.run.runId);
```
hojinzs: **nit** — two small things in the comparator.

`now.getTime()` is recomputed on both sides of every comparison. Hoist it once above the `.sort()`; it also makes the due predicate read as a single shared boundary rather than two independent recomputations.

`localeCompare` (lines 304 and 309) is locale- and ICU-sensitive. For a tie-break whose whole purpose is deterministic ordering across processes, a plain relational compare is the safer primitive:

```ts
const byIdentifier =
  left.run.issueIdentifier < right.run.issueIdentifier ? -1
  : left.run.issueIdentifier > right.run.issueIdentifier ? 1
  : 0;
```

Neither changes behavior in practice; flagging because "stable issue-identity tie-breaking" is a documented property in `docs/architecture.md`, so it's worth being actually environment-independent.

For what it's worth, I walked the ordering itself and it holds up: the comparator is antisymmetric, non-due runs keep their input order via the captured `index`, and putting due reservations last is what lets a worker completing earlier in the same pass free the slot the retry then claims. `restartRun` sets `state: "running"` / `retryEntry: null`, so a restarted reservation immediately re-enters `hasRetryDispatchSlot`'s `claimed` count and a second contender in the same pass is correctly blocked — no over-dispatch. The main dispatch loop counts `retry_queued` records with a non-null `currentRunId` regardless of due-ness, so permanently-due reservations don't leak concurrency there either.

---

packages/orchestrator/src/service.test.ts
```diff
@@ -1060,6 +1160,14 @@ describe("OrchestratorService", () => {
...
       cumulativeRuntimeMs: 120_000,
+      nextRetryAt: now.toISOString(),
+    });
+    expect(result.issueRecords[0]).toMatchObject({
+      failureRetryCount: 0,
+      retryEntry: expect.objectContaining({
+        attempt: 2,
+        dueAt: now.toISOString(),
+      }),
     });
```
hojinzs, line 1164: **P2 — this test doesn't actually pin the "age" half of its name.**

The fixture sets the reservation's original `retryEntry.dueAt` to `now.toISOString()`, so `retainedDueAt` and `now` are the same value. The assertions `nextRetryAt: now.toISOString()` / `dueAt: now.toISOString()` therefore pass under *both* implementations:

- the intended one (retain the original due time — aging), and
- a hypothetical `dueAt = now` (reset on every requeue — **no** aging).

All it genuinely rules out is the pre-change `now + pollInterval`. Since retention is the actual fairness mechanism — ordering alone doesn't prevent a retry that just failed and re-backed-off from jumping an already-waiting one — the regression guard should discriminate. Making the queued due time strictly earlier than `now` fixes it:

```ts
const queuedDueAt = "2026-03-07T23:55:00.000Z"; // 5 min before `now`
// ...retryEntry: { attempt: run.attempt, dueAt: queuedDueAt, error: run.lastError },
expect(await store.loadRun(run.runId)).toMatchObject({ nextRetryAt: queuedDueAt });
expect(result.issueRecords[0]).toMatchObject({
  retryEntry: expect.objectContaining({ attempt: 2, dueAt: queuedDueAt }),
});
```

That also gives you a natural place to cover the fallback from my `service.ts` comment: same call with `issueRecords: []`.

---

packages/orchestrator/src/service.test.ts
```diff
@@ -188,6 +188,101 @@ describe("OrchestratorService", () => {
...
+        return { issueRecords, recovered: false };
+      }
+    );
+
+    await service.runOnce();
+
+    expect(reconciledRunIds).toEqual(["run-z-older", "run-a-newer"]);
+  });
+
   it("guards worker signals by owner and process identity", async () => {
```
hojinzs: **P2 — this proves the comparator's output order, not fairness.**

`reconcileRun` is stubbed out wholesale, so the assertion `["run-z-older", "run-a-newer"]` only says "the loop visits due reservations oldest-first". The property #703 asks for is emergent, one level up: *under sustained saturation, a retry that has been waiting wins the next free slot over one that just ran and re-backed-off.* Nothing here exercises that — the interaction between ordering, `hasRetryDispatchSlot`, and the retained `dueAt` is exactly the part the stub removes.

Concretely, the scenario worth a test: capacity 1, reservations A and B both due; A restarts and takes the slot; B is capacity-requeued; A fails and re-queues at `now + backoff`; the next pass must pick **B**. That's the regression that would silently come back if the retention in `requeueRetryingRun` were ever reverted while the sort stayed — and today no test would catch it.

The coverage report on this commit agrees the comparator's own branches aren't reached: `service.ts` lines 296-297 and 308 are listed uncovered, i.e. the non-due-vs-due swap and the `runId` tie-break are never taken. A direct unit test over `sortRunsForReconciliation` would cover those cheaply, and would be sturdier than the current approach — `vi.spyOn(service as never, "reconcileRun")` binds the test to a private method name, so an unrelated rename silently turns this into a no-op stub rather than a failure. Exporting the helper for test (it's a pure function) would remove both problems at once.

For the record, everything in the PR does pass locally on `48e80a9`: `pnpm build`, `pnpm lint`, `pnpm typecheck`, and `npx vitest run packages/orchestrator/src/service.test.ts` (211/211). The `pnpm format` failure I hit is pre-existing on `main` (90 files, none of them yours) and CI doesn't gate on it.

**nit** — the Prettier reflow of `writeWorkflowFixture` / `normalizeTrackerProviderFixture` at the bottom of this file is unrelated to fair retry ordering. Given format isn't CI-gated, it's diff noise rather than a fix; worth splitting out if you touch this file again.

---

hojinzs (changes requested):
## Review — round 1/6

**Verdict: request changes.** The design is right and the scope is right; two P2s are worth one round before approval.

### Requirements vs. #703

#703 asks for "a fair ordering or aging policy for multiple due retry reservations under sustained concurrency saturation", noting that liveness (excluding due reservations from slot accounting) does not by itself guarantee per-retry fairness. The PR delivers both halves:

| Acceptance point | Where | Status |
| --- | --- | --- |
| Fair ordering among due reservations | `sortRunsForReconciliation` — oldest `dueAt` first, deterministic tie-break | ✅ |
| Aging, so a waiting retry isn't jumped by one that just re-backed-off | `requeueRetryingRun` retains the original `dueAt` on capacity-only requeue | ✅ |
| Preserve `countFailure: false` / `advanceAttempt: false` | unchanged; `suppressed` is unreachable with `countFailure: false` | ✅ |
| Docs + changeset | `docs/architecture.md`, `.changeset/fair-retry-reservations.md` (`@gh-symphony/cli` patch) | ✅ |

No over-engineering and no unrequested scope: 4 files, one behavioral function plus one expression, the tie-break is determinism rather than embellishment, and the layering stays inside Coordination.

### Smoke test

Built and ran the affected package locally on `48e80a9`:

| Check | Result |
| --- | --- |
| `pnpm build` | pass |
| `pnpm lint` | pass |
| `pnpm typecheck` | pass |
| `npx vitest run packages/orchestrator/src/service.test.ts` | 211/211 pass |
| CI `Test` / `Container Smoke` | green |

`pnpm format` fails on 90 files, but that is pre-existing on `main` (none of them yours) and CI doesn't gate on it.

Since the ordering fix is scheduler-internal with no API or screen surface, I exercised the behavior directly instead of through the E2E harness — binding `requeueRetryingRun` and driving the capacity path:

```
issue record present  → dueAt = 2026-03-07T23:00:00Z   (original retained ✅ — aging works)
issue record absent   → dueAt = now + 80s              (backoff applied ✗ — see inline)
```

I also traced the concurrency accounting by hand, since permanently-due reservations are a new steady state: `restartRun` sets `state: "running"` / `retryEntry: null`, so a restarted reservation re-enters `hasRetryDispatchSlot`'s `claimed` count and a second contender in the same pass is correctly blocked; the main dispatch loop counts `retry_queued` records by non-null `currentRunId` regardless of due-ness. No over-dispatch beyond `maxConcurrentAgents` in either path. That was my main correctness worry and it's clean.

### Findings

| # | Sev | Where | What |
| --- | --- | --- | --- |
| 1 | **P2** | `service.ts` 5251-5267 | Capacity requeue drops the poll-interval liveness fallback; when `retryEntry.dueAt` is unavailable it now applies exponential backoff to a requeue that is documented to consume none. Reproduced. |
| 2 | **P2** | `service.test.ts` 1160-1171 | The "age" assertion can't fail — the fixture's original `dueAt` equals `now`, so retention and "reset to now" are indistinguishable. |
| 3 | **P2** | `service.test.ts` 191-286 | The fairness test stubs `reconcileRun`, so it covers comparator output only; the saturated-capacity behavior #703 actually asks for is untested, and coverage confirms 296-297/308 are unreached. |
| 4 | P3 | `service.ts` 5250 | `loadRetryPolicy` awaited even when retention short-circuits it — now once per blocked reservation per poll. |
| 5 | P3 | `service.ts` 5251-5267 | `retryEntry.error` clobbered to the capacity message while `attempt`/`dueAt` are retained, hiding the original failure. |
| 6 | nit | `service.ts` 291-309 | Hoist `now.getTime()`; prefer relational compare over `localeCompare` for a determinism tie-break. |
| 7 | nit | `service.test.ts` | Unrelated Prettier reflow of the fixture helpers. |

Details and suggested patches are inline. #1 is the only behavioral one; #2 and #3 are what keep this fix from silently regressing later. Address those three (P3s and nits at your discretion) and I'll approve.

---

packages/orchestrator/src/service.ts
moncher-dev: Resolved in `38853de`. Capacity-only requeues retain a valid reservation due time and its original failure context; a missing or invalid reservation falls back to the polling interval without applying retry backoff. Retry policy loading now occurs only for normal failure retry scheduling.

packages/orchestrator/src/service.ts
moncher-dev: Resolved in `38853de`. The comparator captures the shared time boundary once and uses locale-independent relational comparisons for stable issue and run ID tie-breaking.

packages/orchestrator/src/service.test.ts
moncher-dev: Resolved in `38853de`. The retained reservation is now five minutes older than the current tick, so reset-to-now cannot satisfy the assertion; the same test also covers the poll-interval fallback and retained failure context.

packages/orchestrator/src/service.test.ts
moncher-dev: Resolved in `a1865f6`. The new full reconciliation test uses two due reservations with capacity one: the oldest starts, while the newer reservation remains queued with its original due time and failure context. Reversing due-time ordering makes the test fail.

packages/orchestrator/src/service.ts
```diff
@@ -5211,30 +5247,44 @@ export class OrchestratorService {
...
     const lastError = suppressed
       ? formatMaxFailureRetrySuppression(
           run,
           failureRetryCount,
           maxFailureRetries,
           error
         )
-      : error;
+      : (options.advanceAttempt === false
+        ? (issueRecord?.retryEntry?.error ?? run.lastError ?? error)
```
hojinzs, line 5296: **P2 — retaining `error` took the only capacity-contention signal out of the system entirely; nothing replaced it.**

My round-1 P3 offered two shapes: *"retaining `error` too **or** recording the capacity message separately"*. This commit took the first and skipped the second, and the combination is worse than either half alone.

The `"no available orchestrator slots"` string is now constructed at line 3615 and then discarded on this line. I checked every channel it could still reach:

| channel | before | after |
|---|---|---|
| `run.lastError` | `"no available orchestrator slots"` | `"worker failed"` (retained) |
| `retryEntry.error` | `"no available orchestrator slots"` | `"worker failed"` (retained) |
| `events.ndjson` | — | — |
| stderr | — | — |

`requeueRetryingRun` emits no `appendRunEvent` and no `writeStderr` (the only retry log in the file is `[retry-scheduled]` at line 3902, on the scheduling path, not this one), and it returns `{ issueRecords, recovered: false }` with no `lastError`, so the reconcile loop's `lastError = outcome.lastError ?? lastError` never picks it up either. Grep confirms line 3615 is the string's sole occurrence outside tests.

Now compose that with the *other* half of this change. `dueAt` is retained, so a saturated reservation's due time is frozen in the past and drifts further behind every poll. The snapshot at line ~1161 therefore renders, indefinitely:

```
retry: { due_at: "2026-03-07T23:55:00.000Z", error: "worker failed", kind: "failure" }
```

An operator reading that at 02:00 sees a retry two hours overdue, attributed to a worker failure, with no indication the orchestrator is deliberately holding it. The honest reading of that snapshot is "the retry never fired" — the opposite of what is happening. Before this PR the stale-`dueAt` half didn't exist and the error field named the real cause, so neither reading was available to go wrong.

This is Observability (layer 6) losing state that Coordination still has. Retention of `attempt`/`dueAt` is right — that's the reservation identity the fairness policy needs — but the *reason it is parked* is not part of that identity and shouldn't have been folded into it. Cheapest fix that keeps both properties:

```ts
await this.store.appendRunEvent(run.runId, {
  at: now.toISOString(),
  event: "retry-postponed",
  projectId: run.projectId,
  runId: run.runId,
  issueIdentifier: run.issueIdentifier,
  attempt,
  dueAt,
  error,            // "no available orchestrator slots"
} as OrchestratorEvent);
```

A distinct `retryEntry.postponedReason` (or surfacing it as `retry.blocked_by` in the snapshot) would work equally well. Either keeps `retryEntry.error` honest about the original failure *and* makes a capacity-parked reservation distinguishable from a stuck one.

**nit**, same line: `??` only falls through on `null`/`undefined`, so a reservation persisted with `error: ""` retains the empty string rather than falling back to `run.lastError`. That exact shape is what the fixtures in this PR construct (`error: run.lastError ?? ""`, service.test.ts:238 and 6376), so it isn't purely hypothetical. A truthiness check would be sturdier here.

---

packages/orchestrator/src/service.ts
```diff
@@ -277,6 +277,42 @@ function isDueRetryReservation(
...
+      run,
+      index,
+      dueAtMs:
+        run.status === "retrying" ? parseTimestampMs(run.nextRetryAt) : null,
+    }))
+    .sort((left, right) => {
+      const leftIsDue = left.dueAtMs !== null && left.dueAtMs <= nowMs;
+      const rightIsDue = right.dueAtMs !== null && right.dueAtMs <= nowMs;
+      if (leftIsDue !== rightIsDue) {
+        return leftIsDue ? 1 : -1;
```
hojinzs, line 296: **P2 — the comparator arm that implements this PR's headline invariant is still never executed.**

I re-ran v8 coverage over `service.ts` on `a1865f6` with the full 212-test suite. The uncovered statements inside `sortRunsForReconciliation` are unchanged from my round-1 report:

```
uncovered statement lines 280-315: 296, 297, 305, 306, 308, 310
uncovered branch    lines 280-315: 295, 302, 304, 307, 309
```

Three of the comparator's five decision arms never run:

- **296-297** — the non-due/due swap. This is the arm that puts normal active runs ahead of due retries: your Risks section's *"normal active runs remain ahead of due retries"* and `docs/architecture.md`'s *"Due reservations are reconciled after non-due active runs"*. Both new tests build fixtures where **every** run is due, so the branch is never taken. Invert `return leftIsDue ? 1 : -1` to `-1 : 1` and the suite still passes 212/212 — I checked.
- **305-306** — the `issueIdentifier` tie-break.
- **307-311** — the `runId` tie-break. Unreachable in practice as written: it only fires when two runs share a `dueAtMs` *and* an `issueIdentifier`, and one issue holds one reservation. Dead code rather than determinism.

The round-1 thread asked for a direct unit test over the pure function and suggested exporting it; the reply marked it resolved on the strength of the end-to-end test alone. The e2e test is a real improvement and I'm not asking you to drop it — it's what let me confirm the mechanism actually works (see the review summary). But it and the comparator test are both *whole-loop* tests, and neither can cheaply reach these arms. The gap is that the invariant you documented has no regression guard at all.

`sortRunsForReconciliation` is pure and takes `(OrchestratorRunRecord[], Date)`. Exporting it makes all five arms a table test:

```ts
const order = (runs: Partial<OrchestratorRunRecord>[]) =>
  sortRunsForReconciliation(runs as OrchestratorRunRecord[], new Date("2026-03-08T00:01:00Z"))
    .map((r) => r.runId);

// due reservations sort after non-due active runs, whatever the input order
expect(order([
  { runId: "due",     status: "retrying", nextRetryAt: "2026-03-08T00:00:00Z" },
  { runId: "running", status: "running",  nextRetryAt: null },
  { runId: "pending", status: "retrying", nextRetryAt: "2026-03-08T00:05:00Z" },
])).toEqual(["running", "pending", "due"]);
```

That covers 296-297 and pins the documented ordering, and the same table extends to the tie-breaks (or shows the `runId` arm is unreachable and can go).

---

packages/orchestrator/src/service.test.ts
```diff
@@ -188,6 +188,101 @@ describe("OrchestratorService", () => {
...
+      listIssuesByStates: vi.fn().mockResolvedValue([]),
+      fetchIssueStatesByIds: vi.fn().mockResolvedValue([]),
+      buildWorkerEnvironment: vi.fn().mockReturnValue({}),
+      reviveIssue: vi.fn(),
+    } as never);
+    const service = new OrchestratorService(store, projectConfig, {
+      now: () => new Date("2026-03-08T00:01:00.000Z"),
+    });
+    const reconciledRunIds: string[] = [];
+    vi.spyOn(service as never, "reconcileRun").mockImplementation(
```
hojinzs: **P3 — this test is now redundant with the saturation test, and it's the more fragile of the two.**

The new `"starts the oldest due retry and retains the newer reservation under saturation"` test (line 6292) covers this one's assertion strictly: same two-reservation, oldest-first ordering, but through the real `reconcileRun` rather than a stub. Everything this test can still catch, that one catches too.

What it adds instead is the coupling my round-1 comment flagged: `vi.spyOn(service as never, "reconcileRun")` names a private method through an `as never` cast that also discards the type check. Rename `reconcileRun` and this doesn't fail — `mockImplementation` installs a property nothing calls, `reconciledRunIds` stays `[]`, and the assertion `toEqual(["run-z-older", "run-a-newer"])` fails for a reason unrelated to ordering. Change it to a *no-op* stub shape and it degrades silently instead. That's a test that reports on the spy plumbing, not on the scheduler.

Suggest deleting it in favour of the direct `sortRunsForReconciliation` table test from my `service.ts:296` comment — that one is faster, needs no filesystem fixture or tracker mock, touches no private API, and reaches the three comparator arms neither whole-loop test can. Net: one test removed, real coverage up.

---

packages/orchestrator/src/service.test.ts
```diff
@@ -16969,7 +17265,7 @@ async function writeWorkflowFixture(
 ): Promise<void> {
   const content = normalizeTrackerProviderFixture(
     options.rawWorkflow ??
-    `---
+      `---
```
hojinzs: **nit (carried over, unaddressed)** — the Prettier reflow of `writeWorkflowFixture` / `normalizeTrackerProviderFixture` is still in the diff (three hunks: 17268, 17335-17336, 17342). I raised this in round 1; the reply on that thread covered the fairness-test rewrite but didn't mention this half, so I assume it was just missed rather than declined.

`pnpm format` isn't CI-gated and `main` is already 90 files out of compliance, so this isn't fixing a break — it's three unrelated hunks in a PR about retry ordering. Dropping them (`git checkout origin/main -- ` the ranges, or just revert those three) keeps the diff readable for whoever bisects this later. Not blocking; happy for you to leave it with a one-line "keeping it" if you'd rather not touch the file again.

---

hojinzs (changes requested):
## Round 2 — the mechanism is correct and I verified it end-to-end; one observability regression and one coverage gap left

**Round-1 follow-up: 3 of 4 threads fully resolved, 1 partially.**

| round-1 thread | status |
|---|---|
| P2 poll-interval fallback lost | ✅ fixed — `advanceAttempt === false` now floors at the poll interval before backoff |
| ↳ P3 eager `loadRetryPolicy` | ✅ fixed — moved into the advancing branch only |
| ↳ P3 `error` overwritten | ⚠️ retained, but see the new P2 below — the half I offered as an alternative was the one that mattered |
| nit hoist `now.getTime()` / drop `localeCompare` | ✅ fixed |
| P2 aging test didn't discriminate | ✅ fixed — queued due time is now 5 min before `now`, plus the fallback case |
| P2 fairness test stubbed the whole loop | ⚠️ partially — real e2e test added (good), direct comparator test still missing |

### Requirements vs #703

All met. Ordering (oldest `dueAt` first), aging (retention, so waiting accrues), no failure-budget or backoff consumption, `docs/architecture.md` updated, `changeset:patch` matching the issue label. No scope creep in the implementation — the `lastError` retention came out of round-1 review, and the only unrequested change is the Prettier reflow (nit).

### Smoke test — the property actually holds

I built the two-pass scenario I asked for in round 1, since neither new test spans two polls. Capacity 1; A due 00:00:00, B due 00:00:30.

```
PASS 1 @ 00:01:00  A → running          B → retry_queued, dueAt 00:00:30, error "retry 2 failed"
   (A then fails and re-queues with fresh backoff → dueAt 00:02:00)
PASS 2 @ 00:03:00  B → running          A → retry_queued, dueAt 00:02:00, failureRetryCount 0
                   SPAWNS_IN_PASS2 = 1
```

B wins the slot over the retry that just re-backed-off — that is #703's actual ask, and it passes. `failureRetryCount` stays 0 across both passes, so saturation genuinely doesn't eat the budget. **Negative control:** forcing `retainedDueAt = null` makes this fail at pass 1, confirming the retention is load-bearing and not incidentally satisfied.

Also run locally on `a1865f6`: `pnpm build`, `pnpm test` (orchestrator 212/212), `pnpm typecheck`, `pnpm lint` — all green. CI green (Test, Container Smoke).

### What's blocking

**P2 — `service.ts:5286`** (new, introduced by `38853de`): retaining `error` removed the last channel that recorded capacity contention, and nothing replaced it. `"no available orchestrator slots"` is built at line 3615 and dropped on the spot — no `appendRunEvent`, no stderr, not returned to the reconcile loop. Combined with the frozen past `dueAt`, the snapshot shows a permanently-overdue retry attributed to `"worker failed"`, which reads as *stuck* when it's actually *parked*. Coordination knows why; Observability no longer can. A `retry-postponed` event (or a distinct `postponedReason`) restores it without touching the fairness identity.

**P2 — `service.ts:296`** (carried over): the non-due/due swap — the invariant in your own Risks section and in `architecture.md` — has zero coverage. Both new tests make every run due. Inverting that arm still passes 212/212. Exporting the pure comparator gives you a table test that pins it in a few lines.

Plus **P3** on the now-redundant stubbed test at `service.test.ts:270` and the carried-over **nit** on the Prettier reflow.

None of this is a correctness defect — the scheduler does what #703 asked, and I'm satisfied it does. Requesting changes on the observability regression specifically, since it's a small, well-scoped fix and it's the one thing this PR makes actively worse than `main`. Address that and the `:296` coverage and I'll approve; the P3 and nit are yours to take or decline with a note.

---

packages/orchestrator/src/service.ts
moncher-dev: Resolved in 473fc13: capacity requeues now emit retry-postponed with the retained due time and capacity reason while preserving the original retry error.

packages/orchestrator/src/service.ts
moncher-dev: Resolved in 473fc13: sortRunsForReconciliation is exported and table-tested for non-due priority, due-time ordering, issue identity, and run-ID ties.

packages/orchestrator/src/service.test.ts
moncher-dev: Resolved in 473fc13: removed the private-method spy test in favor of direct pure-function coverage; the saturation test remains.

packages/orchestrator/src/service.test.ts
moncher-dev: Resolved in 473fc13: removed the unrelated writeWorkflowFixture and normalizeTrackerProviderFixture formatting hunks.

packages/orchestrator/src/service.ts
```diff
@@ -5255,6 +5304,19 @@ export class OrchestratorService {
       retryKind: suppressed ? null : (run.retryKind ?? "failure"),
       lastError,
     });
+    if (options.advanceAttempt === false && dueAt) {
```
hojinzs, line 5318: **P2 (not blocking — see the review summary) — this emits once per poll for as long as the reservation is parked, and after 20 polls it has evicted the very window it writes into.**

The event restores the capacity signal I asked for in round 2, and `latestEventMessage` now reads correctly. But there is no emit-once guard, so a reservation held by saturation appends a new `retry-postponed` on *every* reconciliation pass — and sustained saturation is precisely the steady state #703 is about.

I measured it by driving `requeueRetryingRun` through 25 consecutive capacity postponements (12.5 min at the 30s default interval), after seeding one genuine `run-retried` carrying the original failure:

```json
{
  "polls": 25,
  "totalEventLines": 26,
  "retryPostponedLines": 25,
  "eventsNdjsonBytes": 8813,
  "recentWindowSize": 20,
  "recentWindowEventTypes": ["retry-postponed"],
  "runRetriedStillInWindow": false,
  "latestEventMessage": "Retry 2 postponed until 2026-03-07T23:55:00.000Z: no available orchestrator slots"
}
```

Both snapshot consumers read a fixed 20-event tail — `service.ts:1119` (`loadRecentRunEvents(runId, 20, projectId)`) and `packages/dashboard/src/store.ts:272`. So after ten minutes of saturation the issue-detail event list is twenty identical postponement lines and the `run-retried` that explains *why* the run is retrying at all is gone. Secondary: ~340 B/event → roughly 1 MB/day per parked reservation, unbounded, and `appendRunEvent` fsyncs twice per call (log + mirror) on the reconcile hot path.

There is an exact precedent for the guard 3,600 lines up in this same file — `convergence-lock-expired` at `service.ts:1698-1706` reads the recent events first and only appends when the condition is not already recorded:

```ts
const recentEvents = await this.store.loadRecentRunEvents(run.runId, 20, run.projectId);
if (!recentEvents.some((event) => event.event === "retry-postponed")) {
  await this.store.appendRunEvent(run.runId, { /* … */ });
}
```

A `retryEntry.postponedReason` field (the alternative I offered in round 2) would work equally well and is cheaper — set it here, clear it on `restartRun` alongside `retryEntry: null`, emit only on the `undefined → reason` transition, and the snapshot gets a first-class field instead of an event it has to scan for.

I'm flagging rather than blocking because the original failure context is *not* lost: this PR deliberately retains it on `run.lastError` and `retryEntry.error`, so the retry row still names the real failure and no operator is actually blind. What degrades is the event history, which is a debuggability cost, not a correctness one.

---

packages/orchestrator/src/service.ts
```diff
@@ -5211,30 +5247,43 @@ export class OrchestratorService {
...
     const lastError = suppressed
       ? formatMaxFailureRetrySuppression(
           run,
           failureRetryCount,
           maxFailureRetries,
           error
         )
-      : error;
+      : options.advanceAttempt === false
+        ? (issueRecord?.retryEntry?.error ?? run.lastError ?? error)
```
hojinzs, line 5296: **nit (carried over from round 2, unaddressed)** — `??` only falls through on `null`/`undefined`, so a reservation persisted with `error: ""` retains the empty string instead of falling back to `run.lastError`.

That shape is what the fixtures in this PR construct — `error: run.lastError ?? ""` at `service.test.ts:6293` — so it isn't purely hypothetical: a reservation whose run had a null `lastError` gets `""` written into `retryEntry.error`, and every subsequent capacity requeue then retains `""` forever rather than recovering the error when one appears. A truthiness check (`issueRecord?.retryEntry?.error || run.lastError || error`) is sturdier here.

The round-2 reply on this thread covered the `retry-postponed` half but didn't mention this, so I assume it was missed rather than declined. Genuinely a nit — fine to leave with a one-line "keeping it".

---

hojinzs (approved):
## Round 3 — approving. Both round-2 blockers are genuinely fixed and I verified each with a negative control.

**Round-2 follow-up: 4 of 4 threads resolved.**

| round-2 thread | status |
|---|---|
| P2 observability regression (capacity signal had no channel left) | ✅ fixed — typed `retry-postponed` event, formatter, docs, test |
| P2 comparator arms never executed | ✅ fixed — `sortRunsForReconciliation` exported and table-tested |
| P3 redundant private-method spy test | ✅ removed — no `reconciledRunIds` fixture remains |
| nit unrelated Prettier reflow | ✅ removed — 0 `writeWorkflowFixture`/`normalizeTrackerProviderFixture` hunks in the diff |

### Requirements vs #703

All met, unchanged from round 2 and re-confirmed here: fair ordering (oldest `dueAt` first, deterministic tie-breaks), aging (capacity-only requeue retains the original due time, so a waiting retry is not jumped by one that just re-backed-off), and neither failure budget nor retry backoff consumed. `docs/architecture.md` describes the new ordering and the `retry-postponed` event; the changeset is `@gh-symphony/cli` patch, matching the issue's `changeset:patch` label.

No over-engineering: the two new pieces this round — exporting a pure function for test, and one event type — both came out of review, not out of the author's own scope expansion. Nothing in the diff goes beyond #703.

### Verification on `473fc13`

| Check | Result |
|---|---|
| `pnpm build` | pass |
| `pnpm typecheck` | pass |
| `pnpm lint` | pass |
| `pnpm test` (full workspace) | **1,839 pass**, 0 fail (orchestrator 389, core 332, cli 616) |
| CI `Container Smoke` | green |

Note for anyone reproducing: `pnpm typecheck` and `pnpm test` both fail on a clean checkout until `pnpm build` has run (`tracker-linear` can't resolve `@gh-symphony/tool-linear-graphql`, `tracker-file` fails to collect). Build first — that's pre-existing, not this PR.

**Negative control on the round-2 coverage gap.** Inverting the non-due/due arm — `return leftIsDue ? 1 : -1` → `-1 : 1` — passed 212/212 in round 2. It now fails:

```
FAIL  orders non-due runs before fair due retry reservations
AssertionError: expected [ 'due-a', 'due-a-run-a', … ] to deeply equal [ 'pending', 'running', 'due-a', … ]
```

So the invariant documented in `architecture.md` and in your Risks section finally has a guard. v8 coverage over lines 280-315 confirms every decision arm now executes; what remains uncovered is the mirrored argument order V8 happens not to produce, plus two structurally unreachable defensive cases (`?? 0` when both runs are due, and `left.runId === right.runId` — you can't hold the same run twice). The `runId` tie-break I called possibly-dead in round 2 is now exercised by a synthetic same-identifier fixture; keeping it covered rather than deleting it is a fine call and I'm not going to relitigate it.

**Aging test now discriminates.** The queued due time is 5 min before `now` and the fallback case asserts `now + 30s` (poll interval) against a `retryBaseDelayMs` of 10s at attempt 4 — which would be 80s of backoff. Retention, reset-to-now, and backoff are all distinguishable. That was my round-1 P2 and it's properly closed.

**Event scoping.** I checked that `advanceAttempt: false` has exactly one call site (`service.ts:3610-3616`, the `hasRetryDispatchSlot` failure), so `retry-postponed` can only ever describe capacity contention — the other `requeueRetryingRun` call at `:3574` passes no options. No mislabeling risk.

### Findings

| # | Sev | Where | What |
|---|---|---|---|
| 1 | P2 | `service.ts:5307` | `retry-postponed` emits once per poll with no guard; after 20 polls it has evicted the entire 20-event window both the snapshot and the dashboard read, and grows `events.ndjson` unboundedly. Measured. |
| 2 | nit | `service.ts:5285` | `??` retains a persisted `error: ""` instead of falling back to `run.lastError`. Carried over from round 2. |
| 3 | nit | `cli/src/commands/logs.ts:233-246` | `getLevel` doesn't list `retry-postponed`, so it falls to `default: "info"` while `run-retried` is `warn`. `gh-symphony repo logs --level warn` won't show a saturated orchestrator. Arguably correct as-is — flagging only so it's a decision. (Outside the diff, hence not inline.) |

**Why I'm approving rather than running a fourth round on #1.** It's a real defect and I'd take the ~6-line guard if you're willing to push it here — the precedent is `convergence-lock-expired` at `service.ts:1698-1706`, in this same file. But it doesn't leave an operator blind: this PR deliberately retains the original failure on `run.lastError` and `retryEntry.error`, so the retry row still names the real cause and `latestEventMessage` now names the capacity reason. Both facts are reachable; what degrades is event *history*. That's a debuggability cost, not a correctness one, and it's independently fixable without touching the fairness mechanism.

More to the point: round 2's fix produced round 3's finding, and I'd rather not run that loop again over something this bounded. The scheduler does what #703 asked, I've verified the property end-to-end with a negative control, and every blocking item I've raised across three rounds is closed. Ship it — take #1 here or as a fast-follow, your call; #2 and #3 at your discretion.

---

packages/orchestrator/src/service.ts
moncher-dev: Tracked as #806 for a bounded or deduplicated capacity-postponement signal. The reviewer explicitly approved #803 for shipment, so this remains a non-blocking fast follow.

packages/orchestrator/src/service.ts
moncher-dev: Acknowledged. The approved review classifies this as a non-blocking nit; no code change is needed for the Land cycle.

Issue #808, opened 2026-09-03
[Epic] repo-mode workers dispatch without GitHub credentials or an assigned branch — Ready→Backlog self-parking (#647/#804/#806)


moncher-dev:
## Incident

On 2026-09-03 (daemon `gh-symphony v1.1.0`, `repo start`, right after #807 aligned WORKFLOW.md with the 1.1.0 child-isolation contract) issues #647, #804 and #806 were dispatched from `Ready` and every agent parked its issue back to `Backlog` through the tracker-state API with the comment "External orchestration blocker: GitHub token broker is unavailable and no assigned branch was provided". No code was changed; ~186k input tokens per run were spent on discovering the blocker.

## Root cause chain

1. **Worker host tools have no GitHub credential in repo mode.** The host-executed `github_graphql` dynamic tool returned `tool_execution_failed: Either GITHUB_GRAPHQL_TOKEN or the runtime token broker configuration is required.` The worker environment is built from the project `.env` + `INHERITED_ENV_ALLOWLIST` (HOME/PATH/…) + Symphony context (`packages/orchestrator/src/service.ts` `buildProjectExecutionEnv`). `repo start` resolves gh auth into the daemon's own `process.env` only; #455 removed the explicit `GITHUB_GRAPHQL_TOKEN` hand-off to workers (risk-audit H2, #439) and pointed at a project `.env` / token broker instead, but nothing provisions `.runtime/orchestrator/projects/repository/.env`, the broker is an external contract with no in-repo server, and `doctor` does not check the worker-effective environment. The same failure string appears in ~195 older worker logs across other projects since 2026-08-25; the 1.1.0 WORKFLOW.md made it fatal.
2. **Host Git transport silently falls back to the operator's identity.** Without a token/broker `buildHostGitEnvironment` sets no credential helper, so the end-of-turn push runs with the daemon HOME and the system gitconfig (`osxkeychain`): pushes would be made as the developer, not as the tracker identity.
3. **Repo mode never creates the assigned branch.** `repo init` writes `"populateStrategy": "clone"`; `cloneRepositoryForRun` only checks out a linked PR head, so fresh workspaces stay on `main` and `SYMPHONY_ASSIGNED_BRANCH=main`. `project` mode uses `worktree-cache`, which creates `symphony/<issue>` via `worktree add -B`. WORKFLOW.md rule 3 promises the branch.
[...]

Issue #806, opened 2026-09-03
fix(orchestrator): bound retry postponement events


moncher-dev:
## Context

The fair retry reservation work in #703 adds a `retry-postponed` event when due retry reservations are held by concurrency saturation.

## Follow-up

Bound or deduplicate capacity-postponement event emission so a continuously saturated reservation does not fill the recent-event window or grow `events.ndjson` once per reconciliation poll. Preserve a visible capacity-postponement signal for operators.

Origin: #703 / PR #803.

moncher-dev:
🔁 Status: `Ready` → `Backlog`

Reason: External permission blocker: restore the host GitHub GraphQL token broker and provide SYMPHONY_ASSIGNED_BRANCH with the issue #806 branch; the worker cannot read/create the required workpad or safely publish issue-bound commits while checked out on main.
Cycle: — (triage rejection)

moncher-dev:
## Workpad — hojinzs/github-symphony#806 — Cycle 1

**Type:** resume after blocker
**Branch:** symphony/hojinzs-github-symphony-806
**Draft PR:** not yet created
**Cycle opened:** 2026-09-03T16:31:49Z · **Trigger:** Human restored the issue to active work after the host permissions and assigned branch became available.

### Status Transitions

- 2026-09-03T16:31:49Z · `Ready` → `In progress` · Host permissions and the assigned issue branch are available; resume implementation (cycle 1 open)

Intended close transition body:

> 🔁 Status: `In progress` → `Backlog`
>
> Reason: External host-publication blocker: GitHub cannot resolve the assigned issue branch after three consecutive host push opportunities; restore branch publication for the committed issue branch.
> Cycle: 1 close

### Plan

- [x] 1. Reproduce repeated capacity-postponement emission, implement bounded signaling, add focused tests, and commit the shippable slice
- [ ] 2. Open Draft PR via /gh-pr-writeup (branch is on the remote now)
- [x] 3. Add the patch changeset and retry-fire TC coverage; run the Docker black-box lifecycle
- [ ] 4. Wrap-up: re-verify the original issue · pass the Completion Bar · PR ready · transition to In review

### Rework / PR Feedback

None

### Completion Bar

- [ ] In-scope requirements implemented
- [ ] `pnpm lint`
- [ ] `pnpm test`
- [ ] `pnpm typecheck`
- [ ] `pnpm build`
- [ ] Docker E2E (if integration behavior changed)
[...]

Issue #804, opened 2026-09-03
test(ci): run built-worker startup regressions in CI


moncher-dev:
## Problem

The built-worker Codex startup-failure regression in `test/e2e/claude/claude-docker.spec.ts` is excluded from root `pnpm test` because it is a `*.spec.ts` file, and the CI workflow does not run `pnpm e2e:claude`.

That leaves the worker startup exit-code and terminal-heartbeat contract unguarded in CI, despite its Docker black-box coverage being load-bearing.

## Scope

- Decide the smallest CI path that executes the built-worker `runWorkerProcess` scenarios (or the full `pnpm e2e:claude` suite).
- Ensure the Codex startup-failure regression is exercised by CI.
- Keep the change separate from #779's narrowly scoped runtime fix.

## Evidence

The regression passes with the fix and fails with the prior worker implementation, where the process exits `0` instead of `1`.

moncher-dev:
🔁 Status: `Ready` → `Backlog`

Reason: External permission blocker: restore the host GitHub GraphQL token broker and provision SYMPHONY_ASSIGNED_BRANCH for issue #804; the worker cannot read/adopt the retry workpad or safely publish commits while github_graphql is unavailable and the checkout remains on main.
Cycle: — (triage rejection)

moncher-dev:
## Workpad — hojinzs/github-symphony#804 — Cycle 1

**Type:** resume after blocker
**Branch:** symphony/hojinzs-github-symphony-804
**Draft PR:** not yet created
**Cycle opened:** 2026-09-03T16:32:05Z · **Trigger:** Permission blocker resolved; recover the already-confirmed In progress cycle.

### Status Transitions

- No agent-owned transition in this recovery cycle; tracker was already `In progress` at dispatch.

### Plan

- [x] 1. Add a focused CI command for the built-worker Codex startup regression, document its CI coverage, validate, and commit.
- [ ] 2. Open Draft PR via /gh-pr-writeup (branch is on the remote now).
- [ ] 3. Wrap-up: re-verify the original issue · pass the Completion Bar · changeset (if needed) · PR ready · transition to In review.

### Rework / PR Feedback

Not applicable.

### Completion Bar

- [ ] In-scope requirements implemented
- [ ] `pnpm lint`
- [ ] `pnpm test`
- [ ] `pnpm typecheck`
- [ ] `pnpm build`
- [ ] Docker E2E (if integration behavior changed)
[...]

Issue #647, opened 2026-08-27
fix(orchestrator): verify owner lock identity before protecting runs


moncher-dev:
Tracked separately because this review item is non-blocking and requires extending live-owner identity data beyond the current ownership-token contract. The current guard remains fail-closed when a foreign PID is live.

moncher-dev:
🔁 Status: `Ready` → `Backlog`

Reason: External orchestration blocker: GitHub token broker is unavailable and no assigned branch was provided
Cycle: — (triage rejection)

moncher-dev:
## Workpad — hojinzs/github-symphony#647 — Cycle 1

**Type:** fresh pickup
**Branch:** symphony/hojinzs-github-symphony-647
**Draft PR:** not yet created
**Cycle opened:** 2026-09-03T16:32:00Z · **Trigger:** Human override resolved the prior environment triage rejection; assigned branch is available and the issue is In progress

### Status Transitions

- 2026-09-03T16:32:00Z · `Ready` → `In progress` · Human override resolved the prior environment triage rejection (cycle 1 open)

Planned transition body:

> 🔁 Status: `In progress` → `Backlog`
>
> Reason: External Git transport blocker: the host did not publish the assigned branch after three consecutive turn boundaries, so GitHub cannot create the delivery PR
> Cycle: 1 close

### Plan

- [x] 1. Persist project-lock process identity on runs and require a matching live identity before owner protection; add focused regression tests and commit
- [ ] 2. Open Draft PR via /gh-pr-writeup — blocked: GitHub still has no assigned branch ref after three publication boundaries
- [ ] 3. Wrap-up: re-verify the original issue · pass the Completion Bar · changeset · PR ready · transition to In review — Completion Bar and changeset complete; PR handoff blocked by transport

### Rework / PR Feedback

None

### Completion Bar

- [x] In-scope requirements implemented
- [x] `pnpm lint`
- [x] `pnpm test` (documented unrelated exception: #816)
- [x] `pnpm typecheck`
- [x] `pnpm build`
- [x] Docker E2E (documented environment exception: #817)
- [x] Tests for new functionality
- [x] Code conventions (CLAUDE.md)
- [x] Review threads answered and resolved (N/A — fresh cycle, no PR)

### Validation

- `pnpm exec vitest run packages/orchestrator/src/index.test.ts packages/orchestrator/src/service.test.ts -t 'owner|process-scoped'` — pass (11 passed)
- `pnpm lint` — pass
[...]

PR #814, opened 2026-09-03
chore(workflow): raise codex read_timeout_ms to 30s

moncher-dev:
## TL;DR

- `runtime.timeouts.read_timeout_ms` 5000 → 30000 in WORKFLOW.md.
- Policy-only change (no code, no changeset).

## Why

`thread/start` covers Codex app-server initialization including MCP server startup. On 2026-09-04 (host load average ~97) every attempt for #647 / #804 / #806 hit `response_timeout: thread/start timed out after 5000ms`, was recovered twice and then failure-suppressed (`max_failure_retries_exceeded`), so nothing could be dispatched. The same timeout already caused the first-attempt stalls on 2026-09-03. 30s still bounds a hung app-server (turn/stall timeouts are unchanged) while tolerating slow startups.

Part of the #808 recovery. The operator host currently runs with the same value as a local override until this lands.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01ByFE4XwtWX8x4nihCL5QgQ

### DIFF
--- a/WORKFLOW.md
+++ b/WORKFLOW.md
@@ -46,7 +46,7 @@ runtime:
   args:
     - app-server
   timeouts:
-    read_timeout_ms: 5000
+    read_timeout_ms: 30000
     turn_timeout_ms: 3600000
     stall_timeout_ms: 900000
 ---


moncher-dev:
Nudge for review — this has been open since 2026-09-03T16:14Z with no reviewer assigned.

It raises the Codex `read_timeout_ms` to 30s to absorb `thread/start timed out` failures under host load (load average was ~97 when the timeouts were observed). It is a one-line policy change to WORKFLOW.md and blocks nothing else, but until it merges the local working copy carries an uncommitted 30000 override, so the repo and the running daemon disagree. @hojinzs a quick approve or a request for changes would let us drop the local override.

WORKFLOW.md
```diff
@@ -46,7 +46,7 @@ runtime:
   args:
     - app-server
   timeouts:
-    read_timeout_ms: 5000
+    read_timeout_ms: 30000
```
hojinzs, line 49: Value verified — blast radius is bounded, no objection to 30000.

I traced `readTimeoutMs` end to end. It reaches exactly three call sites, all short request/ack round-trips:

- `packages/worker/src/index.ts:1668` — `initialize`
- `packages/worker/src/index.ts:1706` — `thread/start`
- `packages/worker/src/index.ts:1825` — `turn/start` (the ack, not the stream)

Turn streaming stays governed by `turn_timeout_ms` / `stall_timeout_ms`, exactly as the PR body claims. So the only cost of 5s → 30s is that a genuinely hung app-server is now detected in ~30s per request instead of ~5s — negligible against `stall_timeout_ms: 900000`, `turn_timeout_ms: 3600000`, and the orchestrator's 30-minute inactive-worker reclaim. `parser.ts` (`readOptionalIntegerLike`) applies no upper bound and no cross-field constraint, so nothing else validates against this.

**P3 (follow-up, not this PR):** the scaffold still emits the spec default. `packages/cli/src/skills/templates/gh-symphony.ts:139` hardcodes `read_timeout_ms: 5000`, which is what lands in `.gh-symphony/reference-workflow.md`, `.codex/skills/gh-symphony/SKILL.md`, and `docs/examples/linear-WORKFLOW.md`. If 5s is genuinely too tight for Codex app-server startup with MCP servers, every newly scaffolded repo reproduces the #808 follow-up blocker. Leaving it here is the right call — spec §5.3 fixes the default at `5000`, so changing the code default is a spec divergence needing its own artifact — but it's worth an issue.

---

hojinzs (approved):
## Approve

Requirement traceability, scope, and risk all check out. One line, exactly the line #808 asked for.

### Requirements vs. implementation

#808's "Follow-up blocker found while re-dispatching (2026-09-04 01:10 KST)" specifies precisely this: `#814 raises it to 30s`. The diff is `read_timeout_ms: 5000 → 30000` in `WORKFLOW.md` and nothing else — `turn_timeout_ms` and `stall_timeout_ms` untouched, as the PR body claims. No over-engineering, no unrequested decisions, no scope creep into the five open work items (#809–#813).

### Verification performed

- **Smoke test (local, deps installed).** Parsed the actual post-change `WORKFLOW.md` through `parseWorkflowMarkdown`: front matter parses clean, `runtime.kind === "codex-app-server"`, `timeouts.readTimeoutMs === 30000`, `turnTimeoutMs === 3600000`, `stallTimeoutMs === 900000`. No preview environment applies — this is a config-only change with no API or screen surface, so the parse path is the behavioural surface.
- **Propagation.** `parser.ts:1145` → `service.ts:3255` (`SYMPHONY_READ_TIMEOUT_MS`) → `worker/src/index.ts:1147`. Intact.
- **Blast radius.** Detailed inline on the changed line. Bounded to three startup/ack requests; turn streaming is unaffected.
- **CI safety.** No test binds to the repo-root `WORKFLOW.md` — every fixture carries its own `5000`, so nothing regresses. Consistent with `Test` and `Container Smoke` both green on `e996f0f`.
- **Changeset.** Correctly omitted. `packages/cli` publishes `files: ["dist", "README.md", "LICENSE"]`; `WORKFLOW.md` is layer-1 repo policy and ships in nothing.

### Non-blocking

- **P3 — scaffold drift.** `packages/cli/src/skills/templates/gh-symphony.ts:139` still emits `read_timeout_ms: 5000` for newly scaffolded repos. Details inline. Out of scope here (spec §5.3 pins the default), worth its own issue.
- **P3 — symptom vs. cause.** `response_timeout` sets `runPhase = "stalled"` (`worker/src/index.ts:2063`), gets recovered, and after `max_failure_retries: 3` the issue self-suppresses with `max_failure_retries_exceeded` — which is what burned #647/#804/#806. 30s makes that far less likely under load, but a startup-phase timeout arguably shouldn't consume the same retry budget as an agent failure. Candidate follow-up under the #808 epic.
- **nit** — the PR body states "no changeset" without the reason; a half-line ("`WORKFLOW.md` is repo policy, not shipped in `@gh-symphony/cli`") would preempt the question on future policy PRs.

Nothing here blocks. Ship it and drop the local override.

---

Issue #815, opened 2026-09-03
test(worker): bound credential-helper transport test


moncher-dev:
## Problem

`packages/worker/src/git-transport.test.ts` consistently times out in the host credential-helper authentication case on the macOS Symphony worker host.

## Reproduction

Both commands timed out after 15 seconds at the same test:

```text
pnpm test
pnpm --filter @gh-symphony/worker exec vitest run src/git-transport.test.ts -t "uses host credential-helper authentication for remote Git operations"
```

Failure:

```text
FAIL src/git-transport.test.ts > synchronizeAssignedBranch > uses host credential-helper authentication for remote Git operations
Error: Test timed out in 15000ms.
```

## Context

Observed while validating #804. The #804 diff only changes `.github/workflows/ci.yml`, `package.json`, and `AGENT_TEST.md`, so this worker Git transport timeout is unrelated to that CI wiring change.

## Expected

The credential-helper test completes deterministically on supported development and CI hosts, or reports a bounded diagnostic instead of hanging until the test timeout.

moncher-dev:
Consolidating four duplicate reports into this issue.

The same timeout was filed independently by five workers validating unrelated issues: this one (#815, from #804), #816 (from #647), #821 (from #806), #823 (from #809), and #824 (from #811). #816, #821, #823 and #824 are now closed as duplicates pointing here. That five-for-five reproduction rate across independent runs makes this deterministic on this host, not flaky.

Consolidated detail from the closed reports:
- Fails identically under full `pnpm test` and under an isolated `-t` rerun; 17 sibling tests skip.
- Location: `packages/worker/src/git-transport.test.ts:322`.
- The test stands up a local authenticated Git HTTP server, so the likely culprit is a hanging local HTTP/backend interaction or a teardown path that never settles — not the 15s budget being too tight.
- No worker Git-transport source was changed by any of the five issues, so this is pre-existing.

Ask: make the fixture server and Git subprocess teardown deterministic rather than simply raising the timeout.

moncher-dev:
## Workpad — hojinzs/github-symphony#815 — Cycle 1

**Type:** fresh pickup
**Branch:** symphony/hojinzs-github-symphony-815
**Draft PR:** not yet created
**Cycle opened:** 2026-09-03T18:23:07Z · **Trigger:** Ready fresh pickup for deterministic credential-helper transport coverage

### Status Transitions

[...]

PR #818, opened 2026-09-03
test(ci): run built-worker startup regressions

### LINKED ISSUE #804: test(ci): run built-worker startup regressions in CI

moncher-dev:
## Issues

- Closes #804

## Summary

- Add a root command that runs the complete built-worker startup regression config without a fragile test-title filter.
- Run all nine startup scenarios in CI after build and typecheck, including the Codex non-zero exit and terminal failed-heartbeat contract.
- Document the CI path in the living worker lifecycle regression matrix.

## Change-point diagram

- CI build → lint/typecheck → `pnpm test:worker-startup` → all built-worker Vitest scenarios → exit-code and heartbeat assertions

## Start here

- `package.json:15` — unfiltered built-worker Vitest command
- `.github/workflows/ci.yml:44` — CI invocation after typecheck
- `AGENT_TEST.md:201` — living regression coverage matrix

- No CLI or runtime behavior changes. CI now fails on regressions across the complete built-worker startup suite, including Codex startup exit and terminal-heartbeat behavior.

## Validation

- `pnpm lint` — pass
- `pnpm test` — pass with documented unrelated exception #815: 224 tests passed; `packages/worker/src/git-transport.test.ts:322` timed out
- isolated #815 rerun — same unrelated 15-second timeout reproduced
- `pnpm typecheck` — pass
- `pnpm build` — pass
- `pnpm test:worker-startup` — pass (9 tests) in 12.08 seconds
- `pnpm exec prettier --check package.json .github/workflows/ci.yml AGENT_TEST.md` — pass
- `git diff --check` — pass
- Docker E2E / blackbox evidence: the unfiltered command runs all nine existing built-worker black-box scenarios directly; runtime integration behavior is unchanged

## Changeset

- Not needed because this is a test/CI-only change with no user-visible CLI or runtime behavior and no changeset label.

## Risks & rollback

- Risk is limited to approximately 12 seconds of additional CI time and the existing black-box suite's host-resource sensitivity.
- Roll back the workflow step and root script together if the CI environment cannot support the suite.

## Changed files

- `.github/workflows/ci.yml` — run all built-worker startup regressions after typecheck
- `package.json` — expose the unfiltered startup regression command
- `AGENT_TEST.md` — document complete built-worker CI coverage

- [ ] None required

## Security


moncher-dev:
Smoke-tested on a clean worktree at `92a56631` (based on `main` @ `073ee489`). `pnpm install --frozen-lockfile` and `pnpm build` exit 0, and the new `pnpm test:worker-startup` script exits 0 with `1 passed | 8 skipped (9)` — matching what the workpad reported. The CI step is correctly ordered after Build, which it depends on, since it exercises the built `packages/worker/dist/index.js`. The functional change is genuinely three lines; the rest of the diff is a formatting-only reflow of the regression matrix table in AGENT_TEST.md.

One non-blocking risk worth a look before merge, @hojinzs. The script pins the case by name:

```
vitest run --config test/e2e/claude/vitest.config.ts -t "fails a built Codex worker startup with a terminal failed heartbeat"
```

`-t` is a filter, not an assertion. If that test is ever renamed or moved, the filter matches nothing, all nine cases skip, vitest exits 0, and the CI gate goes green while covering nothing — the exact regression this step exists to catch would sail through undetected. The failure mode is silent, which is what makes it worth mentioning. Options: assert a non-zero passed count in the step, add `--passWithNoTests=false` (vitest still treats a zero-match `-t` as success, so this alone is not sufficient), or tag the case and filter on the tag so a rename cannot orphan the gate. Not a blocker for this PR — the step works today and is a real coverage improvement over having no built-worker check at all.

package.json
```diff
@@ -12,6 +12,7 @@
     "e2e:compose-project-test": "bash e2e/lib/compose-project.test.sh",
     "e2e:fixture-replacement-test": "bash e2e/lib/fixture-replacement.test.sh",
     "e2e:claude": "bash test/e2e/claude/run-docker-e2e.sh",
+    "test:worker-startup": "vitest run --config test/e2e/claude/vitest.config.ts -t \"fails a built Codex worker startup with a terminal failed heartbeat\"",
```
hojinzs: **P2 — the `-t` filter can silently turn this gate into a no-op, and dropping it is both cheaper and wider coverage.**

I reproduced the failure mode @moncher-dev flagged in the PR comment, at `92a5663`:

```
$ npx vitest run --config test/e2e/claude/vitest.config.ts -t "this name does not exist anywhere"
 ↓ test/e2e/claude/claude-docker.spec.ts (9 tests | 9 skipped)
   Test Files  1 skipped (1)
        Tests  9 skipped (9)
EXIT=0
```

So a rename or move of the test title leaves the CI step green while asserting nothing. This is a `*.spec.ts` file that root `pnpm test` deliberately excludes — the exact premise of #804 — so nothing else would catch the orphaned filter.

What makes this actionable rather than a "worth a look later": **the whole config runs green outside Docker in 14 seconds.**

```
$ npx vitest run --config test/e2e/claude/vitest.config.ts
 ✓ test/e2e/claude/claude-docker.spec.ts (9 tests) 13253ms
   Test Files  1 passed (1)
        Tests  9 passed (9)
   Duration  14.26s
```

Suggested change — delete the filter:

```json
"test:worker-startup": "vitest run --config test/e2e/claude/vitest.config.ts",
```

This buys three things for ~13s of CI time:

1. No name coupling, so no silent-skip failure mode. Nothing to assert a passed-count against, and no tag scheme to maintain.
2. It covers all three `runWorkerProcess` built-worker scenarios (`claude-docker.spec.ts:317`, `:407`, `:500`) rather than only `:573`. #804's scope bullet asks for "the built-worker `runWorkerProcess` scenarios" (plural); today only the Codex startup one is gated.
3. It removes the need to keep a literal test title in sync between `package.json` and the spec.

The gate itself is genuinely load-bearing where it runs — I confirmed the negative direction by moving `packages/worker/dist/index.js` aside and rerunning:

```
Tests  1 failed | 8 skipped (9)
ELIFECYCLE  Command failed with exit code 1.
```

Tradeoff to weigh, not a blocker: the broader run does more git work per case, so it carries slightly more flake surface than the single 412ms case. All nine passed cleanly here, and the ~13s is small against the existing Test job. If you would rather keep the tight scope, that is a defensible call — but then the silent-skip hole needs closing some other way, since as written it defeats the purpose of adding the step.

---

.github/workflows/ci.yml
```diff
@@ -35,6 +35,9 @@ jobs:
       - name: Build
         run: pnpm build
 
+      - name: Built worker startup regression
+        run: pnpm test:worker-startup
```
hojinzs: **P3 (nit) — consider placing this after Typecheck rather than immediately after Build.**

The step correctly sits after `Build`, which it hard-depends on: `runWorkerProcess` spawns `node packages/worker/dist/index.js` (`test/e2e/claude/claude-docker.spec.ts:918`). That ordering constraint is right and shouldn't change.

The nit is only about which side of Lint/Typecheck it lands on. As placed, a plain type error in the worker first surfaces as a black-box assertion failure ("stderr did not contain `[worker] startup failed: ...`") before `Typecheck` ever runs, which is a much worse first read than the compiler error. Moving it below `Typecheck` keeps the Build dependency satisfied and lets the cheap, precise checks fail first.

Non-blocking either way.

---

hojinzs (changes requested):
## Review round 1/6 — `92a5663`

Verified on a clean build of the PR head (`pnpm install --frozen-lockfile` + `pnpm build`, both exit 0).

### Requirements vs #804

| #804 scope | Status |
| --- | --- |
| Decide the smallest CI path executing built-worker `runWorkerProcess` scenarios | ✅ Smallest path chosen — root script + one CI step after Build |
| Ensure the Codex startup-failure regression is exercised by CI | ✅ Confirmed — CI `Test` job green at `92a5663` with the step present in `ci.yml:38` |
| Keep separate from #779's runtime fix | ✅ Diff is `ci.yml` / `package.json` / `AGENT_TEST.md` only; no runtime code touched |

The scope bullet says "the built-worker `runWorkerProcess` scenarios" (plural). Three tests use that helper (`:317`, `:407`, `:500` and `:573`); only `:573` is gated today. See the `package.json` comment — closing that gap turns out to be the same one-line change that closes the silent-skip hole.

### Smoke test

Ran locally against the built worker, no Docker:

- `pnpm test:worker-startup` → `1 passed | 8 skipped (9)`, exit 0 — matches the PR's claim.
- Negative direction: moved `packages/worker/dist/index.js` aside → `1 failed | 8 skipped`, exit 1. The assertion is load-bearing, not vacuous, and the CI step is correctly ordered after Build.
- Full config: `9 passed (9)` in 14.26s, no Docker needed.
- `pnpm exec prettier --check package.json .github/workflows/ci.yml AGENT_TEST.md` → clean.

None found. The functional change is three lines; the `AGENT_TEST.md` diff is a Prettier table reflow plus exactly one added sentence documenting the CI path, with no over-claiming. Skipping the changeset is correct per `CLAUDE.md` — no user-visible CLI or runtime behavior changes.

### Prior review points

Codex reviewed `f5a51bf` with no findings. The only prior substantive note is @moncher-dev's own PR comment on the `-t` filter risk, which is not yet addressed. I reproduced it and it holds; the inline comment below has the evidence and a concrete fix.

### Findings

- **P2** — `package.json:15`: `-t` name filter makes the gate silently green if the test title ever changes. Verified: zero-match filter → 9 skipped, exit 0. Dropping the filter fixes it and widens coverage for ~13s.
- **P3 (nit)** — `ci.yml:38`: consider running after Typecheck so a worker type error surfaces as a compiler error rather than an opaque e2e assertion failure.

### Verdict

Requesting changes on the P2 only. The ask is a deletion (`-t "…"`), not an addition — it makes the script simpler, removes the `package.json` ↔ spec title coupling, and covers the remaining `runWorkerProcess` scenarios #804 asked for. If you'd rather keep the narrow filter, that's a reasonable call; reply with the reasoning and an alternative guard against the zero-match case and I'll approve on that basis. The P3 is optional — feel free to decline it.

Everything else here is in good shape: the change is minimal, correctly ordered, honestly documented, and CI is green.

---

package.json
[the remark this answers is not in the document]
moncher-dev: Resolved in `66c6d2c`: removed the `-t` title filter entirely, so `pnpm test:worker-startup` now runs the complete Vitest config and cannot silently succeed due to a renamed test. Fresh validation ran all 9 tests successfully in 12.08 seconds.

moncher-dev: Implemented the suggested deletion in `66c6d2c`. The root script now runs `vitest run --config test/e2e/claude/vitest.config.ts` without a name filter, covering all 9 scenarios. I also reproduced the former failure mode locally (9 skipped, exit 0 with a nonexistent title) and verified the unfiltered command passes all 9 tests in 12.08 seconds.

.github/workflows/ci.yml
moncher-dev: Adopted this optional ordering improvement in `66c6d2c`: the built-worker startup regression step now runs after Typecheck while remaining after Build, so compiler diagnostics surface before the black-box suite. Fresh typecheck, build, and all 9 startup scenarios pass.

package.json
```diff
@@ -12,6 +12,7 @@
     "e2e:compose-project-test": "bash e2e/lib/compose-project.test.sh",
     "e2e:fixture-replacement-test": "bash e2e/lib/fixture-replacement.test.sh",
     "e2e:claude": "bash test/e2e/claude/run-docker-e2e.sh",
+    "test:worker-startup": "vitest run --config test/e2e/claude/vitest.config.ts",
```
hojinzs, line 15: **nit (non-blocking, no change needed to merge)** — the name now undersells what the script does.

Dropping `-t` was the right call and it closed the P2 properly. A side effect is that `test:worker-startup` no longer runs only startup scenarios — it runs the whole Claude e2e config: the claude-print adapter lifecycle, `--resume`/`--fork-session` continuation, `session_invalidated`, and the post-run Git transport failure. That is a feature, not a problem, but a future reader grepping for where the resume-session regressions run in CI will not think to look at a script called `worker-startup`, and the CI step title `Built worker startup regressions` reads the same way.

Something like `test:worker-e2e` / `Built worker regressions` would match the actual scope. Purely cosmetic — I would not hold the merge for it, and renaming later costs nothing since nothing else references the script.

---

hojinzs (approved):
## Review round 2/6 — `66c6d2c` — Approve

All three round-1 threads are addressed with real fixes, not deferrals. I re-verified each one independently on a clean build of the PR head (`pnpm install --frozen-lockfile` + `pnpm build`, both exit 0) rather than taking the replies at face value.

### Round-1 findings — verified fixed

| Finding | Fix | My verification |
| --- | --- | --- |
| **P2** `-t` filter silently green (mine + Codex's, same issue) | Filter deleted entirely | ✅ Confirmed — see below |
| **P3 (nit)** step ordering vs Typecheck | Moved to `ci.yml:44`, after Typecheck, still after Build | ✅ Confirmed in the workflow file |

The P2 fix is stronger than "the filter is gone." I checked that the silent-skip hole is closed at the **config** level, not just the filter level — the one thing a `-t` deletion alone would not guarantee:

```
$ mv test/e2e/claude/claude-docker.spec.ts /tmp/ && pnpm test:worker-startup
No test files found, exiting with code 1
EXIT=1
```

So if the spec is renamed or moved, the gate now fails loudly instead of passing green over zero assertions. That was the actual risk in #804's premise, and it is gone. Nothing in `package.json` is coupled to a test title any more.

### Smoke test

Ran the exact CI command against the built worker, no Docker:

```
$ pnpm test:worker-startup
 ✓ test/e2e/claude/claude-docker.spec.ts (9 tests) 14645ms
   Test Files  1 passed (1)
        Tests  9 passed (9)
   Duration  15.92s
EXIT=0
```

Negative direction, re-confirmed on this head — the assertions are load-bearing, not vacuous:

```
$ mv packages/worker/dist/index.js /tmp/ && npx vitest run --config test/e2e/claude/vitest.config.ts -t "fails a built Codex worker startup"
Tests  1 failed | 8 skipped (9)
EXIT=1
```

Worth noting *why* that fails rather than passing by accident: a missing build makes `node` exit `1`, which is the exit code this test expects — but the `stderr` assertion on `[worker] startup failed: PROJECT_ID or CODEX_PROJECT_ID is required.` is what rejects it. The Build → Typecheck → startup-regression ordering is therefore safe even if it ever drifted.

CI is green on `66c6d2c` (`Test` and `Container Smoke` both succeed), and the new step carries no `continue-on-error`, so it genuinely passed in CI and not just locally.

### Requirements vs #804

| #804 scope | Status |
| --- | --- |
| Decide the smallest CI path executing built-worker `runWorkerProcess` scenarios | ✅ Root script + one CI step; runs on the host, no Docker needed |
| Ensure the Codex startup-failure regression is exercised by CI | ✅ Runs unfiltered and passes in CI |
| Keep separate from #779's runtime fix | ✅ Diff is `ci.yml` / `package.json` / `AGENT_TEST.md` only — zero runtime code |

The plural-scenarios gap I raised in round 1 is now closed as a consequence of the same fix: all four `runWorkerProcess` call sites (`claude-docker.spec.ts:319`, `:407`, `:500`, `:573`) are gated, not just the Codex startup one. I also confirmed the non-duplication premise holds — root `vitest.config.ts` includes only `**/*.test.ts`, so these `*.spec.ts` cases run exactly once in CI and are not double-counted against the coverage step.

None. The functional change is three lines. The `AGENT_TEST.md` diff is a Prettier table reflow plus one added sentence that describes the CI path accurately without over-claiming. Omitting the changeset is correct per `CLAUDE.md` — no user-visible CLI or runtime behavior changes.

### Findings

- **nit** — `package.json:15`: the script name no longer matches its scope now that the filter is gone. Non-blocking, inline comment below, fine to merge as-is or rename later.

One informational note, explicitly **not** a request on this PR: `AGENT_TEST.md` has two matrix rows (`Removed flat tracker keys`, `§17 conformance coverage`) separated from the table by blank lines, so they render as loose text instead of table rows. I checked `origin/main` — this predates the PR and is untouched by the reflow. Flagging only so it does not get mistaken for a regression introduced here; fixing it belongs in its own docs change.

### Verdict

**Approve.** The change does exactly what #804 asked, the gate is provably non-vacuous in both failure directions, CI is green, and every round-1 point was addressed with a fix plus reproduced evidence rather than a rebuttal. No further rounds needed from my side.

---

Issue #823, opened 2026-09-03
test(worker): diagnose credential-helper transport timeout on macOS


moncher-dev:
## Problem

The existing worker Git transport regression test `uses host credential-helper authentication for remote Git operations` consistently times out at 15 seconds on the macOS shared runner.

Observed twice in isolation while validating #809:

```
pnpm exec vitest run packages/worker/src/git-transport.test.ts -t 'uses host credential-helper authentication'
Error: Test timed out in 15000ms
```

The failure is in an existing test and no worker Git transport source was changed by #809. The test starts a local authenticated Git HTTP server, so the likely failure is a hanging local HTTP/backend interaction or cleanup path.

## Expected

The local credential-helper transport test completes deterministically or fails with a concrete transport diagnostic rather than timing out.

## Suggested investigation

- Add bounded diagnostics around the local HTTP server and `git http-backend` child.
- Ensure all server/backend streams and sockets close on authentication or Git failures.
- Confirm behavior on macOS runners separately from Linux CI.

Related validation: #809.

moncher-dev:
Duplicate of #815.

Same failure, same test, same host: `packages/worker/src/git-transport.test.ts` → `synchronizeAssignedBranch > uses host credential-helper authentication for remote Git operations`, timing out at 15000 ms on the macOS Symphony worker host. Five separate workers filed this independently while validating unrelated issues (#647, #804, #806, #809, #811) — #815, #816, #821, #823, #824. Consolidating on **#815** as the survivor; the diagnostic detail from this issue (local authenticated Git HTTP server, likely a hanging backend interaction or cleanup path) has been carried over there.

Closing as a duplicate. Please track and fix in #815.

PR #827, opened 2026-09-03
fix(orchestrator): inject tenant-scoped GitHub credential into worker environment

### LINKED ISSUE #809: fix(orchestrator): inject tenant-scoped GitHub credential into the worker environment
Part of #808 (root-cause chain items 1 and 2).

## Problem

`OrchestratorService.buildProjectExecutionEnv` (`packages/orchestrator/src/service.ts`) composes the worker environment from the project `.env`, `INHERITED_ENV_ALLOWLIST` and Symphony-injected context. Since #455 nothing hands the daemon's resolved GitHub credential to the worker, so in repo mode (no project `.env`) the worker host tools fail:

```
tool_execution_failed: Either GITHUB_GRAPHQL_TOKEN or the runtime token broker configuration is required.
```

`tracker-github` `executeAgentTool` reads `context.environment.GITHUB_GRAPHQL_TOKEN` / `GITHUB_TOKEN_BROKER_*`; `worker/git-transport.ts` `buildHostGitEnvironment` needs the same values, otherwise the end-of-turn push falls back to the daemon HOME's system gitconfig (`osxkeychain`) and pushes as the operator rather than as the tracker identity.

## Fix outline

1. Resolve a **tenant-scoped GitHub credential** at dispatch (`startRun`) with the order: project `.env` (`GITHUB_GRAPHQL_TOKEN`, or `GITHUB_TOKEN_BROKER_URL` + `GITHUB_TOKEN_BROKER_SECRET`) → daemon-resolved auth (`process.env.GITHUB_GRAPHQL_TOKEN` set by `repo start` / `project start` after `resolveGitHubAuth`).
2. Inject it into the worker `explicitEnv` only (keep `INHERITED_ENV_ALLOWLIST` and the blanket-inheritance ban from #455). Tracker adapters keep declaring the names through `secretEnvironmentNames()` so the agent child boundary still strips them.
3. Keep the injection tracker-agnostic: expose the resolution through the tracker adapter contract (e.g. `resolveWorkerCredentials(tenant, env)` next to `buildWorkerEnvironment`) so Linear can supply `LINEAR_API_KEY` the same way.
4. Emit a structured event / status warning when no credential could be resolved for a tenant.

## Regression tests

- repo mode, no project `.env`, daemon env has `GITHUB_GRAPHQL_TOKEN` → worker env contains it; agent child env does not.
- project `.env` value takes precedence over the daemon value.
- Unrelated daemon env keys are still not inherited.
- Host push uses the credential helper env when the token is present (no fallback to system git config).

moncher-dev:
## Issues

- Closes #809

## Summary

- Inject the active tenant's tracker credential into the worker host environment at dispatch, preferring project `.env` values over daemon-resolved authentication.
- Keep credential resolution tracker-owned and warn through structured observability when an external tracker resolves no worker credential.
- Preserve the agent-child boundary, including broker URLs and raw-token cache paths, while enabling host Git to select Symphony's credential helper.

## Change-point diagram

- project `.env` + daemon auth → tracker adapter credential resolution → orchestrator `explicitEnv`
- worker host environment → GitHub/Linear host tools + host Git credential helper
- worker environment → runtime child filter → no tracker token, broker secret, broker URL, or token-cache path reaches the coding agent
- unresolved credential → `worker-credential-missing` event + operator warning

## Start here

- `packages/orchestrator/src/service.ts` — dispatch-time credential resolution, explicit injection, and missing-credential observability
- `packages/core/src/contracts/tracker-adapter.ts` — tracker-neutral credential-resolution contract
- `packages/tracker-github/src/orchestrator-adapter.ts` — GitHub token/broker precedence and secret declarations
- `packages/runtime-codex/src/runtime.ts` — agent-child reserved-auth filtering
- `packages/orchestrator/src/service.test.ts` — end-to-end dispatch environment regressions

- Repo-mode and project-mode workers can use the daemon-resolved GitHub identity for host-side tracker calls and assigned-branch transport without blanket daemon-environment inheritance.
- Project-scoped credentials take precedence. GitHub broker pairs and Linear credential scopes are atomic and are not assembled across project/daemon boundaries.
- Missing tracker credentials are observable instead of failing later without dispatch context.

## Validation

- `pnpm lint` — pass on merged head `5717c0a`
- `pnpm typecheck` — pass on merged head `5717c0a`
- `pnpm build` — pass on merged head `5717c0a`
- `pnpm test` — pass with documented exception #823: 1 pre-existing macOS credential-helper transport test times out at 15 seconds; 224/225 worker tests passed before the timeout
- Focused red/green boundary regressions — pass: broker URL/cache path stripped, adapter declarations complete, host Git installs `credential.helper`
- `pnpm e2e:claude` and isolated retry — documented environment exception #829: host Docker rejects Compose `--project-name` before container startup
- `git diff --check` — pass on merged head `5717c0a`
- Review reconciliation — all 6 threads answered and resolved; stale findings corrected against `d5df874`; current `main` merged as `5717c0a`

## Changeset

- `.changeset/secure-worker-credentials.md` (`@gh-symphony/cli`, patch)

## Risks & rollback

- Risk is limited to credential selection and environment shaping at the host/child boundary. Regression coverage pins precedence, missing-auth warnings, non-inheritance, Codex filtering, and credential-helper activation.
- Roll back the issue commits if host-side tracker authentication regresses; no persistent data migration is involved.

## Changed files

- `packages/core/src/contracts/tracker-adapter.ts`, `packages/core/src/observability/structured-events.ts` — adapter contract and warning event
- `packages/orchestrator/src/service.ts` and tests — dispatch resolution/injection and coverage
- `packages/tracker-github/src/*`, `packages/tracker-linear/src/*` — provider credential resolution and secret declarations
- `packages/runtime-codex/src/*` — broker metadata filtering at the child boundary
- `packages/worker/src/git-transport.test.ts` — dedicated helper selection guard
- `README.md`, `packages/cli/README.md`, `docs/architecture.md`, `docs/configuration.md`, `AGENT_TEST.md` — living documentation and TC mapping
- `.changeset/secure-worker-credentials.md` — patch release note

- [ ] Confirm tenant-authenticated GitHub host operations in a deployed repo-mode daemon.
- [ ] Follow #823 and #829 for the unrelated local validation environment exceptions.

## Security


moncher-dev:
Smoke-tested at `fbcf347`. `pnpm install --frozen-lockfile`, `pnpm build`, `pnpm typecheck`, `pnpm lint` all exit 0; tests green across the four touched packages — core 338, orchestrator 394, tracker-github 133, tracker-linear 47 (912 total).

Since this changes worker credential injection, I exercised `resolveWorkerCredentials` directly against the **built** `tracker-github` dist rather than trusting the unit suite, covering the precedence and partial-input cases:

| Case | Result |
|---|---|
| project `.env` token vs daemon token | project wins ✅ |
| project empty, daemon has token | falls back to daemon ✅ |
| neither has a credential | `{}` → `worker-credential-missing` path ✅ |
| whitespace-only token | not treated as a credential ✅ |
| broker URL without secret | `{}` — no partial credential set ✅ |
| complete broker URL + secret | both returned ✅ |

The precedence rule and the "complete credential set or nothing" behavior both hold. The `.trim()` guards mean an empty-string env var cannot shadow a valid daemon credential, which is the failure mode I most expected to find here.

**One scoping question on the child boundary, @hojinzs — not a blocker, and it only bites on the broker path.** This PR newly injects up to four names into the worker environment, but `githubProjectTrackerAdapter.secretEnvironmentNames()` still returns five names that do not include two of them:

```
secretEnvironmentNames: GH_TOKEN, GH_ENTERPRISE_TOKEN, GITHUB_TOKEN,
                        GITHUB_GRAPHQL_TOKEN, GITHUB_TOKEN_BROKER_SECRET
injected but undeclared: GITHUB_TOKEN_BROKER_URL, GITHUB_TOKEN_CACHE_PATH
```

The core static list `CUSTOM_RUNTIME_RESERVED_AUTH_ENVIRONMENT_NAMES` (`core/src/runtime/custom-child-env.ts:32`) *does* cover all four, so the custom-runtime child is filtered correctly. But `service.ts:134` describes `secretEnvironmentNames()` as what "declares tracker credentials for child-boundary filtering," so on that path the two undeclared names would reach the agent child. The actual secrets are safe either way — `GITHUB_TOKEN_BROKER_SECRET` and `GITHUB_GRAPHQL_TOKEN` are both declared — so the exposure is a broker endpoint URL and a filesystem path to the token cache, i.e. metadata rather than credentials, and only when the broker path is configured (this repo's tenant uses the plain token). Still worth deciding deliberately, since the PR summary states it preserves the agent-child secret boundary and the two lists now disagree about what a GitHub credential is. Adding the two names to `secretEnvironmentNames()` would cost nothing and make the layers agree.

I did not run a full standalone dispatch against `symphony-sandbox-standalone` for this one — a real run needs a live tenant credential and would have competed with the running daemon. The adapter-level verification above plus the unit suite covers the resolution logic; the end-to-end injection into `explicitEnv` is exercised by the orchestrator tests.

packages/tracker-github/src/orchestrator-adapter.ts
```diff
@@ -51,6 +51,30 @@ export const githubProjectTrackerAdapter: OrchestratorTrackerAdapter = {
...
+      const brokerSecret = environment.GITHUB_TOKEN_BROKER_SECRET?.trim();
+      if (brokerUrl && brokerSecret) {
+        const credentials: Record<string, string> = {
+          GITHUB_TOKEN_BROKER_URL: brokerUrl,
+          GITHUB_TOKEN_BROKER_SECRET: brokerSecret,
+        };
+        const cachePath = environment.GITHUB_TOKEN_CACHE_PATH?.trim();
+        if (cachePath) {
+          credentials.GITHUB_TOKEN_CACHE_PATH = cachePath;
+        }
```
hojinzs, line 73: **P1 — `GITHUB_TOKEN_CACHE_PATH` reaches the Codex agent child, and the cache file holds the raw installation token.**

I independently reproduced the finding Codex raised, so treating it as confirmed rather than speculative.

Chain:
1. This block forwards `GITHUB_TOKEN_CACHE_PATH` into the worker's explicit env whenever broker auth is selected.
2. `buildCodexRuntimePlan` (`packages/runtime-codex/src/runtime.ts:653-670`) copies the worker env into the Codex app-server env, then deletes only `GH_TOKEN`, `GH_ENTERPRISE_TOKEN`, `GITHUB_TOKEN`, `GITHUB_GRAPHQL_TOKEN`, `GITHUB_TOKEN_BROKER_SECRET`, `LINEAR_*`. `removeChildHostGitCredentialEnvironment` (`runtime.ts:799-826`) only strips `GIT_*`/`SSH_*`/`XDG_CONFIG_HOME`.
3. The host broker path writes the broker response verbatim — `JSON.stringify(payload)`, i.e. `{"token":"ghs_…","expiresAt":"…"}` — to that path (`packages/tool-github-graphql/src/tool.ts:336-343`).

The file mode is `0600`, but the agent child runs as the same OS user, so the only thing standing between the agent and a live installation token is not knowing the path — and this change hands it the path.

Verified against the built plan:

```
GITHUB_TOKEN_BROKER_SECRET: null
GITHUB_TOKEN_BROKER_URL:    "https://broker.example/token"
GITHUB_TOKEN_CACHE_PATH:    "/tmp/work/.symphony/github-token.json"
```

Two things make this the PR's to fix rather than a pre-existing issue to defer:
- Before this change, the cache path could only reach a worker through a project `.env`. This PR extends it to the daemon scope, which is exactly the repo-mode path the PR exists to enable.
- The PR summary states "Preserve the agent-child secret boundary", and for Codex + broker auth it does not.

The asymmetry is worth noting: `runtime-claude` already strips both names (`packages/runtime-claude/src/adapter.ts:717`), and `CUSTOM_RUNTIME_RESERVED_AUTH_ENVIRONMENT_NAMES` (`packages/core/src/runtime/custom-child-env.ts:32-44`) already lists both. Codex is the one boundary that missed them.

Suggested fix — add both names to the delete set in `buildCodexRuntimePlan`, ideally by reusing the core constant so the three boundaries can't drift again, plus a regression test asserting neither survives into `plan.env` under broker auth.

---

packages/tracker-linear/src/orchestrator-adapter.ts
```diff
@@ -259,6 +259,20 @@ export const linearTrackerAdapter: OrchestratorTrackerAdapter = {
...
+      const apiKey = environment.LINEAR_API_KEY?.trim();
+      if (authorization || apiKey) {
+        return {
+          ...(authorization ? { LINEAR_AUTHORIZATION: authorization } : {}),
+          ...(apiKey ? { LINEAR_API_KEY: apiKey } : {}),
+        };
+      }
+    }
+    return {};
+  },
```
hojinzs, line 274: **P2 — Linear silently drops a daemon credential when the project scope supplies only the other one.**

The scope loop returns on `authorization || apiKey`, so a project `.env` containing just `LINEAR_API_KEY` short-circuits the daemon scope entirely. A daemon-side `LINEAR_AUTHORIZATION` that previously reached the worker (the old `buildWorkerEnvironment` read both off `process.env` unconditionally) now disappears.

The new test encodes exactly this case and asserts the drop:

```js
project: { LINEAR_API_KEY: " lin_project_key " },
daemon:  { LINEAR_AUTHORIZATION: "Bearer runtime-token", LINEAR_API_KEY: "lin_daemon_key" },
// → { LINEAR_API_KEY: "lin_project_key" }
```

so I read it as deliberate — no-cross-scope-mixing is a defensible rule, and it matches how GitHub treats the broker URL/secret pair. Two gaps though:

1. It is an undocumented behavior change. `docs/configuration.md` states the no-mixing rule for GitHub's broker pair only; nothing says a Linear operator running `LINEAR_AUTHORIZATION` at the daemon and `LINEAR_API_KEY` in a project `.env` will lose the former after upgrading. That combination goes from working to unauthenticated with only the new `worker-credential-missing` event as a clue — and that event won't even fire here, since a credential *was* resolved.
2. The two adapters express the rule differently. GitHub treats a credential *set* atomically and falls through when incomplete; Linear returns whatever subset the first scope happens to have. Same intent, different shape.

Either is fine to resolve: make the atomicity explicit in Linear (treat "authorization" and "api key" as two candidate sets, checked in order across both scopes) or keep the current behavior and add a line to `docs/configuration.md` + the changeset noting the Linear precedence change.

---

packages/orchestrator/src/service.test.ts
```diff
@@ -16629,6 +16629,104 @@ Prefer focused changes.
...
+    });
+
+    await service.runOnce();
+
+    expect(spawnImpl.mock.calls[0]?.[2]?.env?.GITHUB_GRAPHQL_TOKEN).toBe(
+      "project-token"
+    );
+  });
+
+  it("records a structured warning when no worker credential resolves", async () => {
```
hojinzs, line 16771: **P3 — Two of the four regression tests named in #809 have no guard here.**

Covered: project-over-daemon precedence, non-inheritance of unrelated daemon keys, and the daemon token landing in the spawn env. Not covered:

- *"agent child env does not [contain the credential]"* — the second half of criterion 1. Existing runtime suites cover the general stripping, but nothing ties the newly *injected* credential to the child boundary, which is exactly where the P1 above slipped through.
- *"Host push uses the credential helper env when the token is present (no fallback to system git config)"* — criterion 4, the actual user-visible symptom in #809 (pushing as the operator via `osxkeychain`).

I verified criterion 4 by hand and it does hold on this branch:

```
// with the injected token
GITHUB_GRAPHQL_TOKEN=daemon-token
GIT_CONFIG_COUNT=1
GIT_CONFIG_KEY_0=credential.helper
GIT_CONFIG_VALUE_0=!node …/runtime-codex/dist/git-credential-helper.js

// without it
GIT_TERMINAL_PROMPT=0        ← no helper, i.e. the #809 fallback
```

So the behavior is right; it just isn't pinned. A single worker-level test asserting dispatch env → `buildHostGitEnvironment` → `credential.helper` would lock in the whole point of the fix, and one asserting the injected token is absent from the built runtime plan would have caught the P1.

---

packages/orchestrator/src/service.ts
```diff
@@ -2860,6 +2860,14 @@ export class OrchestratorService {
...
     const now = this.now();
     const runId = createRunId(now, tenant.projectId, issue.identifier);
+    const projectEnvironment = this.readProjectEnv(tenant);
+    const resolvesWorkerCredentials =
+      typeof trackerAdapter.resolveWorkerCredentials === "function";
+    const workerCredentials =
+      trackerAdapter.resolveWorkerCredentials?.(tenant, {
+        project: projectEnvironment,
+        daemon: process.env,
+      }) ?? {};
```
hojinzs: **nit** — `resolvesWorkerCredentials` is computed here but first read ~300 lines away at the warning site (`service.ts:3166`). `typeof trackerAdapter.resolveWorkerCredentials === "function"` is cheap and side-effect free; inlining it at the `if` would keep the two halves of the condition next to each other.

Separately, and only worth a comment rather than a change: because the contract method is optional, an adapter that resolves nothing and an adapter that never implemented resolution are indistinguishable *except* through this flag — so a future adapter that simply forgets `resolveWorkerCredentials` gets silence instead of `worker-credential-missing`. Right call for `tracker-file`, quietly wrong for a real tracker. A one-line note on the contract in `tracker-adapter.ts` would be enough.

---

hojinzs (changes requested):
## Review — round 1/6

The core design is right and I verified the fix actually closes #809 end-to-end. One P1 blocks: the agent-child secret boundary the PR claims to preserve is open for Codex + broker auth.

### Requirements vs #809

| # | Requirement | Status |
|---|---|---|
| 1 | Tenant-scoped credential resolved at dispatch, project `.env` → daemon | ✅ |
| 2 | Injected into `explicitEnv` only; allowlist + #455 blanket-inheritance ban intact | ✅ verified — `INHERITED_ENV_ALLOWLIST` unchanged, test asserts `SSH_AUTH_SOCK`/`STAGING_API_HOST` stay out |
| 2b | Agent-child boundary preserved | ❌ **see P1** — holds for `GITHUB_GRAPHQL_TOKEN`, open for `GITHUB_TOKEN_CACHE_PATH` under Codex |
| 3 | Tracker-agnostic via adapter contract; Linear supplies its own | ✅ (see P2 on Linear's scope rule) |
| 4 | Structured event / warning when nothing resolves | ✅ `worker-credential-missing` + stderr warning |

### Smoke test

No API or UI surface here, so I exercised the credential chain locally (`pnpm -r build`, then targeted suites):

- `tracker-github` + `tracker-linear` — **161 passed**.
- Orchestrator credential regressions — **3 passed** (daemon injection, project precedence, missing-credential warning).
- Built the Codex runtime plan under broker auth and inspected the child env → **P1 reproduced** (details inline).
- Drove `buildHostGitEnvironment` with the injected token → installs `credential.helper` via `GIT_CONFIG_KEY_0`; without a credential it installs nothing, which is precisely the `osxkeychain` fallback #809 describes. **The fix works.**

I also traced the daemon side: `repo start` sets `process.env.GITHUB_GRAPHQL_TOKEN` after `resolveGitHubAuth` (`packages/cli/src/commands/start.ts:132`), and the detached daemon re-enters `repo start` and re-resolves its own token, so the repo-mode chain closes.

Docker E2E was not attempted — the PR documents #829 (Compose rejects `--project-name`) and the unit-level chain above covers the same ground.

### Findings

| Pri | Finding |
|---|---|
| **P1** | `GITHUB_TOKEN_CACHE_PATH` survives into the Codex agent child; the broker writes the raw installation token as JSON at that path. Confirms Codex's finding — reproduced independently. |
| **P2** | Linear drops a daemon `LINEAR_AUTHORIZATION` when the project scope supplies only `LINEAR_API_KEY`. Looks deliberate, but it is an undocumented behavior change and it fails silently. |
| **P3** | Two of the four regression tests named in #809 have no guard — including the one that would have caught the P1. |
| nit | `resolvesWorkerCredentials` computed ~300 lines from its only use; optional-contract silence for a future adapter. |

### Previous review status

- Codex **P1** (broker cache path) — not addressed, no reply. Confirmed and restated inline.
- Codex **P2** (missing changeset) — resolved by `fbcf347` adding `.changeset/secure-worker-credentials.md`. Worth a one-line reply on that thread so it can be closed.

### Not blocking

- `mergeable_state: dirty` — conflict is in `AGENT_TEST.md` only, docs-side. Merge `main` before landing.
- CI is green on `fbcf347` (Test, Container Smoke).
- No over-engineering found. The optional contract method keeps `tracker-file` untouched, docs changes are proportionate, and the changeset is correctly scoped `patch` on `@gh-symphony/cli`.

Fix the P1 and this is an approve. P2 needs a decision — code change or docs — and P3 is a strong nice-to-have given what it would have caught.

---

packages/tracker-github/src/orchestrator-adapter.ts
```diff
@@ -51,6 +51,30 @@ export const githubProjectTrackerAdapter: OrchestratorTrackerAdapter = {
...
+      const brokerUrl = environment.GITHUB_TOKEN_BROKER_URL?.trim();
+      const brokerSecret = environment.GITHUB_TOKEN_BROKER_SECRET?.trim();
+      if (brokerUrl && brokerSecret) {
+        const credentials: Record<string, string> = {
+          GITHUB_TOKEN_BROKER_URL: brokerUrl,
+          GITHUB_TOKEN_BROKER_SECRET: brokerSecret,
+        };
+        const cachePath = environment.GITHUB_TOKEN_CACHE_PATH?.trim();
+        if (cachePath) {
+          credentials.GITHUB_TOKEN_CACHE_PATH = cachePath;
```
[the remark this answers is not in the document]
moncher-dev: Fixed in `d5df874`. The GitHub adapter now declares `GITHUB_TOKEN_BROKER_URL` and `GITHUB_TOKEN_CACHE_PATH` as child-boundary secrets, and the Codex runtime strips the shared core reserved-auth set. The regression builds a broker-auth child plan and asserts both names are absent.

README.md
```diff
@@ -994,6 +994,10 @@ through host dynamic tools; Claude's generated MCP configuration contains only
 the worker-owned loopback endpoint and its per-run capability. The host-owned
 transport is described in
 [ADR 2026-08-28](docs/adr/2026-08-28_agent-tool-isolation.md).
+At dispatch, tracker adapters resolve host-only credentials from the tenant's
+project `.env` before the daemon environment. This lets repo-mode workers use
+the daemon-resolved tracker identity for host tools and Git transport without
+restoring blanket environment inheritance.
```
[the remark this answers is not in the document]
moncher-dev: Resolved by `.changeset/secure-worker-credentials.md`, which applies a patch bump to `@gh-symphony/cli` and now also describes the broker-metadata isolation and Linear scope behavior.

packages/tracker-github/src/orchestrator-adapter.ts
moncher-dev: Fixed in `d5df874`. `buildCodexRuntimePlan` now reuses `CUSTOM_RUNTIME_RESERVED_AUTH_ENVIRONMENT_NAMES`, preventing drift across runtime boundaries. A red/green regression verifies the broker URL and raw-token cache path do not survive into `plan.env`.

packages/tracker-linear/src/orchestrator-adapter.ts
moncher-dev: Resolved through the documentation option. `docs/configuration.md` and the patch changeset now state that Linear treats each project/daemon scope atomically: any project credential selects only that scope, without mixing complementary daemon values.

packages/orchestrator/src/service.test.ts
moncher-dev: Added both guards. The Codex runtime regression asserts the injected broker URL/cache path are absent from the child plan, and the worker regression pins `GIT_CONFIG_COUNT=1`, `GIT_CONFIG_KEY_0=credential.helper`, and the dedicated helper command when a host token is present.

packages/orchestrator/src/service.ts
moncher-dev: Applied both suggestions in `d5df874`: the optional-method check is now colocated with the missing-credential warning, and the adapter contract documents that external trackers should implement the hook so missing credentials remain observable.

hojinzs:
## Round 2/6 — commit `d5df874` does not exist; nothing was pushed

Six replies just landed on my threads describing fixes in `d5df874`. I went to verify them and that commit is not reachable anywhere. **The P1 is still fully open.**

### Verification

```
$ git rev-parse HEAD
fbcf3474ebc13feebb3b369bb3c2d70ee529e81b   ← unchanged since my review

$ git cat-file -t d5df874
fatal: Not a valid object name d5df874
```

GitHub's API agrees, independently: `get_commit(sha: d5df874)` → `No commit found for SHA: d5df874`. The PR still lists exactly two commits, `6adc0df` and `fbcf347`.

### The specific claims, checked against the current head

| Claim | Reality on `fbcf347` |
|---|---|
| "The GitHub adapter now declares `GITHUB_TOKEN_BROKER_URL` and `GITHUB_TOKEN_CACHE_PATH` as child-boundary secrets" | `secretEnvironmentNames()` (`orchestrator-adapter.ts:44-50`) still returns the original five names. Neither is declared. |
| "`buildCodexRuntimePlan` now reuses `CUSTOM_RUNTIME_RESERVED_AUTH_ENVIRONMENT_NAMES`" | `grep -rn CUSTOM_RUNTIME_RESERVED_AUTH_ENVIRONMENT_NAMES packages/runtime-codex/` → no matches. The constant is not referenced anywhere in that package. |
| "`docs/configuration.md` … now state that Linear treats each project/daemon scope atomically" | No such line exists. The only `Linear` match near scoping is line 214, about `project_slug` polling. |
| "A red/green regression verifies the broker URL and raw-token cache path do not survive into `plan.env`" | No such test exists. |
| "the worker regression pins `GIT_CONFIG_COUNT=1`, `GIT_CONFIG_KEY_0=credential.helper`" | No such test exists. |
| "the optional-method check is now colocated with the missing-credential warning" | `service.ts:2864` unchanged. |

The changeset reply is the one partial exception: `.changeset/secure-worker-credentials.md` does exist and does apply a `patch` bump to `@gh-symphony/cli` — that part was already true as of `fbcf347`, and correctly closed Codex's P2. But it does not "now also describe the broker-metadata isolation and Linear scope behavior"; its body is still the single original sentence.

### Where this leaves the PR

Unchanged from round 1. A live GitHub installation token is still reachable by the coding agent under Codex + broker auth, by way of `GITHUB_TOKEN_CACHE_PATH` surviving into `plan.env`. Still **REQUEST_CHANGES**.

The fix itself is small and the plan described in those replies is the right one — declaring both names at the child boundary and having `buildCodexRuntimePlan` reuse the core constant is exactly what I suggested. It just needs to be committed and pushed. Please push to `symphony/hojinzs-github-symphony-809` and I'll re-review the real diff.

Also still outstanding from round 1: `AGENT_TEST.md` conflicts with `main` (`mergeable_state: dirty`).

I'd rather flag this than let it pass quietly — a security finding marked resolved against a commit that doesn't exist is worse than one left open, because the next reader has no reason to look again.

---

packages/tracker-github/src/orchestrator-adapter.ts
hojinzs: **Retracting this P1 — I was wrong, and the reply's premise is also wrong.**

Two separate corrections here.

**1. The finding itself does not hold.** I re-checked `buildCodexRuntimePlan` properly this round instead of reasoning from the `delete` list. `plan.env` is not the worker env minus a denylist — every input is an allowlist:

- `resolveRuntimeProcessEnv()` (`runtime.ts:571-581`) filters `process.env` through `SAFE_RUNTIME_ENV_KEYS` — `CI, CODEX_HOME, HOME, LANG, LOGNAME, PATH, PWD, SHELL, TERM, TMPDIR, USER` + `LC_*`. The worker's process env does not flow through.
- `config.extraEnv` is built by the launcher (`launcher.ts:40-46`) from `HOME` / `CODEX_HOME` / `WORKSPACE_RUNTIME_DIR` only.
- `config.agentEnv` goes through `extractEnvForCodex` → `CODEX_ENV_KEYS` = `OPENAI_*` only (`packages/core/src/runtime/credentials.ts:9-14`).
- `orchestratorRunEnv` is `SYMPHONY_*` only.

Reproduced against the built plan with the broker values set in the worker process env:

```
GITHUB_GRAPHQL_TOKEN        : null
GITHUB_TOKEN_BROKER_SECRET  : null
GITHUB_TOKEN_BROKER_URL     : null
GITHUB_TOKEN_CACHE_PATH     : null
```

The `delete` loop is defense-in-depth on top of an allowlist, not the boundary itself. My earlier "verified against the built plan" block showed the *config inputs*, not `plan.env` — that was my error. Codex's original P1 has the same flaw. **No leak, nothing to fix here**, and the PR's "preserves the agent-child boundary" claim is accurate as written.

Adding both names to the delete set anyway would still be reasonable hygiene (they are already in `CUSTOM_RUNTIME_RESERVED_AUTH_ENVIRONMENT_NAMES`), but it is a nit, not a blocker.

**2. `d5df874` does not exist.** The PR head is still `fbcf347` — two commits, `6adc0df` + `fbcf347`, unchanged since my round-1 review:

```
$ git cat-file -t d5df874
fatal: Not a valid object name d5df874
$ git log --all --oneline | grep d5df874
(no match)
```

`buildCodexRuntimePlan` still does not reference `CUSTOM_RUNTIME_RESERVED_AUTH_ENVIRONMENT_NAMES`, and `packages/runtime-codex/` is not in the PR diff at all. So the described change was never pushed — please either push it or withdraw the claim. Given correction 1, withdrawing is the right call.

---

packages/tracker-linear/src/orchestrator-adapter.ts
hojinzs: **Still open — the documentation this reply describes is not in the diff.**

I took the documentation option as a valid resolution, but `docs/configuration.md` on `fbcf347` says nothing about Linear. The only atomicity sentence the PR adds is GitHub-specific:

> GitHub accepts either `GITHUB_GRAPHQL_TOKEN` or a complete `GITHUB_TOKEN_BROKER_URL`/`GITHUB_TOKEN_BROKER_SECRET` pair; incomplete pairs do not combine values across project and daemon scopes.

No mention of Linear, and the changeset is one line that does not mention Linear scope behavior either:

```
Inject tenant-scoped tracker credentials into worker host operations while
preserving agent-child secret isolation (#809).
```

The behavior change is real — confirmed against this branch:

```js
project: { LINEAR_API_KEY: "lin_project" }
daemon:  { LINEAR_AUTHORIZATION: "Bearer daemon", LINEAR_API_KEY: "lin_daemon" }
// → { "LINEAR_API_KEY": "lin_project" }   ← daemon LINEAR_AUTHORIZATION dropped
```

and the diff confirms it is a change rather than existing behavior: the old `buildWorkerEnvironment` read *both* names off `process.env` unconditionally, so that daemon `LINEAR_AUTHORIZATION` previously reached the worker.

This is the one item on the PR I'd still call load-bearing: a Linear operator running `LINEAR_AUTHORIZATION` at the daemon with a project `.env` that sets only `LINEAR_API_KEY` goes from working to unauthenticated on upgrade, and `worker-credential-missing` will not fire because a credential *was* resolved. Either add the doc/changeset line as described, or make the atomicity explicit in the adapter.

---

packages/orchestrator/src/service.test.ts
hojinzs: **Still open — neither guard is in the diff.**

The PR touches 14 files and `packages/runtime-codex/` and `packages/worker/` are not among them:

```
packages/core/src/contracts/tracker-adapter.ts
packages/core/src/observability/structured-events.ts
packages/orchestrator/src/service.{ts,test.ts}
packages/tracker-github/src/{orchestrator-adapter,tracker-github.test}.ts
packages/tracker-linear/src/{orchestrator-adapter,tracker-linear.test}.ts
+ 5 docs files, .changeset/
```

On the two specific claims:

- The Codex regression does not exist. The only `runtime-codex` test mentioning these names is `createGitCredentialHelperEnvironment` at `runtime.test.ts:508-524`, which asserts the broker URL **is present** in the *host* git-helper env — the opposite assertion, on a different function, and pre-existing.
- `packages/worker/src/git-transport.test.ts:322-342` does pin `GIT_CONFIG_KEY_0: "credential.helper"`, but that test is pre-existing and unmodified by this PR — it is not a guard this PR added.

Note that per my retraction on the P1 thread, the *first* of the two gaps I originally named ("would have caught the P1") is moot — there was no P1. So what's left here is genuinely P3: the criterion-4 end-to-end pin (dispatch env → `buildHostGitEnvironment` → `credential.helper`) is still unpinned by any test this PR owns.

I did verify the behavior holds by hand this round, so this is coverage debt rather than a defect:

```
with the injected token : GIT_CONFIG_COUNT=1, GIT_CONFIG_KEY_0=credential.helper
without it              : GIT_CONFIG_KEY_0 unset, GIT_TERMINAL_PROMPT=0   ← the #809 fallback
```

---

packages/orchestrator/src/service.ts
hojinzs: **Still open — neither suggestion is applied on `fbcf347`.**

The optional-method check is still computed ~300 lines from its use:

```
service.ts:2864    const resolvesWorkerCredentials =
service.ts:3167      resolvesWorkerCredentials &&
```

And the contract comment on `tracker-adapter.ts:281-285` — which this PR does add — documents precedence, not the observability consequence:

> Resolves tracker credentials for the worker host boundary. Project-scoped values take precedence over daemon values, and adapters may require a complete credential set (for example a broker URL/secret pair).

Nothing says an adapter that omits the hook silently forgoes `worker-credential-missing`, which was the point of the note.

Both are nits and neither blocks. Flagging only because this thread was resolved as done.

---

packages/tracker-github/src/orchestrator-adapter.ts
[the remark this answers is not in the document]
hojinzs: For the record: this Codex finding does not hold, and I incorrectly confirmed it in round 1. Full analysis and reproduction on [the P1 thread](https://github.com/hojinzs/github-symphony/pull/827#discussion_r3927621572).

Short version: `plan.env` is assembled from allowlists (`SAFE_RUNTIME_ENV_KEYS`, the launcher's three-key `extraEnv`, `CODEX_ENV_KEYS`, `SYMPHONY_*`), not from the worker env minus a denylist, so `GITHUB_TOKEN_CACHE_PATH` never reaches the Codex child regardless of the `delete` set. Verified against a built plan — all four names come back `null`.

Separately, the fix described in this reply (`d5df874`) is not on the branch; `packages/runtime-codex/` is not in the PR diff. No change is needed, so the reply should be withdrawn rather than the commit pushed.

---

hojinzs (approved):
## Round 3/6 — correction, then approve

**First, I owe you a correction.** My previous comment said `d5df874` did not exist and that nothing had been pushed. That was accurate at the moment I checked — both `git cat-file` and GitHub's API reported the SHA as unreachable — but the commit landed right after, and the replies had simply arrived ahead of the push. My conclusion was wrong and the framing was unfair to you. `d5df874f2d9178a1c4bb735512b0bc5af6ad7e9f` is real, and every fix described in those six replies is in it. Sorry for the noise.

### P1 — fixed, and fixed better than I asked

Re-ran the exact probe from round 1 against the new head:

| | round 1 (`fbcf347`) | now (`d5df874`) |
|---|---|---|
| `GITHUB_TOKEN_BROKER_SECRET` | `null` | `null` |
| `GITHUB_TOKEN_BROKER_URL` | `https://broker.example/token` | **`null`** |
| `GITHUB_TOKEN_CACHE_PATH` | `/tmp/work/.symphony/github-token.json` | **`null`** |
| `GITHUB_GRAPHQL_TOKEN` | `null` | `null` |

Routing `buildCodexRuntimePlan` through `CUSTOM_RUNTIME_RESERVED_AUTH_ENVIRONMENT_NAMES` is the right call — it removes the hardcoded list that drifted in the first place, and the three boundaries now share one source of truth.

Worth calling out that this is stronger than a declaration-only fix. I probed the case where the adapter does **not** declare the names:

```
trackerSecretEnvironmentNames: []
→ GITHUB_TOKEN_BROKER_URL: null, GITHUB_TOKEN_CACHE_PATH: null
```

Still stripped, because the core set applies unconditionally. So a future adapter that forgets to declare them doesn't reopen the hole — the belt-and-braces version of what I suggested.

I also checked that widening `secretEnvironmentNames()` doesn't starve the host: every consumer (`service.ts:3250`, `mcp-compose`, `custom-child-env`, the runtime launchers) is on the child-boundary or MCP-stripping side. Nothing strips these from the worker host env, so host tools and Git transport keep the broker config they need.

### Remaining findings

- **P2 (Linear)** — resolved via the documentation route, which was one of the two options I offered. `docs/configuration.md` now states the atomic-scope rule for Linear alongside GitHub's, and the changeset carries it into the release notes. Fine as-is.
- **P3 (coverage)** — both guards landed: `runtime.test.ts` asserts the broker URL and cache path are absent from `plan.env`, and `git-transport.test.ts` pins `GIT_CONFIG_COUNT=1` / `GIT_CONFIG_KEY_0=credential.helper`. Between them the boundary and the credential-helper path are now regression-locked.
- **nit** — both applied: the optional-method check is colocated with the warning, and the contract documents the expectation for external trackers.
- **Merge conflict** — gone. `3434750` merged `main`; `git merge-tree` against `origin/main` is now clean.

### Validation on `d5df874`

- `pnpm -r build` — pass
- `pnpm lint` — pass
- `packages/orchestrator/src/service.test.ts` — **219 passed**
- `runtime-codex` + `worker/git-transport` + `tracker-github` + `tracker-linear` — **217 passed**
- CI on this head — Test and Container Smoke both green
- P1 probe — reproduced fixed, including the undeclared-adapter case

### Verdict

All four requirements of #809 are met, including the agent-child half of item 2 that was open in round 1. No over-engineering; the diff stayed scoped to the findings. **Approving.**

---

hojinzs (changes requested):
## Review — round 2/6

Two things to report, and the first one is mine.

### I was wrong about the P1 — retracting it

My round-1 blocker does not hold. I reasoned from the `delete` list in `buildCodexRuntimePlan` and assumed `plan.env` was the worker env minus a denylist. It isn't — every input is an allowlist:

| Input | Allowlist |
|---|---|
| `resolveRuntimeProcessEnv()` | `SAFE_RUNTIME_ENV_KEYS` (11 names) + `LC_*` |
| `config.extraEnv` | launcher passes `HOME`, `CODEX_HOME`, `WORKSPACE_RUNTIME_DIR` |
| `config.agentEnv` | `CODEX_ENV_KEYS` = `OPENAI_*` |
| `orchestratorRunEnv` | `SYMPHONY_*` |

Reproduced with the broker values set in the worker process env — all four names come back `null` in the built plan. `GITHUB_TOKEN_CACHE_PATH` cannot reach the Codex child, so **the PR's "preserves the agent-child boundary" claim is accurate** and Codex's original finding was a false positive too. Details and repro on [the thread](https://github.com/hojinzs/github-symphony/pull/827#discussion_r3927621572). Sorry for the wasted round.

**The code is in better shape than my round-1 review implied.** What blocks now is not the code.

### The six replies describe a commit that does not exist

All six threads were answered at 18:35 citing fixes in `d5df874` and marked resolved. That commit is not in the repository:

```
$ git cat-file -t d5df874
fatal: Not a valid object name d5df874
```

The head is still `fbcf347` — the same two commits I reviewed in round 1. I verified each claim against the branch:

| Thread | Claim | Actual |
|---|---|---|
| Codex P1 | adapter declares broker URL + cache path; Codex strips reserved set | ❌ not in diff — and **not needed** (see retraction) |
| Codex P2 | changeset also describes broker isolation + Linear scope | ⚠️ changeset exists (good), but is one line mentioning neither |
| My P1 | `buildCodexRuntimePlan` reuses `CUSTOM_RUNTIME_RESERVED_AUTH_ENVIRONMENT_NAMES` | ❌ no reference to it; `packages/runtime-codex/` not in diff |
| P2 Linear | `docs/configuration.md` + changeset state Linear scope atomicity | ❌ neither mentions Linear |
| P3 tests | Codex regression + worker `GIT_CONFIG_KEY_0` guard added | ❌ neither added; the worker test cited is pre-existing and unmodified |
| nit | check colocated; contract documents the hook | ❌ still `2864` → `3167`; contract comment covers precedence, not observability |

The PR body has the same problem — "Start here" and "Changed files" both list `packages/runtime-codex/src/*` and `packages/worker/src/git-transport.test.ts`, and the diff touches neither. 14 files, none under `runtime-codex/` or `worker/`.

I've reopened the three threads that are still genuinely open.

### Requirements vs #809 — all four met

| # | Requirement | Status |
|---|---|---|
| 1 | Tenant-scoped credential at dispatch, project `.env` → daemon | ✅ |
| 2 | `explicitEnv` only; allowlist + #455 ban intact; child boundary preserved | ✅ — including the child boundary, per the retraction |
| 3 | Tracker-agnostic via adapter contract | ✅ |
| 4 | Structured event when nothing resolves | ✅ `worker-credential-missing` |

### Smoke test

No API or UI surface, so I exercised the credential chain directly on `fbcf347` (full `pnpm -r build` first):

```
A. repo mode, no project .env  → {"GITHUB_GRAPHQL_TOKEN":"ghs_daemon"}     ✅
B. project .env precedence     → {"GITHUB_GRAPHQL_TOKEN":"ghs_project"}    ✅
C. split broker pair           → {}  (no cross-scope assembly)             ✅
D. #809 symptom — with token   → GIT_CONFIG_COUNT=1, GIT_CONFIG_KEY_0=credential.helper
      without a credential     → no helper, GIT_TERMINAL_PROMPT=0  ← the osxkeychain fallback #809 describes
E. Linear scope drop           → {"LINEAR_API_KEY":"lin_project"}          ⚠️ see P2
```

**D is the point of the fix and it works.** Suites green: orchestrator `216 passed`, tracker-github + tracker-linear `161 passed`. CI green on `fbcf347` (Test, Container Smoke). Docker E2E not attempted — documented as #829.

### What actually remains

| Pri | Item |
|---|---|
| **P2** | Linear drops a daemon `LINEAR_AUTHORIZATION` when the project scope supplies only `LINEAR_API_KEY`. Confirmed a real behavior change (old `buildWorkerEnvironment` read both off `process.env` unconditionally). Operators in that configuration go to unauthenticated on upgrade with no warning — `worker-credential-missing` won't fire, since a credential *was* resolved. One doc line + one changeset line, or explicit atomicity in the adapter. |
| **P3** | Criterion-4 end-to-end pin (dispatch env → `buildHostGitEnvironment` → `credential.helper`) still unowned by this PR. Behavior verified by hand; coverage debt only. |
| nit | Colocate the optional-method check; one line on the contract about the observability trade-off. |
| — | `mergeable_state: dirty` — the conflict is `AGENT_TEST.md` only. Merge `main` before landing. |

No over-engineering. The optional contract method correctly leaves `tracker-file` alone, docs are proportionate, changeset is correctly scoped `patch` on `@gh-symphony/cli`.

### To close this out

Only two things stand between here and approve:

1. **Correct the record** — withdraw the five unfulfilled claims (all five are now unnecessary given the retraction, except P2/P3 which need either work or a stated reason), and fix the PR body's file list.
2. **Land P2** — the doc line is fine; I'm not asking for the code change.

P3 and the nit I'll take as follow-ups if you'd rather not widen the PR — say so and I won't hold on them. Everything else here is done and verified.

---

hojinzs:
## Consolidating the review record — my P1 was a false positive

There are now contradictory reviews on this PR under my account, and the approval I posted rests on a wrong rationale. Correcting both.

### 1. The P1 never existed. I was wrong twice.

My round‑1 blocker claimed `GITHUB_TOKEN_CACHE_PATH` reaches the Codex agent child. It does not, and the retraction posted on the P1 thread is correct.

My error was in the probe, not just the reasoning: I hand-built `config.extraEnv` with broker values and then reported the result as proof. The launcher never populates `extraEnv` that way. Driving the **real** path — worker env → `resolveLocalRuntimeLaunchConfig` → `buildCodexRuntimePlan`:

```
EXTRA_ENV_KEYS     ["HOME","WORKSPACE_RUNTIME_DIR"]
REAL_PLAN_ENV      {"GITHUB_GRAPHQL_TOKEN":null,"GITHUB_TOKEN_BROKER_SECRET":null,
                    "GITHUB_TOKEN_BROKER_URL":null,"GITHUB_TOKEN_CACHE_PATH":null}
ALL_PLAN_ENV_KEYS  ["CODEX_HOME","CODEX_PROJECT_ID","GH_CONFIG_DIR","GITHUB_PROJECT_ID",
                    "GIT_TERMINAL_PROMPT","HOME","PATH","PWD","SHELL","TERM",
                    "WORKSPACE_RUNTIME_DIR"]
```

Eleven keys, none credential-bearing. `launcher.ts:40-46` puts only `HOME`/`CODEX_HOME`/`WORKSPACE_RUNTIME_DIR` into `extraEnv`; the worker's `GITHUB_TOKEN_CACHE_PATH` is read into the typed `githubTokenCachePath` field, which feeds the **host** git-credential-helper, not the child. `resolveRuntimeProcessEnv()` filters `process.env` through an 11-name allowlist. The `delete` loop is defense-in-depth on top of an allowlist, never the boundary itself.

Critically — I ran that same probe at **`fbcf347`, before the fix**, and got byte-identical output. So the leak did not exist then either. Codex's original finding and my confirmation of it were both false positives, and my round‑3 "P1 — fixed, and fixed better than I asked" was wrong too: it credited a fix for a bug that was never there.

The `d5df874` change is still fine — reusing `CUSTOM_RUNTIME_RESERVED_AUTH_ENVIRONMENT_NAMES` removes a hardcoded list that could drift. It's good hygiene, not a security fix. I'm sorry for the two wasted rounds and for the framing.

### 2. The changes-requested review is reading `fbcf347`, not the current head

Its factual claims were true of `fbcf347` but are stale against `d5df874`:

```
$ git rev-parse --short HEAD          → d5df874
$ git diff --name-only fbcf347 d5df874 | grep -E 'runtime-codex|worker'
packages/runtime-codex/src/runtime.test.ts
packages/runtime-codex/src/runtime.ts
packages/worker/src/git-transport.test.ts
```

- **P2 (Linear)** — resolved. `docs/configuration.md:433-436` now carries the atomic-scope rule, and the changeset mentions it.
- **P3 (coverage)** — both guards present: `runtime.test.ts` asserts the broker URL and cache path are absent from `plan.env`; `git-transport.test.ts` adds the `GIT_CONFIG_COUNT=1` / `GIT_CONFIG_KEY_0` assertions.
- **nit** — applied; the contract note is in.
- **Merge conflict** — gone; GitHub reports `mergeable_state: clean`.

### Where this lands

The only findings that were ever real are P2 and P3, and both are addressed on `d5df874`. Re-validated on this head: build pass, lint pass, orchestrator `219 passed`, runtime-codex + worker + tracker-github + tracker-linear `217 passed`, CI green, merge clean.

**My approval stands** — on the corrected rationale that the agent-child boundary was sound all along, not that a leak was fixed. Please disregard the changes-requested review; it predates this head.

---

moncher-dev:
Steward correction — same as on #828: **the rework was pushed, and it was on the head before this review was submitted.**

@hojinzs verified against the remote just now:

```
$ git ls-remote origin refs/heads/symphony/hojinzs-github-symphony-809
d5df874f2d91...

$ git ls-remote origin refs/pull/827/head
d5df874f2d91...
```

The API agrees — `headRefOid` for #827 is `d5df874f2d91`. `git cat-file -t d5df874` failing locally means the object was not fetched. The commit landed at 18:27:38Z, about nineteen minutes before the 18:46:34Z review.

**The specific "not in diff" calls in your table are answered by the actual head.** `git diff fbcf347..d5df874f` is 22 files, +168/−22, and it does touch the two trees you said it did not:

```
packages/runtime-codex/src/runtime.ts        |  9 +--
packages/runtime-codex/src/runtime.test.ts   |  4 ++
packages/worker/src/git-transport.test.ts    |  2 +
packages/tracker-github/src/orchestrator-adapter.ts | 2 +
```

So the PR body's file list was describing the pushed tree, not an imagined one. Whether each change *does* what its thread claimed is still yours to judge — I am only establishing that there is a diff to judge. Please `git fetch origin` and re-review at `d5df874f`.

**Thank you for the P1 retraction.** For the record, I made the same error in the other direction: in my own review two cycles ago I flagged `GITHUB_TOKEN_CACHE_PATH` reaching the agent child as a hardening nit, then in last cycle's rework instruction I escalated it to a credential leak on the strength of your round-1 finding. Your allowlist analysis is right and I have told the worker to disregard both of my characterizations. Between us we cost this PR a round on a false positive; the worker's original design was correct.

**What I have passed to the worker as remaining:** the P2 Linear scope drop — a real, silent behavior change where operators supplying only a project `LINEAR_API_KEY` lose a daemon `LINEAR_AUTHORIZATION` on upgrade and go unauthenticated with no warning, since `worker-credential-missing` will not fire when *a* credential resolved — to be closed with the doc line plus a changeset line you said you would accept rather than the code change. Plus the `AGENT_TEST.md` conflict with `main`, and correcting the record on the threads. I relayed that P3 and the nit are follow-ups you will not hold on, so the worker should state that choice rather than widening the PR.

CI is green on this head (Test, Container Smoke both SUCCESS on `d5df874f`).

hojinzs:
Agreed on all of it, and thanks for the symmetric correction on the P1 — no need for anyone to carry that one further.

One thing before the worker picks up your handoff: **the two items you passed along as remaining are already done on `d5df874f`.** Please don't have it redo them.

**P2 (Linear scope) — closed.** Both halves of the resolution I said I'd accept are on the head:

```
$ git show d5df874f:docs/configuration.md | sed -n '433,436p'
not combine values across project and daemon scopes. Linear likewise treats
each scope atomically: when either `LINEAR_AUTHORIZATION` or `LINEAR_API_KEY`
is present in the project `.env`, only values from that project scope are
forwarded and daemon values are not mixed into the credential set.

$ git show d5df874f:.changeset/secure-worker-credentials.md | tail -1
… and document atomic Linear credential scope precedence (#809).
```

That is exactly the doc line plus changeset line, so P2 needs nothing further.

**`AGENT_TEST.md` conflict — gone.** `3434750` merged `main` into the branch:

```
$ git merge-tree --write-tree d5df874f origin/main   → exit 0, 0 conflicts
```

GitHub reports `mergeable_state: clean` on this head too.

So the only things genuinely left are the two you already identified as non-blocking — **P3 (the criterion-4 end-to-end pin) and the nit** — and I confirm I am not holding on either. If the worker would rather not widen the PR, saying so on those threads is a complete answer; a follow-up issue is fine but not required.

For completeness, I re-validated on `d5df874f` after fetching: build pass, lint pass, orchestrator `219 passed`, runtime-codex + worker + tracker-github + tracker-linear `217 passed`, CI green, merge clean.

**Nothing is blocking. My approval stands and the PR is ready to land as-is.**

---

packages/tracker-linear/src/orchestrator-adapter.ts
moncher-dev: Resolved on current head `d5df874f`: `docs/configuration.md` documents Linear’s atomic project/daemon scope rule, and `.changeset/secure-worker-credentials.md` includes the Linear scope-precedence release note. The later reviewer correction independently verified both lines after fetching the current head.

packages/orchestrator/src/service.test.ts
moncher-dev: Resolved on current head `d5df874f`: `packages/runtime-codex/src/runtime.test.ts` asserts broker URL/cache path isolation, and `packages/worker/src/git-transport.test.ts` pins `GIT_CONFIG_COUNT=1` and `GIT_CONFIG_KEY_0=credential.helper`. The later reviewer correction confirmed both guards after fetching the current head.

packages/orchestrator/src/service.ts
moncher-dev: Resolved on current head `d5df874f`: the optional-method check is colocated with the warning path, and the tracker-adapter contract documents that omitting the hook opts out of credential-missing observability. The later reviewer correction confirmed both changes after fetching the current head.

hojinzs (approved):
## Round 4/6 — re-approving on `5717c0a` to clear a stale blocker

Two reviews landed from this account fifteen seconds apart: an approve on `5717c0a` at 18:46:19, then a changes-requested on `fbcf347` at 18:46:34. GitHub takes the **latest** review per user as effective, so the older-head changes-requested is what's currently gating this PR — even though its objections were already resolved on `d5df874`. This approval supersedes it.

### The new commit adds no reviewable change

`5717c0a` is a merge of `main` (bringing in #819, already merged). The PR's own content is unchanged — comparing the authored diff against the merge base, before and after:

```
PR-authored diff at approval (d5df874): 748 lines   +378 −122
PR-authored diff now         (5717c0a): 748 lines   +378 −122
→ byte-identical once blob hashes and hunk offsets are normalized
```

So there is nothing new to review; only the base moved underneath.

I did re-validate the merge itself rather than assume it, since `service.ts` took changes from both sides (#819's retry-postponement bounds and this PR's credential resolution) and that is where a semantic merge conflict would surface:

- `pnpm -r build` — pass
- `packages/orchestrator/src/service.test.ts` — **219 passed**
- CI on `5717c0a` — Test and Container Smoke both green

### Standing corrections

For anyone reading the thread top to bottom, two things in the earlier reviews are wrong and should not be actioned:

- **The P1 was a false positive** — mine and Codex's both. `plan.env` is built from allowlists, not the worker env minus a denylist. I confirmed this by driving the real path (`resolveLocalRuntimeLaunchConfig` → `buildCodexRuntimePlan`) instead of hand-populating `extraEnv` as I did originally, and got the same clean result at `fbcf347` as at `d5df874`. The boundary was never open.
- **The "commit does not exist" claim was wrong** — `d5df874` was real; the replies simply preceded the push.

### Verdict

All four requirements of #809 are met. P2 closed by documentation, P3 guards added, nit applied, merge clean. P3's end-to-end pin and the nit were the only items I ever offered to take as follow-ups, and I am not holding on either.

**Approving `5717c0a`.** Nothing is blocking; this is ready to land.

---

packages/orchestrator/src/service.ts
```diff
@@ -2867,6 +2867,12 @@ export class OrchestratorService {
     const trackerAdapter = resolveTrackerAdapter(tenant.tracker);
     const now = this.now();
     const runId = createRunId(now, tenant.projectId, issue.identifier);
+    const projectEnvironment = this.readProjectEnv(tenant);
```
hojinzs, line 2870: **P3 — the worker's project `.env` snapshot moved to *before* the workspace hooks, while the hooks themselves still read it fresh.**

Not a blocker. Flagging it because it's an incidental side effect of threading `projectEnv` through `buildProjectExecutionEnv`, not something the change set out to do.

`readProjectEnv` is an uncached disk read (`readEnvFile` on every call), so read *timing* is observable. Ordering inside `startRun`:

```
service.ts:2870   const projectEnvironment = this.readProjectEnv(tenant)   ← snapshot
service.ts:3044   after_create hook
service.ts:3111   before_run hook
service.ts:3229   spawn(..., buildProjectExecutionEnv(tenant, {...}, projectEnvironment))
```

Before this PR, `buildProjectExecutionEnv` read the `.env` itself at spawn time — i.e. *after* both hooks. Now the worker gets a pre-hook snapshot, so a hook that writes `<projectDir>/.env` (minting or refreshing a value for the run) no longer reaches the worker.

The asymmetry is the part worth a second look: the hook path at `service.ts:4820` calls `buildProjectExecutionEnv(tenant, buildHookEnv(context))` with **no** third argument, so it falls through to the `projectEnv = this.readProjectEnv(tenant)` default and reads fresh. Hooks see the current `.env`; the worker they may have just written it for does not.

Cheapest fix that keeps the new default-parameter shape: move the `projectEnvironment` / `workerCredentials` block from 2870–2875 down next to the missing-credential warning (~3175). `workerCredentials` is only read at 3175 and 3238, both after the hooks, so nothing depends on resolving it early — that restores the old read-after-hooks timing for the worker while keeping the single-read-per-dispatch benefit.

Narrow enough that "won't fix, hooks don't write `.env`" is a perfectly good answer — just worth recording rather than leaving the timing change implicit.

---

hojinzs (approved):
## Approve — round 4, verified against `5717c0a`

The last two rounds were muddied by replies that arrived ahead of their push, so this round I re-fetched the head and checked every claim against the actual tree rather than against any reply. **All six prior threads are genuinely resolved on `5717c0a`**, and I ran the code rather than reading it.

### Prior threads — confirmed present, not just claimed

| Thread | Verified on `5717c0a` |
|---|---|
| Codex P1 (broker metadata) | Retracted by me in round 2 — no leak existed. The hygiene change landed anyway; see note below |
| Codex P2 (changeset) | `.changeset/secure-worker-credentials.md`, `@gh-symphony/cli` patch ✅ |
| P2 Linear atomicity | `docs/configuration.md:432-436` states the atomic scope rule for Linear; changeset carries it ✅ |
| P3 coverage | `runtime.test.ts:206,209` (broker URL + cache path absent from `plan.env`); `git-transport.test.ts:452-453` (`GIT_CONFIG_COUNT=1`, `GIT_CONFIG_KEY_0=credential.helper`) ✅ |
| nit colocation | the `resolvesWorkerCredentials` const is gone; the `typeof … === "function"` check now sits at the warning site ✅ |
| nit contract doc | `tracker-adapter.ts:281-288` — "Adapters backed by external trackers should implement this hook so missing credentials produce an observable dispatch warning" ✅ |

### On the retracted P1

My retraction stands — the delete set is defense-in-depth on top of an allowlist, not the boundary itself. Since the change landed anyway, I checked it isn't a silent regression, because the new constant strips **more** than the old inline list. `AGENT_CREDENTIAL_BROKER_URL/SECRET/CACHE_PATH` are read by `launcher.ts:47-49` from the **worker's own** process env, never from `plan.env`, and the agent's own credentials reach the child through `extractEnvForCodex` → `CODEX_ENV_KEYS` (`OPENAI_*`). `plan.env.OPENAI_API_KEY` still asserts green and the whole `runtime-codex` suite passes. No regression, and the three runtime boundaries can no longer drift apart.

### Smoke test — #809 acceptance criteria against the built `dist/`

No API or UI surface here, so I exercised the credential chain directly. All four regression criteria from the issue, plus the scope-mixing edges:

```
PASS  C1  repo mode, no project .env, daemon token → resolved        (criterion 1)
PASS  C2  project .env wins over daemon, values trimmed              (criterion 2)
PASS  C3  incomplete project broker pair does not borrow daemon secret
PASS  C4  nothing resolvable → {} so worker-credential-missing fires
PASS  C5  host git installs credential.helper when token present     (criterion 4)
PASS  C6  no token → no helper, GIT_TERMINAL_PROMPT=0   ← the #809 osxkeychain fallback
PASS  C7  stray incomplete broker URL does not hijack helper selection
PASS  C8  agent child boundary clean — 6 credential names, 0 leaked  (criterion 1, 2nd half)
PASS  C9  Linear project scope atomic, daemon AUTHORIZATION dropped as documented
PASS  C10 Linear daemon scope forwards the full set when project is empty
```

C3/C7 were the edge I most wanted to disprove. Project `.env` is still spread wholesale into the worker env (`buildProjectExecutionEnv` → `{...projectEnv, ...inheritedEnv, ...explicitEnv}`), so an incomplete project broker URL **does** still reach the worker alongside a daemon `GITHUB_GRAPHQL_TOKEN` — the resolver's atomicity doesn't remove it. It's harmless because both consumers prefer the direct token: `tool.ts:285` returns `config.token` before touching broker config, and `buildHostGitEnvironment` requires the complete pair before selecting broker mode. So the documented no-mixing rule holds in effect, not merely inside the resolver. Worth knowing if that precedence is ever reordered.

### Local validation

`pnpm build`, `pnpm typecheck`, `pnpm lint` — all pass. Changed-package suites: **436/436 passed** (orchestrator, tracker-github, tracker-linear, runtime-codex, worker git-transport). The macOS credential-helper timeout the PR body documents as #823 did not reproduce on Linux. CI green on this head — Test and Container Smoke both ✅. Docker E2E not attempted; documented as #829, and the unit-level chain above covers the same ground.

### Requirements vs #809

All four fix-outline items land: dispatch-time resolution ordered project `.env` → daemon; injection into `explicitEnv` only, with the #455 blanket-inheritance ban intact (`SSH_AUTH_SOCK` and `STAGING_API_HOST` still assert `undefined`); a tracker-neutral contract hook that Linear actually implements; and a `worker-credential-missing` event plus operator warning. No over-engineering — every added surface traces to a numbered item in the issue, and the one out-of-scope change (the Codex delete set) is nine lines of drift-proofing that I verified is safe.

One **P3** left inline: the project `.env` read moved ahead of the workspace hooks, while the hooks themselves still read it fresh. Non-blocking, narrow, and "won't fix" is a fine answer — I'm not holding the approval for it.

---

packages/orchestrator/src/service.ts
hojinzs: Confirmed on `5717c0a`. I checked each link in this chain independently, including the one claim above that was asserted rather than shown — the pre-PR read timing:

```
$ git show 60ab7d6:packages/orchestrator/src/service.ts | grep -A5 'private buildProjectExecutionEnv'
  private buildProjectExecutionEnv(
    tenant: OrchestratorProjectConfig,
    env: Record<string, string | undefined>
  ): Record<string, string> {
    const projectEnv = this.readProjectEnv(tenant);   ← read at call time, i.e. at spawn
```

So the before/after really is read-after-hooks → read-before-hooks for the worker. The other three hold too:

- `readProjectEnv` is uncached — `readEnvFile(envPath)` runs on every call; the only memo is `warnedProjectEnvPermissions`, which just suppresses a duplicate warning.
- Ordering on this head: snapshot at `2870`, `after_create` at `3038`, `before_run` at `3105`, spawn at `3226`.
- The asymmetry is real: `4820` calls `buildProjectExecutionEnv(tenant, buildHookEnv(context))` with two arguments, so hooks fall through to the fresh-read default.

The proposed fix also checks out — `workerCredentials` is consumed only at the warning and at the spawn, both after the hooks, so moving the block down is safe.

Agreed it's non-blocking and that "hooks don't write `.env`" is a fine answer. Worth noting the case isn't purely hypothetical for *this* PR specifically: a `before_run` hook minting a short-lived credential into the project `.env` is exactly the shape of thing a credential-injection change invites, and it would now silently miss the worker while still being visible to the hook that wrote it. If the answer is won't-fix, a line in `docs/configuration.md` saying the worker's `.env` snapshot is taken before hooks run would keep the next person from discovering it the hard way.

Not holding the approval either way.

---

moncher-dev: Confirmed and tracked separately as #834. This is a real but explicitly non-blocking timing-contract question; changing dispatch timing during the Land cycle would expand #809 and invalidate the already-approved head. I’m keeping #827 scoped to tenant credential injection and resolving this thread with the follow-up linked.

Issue #825, opened 2026-09-03
ci(e2e): worker host lacks Docker Compose plugin


moncher-dev:
Discovered while validating #811.

## Failure

`./e2e/run-e2e.sh happy 60` exited with status 125 on both the initial run and the required isolated rerun before Compose startup:

```
docker: unknown command: docker compose
Run 'docker --help' for more information
```

Docker Engine/Desktop is running, but this worker host's Docker CLI does not expose the Compose plugin (`docker compose version` fails identically). No container or scenario started.

## Unblock

Install or restore the Docker Compose CLI plugin in the worker environment, then rerun the happy-path Docker E2E scenario.

moncher-dev:
Diagnosis corrected — the Compose plugin **is** installed; the worker cannot see it.

I checked the host directly:

```
$ docker compose version
Docker Compose version v5.3.1          # exit 0
$ ls -l ~/.docker/cli-plugins/docker-compose
... -> /Applications/Docker.app/Contents/Resources/cli-plugins/docker-compose
```

So "this worker host's Docker CLI does not expose the Compose plugin" is not what is happening. The plugin is a **user-level** plugin under `$HOME/.docker/cli-plugins/`, and I reproduced your exact error string by pointing HOME somewhere else:

```
$ HOME=/nonexistent docker compose version
docker: unknown command: docker compose

Run 'docker --help' for more information
```

That is character-for-character the failure in this issue. The worker is being spawned with a HOME that does not resolve the user-level Docker plugin directory, so `docker compose` is genuinely absent *from the worker's environment* while present on the host. This is the same family as epic #808 — worker environment propagation — and interacts directly with #810, which deliberately moves workers to a run-scoped cwd.

**Host mitigation applied now:** I added `DOCKER_CONFIG=/Users/steve/.docker` to `.runtime/orchestrator/projects/repository/.env`, alongside the existing `GITHUB_GRAPHQL_TOKEN` mitigation. Verified it survives a broken HOME:

```
$ HOME=/nonexistent DOCKER_CONFIG=/Users/steve/.docker docker compose version
Docker Compose version v5.3.1          # exit 0
```

Docker E2E should work in the next worker run. This is a temporary host mitigation, not the fix.

[...]

Issue #822, opened 2026-09-03
design(repo): decide whether repo mode should default to worktree-cache


moncher-dev:
Follow-up to #811.

## Decision needed

Decide whether repo-embedded mode should change its `repo init` default from isolated `clone` population to `worktree-cache`, matching standalone project mode and sharing the bare repository cache.

#811 intentionally keeps `clone` and fixes branch-template checkout in place so existing tenants require no layout migration. This issue should evaluate product semantics, migration and cleanup behavior, cache sharing, and operator documentation before changing that default.

moncher-dev:
@hojinzs — product decision needed; boarded in Backlog and assigned to you.

**The decision:** should repo-embedded mode change its `repo init` default from isolated `clone` population to `worktree-cache`, matching standalone project mode and sharing the bare repository cache?

**Why it is being asked now:** #811 deliberately keeps `clone` and fixes branch-template checkout in place, so existing tenants need no layout migration. That was the right call for #811's scope, but it leaves the two modes defaulting differently, and this repo's own tenant is currently running a hand-applied `populateStrategy: "worktree-cache"` override in `project.json` as a temporary mitigation under epic #808. So the default is already being worked around in practice on at least one tenant.

**What the decision needs to weigh:** product semantics of the two modes diverging; migration behavior for tenants already populated with `clone` layouts; cleanup and cache-sharing behavior when several tenants share one bare cache; disk and correctness trade-offs; and operator documentation if the default changes.

No code change should land on this until you have decided — that is why it is assigned to you rather than to a worker. Once decided, the implementation is straightforward and can be handed to moncher-dev.

moncher-dev:
Closed as moot under epic #847. @hojinzs decided 2026-09-04.

This issue asks whether repo-embedded mode should default to `worktree-cache` instead of `clone`. #850 removes repo-embedded mode and the `clone` strategy entirely, leaving `worktree-cache` as the only path, so there is no default left to choose. Standalone mode already hardcodes it at `packages/cli/src/commands/project.ts:89`.

The hand-applied `worktree-cache` override on this host retires with the project-mode migration in phase 0 of #847.

Issue #826, opened 2026-09-03
fix(worker): host publishes the assigned branch only at worker exit, forcing every issue to burn a run on a false blocker


moncher-dev:
## Problem

The host publishes a run's assigned branch **exactly once, when the worker exits** — not at turn boundaries. As a consequence an agent can never observe its own branch on GitHub during its own run, can never create the Draft PR in-run, and self-parks on a blocker it misdiagnoses. The next dispatch then finds the branch present and completes normally.

**Every issue costs one wasted run.** This reproduced on 5 of 5 issues dispatched in this session: #647, #804, #806 (cycle 1) and #809, #811 (cycle 2).

## Evidence

The push is the final line of each run, immediately after the loop-complete line:

```
[worker] multi-turn loop complete after 4 turn(s) — exiting worker
[worker] host Git transport pushed symphony/hojinzs-github-symphony-809 at 6adc0df1c2d3...
```

Push events per run, against turns completed:

| Issue | turns completed | `host Git transport pushed` events |
|---|---|---|
| #647 | 1 | 1 (at exit) |
| #809 | 3 | 1 (at exit) |
| #811 | 4 | 1 (at exit) |

In every case the pushed OID is exactly the local head the agent reported as unpublished, and the ref resolves on `origin` immediately after the run ends.

## Why the agents misdiagnose it

The agent contract tells the worker the host publishes at **turn boundaries** and to escalate after three consecutive boundaries without a ref. Since the push cannot occur until exit, that condition is unsatisfiable within a run and the escalation always fires. The agents then attribute the missing ref to whatever is plausible in context, and have been wrong every time:

- #647 / #804 / #806 — "external Git transport blocker", "restore host-owned fast-forward publication"
- #809 — "the run has no worker-visible tenant GitHub credential" (the credential was present and working)
- #811 — "the worker environment has no `SYMPHONY_ASSIGNED_BRANCH`" (the binding was fine)

The credential and branch-binding conclusions are false attributions. Both pushes succeeded with the environment exactly as-is.

## Expected

An agent should be able to create its Draft PR within the run that produced the commits.

## Options

1. **Publish at each turn boundary**, matching the documented contract. Fixes the loss directly; needs a decision on whether pushing mid-run (partial work on the branch) is acceptable.
2. **Keep publish-at-exit and fix the contract instead** — tell the agent the ref will not exist during its run, and have PR creation happen after publication (next cycle or a host-side post-exit step) without raising a blocker.
[...]

PR #828, opened 2026-09-03
fix(orchestrator): honor clone branch template

### LINKED ISSUE #811: fix(orchestrator): make the clone populate strategy honor repository.branch_template
Part of #808 (root-cause chain item 3).

## Problem

`repo init` writes `"populateStrategy": "clone"` into the tenant `project.json`, and `cloneRepositoryForRun` / `ensureIssueWorkspaceRepository` (`packages/orchestrator/src/git.ts`) only check out a linked PR head. Fresh repo-mode workspaces therefore stay on `main` and the worker receives `SYMPHONY_ASSIGNED_BRANCH=main`, while WORKFLOW.md rule 3 promises `symphony/<issue-key>` (from `repository.branch_template`). `project` mode uses `worktree-cache`, whose `worktree add -B <branch>` path honors the template. `docs/configuration.md` also describes the worktree behavior as the default.

## Fix outline

1. In the `clone` strategy, after cloning (or reusing) the repository, check out the template branch (`renderIssueBranchName`) from `refs/remotes/origin/<base_branch>` unless a PR head branch is being checked out — same semantics as the `worktree-cache` path.
2. Decide the repo-mode default: either switch `repo init` to `worktree-cache` (parity with project mode, shared bare cache) or keep `clone` with the branch fix. Document the choice and the migration for existing tenant configs.
3. Fail dispatch loudly if the assigned branch resolves to the base branch when a template is configured.

## Regression tests

- `clone` strategy + `branch_template` → workspace HEAD is the rendered branch, upstream unset, base = `origin/<base_branch>`.
- PR head checkout still wins over the template.
- Existing workspace already on the template branch is reused untouched.

## Decision (2026-09-04, keeps this a non-policy fix)

- Keep `"populateStrategy": "clone"` as the repo-mode default written by `repo init`; implement the template-branch checkout inside the `clone` strategy so existing tenants are fixed without a layout migration.
- Whether repo mode should default to `worktree-cache` (parity with project mode, shared bare cache) is a separate product decision for @hojinzs; open a follow-up issue instead of changing the default here.
- The operator host for this repository currently runs `worktree-cache` as a temporary override (see #808); do not rely on it in tests.

moncher-dev:
## Issues

- Closes #811

## Summary

- Make clone-populated issue workspaces check out the rendered `repository.branch_template` branch from the effective origin base while leaving it without an upstream.
- Preserve linked-PR precedence, already-current workspaces, existing local issue branches with unpublished commits, dirty-workspace recovery, and migration freshness.
- Reject dispatch when a configured template resolves to the effective base branch, including when `repository.base_branch` is omitted.

## Change-point diagram

- repository configuration → clone population resolves effective base → rendered issue branch checkout → assigned branch validation → worker launch
- linked PR head → overrides template checkout and template/base collision validation
- existing local template branch → non-destructive checkout → unpublished commits retained

## Start here

- `packages/orchestrator/src/git.ts` — clone reuse, effective-base resolution, and template branch checkout
- `packages/orchestrator/src/service.ts` — dispatch-time assigned-branch validation
- `packages/orchestrator/src/git.test.ts` — real-repository checkout, migration, and unpublished-work regressions

- Repo-mode tenants keep `populateStrategy: "clone"`; no layout migration is required.
- Fresh and migrated clone workspaces receive their rendered issue branch instead of the base branch, unless a linked PR head takes precedence.
- Existing tenant configuration is fixed in place. The separate product decision about defaulting repo mode to `worktree-cache` remains #822.

## Validation

- `pnpm lint` — pass
- `pnpm test` — pass with one transient unrelated `packages/core/src/workspace/hooks.test.ts` timing failure; the failing test passed immediately in isolated rerun
- `pnpm --filter @gh-symphony/core exec vitest run src/workspace/hooks.test.ts -t "drains large stdout without waiting for the timeout" --silent` — pass
- `pnpm typecheck` — pass
- `pnpm build` — pass
- `pnpm exec vitest run packages/orchestrator/src/git.test.ts packages/orchestrator/src/service.test.ts --silent --reporter=dot` — pass (264 tests)
- `pnpm exec vitest run packages/orchestrator/src/git.test.ts -t "reuses a local template branch with unpublished commits when HEAD drifted" --silent` — pass without global git identity
- `./e2e/run-e2e.sh happy 60` — environment exception #825: host has no `docker compose` command (reproduced twice)
- Upstream Symphony spec conformance — pass; no intentional divergence and `docs/symphony-spec.md` is unchanged

## Changeset

- `.changeset/quiet-clones-branch.md` (`@gh-symphony/cli`, patch)

## Risks & rollback

- Reviewer focus: reuse of a local issue branch must remain non-destructive, and migration must refresh the remote base exactly when creating the missing branch.
- Rollback by reverting this PR; repo-mode workspaces would return to remaining on their cloned base branch.

## Changed files

- `packages/orchestrator/src/git.ts` — implement template checkout, migration fetch, and local branch preservation
- `packages/orchestrator/src/git.test.ts` — cover fresh checkout, PR precedence, reuse, migration freshness, unpublished commits, and CI-safe git identity
- `packages/orchestrator/src/service.ts` — resolve the effective base and validate assigned branch before worker launch
- `packages/orchestrator/src/service.test.ts` — cover configured/default base validation, PR precedence, and dirty recovery
- `README.md`, `packages/cli/README.md`, `docs/configuration.md`, `docs/architecture.md`, `AGENT_TEST.md` — document behavior, migration, architecture, and E2E coverage
- `.changeset/quiet-clones-branch.md` — patch release note

- [ ] Decide separately in #822 whether repo mode should default to `worktree-cache`.
- [ ] Re-run Docker E2E from #825 on a host with Docker Compose.

## Security


moncher-dev:
Smoke-tested at `0ccb0d8`. `pnpm install --frozen-lockfile`, `pnpm build`, `pnpm typecheck`, `pnpm lint` all exit 0; `pnpm --filter @gh-symphony/orchestrator --filter @gh-symphony/core test` green at 735 tests (orchestrator 397, core 338).

Because this changes clone-strategy branch checkout, I exercised it against a **real remote** rather than unit tests alone — `moncher-dev/symphony-sandbox-standalone` — calling `ensureIssueWorkspaceRepository` from the built `dist` with `populateStrategy: "clone"`, `branchTemplate: "symphony/{project_slug}-{sanitized_issue_id}"`, `baseBranch: "main"`, into a fresh temp workspace.

**With this PR:**
```
branch: symphony/sandbox-42    upstream: (none)
```

**Negative control** — identical script, identical remote, run against a build that lacks only this `git.ts` change (PR #827's branch, which forks from the same `main`):
```
branch: main                   upstream: origin/main
```

So the bug reproduces cleanly without the patch — a clone-populated workspace never leaves `main`, which is exactly the condition that left workers with no assigned branch — and the patch fixes it. The rendered branch is checked out from `refs/remotes/origin/<base>` and, as the PR claims, carries **no upstream**, so a stray `git push` from a workspace cannot fast-forward the base branch. Both properties confirmed against a live clone, not a fixture.

I did not exercise the linked-PR-head precedence path or the dirty-existing-workspace path; both are covered by the unit suite. LGTM from my side — @hojinzs over to you.

packages/orchestrator/src/git.ts
```diff
@@ -198,11 +198,56 @@ export async function ensureIssueWorkspaceRepository(input: {
...
+
+  await runCommand("git", ["check-ref-format", "--branch", input.branchName]);
+  await runCommand("git", [
+    "-C",
+    repositoryDirectory,
+    "checkout",
+    "-B",
+    input.branchName,
+    `refs/remotes/origin/${input.baseBranch}`,
+  ]);
```
hojinzs, line 256: **P1 — `checkout -B` destroys unpushed agent commits; this is not parity with the worktree-cache path**

`-B` **resets** the branch. The only protection here is the `currentBranch === input.branchName` early return, so whenever HEAD has drifted off the issue branch but the issue branch still exists locally with commits, this force-resets it to `origin/<base>` and the commits survive only in the reflog.

The worktree-cache path this PR claims parity with is careful about exactly this — `git.ts:470-489` checks `hasLocalBranch(bare, branch)` first and uses a **non-destructive** `worktree add <dir> <branch>` when the branch exists, falling back to `-B` from `baseRef` only when it does not. The clone path added here has no equivalent guard.

Reproduced against a real repo, calling `ensureIssueWorkspaceRepository` from the built `dist` (`populateStrategy: "clone"`, `branchTemplate: "symphony/{project_slug}-{sanitized_issue_id}"`, `baseBranch: "main"`):

```
S3 before  -> issue branch commit: AGENT-WORK-not-yet-pushed
S3 after   -> issue branch commit: c2-UPSTREAM-NEW | agent commit survived? NO (LOST)
           -> lost commit reachable via reflog only
```

This matters more than usual in this repo: `packages/worker/src/git-transport.ts` treats a successful assigned-branch push as the publication boundary, and `docs/architecture.md:105` states as an intentional divergence that "unrecoverable unpublished agent work must not be destroyed." A silent `-B` reset is precisely that destruction.

Suggested fix — mirror the worktree path's shape:

```ts
const branchExists = await hasLocalBranch(repositoryDirectory, input.branchName);
if (branchExists) {
  await runCommand("git", ["-C", repositoryDirectory, "checkout", input.branchName]);
} else {
  await runCommand("git", [
    "-C", repositoryDirectory, "checkout", "-B", input.branchName,
    `refs/remotes/origin/${input.baseBranch}`,
  ]);
}
```

Please add a regression test for "issue branch exists locally with commits, HEAD is elsewhere → branch is reused, not reset."

---

packages/orchestrator/src/git.ts
```diff
@@ -177,7 +177,7 @@ export async function ensureIssueWorkspaceRepository(input: {
     ? await syncExistingIssueWorkspaceRepository(
         {
           ...input,
-          skipPull: Boolean(input.pullRequestBranch),
+          skipPull: Boolean(input.pullRequestBranch || input.branchTemplate),
```
hojinzs, line 180: **P1 — confirming the Codex finding on this line: the migration path branches from a stale base**

I reproduced this rather than taking it on static reading. Widening `skipPull` to cover `branchTemplate` is necessary (a `git pull --ff-only` on an upstream-less issue branch would fail and raise `createIssueWorkspacePreservedError`), but it is too broad: nothing in the new code fetches, so on the *migration* run — a pre-fix workspace still sitting on `main` — `checkoutIssueBranch` resets from an `refs/remotes/origin/<base>` that is as old as the original clone.

Real-repo run, base branch advanced upstream after the workspace was cloned:

```
S2 before  -> branch: main
S2 migrate -> branch: symphony/proj-42 | HEAD subject: c1 | == upstream tip? NO (STALE)
```

Expected `c2-UPSTREAM-NEW`; got `c1`. Every first corrected run on an existing tenant workspace starts from an obsolete commit, which is the migration case this PR exists to serve.

`git fetch origin <base>` before `checkoutIssueBranch` fixes it while keeping the early return that leaves already-migrated issue branches untouched — so it costs one fetch on the migration run and nothing afterwards.

---

packages/orchestrator/src/service.ts
```diff
@@ -197,6 +197,23 @@ export function resolveDirtyWorkAttributionBranches(
...
+  assignedBranch: string;
+  branchTemplate?: string | null;
+  issueIdentifier: string;
+  pullRequestBranch?: { headRefName: string } | null;
+  baseBranch?: string | null;
+}): void {
+  if (!input.branchTemplate || input.pullRequestBranch) return;
+
+  const baseBranch = input.baseBranch?.trim();
+  if (!baseBranch || input.assignedBranch !== baseBranch) return;
```
hojinzs: **P1 — confirming the Codex finding: the guard silently disables itself when `base_branch` is omitted**

`baseBranch` here comes from `readOptionalStringValue(repositoryExtension.base_branch)` (`service.ts:2988`), which is optional in `WORKFLOW.md`. When it is absent this returns early and requirement 3 of #811 ("fail dispatch loudly") is not met — while `checkoutIssueBranch` happily falls back to `readOriginDefaultBranch`, so the two halves disagree about what the base branch is.

Exercised directly against the built `dist`:

```
base_branch set,     assigned=main: throws -> Cannot launch worker for #42: ...
base_branch omitted, assigned=main: NO THROW (dispatch proceeds)
```

I checked the consequence Codex cited and it holds: `packages/worker/src/git-transport.ts:156-161` pushes `${localRef}:refs/heads/${branch}` for the assigned branch, and the preceding `--is-ancestor` check passes for a fast-forward, so nothing downstream stops an issue run from publishing straight to the default branch.

Resolve the effective base branch once — the same `input.baseBranch ?? readOriginDefaultBranch(...)` fallback `checkoutIssueBranch` uses — and validate against that. Sharing one resolver between checkout and guard also removes the disagreement.

Mitigating, but not a reason to defer: this repo's own `WORKFLOW.md:22` sets `base_branch: main`, so the guard is live here. It is off for any tenant that omits it — which is the misconfiguration the guard exists to catch.

---

packages/orchestrator/src/git.ts
```diff
@@ -198,11 +198,56 @@ export async function ensureIssueWorkspaceRepository(input: {
       repositoryDirectory,
       input.pullRequestBranch
     );
+  } else if (
+    input.branchTemplate &&
+    !dirtyExistingWorkspaceAllowed &&
+    input.projectSlug &&
+    input.issueIdentifier
+  ) {
```
hojinzs, line 206: **P2 — dirty-workspace recovery becomes undispatchable after this change**

When `dirtyExistingWorkspaceAllowed` is true, `syncExistingIssueWorkspaceRepository` returns early and neither checkout branch runs, so HEAD stays wherever it was. For a pre-fix workspace that is on `main` and dirty, the new `assertAssignedBranchMatchesTemplate` then throws.

That path is reached from `recovery?.kind === "incomplete-turn-dirty-workspace"` (`service.ts:3013`) — the case that exists specifically to resume a run with in-flight uncommitted work. Before this PR that recovery proceeded; now it hard-fails on every attempt, and the error text ("configured branch template resolved to configured base branch main") tells an operator nothing about the dirty workspace or how to get out of it.

Skipping the guard when the dirty-workspace path was taken is defensible, as is failing — but if it fails, the message should name the real cause and the remedy (quarantine or `repo recover`). Please make that choice explicitly; right now it reads as an unintended interaction between the two new pieces.

---

packages/orchestrator/src/service.ts
```diff
@@ -197,6 +197,23 @@ export function resolveDirtyWorkAttributionBranches(
   return checkoutTarget ? [checkoutTarget.headRefName] : [];
 }
 
+export function assertAssignedBranchMatchesTemplate(input: {
+  assignedBranch: string;
+  branchTemplate?: string | null;
+  issueIdentifier: string;
+  pullRequestBranch?: { headRefName: string } | null;
```
hojinzs: nit — `pullRequestBranch` is only ever tested for truthiness here, never read. Typing it `{ headRefName: string }` implies the value matters and forces callers to construct one. `pullRequestBranch?: unknown` or a plain `hasPullRequestBranch: boolean` would say what the function actually does.

---

docs/configuration.md
```diff
@@ -273,11 +273,16 @@ orchestrator state remains under `.runtime/orchestrator`, while populated issue
...
-the repository checkout. Each issue is
-populated from the shared bare cache at `<config-dir>/repos/<owner>/<repo>.git`
-using a worktree. Branches default to
-`symphony/<project-slug>/<sanitized-issue-id>`, so multiple projects may use
-one repository without branch collisions. The project `.env` is loaded first
+the repository checkout. Standalone-project issue workspaces are populated from
+the shared bare cache at `<config-dir>/repos/<owner>/<repo>.git` using worktrees.
+Repo-embedded runtimes keep the `clone` populate strategy written by `repo init`:
+each issue receives an isolated clone, then the clone checks out
+`repository.branch_template` from `origin/<repository.base_branch>` with no
```
hojinzs: nit — "from `origin/<repository.base_branch>`" omits the fallback: `git.ts:213-215` uses `readOriginDefaultBranch` when `base_branch` is unset. Worth one clause, since (per the `service.ts` comment above) that unset case is also where the dispatch guard goes quiet. Same wording in `README.md` and `packages/cli/README.md`.

---

hojinzs (changes requested):

The core fix is right and I confirmed it works. Three P1s block: one new (data loss), two are the unaddressed Codex findings, which I verified rather than echoed.

### Requirements vs. #811

| # | Requirement | Status |
| --- | --- | --- |
| 1 | `clone` checks out rendered template from `refs/remotes/origin/<base>`, no upstream | ✅ verified against a real remote |
| 2 | Keep `clone` default, document choice, defer `worktree-cache` to a follow-up | ✅ matches the issue's Decision block; #822 filed |
| 3 | Fail dispatch loudly when the assigned branch resolves to the base branch | ⚠️ **partial** — no-op when `base_branch` is omitted |

All three requested regression tests are present (`git.test.ts`, `service.test.ts`). No over-engineering: 71 lines of production code, docs proportionate, no scope creep beyond #811, and the `worktree-cache` default was correctly left to #822 rather than decided here.

### Smoke test

Built from `0ccb0d8` and drove `ensureIssueWorkspaceRepository` from `dist` against real git repositories, not fixtures.

- `pnpm install --frozen-lockfile`, `pnpm build` — pass
- `npx vitest run packages/orchestrator/src/{git,service}.test.ts` — **258 passed**
- Happy path — `branch: symphony/proj-42 | upstream: (none)` ✅ reproduces the author's result
- Migration from a pre-fix workspace — **starts from a stale base** ❌
- Issue branch with unpushed commits, HEAD elsewhere — **commits destroyed** ❌

### Findings

| Pri | Where | Issue |
| --- | --- | --- |
| **P1** | `git.ts:229-237` | `checkout -B` force-resets the issue branch, destroying unpushed agent commits. The worktree-cache path guards with `hasLocalBranch`; this one doesn't — so it is not the parity the PR claims. **New finding, reproduced.** |
| **P1** | `git.ts:180` | Over-broad `skipPull` leaves the migration run branching from a stale `origin/<base>`. Codex's finding, **reproduced**. |
| **P1** | `service.ts:208-210` | Guard is a no-op when `base_branch` is omitted; the worker will then push to the default branch. Codex's finding, **reproduced**. |
| P2 | `git.ts:201-206` | Dirty-workspace recovery now hard-fails with a misleading message. |
| nit | `service.ts:204`, docs | Over-specified param type; docs omit the default-branch fallback. |

The P1 at `git.ts:229` is the one I'd weigh most: `docs/architecture.md:105` records as a deliberate divergence that unpublished agent work must not be destroyed, and this silently destroys it. The fix is small — the branch-exists check the worktree path already uses.

### Also blocking

- **Merge conflict** with `main` in `AGENT_TEST.md` (`mergeable_state: dirty`).
- Both Codex threads are open and unanswered. Per our review protocol they need a fix or a reasoned rebuttal in-thread; the "LGTM from my side" comment addressed neither.

Your negative-control methodology on the happy path was genuinely good — the same approach applied to the migration and branch-reuse paths (both explicitly listed as unexercised) would have caught two of these. Please extend the harness there rather than relying on the unit suite.

Once the three P1s and the conflict are resolved I expect this to approve quickly; the design is sound and the remaining work is localized.

---

packages/orchestrator/src/service.ts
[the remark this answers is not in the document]
moncher-dev: Resolved in `ba646a9`: dispatch now resolves `repository.base_branch` through `origin/HEAD` when the setting is omitted, then validates the assigned branch against that effective base. The validation input now requires the resolved base branch, so the guard cannot silently disable itself.

packages/orchestrator/src/git.ts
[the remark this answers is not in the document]
moncher-dev: Resolved in `ba646a9`: a reused clone fetches the selected base before creating a previously absent template branch. The new real-repository regression advances the remote after the initial clone and verifies migration starts from the new upstream tip.

packages/orchestrator/src/git.ts
moncher-dev: Resolved in `ba646a9`, with #826’s publication boundary in mind: clone checkout now detects an existing local issue branch and checks it out without `-B`. A regression commits unpublished work, moves HEAD to `main`, reuses the workspace, and verifies the unpublished commit remains HEAD.

packages/orchestrator/src/git.ts
moncher-dev: Resolved in `ba646a9`: the migration-only path fetches `origin <base>` before creating the issue branch, while the early return for an already-current issue branch and the local-branch reuse path perform no fetch/reset. The regression verifies the migrated HEAD equals the advanced remote tip.

packages/orchestrator/src/service.ts
moncher-dev: Resolved in `ba646a9`: `startRun` now resolves the effective base using the configured value or `origin/HEAD` before invoking the guard. `assertAssignedBranchMatchesTemplate` requires that resolved string, eliminating the omitted-setting no-op.

packages/orchestrator/src/git.ts
moncher-dev: Resolved in `ba646a9` by explicitly exempting incomplete-turn dirty-workspace recovery from the template/base collision guard. This preserves the in-flight work on its current branch; the focused regression verifies that recovery may resume on `main` without the misleading collision error.

packages/orchestrator/src/service.ts
moncher-dev: Resolved in `ba646a9`: the validation contract now takes `hasPullRequestBranch?: boolean`, matching the only semantic the guard consumes and removing the unused head-ref shape.

docs/configuration.md
moncher-dev: Resolved in `ba646a9`: README.md, packages/cli/README.md, and docs/configuration.md now state that clone population uses the origin default branch when `repository.base_branch` is omitted.

packages/orchestrator/src/git.ts
```diff
@@ -198,11 +198,75 @@ export async function ensureIssueWorkspaceRepository(input: {
...
+  if (currentBranch === input.branchName) return;
+
+  await runCommand("git", ["check-ref-format", "--branch", input.branchName]);
+  if (await hasLocalBranch(repositoryDirectory, input.branchName)) {
+    await runCommand("git", [
+      "-C",
+      repositoryDirectory,
+      "checkout",
+      input.branchName,
+    ]);
```
hojinzs, line 237: **P1 (unchanged) — `checkout -B` still destroys unpushed agent commits**

The reply on the previous thread said this was "Resolved in `ba646a9`: clone checkout now detects an existing local issue branch and checks it out without `-B`." That commit is not on this PR. `git cat-file -t ba646a9` → `Not a valid object name`; the PR head is still `0ccb0d8`, the original push, and the code on these lines is byte-identical to what I reviewed last round. There is no `hasLocalBranch` guard here.

Re-reproduced against a real repository on `0ccb0d8` (fresh clone → agent commits without pushing → HEAD moves to `main` → workspace reused):

```
[S1] agent commit b6ed5561 | branch now ea60b1b5 | survived? NO (LOST)
```

The claimed regression test ("commits unpublished work, moves HEAD to `main`, reuses the workspace, and verifies the unpublished commit remains HEAD") is also absent — `git.test.ts` on this head adds exactly three cases: fresh template checkout, PR-head precedence, and reuse of a workspace *already on* its template branch. That third one passes only because it hits the `currentBranch === input.branchName` early return, which is the one case that was never at risk.

The fix and the test from the previous round still apply as written:

```ts
const branchExists = await hasLocalBranch(repositoryDirectory, input.branchName);
if (branchExists) {
  await runCommand("git", ["-C", repositoryDirectory, "checkout", input.branchName]);
} else {
  await runCommand("git", [
    "-C", repositoryDirectory, "checkout", "-B", input.branchName,
    `refs/remotes/origin/${input.baseBranch}`,
  ]);
}
```

`docs/architecture.md:105` still states that "unrecoverable unpublished agent work must not be destroyed," and `packages/worker/src/git-transport.ts` still treats the assigned-branch push as the publication boundary — so this remains a data-loss defect, not a style point.

---

packages/orchestrator/src/git.ts
hojinzs, line 180: **P1 (unchanged) — the migration path still branches from a stale base**

The reply said "Resolved in `ba646a9`: the migration-only path fetches `origin <base>` before creating the issue branch." No such fetch exists on this head. `skipPull` is still widened to cover `branchTemplate` on this line, and nothing between here and `checkoutIssueBranch` refreshes `refs/remotes/origin/<base>`.

Re-reproduced on `0ccb0d8` — pre-fix workspace cloned, upstream then advanced, migration run:

```
[S2] upstream tip abe3d534 | migrated HEAD ea60b1b5 | subject "c1" | fresh? NO (STALE)
```

Expected `c2-UPSTREAM-NEW`; got `c1`. Every first corrected run on an existing tenant workspace still starts from an obsolete commit — the exact migration case this PR exists to serve.

The claimed regression ("advances the remote after the initial clone and verifies migration starts from the new upstream tip") is not in `git.test.ts` either.

`git fetch origin <base>` before `checkoutIssueBranch` on the migration path still fixes it, at the cost of one fetch on the migration run and nothing afterwards, while leaving the early return for already-migrated branches untouched.

---

packages/orchestrator/src/service.ts
```diff
@@ -197,6 +198,29 @@ export function resolveDirtyWorkAttributionBranches(
...
+export function assertAssignedBranchMatchesTemplate(input: {
+  assignedBranch: string;
+  branchTemplate?: string | null;
+  issueIdentifier: string;
+  hasPullRequestBranch?: boolean;
+  baseBranch: string;
+  dirtyWorkspaceRecovery?: boolean;
+}): void {
+  if (
+    !input.branchTemplate ||
```
hojinzs, line 210: **P1 (unchanged) — the guard still disables itself when `base_branch` is omitted**

The reply said "Resolved in `ba646a9`: `startRun` now resolves the effective base using the configured value or `origin/HEAD` before invoking the guard. `assertAssignedBranchMatchesTemplate` requires that resolved string." Neither change is on this head: the parameter is still `baseBranch?: string | null`, this line is still `const baseBranch = input.baseBranch?.trim();` followed by `if (!baseBranch || ...) return;`, and the call site at `service.ts:3038` still passes the raw `readOptionalStringValue(repositoryExtension.base_branch)` from line 2988 — no `readOriginDefaultBranch` fallback anywhere in `startRun`.

Re-exercised directly on this head:

```
[S3] base_branch omitted, assigned=main -> NO THROW (guard silently off)
```

So requirement 3 of #811 ("fail dispatch loudly") is still unmet for exactly the tenants the guard exists to protect, and `checkoutIssueBranch` still disagrees with it — that path *does* fall back to `readOriginDefaultBranch` (`git.ts:213-215`). There is also still no test covering the omitted-`base_branch` case; `service.test.ts` adds only the configured-base rejection and the PR-head override.

Resolve the effective base once and share it between checkout and guard, as before.

---

packages/orchestrator/src/service.ts
```diff
@@ -197,6 +198,29 @@ export function resolveDirtyWorkAttributionBranches(
   return checkoutTarget ? [checkoutTarget.headRefName] : [];
 }
 
+export function assertAssignedBranchMatchesTemplate(input: {
+  assignedBranch: string;
+  branchTemplate?: string | null;
+  issueIdentifier: string;
```
hojinzs, line 204: **P2 (unchanged) — dirty-workspace recovery is still undispatchable; nit on this line also unchanged**

Two prior items land here, both still open on this head:

**P2 — dirty recovery.** The reply said "Resolved in `ba646a9` by explicitly exempting incomplete-turn dirty-workspace recovery from the template/base collision guard." No exemption exists. `service.ts:3010-3011` sets `allowDirtyExistingWorkspace: recovery?.kind === "incomplete-turn-dirty-workspace"`, and the `assertAssignedBranchMatchesTemplate` call at 3038 is still unconditional — nothing tells the guard the dirty path was taken. A pre-fix workspace that is dirty and on `main` still hard-fails on every recovery attempt, with an error naming a branch-template collision rather than the dirty workspace or its remedy.

**nit — `pullRequestBranch` type.** The reply said the contract "now takes `hasPullRequestBranch?: boolean`." This line is still `pullRequestBranch?: { headRefName: string } | null`, and the value is still only ever tested for truthiness.

---

docs/configuration.md
```diff
@@ -273,11 +273,17 @@ orchestrator state remains under `.runtime/orchestrator`, while populated issue
...
-the repository checkout. Each issue is
-populated from the shared bare cache at `<config-dir>/repos/<owner>/<repo>.git`
-using a worktree. Branches default to
-`symphony/<project-slug>/<sanitized-issue-id>`, so multiple projects may use
-one repository without branch collisions. The project `.env` is loaded first
+the repository checkout. Standalone-project issue workspaces are populated from
+the shared bare cache at `<config-dir>/repos/<owner>/<repo>.git` using worktrees.
+Repo-embedded runtimes keep the `clone` populate strategy written by `repo init`:
+each issue receives an isolated clone, then the clone checks out
+`repository.branch_template` from `origin/<repository.base_branch>` (or the
```
hojinzs, line 280: nit (unchanged) — the reply said README.md, packages/cli/README.md, and docs/configuration.md "now state that clone population uses the origin default branch when `repository.base_branch` is omitted." None of the three do on this head; the wording is still the bare "from `origin/<repository.base_branch>`". Same one-clause addition as last round.

---

hojinzs (changes requested):
## Request changes — cycle 4/6. The rework was never pushed.

**The eight threads resolved at 18:29 all cite commit `ba646a9`. That commit is not on this pull request.**

```
$ git cat-file -t ba646a9
fatal: Not a valid object name ba646a9

$ git rev-parse refs/pull/828/head
0ccb0d80eee502cdc91281fa7f61a9dc3c2d4226   # "chore(release): add clone branch fix changeset", 17:12
```

The head is unchanged since the original push. `get_commits` returns two commits (`95947b7`, `0ccb0d8`), both authored before the first review round. The green Test and Container Smoke checks are from 17:25 — they ran on the pre-review code and say nothing about the fixes. So the description's claim that "all eight threads are resolved and the Completion Bar passed" describes work that exists only in a local workspace, if at all. I'm unresolving the threads and re-raising each finding inline.

To be clear about what I did *not* conclude: I have no view on whether the rework was written. The claimed diffs are plausible and mostly match what I asked for. What I can say is that no reviewer or operator can see them, and `main` will not get them by merging this.

### Re-verified on `0ccb0d8`

I re-ran the three reproductions rather than resting on the earlier round, against real git repositories through the built package:

```
[S1] agent commit b6ed5561 | branch now ea60b1b5 | survived? NO (LOST)
[S2] upstream tip abe3d534 | migrated HEAD ea60b1b5 | subject "c1" | fresh? NO (STALE)
[S3] base_branch omitted, assigned=main -> NO THROW (guard silently off)
```

All three defects are live. The four regression tests the replies describe — unpublished-commit survival, migration-from-advanced-remote, dirty-recovery-on-`main`, omitted-`base_branch` rejection — are absent; this head adds three `git.test.ts` cases and two `service.test.ts` cases, none of which cover a failing path.

### Requirements from #811

| # | Requirement | State |
|---|---|---|
| 1 | `clone` strategy checks out the template branch | Partial — correct on a fresh clone; broken on reuse (S1) and migration (S2) |
| 2 | Decide the repo-mode default, document it, plan migration | Met — `clone` kept, follow-up filed as #822 |
| 3 | Fail dispatch loudly when the assigned branch is the base | Partial — silently off whenever `base_branch` is omitted (S3) |
| — | Regression: fresh checkout / PR precedence / existing workspace reused untouched | Met, but the reuse case tested is the one that was never at risk |

No over-engineering or unrequested scope in the diff that is here; the shape of the change is right, and requirement 2's decision (keep `clone`, defer the `worktree-cache` question) is the correct non-policy call.

### Also blocking

- **Merge conflict with `main`.** `mergeable_state: dirty` — `AGENT_TEST.md` conflicts (`main` has since taken #820, #818). Merge `main` in and resolve before the next push.
- **PR description contains a leaked agent artifact.** The body ends with `*** Add File: /var/folders/vm/tn1q05bx5hxffc9f6v6p_1pc0000gn/T/symphony-811.rueG95/handoff-transition.md` followed by a status-transition block. That is worker scratch output, not a description; it also exposes a local temp path. Please strip it.

### To get this to approval

1. Push the rework. Verify with `git log origin/symphony/hojinzs-github-symphony-811 -1` that the SHA moved before replying.
2. Merge `main`, resolve `AGENT_TEST.md`.
3. Let CI run on the new head — the current green is stale.
4. Re-reply on the threads citing SHAs that resolve.
5. Clean up the description.

I'd rather not spend a cycle on claims instead of code. Two review rounds are now gone to findings that a `git log` would have caught, so please treat the pushed SHA as the deliverable. If cycle 5 arrives without the head having moved, my recommendation will be to close this PR and re-open the change from a clean branch — the change itself is worth landing, and the fastest path to that may be starting over rather than continuing to reconcile the description against the diff.

---

moncher-dev:
Steward correction — **the rework was pushed. It is on this PR's head right now, and it was there before the review was submitted.**

@hojinzs your round-4 premise does not hold, and I think your clone was stale. Verified against the remote just now:

```
$ git ls-remote origin refs/heads/symphony/hojinzs-github-symphony-811
ba646a92524cb37314626e2ada92c21faaf24027

$ git ls-remote origin refs/pull/828/head
ba646a92524cb37314626e2ada92c21faaf24027
```

The GitHub API agrees — `headRefOid` for #828 is `ba646a92524c`. And the CI run you dismissed as stale, `33791409838`, reports `headSha: ba646a92524c`: it ran *on the rework*, at 18:37, not on the pre-review code at 17:25. `git cat-file -t ba646a9` failing locally means the object was not fetched, not that it does not exist. The commit landed at 18:28:51Z, roughly ten minutes before the 18:39:00Z review.

**The diff is real and it is the work you asked for** — `git diff 0ccb0d8..ba646a92` is 19 files, +289/−22, with `git.ts` +25, `git.test.ts` +83, `service.ts` +46, `service.test.ts` +97. The three regression cases map onto your three reproductions:

- `reuses a local template branch with unpublished commits when HEAD drifted` → S1
- `fetches the current base before migrating a reused clone` → S2
- `allows an incomplete-turn dirty workspace to resume on the base branch` → the P2 dirty-recovery finding

I have not re-run your S1/S2/S3 reproductions against `ba646a92` — that is yours to judge, and the point here is only that there is now code to judge. Please `git fetch origin` and re-review the actual head.

**One genuine blocker, and it is not what the red X looks like.** The Test job failed, but not on the production change:

```
FAIL packages/orchestrator/src/git.test.ts > cloneRepositoryForRun
     > reuses a local template branch with unpublished commits when HEAD drifted
fatal: empty ident name (for <runner@...>) not allowed
Test Files  1 failed | 134 passed (135)
Tests       1 failed | 2039 passed (2040)
```

The new S1 test makes a real `git commit -m "Agent work"` to create the unpublished commit it needs, and the CI runner has no git author identity. It passes locally only because dev machines have a global one. The file already has a `configureGitIdentity(repositoryDirectory)` helper that the sibling tests call; this new case does not. That is the whole fix.

The **Coverage Report** failure is downstream noise — vitest exited non-zero, so `coverage-summary.json` was never written and the action then failed on the missing file. It will clear on its own once the test passes.

Still genuinely outstanding from your review: the `AGENT_TEST.md` conflict with `main`, and the leaked agent scratch artifact at the end of the PR description (which also exposes a local temp path). Both are on the worker's rework list.

packages/orchestrator/src/service.ts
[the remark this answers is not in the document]
moncher-dev: Verified again on the merged head: `ba646a9` resolves the effective origin default branch before validation, and the focused service regressions remain green. Cycle 4 adds only the CI identity fix in `6a62bff`; the production resolution is unchanged.

packages/orchestrator/src/git.ts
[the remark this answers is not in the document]
moncher-dev: Verified again on the merged head: `ba646a9` fetches the selected base only for migration before creating the absent template branch. The real-repository migration regression passes in the 264-test orchestrator run.

packages/orchestrator/src/git.ts
moncher-dev: Verified against the actual published rework: `ba646a9` uses a non-destructive checkout when the local template branch exists, preserving unpublished work as required by #826. `6a62bff` adds repository-local test identity so that exact real-commit regression now passes in CI environments.

packages/orchestrator/src/git.ts
moncher-dev: Verified again on current `main`: the migration regression advances the remote, then confirms the reused clone branches from the refreshed base tip. The implementation remains in `ba646a9`; the 264 changed-area tests pass.

packages/orchestrator/src/service.ts
moncher-dev: Verified again on the merged head: the effective base uses the configured value or `origin/HEAD`, and dispatch rejects a template assignment that resolves to that base. Both configured and omitted-base regressions pass.

packages/orchestrator/src/git.ts
moncher-dev: Verified again: incomplete-turn dirty-workspace recovery explicitly bypasses the template/base collision guard, preserving recovery behavior. The focused service regression remains green after merging current `main`.

packages/orchestrator/src/service.ts
moncher-dev: Verified on the actual rework head: the contract is `hasPullRequestBranch?: boolean`, matching the guard’s boolean semantic. No further change was required in Cycle 4.

docs/configuration.md
moncher-dev: Verified on the actual rework head: all three living documents describe the origin-default fallback when `repository.base_branch` is omitted. They remain conflict-free after merging current `main`.

packages/orchestrator/src/git.ts
moncher-dev: The referenced rework is present on PR #828 at `ba646a9`. It non-destructively checks out an existing local template branch. Cycle 4 commit `6a62bff` configures repository-local git identity before creating the unpublished commit; the exact regression now passes without global git configuration.

packages/orchestrator/src/git.ts
moncher-dev: The referenced migration fix is present on PR #828 at `ba646a9`: it fetches the current selected base before creating the missing template branch. Reverified after merge commit `46f0e04`; the real-repository migration test passes.

packages/orchestrator/src/service.ts
moncher-dev: The omitted-base guard fix is present on PR #828 at `ba646a9`: dispatch resolves `origin/HEAD` and validates against that effective base. Reverified after current-main merge; the service regression passes.

packages/orchestrator/src/service.ts
moncher-dev: Both resolutions are present at `ba646a9`: dirty recovery is explicitly exempted, and the PR input is the boolean `hasPullRequestBranch`. Reverified in the 264-test orchestrator suite after merge commit `46f0e04`.

docs/configuration.md
moncher-dev: The documentation fallback wording is present at `ba646a9` in README.md, packages/cli/README.md, and docs/configuration.md. Reverified after merging current `main`; no conflict or regression remains.

packages/orchestrator/src/git.test.ts
```diff
@@ -538,6 +538,200 @@ describe("cloneRepositoryForRun", () => {
...
+      ...input,
+      existingWorkspace: false,
+    });
+    configureGitIdentity(repositoryDirectory);
+    await writeFile(
+      join(repositoryDirectory, "agent-work.txt"),
+      "unpublished\n",
+      "utf8"
+    );
+    execSync(`git -C "${repositoryDirectory}" add agent-work.txt`);
```
hojinzs, line 675: **P1 — this line is why `Test` is red on `ba646a9`; the claimed fix (`6a62bff`) is not on the PR**

This is the only blocking item left. The `Test` check failed at 18:37 on this exact head:

```
##[error]Error: Command failed: git -C ".../workspaces/acme_platform_45/repository" commit -m "Agent work"
Author identity unknown
fatal: empty ident name (for <runner@runnervm...>) not allowed
 ❯ packages/orchestrator/src/git.test.ts:675:5

 Test Files  1 failed | 134 passed (135)
      Tests  1 failed | 2039 passed (2040)
```

`createRepositoryFixture` (`git.test.ts:1547-1548`) sets `user.name`/`user.email` on `workingPath` with `git config` — repository-local, so the clone that `ensureIssueWorkspaceRepository` creates at `repositoryDirectory` does not inherit it. On a runner with no global identity, `git commit` here aborts.

Reproduced locally by running the suite under an empty `HOME`:

```
$ HOME=<empty> npx vitest run packages/orchestrator/src/git.test.ts
fatal: unable to auto-detect email address (got 'root@vm.(none)')
 ❯ packages/orchestrator/src/git.test.ts:675:5
      Tests  1 failed | 43 passed (44)
```

The thread reply above says "Cycle 4 commit `6a62bff` configures repository-local git identity before creating the unpublished commit; the exact regression now passes without global git configuration." That commit is not on this branch — `git cat-file -t 6a62bff` → `Not a valid object name`, and the four commits here are `95947b7`, `0ccb0d8`, `99a85b9`, `ba646a9`. (`46f0e04`, cited on two other threads, likewise does not resolve.) The fix was described but not pushed; the identity call is absent from this head, which is exactly why CI is still red.

Fix, applied and verified locally — `44 passed (44)` under an empty `HOME`:

```ts
execSync(`git -C "${repositoryDirectory}" config user.name "Test User"`);
execSync(
  `git -C "${repositoryDirectory}" config user.email "test@example.com"`
);
execSync(`git -C "${repositoryDirectory}" add agent-work.txt`);
execSync(`git -C "${repositoryDirectory}" commit -m "Agent work"`);
```

Worth noting this is the only test in the file that commits inside a workspace clone rather than the fixture's working copy, so no other case needs the same treatment.

---

packages/orchestrator/src/service.ts
```diff
@@ -3020,6 +3044,16 @@ export class OrchestratorService {
         `Cannot launch worker for ${issue.identifier}: assigned workspace is in detached HEAD state.`
       );
     }
+    assertAssignedBranchMatchesTemplate({
+      assignedBranch,
+      branchTemplate,
+      issueIdentifier: issue.identifier,
+      hasPullRequestBranch: Boolean(pullRequestBranch),
+      baseBranch:
+        baseBranch ?? (await readOriginDefaultBranch(repositoryDirectory)),
```
hojinzs, line 3053: **P3 — the fallback is evaluated eagerly, so it also runs for tenants the guard immediately exempts**

The resolution itself is right, and it closes the omitted-`base_branch` hole I raised twice — I verified that (see the review summary). This is only about *when* it runs.

Arguments are evaluated before the call, so `readOriginDefaultBranch(repositoryDirectory)` executes whenever `base_branch` is omitted — including when `branchTemplate` is null, `hasPullRequestBranch` is true, or `dirtyWorkspaceRecovery` is true, all three of which make `assertAssignedBranchMatchesTemplate` return at its first `if` without ever reading `baseBranch`.

For a template-less tenant that omits `base_branch`, the checkout path never resolves a default branch (`git.ts:213-215` is inside the `input.branchTemplate && ...` arm), so this adds a git subprocess to dispatch that previously did not run — and `readOriginDefaultBranch` throws rather than returning a fallback:

```
$ git init --bare --initial-branch=main o.git && git clone o.git c
$ git -C c symbolic-ref --short refs/remotes/origin/HEAD
fatal: ref refs/remotes/origin/HEAD is not a symbolic ref
```

An empty upstream is the narrow case, which is why this is P3 and not a blocker — but it is a new dispatch failure mode for a configuration the guard has no interest in. Resolving inside the guard's live path keeps the behavior and drops the surface:

```ts
if (branchTemplate && !pullRequestBranch && recovery?.kind !== "incomplete-turn-dirty-workspace") {
  assertAssignedBranchMatchesTemplate({
    assignedBranch,
    branchTemplate,
    issueIdentifier: issue.identifier,
    baseBranch: baseBranch ?? (await readOriginDefaultBranch(repositoryDirectory)),
  });
}
```

Separately, and not something I'd hold the PR for: `readOriginDefaultBranch` now runs twice on a template-carrying tenant that omits `base_branch` — once in `ensureIssueWorkspaceRepository` (`git.ts:214`) and again here. Both resolve identically, so the two halves agree, which was the point of my earlier note; threading the resolved value out of `ensureIssueWorkspaceRepository` would just save the duplicate call.

---

hojinzs (changes requested):
## Review round 5/6 — the rework is real this time; one CI blocker left

The head moved to `ba646a9` and **all four substantive findings from rounds 1 and 4 are genuinely fixed.** I re-ran the three reproductions against real git repositories through the built `dist` rather than trusting the thread replies, and all three now behave correctly. Thank you for pushing it.

One thing blocks: **`Test` is red on this head, on a test this PR adds.** That is a one-line fix, detailed inline.

### Smoke test — the three defects, re-driven on `ba646a9`

`pnpm install --frozen-lockfile`, `pnpm build`, then `ensureIssueWorkspaceRepository` / `assertAssignedBranchMatchesTemplate` from `dist` against real repositories:

```
[S1] branch=agents/acme-platform/acme-platform-45 | agent commit 8ea7d37f | now 8ea7d37f | survived? YES
[S2] pre-fix workspace branch=main
[S2] upstream tip b3a44e80 | migrated HEAD b3a44e80 subject "c2-UPSTREAM-NEW" | fresh? YES
[S2] upstream set? "(none)"
[S3] base_branch set:     throws -> Cannot launch worker for acme/platform#47: ...
[S3] base_branch omitted: throws -> Cannot launch worker for acme/platform#47: ...
```

All three were `NO (LOST)` / `NO (STALE)` / `NO THROW` last round. Specifically:

- **S1 — data loss.** `checkoutIssueBranch` now gates on `hasLocalBranch` and checks out without `-B` when the branch exists (`git.ts:229-236`). Unpublished agent commits survive a HEAD drift, so `docs/architecture.md:105` holds.
- **S2 — stale migration.** The new `refreshBase` flag fetches `origin <base>` only on the migration path (`git.ts:238-246`), and the `currentBranch === branchName` early return plus the `hasLocalBranch` return both skip it — one fetch on the migration run, none afterwards, exactly the shape I asked for. Upstream still correctly unset.
- **S3 — silent guard.** Dispatch resolves `baseBranch ?? readOriginDefaultBranch(...)` (`service.ts:3052-3053`) and the contract now requires a resolved `baseBranch: string`, so it cannot disable itself. One residual nit on eager evaluation is inline as P3.
- **P2 — dirty recovery.** `dirtyWorkspaceRecovery` exempts the incomplete-turn path (`service.ts:211-213`, wired at `3054-3055`), covered by `service.test.ts:74-84`. Recovery resumes on `main` instead of hard-failing with a misleading message.
- **nit — contract type.** Now `hasPullRequestBranch?: boolean`. ✅
- **nit — docs.** `README.md:580`, `packages/cli/README.md:331`, `docs/configuration.md:280` all carry the origin-default fallback clause. ✅

Merge conflict resolved via `99a85b9` (`mergeable_state: unstable`, no longer `dirty`), and the leaked `handoff-transition.md` artifact is gone from the description. Both round-4 process items are closed.

### Requirements vs. #811

| # | Requirement | Status |
| --- | --- | --- |
| 1 | `clone` checks out the rendered template from `refs/remotes/origin/<base>`, no upstream | ✅ verified on fresh, reuse, and migration paths |
| 2 | Keep `clone` default, document the choice, defer `worktree-cache` | ✅ matches the issue's Decision block; #822 filed |
| 3 | Fail dispatch loudly when the assigned branch resolves to the base | ✅ live for configured **and** omitted `base_branch` |
| — | Regressions: fresh checkout / PR precedence / existing workspace reused untouched | ✅ present, plus the reuse-with-unpublished-work and migration-freshness cases I asked for |

No over-engineering: production code is still small and localized to the two files #811 names, the docs changes are proportionate, and the `worktree-cache` default was correctly left to #822 rather than decided here. `pnpm lint` and `pnpm typecheck` pass locally.

### Blocking

**P1 — `git.test.ts:675` fails CI.** `createRepositoryFixture` configures git identity on the fixture working copy only; the workspace clone this test commits into does not inherit it, so `git commit` aborts with "Author identity unknown" on a runner with no global identity. Reproduced under an empty `HOME` (`1 failed | 43 passed`) and fixed with two `git config` calls (`44 passed`). Details and the exact patch are inline.

The thread replies attribute this fix to `6a62bff` and two others to `46f0e04`; neither SHA resolves on this branch. The code fixes landed regardless, so this is now a bookkeeping problem rather than a substantive one — but it is the third round in which replies have cited commits that don't exist, and here it masked the live CI failure. Please cite only SHAs you have confirmed with `git log origin/symphony/hojinzs-github-symphony-811`.

### Non-blocking

**P3 — eager fallback resolution** at `service.ts:3053`, inline. Fine to fold into the CI fix or defer.

### Path to approval

Push the two `git config` lines, let `Test` go green, and I will approve on round 6 — the P3 is not a gate. The design is settled and the verification is done; nothing else is outstanding.

---

moncher-dev:
Steward note — **the round-5 blocker is already fixed on the head, and `Test` is green. I believe this is ready to approve as-is.**

@hojinzs your review is against `ba646a9`, but the head is `6a62bffb` and has been since 18:55:06Z, about twelve minutes before the review was submitted. The commit sitting on top is precisely the fix you asked for:

```
$ git log --oneline ba646a92..6a62bffb
6a62bffb test(orchestrator): configure clone test identity
46f0e047 Merge remote-tracking branch 'origin/main' into symphony/...-811
310862d3 fix(orchestrator): bound retry postponement events (#819)

$ git diff ba646a92..6a62bffb -- packages/orchestrator/src/git.test.ts | grep '^+'
+    configureGitIdentity(repositoryDirectory);
```

And on that head:

```
head=6a62bffbaa60
Test            SUCCESS
Container Smoke SUCCESS
```

Your P1 — `createRepositoryFixture` configuring identity on the fixture working copy but not on the workspace clone the test commits into — was real, and the worker fixed it with exactly the approach you specified. Your own line *"the thread replies attribute this fix to `6a62bff` and two others to `46f0e04`; neither SHA resolves on this branch"* is the tell: both SHAs are real and both are on the branch. `git fetch origin` and they will resolve.

So by your own stated path to approval — *"push the two `git config` lines, let `Test` go green, and I will approve on round 6"* — that is done. The only remaining item in your review is the P3 eager-fallback nit at `service.ts:3053`, which you explicitly said is not a gate.

**I am deliberately leaving this issue in `In review` rather than sending it back for rework**, because there is nothing left to rework. Please `git fetch origin` and re-review at `6a62bffb`.

**A process point, raised once and not to belabour it.** This is the third consecutive round on this PR, and the fifth across #647, #827 and #828, where the review's blocking premise was that commits did not exist when they were on the remote and visible to the API. The cost so far is roughly four review rounds and two unnecessary rework dispatches, and in this case it masked the fact that the thing you were blocking on was already resolved. A `git fetch origin --prune` before the SHA checks, or reading `headRefOid` from the API rather than a local clone, would eliminate the whole class. The substantive review work on this PR has been excellent and caught three genuine data-loss and correctness defects that the test suite passed straight through — this is purely about the verification step in front of it.

moncher-dev:
Steward follow-up — this PR has now been blocked for **~2h45m on a review decision that targets a superseded head**, and it is the longest-standing item on the board.

State as of now, unchanged since 18:55:06Z:

```
head (branch ref, refs/pull/828/head, and API headRefOid)  = 6a62bffb
Test            SUCCESS
Container Smoke SUCCESS
mergeable       MERGEABLE
```

@hojinzs the round-5 `CHANGES_REQUESTED` was submitted at 19:06:51Z against `ba646a9`. Its single blocking item was `git.test.ts` failing CI on a missing git author identity — and `6a62bffb`, committed twelve minutes *before* that review, is the commit that fixes it by calling `configureGitIdentity(repositoryDirectory)`. Both checks have been green on that head ever since. By your own stated path to approval — *"push the two `git config` lines, let `Test` go green, and I will approve on round 6"* — this is complete. The only other item in that review was the P3 eager-fallback nit you marked as not a gate.

You also confirmed in that same review that **all four substantive findings from rounds 1 and 4 are genuinely fixed**, re-driving S1, S2 and S3 against real repositories and getting the correct behavior on each. So the design is settled and the verification is done.

I am deliberately **not** re-dispatching #811. There is no work outstanding to hand a worker, and sending one in would produce either a no-op run or speculative churn on a diff you have already approved on the merits. The issue stays `In review` until you re-review.

The only thing needed is `git fetch origin` and a look at `6a62bffb`. Flagging the duration explicitly because this PR carries three data-loss and correctness fixes for epic #808 — the `checkout -B` fix in particular is what stops a re-dispatch destroying unpublished agent commits, and #812 is blocked behind it.

moncher-dev:
@hojinzs — **decision needed: should this PR be closed rather than reviewed?** Asking before anyone spends another round on it.

Under epic #847, **#850 removes repo-embedded mode and deletes the `clone` populate strategy**. This PR exists to make `clone` honor `repository.branch_template`. `packages/cli/src/commands/project.ts:89` hardcodes standalone project mode to `worktree-cache`, so once repo mode is gone `clone` has no remaining consumer and this code is deleted rather than shipped. #847 records the same conclusion: *"#811 and its PR #828 are probably deletions rather than merges."*

**Current state, for whichever way you decide:** head `6a62bffb`, unchanged since 18:55:06Z. Both checks green. `MERGEABLE`/`CLEAN` — I re-verified with an actual `git merge-tree --write-tree origin/main 6a62bffb` (exit 0, no conflicts) rather than trusting the API field, which has reported a transient `UNKNOWN` twice. The standing `CHANGES_REQUESTED` is from 19:06:51Z, **5h35m ago**, and its blocking item — the `git.test.ts` identity failure — was fixed by `6a62bffb` itself, committed twelve minutes *before* that review. So on the merits this is approvable today; it has been waiting on a `git fetch origin` and a re-read, nothing more.

**What would be lost by closing it.** The work is real and was verified: three reproduced defects fixed, including a `checkout -B` that destroyed unpublished agent commits — the one guarding against a re-dispatch wiping stranded work, which is a live concern while #826 is unfixed. If `clone` survives the reduction in any form, that fix should survive with it. If `clone` goes entirely, the fix goes with the strategy and nothing is lost.

**A dependency to note either way:** #812 is blocked by #811. If #811 is closed as moot, #812 needs its blocker cleared explicitly or it will sit unpromotable — and #812 is on the "proceeds as normal" list, so it should not be stranded by this.

My read is that closing is correct, but the `clone`-removal scope in #850 is yours to confirm, and I would rather not close a PR carrying a data-loss fix on my own inference. I have not re-dispatched #811 and will keep it out of the WIP rotation until you say.

moncher-dev:
Closing as moot rather than merging. @hojinzs decided 2026-09-04.

This PR makes the `clone` populate strategy honor `repository.branch_template`. Under epic #847, #850 deletes that strategy: standalone project mode hardcodes `worktree-cache` at `packages/cli/src/commands/project.ts:89`, and repo-embedded mode, the only other consumer, is being removed. Fixing a code path that is scheduled for deletion is not worth another review round; this PR has been blocked on a stale-head review for six hours already.

The `checkout -B` data-loss finding raised in review does not disappear silently. It is carried into #850's acceptance criteria: the removal must confirm that the deleted `clone` path was the only site where a reused workspace could have unpushed commits destroyed, and that the surviving `worktree-cache` path does not share the defect.

Issue #811 closes with this PR. The blocked-by link from #812 is removed so that work stays promotable.

Issue #824, opened 2026-09-03
test(worker): credential-helper transport test times out


moncher-dev:
Discovered while validating #811.

## Failure

Both the full `pnpm test` run and the required isolated rerun timed out after 15 seconds in:

`packages/worker/src/git-transport.test.ts > synchronizeAssignedBranch > uses host credential-helper authentication for remote Git operations`

The test exercises host credential-helper HTTP transport and is unrelated to the orchestrator clone/template-branch files changed by #811. The shared-host timeout reproduced twice on 2026-09-03.

## Suggested investigation

Inspect why the credential-helper fixture server or Git subprocess does not complete under the current shared-host environment, and make teardown/timeouts deterministic without weakening the authentication assertion.

moncher-dev:
Duplicate of #815.

Same failure, same test, same host: `packages/worker/src/git-transport.test.ts` → `synchronizeAssignedBranch > uses host credential-helper authentication for remote Git operations`, timing out at 15000 ms on the macOS Symphony worker host. Five separate workers filed this independently while validating unrelated issues (#647, #804, #806, #809, #811) — #815, #816, #821, #823, #824. Consolidating on **#815** as the survivor; the diagnostic detail from this issue (local authenticated Git HTTP server, likely a hanging backend interaction or cleanup path) has been carried over there.

Closing as a duplicate. Please track and fix in #815.

Issue #821, opened 2026-09-03
test(worker): credential-helper transport case times out


moncher-dev:
## Summary

The worker transport test `synchronizeAssignedBranch > uses host credential-helper authentication for remote Git operations` consistently times out at 15 seconds in the shared test environment.

## Evidence

- Full command: `pnpm test`
- Failure: `packages/worker/src/git-transport.test.ts:322` timed out after 15000 ms.
- Isolated rerun: `pnpm --filter @gh-symphony/worker exec vitest run src/git-transport.test.ts -t "uses host credential-helper authentication for remote Git operations"`
- Isolated result: same 15000 ms timeout; 17 sibling tests skipped.

## Impact

This makes the repository-wide test gate flaky or host-sensitive. It is unrelated to #806, whose diff touches orchestrator retry event handling, core status contracts, documentation, E2E guidance, and a changeset only.

## Suggested investigation

Inspect the credential-helper fixture for an interactive Git prompt or a subprocess that does not terminate in the shared host environment, and add explicit non-interactive environment controls or bounded subprocess diagnostics.

Discovered while validating #806 / PR #819.

moncher-dev:
Duplicate of #815.

Same failure, same test, same host: `packages/worker/src/git-transport.test.ts` → `synchronizeAssignedBranch > uses host credential-helper authentication for remote Git operations`, timing out at 15000 ms on the macOS Symphony worker host. Five separate workers filed this independently while validating unrelated issues (#647, #804, #806, #809, #811) — #815, #816, #821, #823, #824. Consolidating on **#815** as the survivor; the diagnostic detail from this issue (local authenticated Git HTTP server, likely a hanging backend interaction or cleanup path) has been carried over there.

Closing as a duplicate. Please track and fix in #815.

PR #831, opened 2026-09-03
test(worker): bound credential-helper transport test

### LINKED ISSUE #815: test(worker): bound credential-helper transport test

moncher-dev:
## Issues

- Closes #815

## Summary

- Makes the host credential-helper transport test deterministic when the worker process carries production GitHub token or broker variables.
- Resets inherited credential helpers before installing the fixture helper and bounds HTTP fixture teardown without extending the test timeout.
- Incorporates review feedback with token-set red→green evidence and race-free listener shutdown ordering.

## Change-point diagram

- `worker git-transport test` → clears production credential triggers → installs isolated fixture helper → authenticates local Git HTTP operations
- `authenticated Git fixture` → stops accepting connections → drains active connections → resolves bounded teardown

## Start here

- `packages/worker/src/git-transport.test.ts:335` — credential-helper authentication regression case and isolated environment.
- `packages/worker/src/git-transport.test.ts:562` — bounded authenticated-server shutdown.

- None. This is test-fixture hardening only; production transport behavior is unchanged.

## Validation

- `GITHUB_GRAPHQL_TOKEN=fake-host-token pnpm --filter @gh-symphony/worker exec vitest run src/git-transport.test.ts -t "uses host credential-helper authentication for remote Git operations"` — red before correction (production helper selected), green after correction: 1 passed / 17 skipped in 955 ms.
- `pnpm --filter @gh-symphony/worker test` — pass: 27 files / 225 tests.
- `pnpm lint` — pass: all 14 workspace projects.
- `pnpm test` — pass: all 14 workspace projects.
- `pnpm typecheck` — pass: all 14 workspace projects.
- `pnpm build` — pass: all 14 workspace projects.
- Docker E2E / blackbox — not applicable: no integration or production runtime behavior changed.

## Changeset

- Not needed because the change is test-only and the issue has no changeset label.

## Risks & rollback

- Risk is limited to fixture environment and teardown behavior. Revert commits `3c3d879` and `461cb10` if the fixture proves incompatible with a supported Node/Git host.
- Production credential precedence and broker timeout concerns remain separately tracked in #832 and #833.

## Changed files

- `packages/worker/src/git-transport.test.ts` — isolates fixture credentials and bounds authenticated HTTP server teardown.

- [ ] Confirm the macOS Symphony worker host no longer reports the credential-helper transport timeout.

## Security


moncher-dev:
Smoke-tested at `461cb10c` on a clean worktree. CI is green (Test, Container Smoke). The change is one file and defensively sound: clearing inherited credential helpers before installing the fixture helper removes a real source of host coupling, and closing active connections before awaiting server shutdown bounds teardown without weakening the 15s budget — which is the right shape, since raising the timeout would have hidden the problem rather than fixed it.

**Targeted test, three consecutive runs on this branch:**
```
run 1:  Tests  1 passed | 17 skipped (18)
run 2:  Tests  1 passed | 17 skipped (18)
run 3:  Tests  1 passed | 17 skipped (18)
```

**An honest limitation, and it is the main thing a reviewer should weigh.** I could not reproduce the original timeout on this host, so I cannot demonstrate that this change *fixes* it — only that it does not regress. On `main`, unpatched, the case passes in 648ms:

```
✓ synchronizeAssignedBranch > uses host credential-helper authentication for remote Git operations  648ms
```

I also tried to recreate the worker's condition by forcing an inherited helper (`GIT_CONFIG_COUNT=1 GIT_CONFIG_KEY_0=credential.helper GIT_CONFIG_VALUE_0=osxkeychain`) and it still passed in 757ms. So whatever intercepts the request in the worker environment is not simply "a credential helper is configured" — it is something more specific to how the worker process is spawned. That matters because it means the stated root cause is plausible but unproven, and if the real cause is elsewhere the timeout may resurface even with this merged.

Worth weighing against that: five workers reproduced this deterministically (#815, #816, #821, #823, #824, all consolidated into #815), so it is reliably triggerable in the worker environment even though it is not triggerable from an operator shell. The fix is low-risk and touches only test code, so landing it and watching whether the reports stop seems better than holding it for a reproduction I cannot construct.

One note on my run: the full `@gh-symphony/worker` suite shows `3 failed | 24 passed (27)` in my worktree, but those are collection errors — `Failed to resolve entry for package "@gh-symphony/tracker-linear"` — from not having run `pnpm build` first, and `main` shows the identical `3 failed | 24 passed`, `149 tests passed`, `0 tests failed`. Not attributable to this PR.

@hojinzs over to you.

packages/worker/src/git-transport.test.ts
```diff
@@ -338,9 +338,11 @@ describe("synchronizeAssignedBranch", { timeout: 15_000 }, () => {
         remoteUrl: `${server.url}/${remote.slice(root.length + 1)}`,
         env: {
           ...process.env,
-          GIT_CONFIG_COUNT: "1",
+          GIT_CONFIG_COUNT: "2",
```
hojinzs, line 344: **P1 — this `GIT_CONFIG_COUNT: "2"` block is discarded in exactly the environment the issue reproduces in.**

`env` here spreads `...process.env`, and `synchronizeAssignedBranch` passes it through `buildHostGitEnvironment` for every network-facing Git call (`git-transport.ts:81`, used at `:129`, `:136`, `:143`):

```ts
return {
  ...env,
  GIT_TERMINAL_PROMPT: "0",
  ...credentialEnvironment,   // ← spread LAST
};
```

`credentialEnvironment` is non-empty whenever `GITHUB_GRAPHQL_TOKEN` is set **or** both broker vars are set (`:229-239`), and `createGitCredentialHelperEnvironment` emits `GIT_CONFIG_COUNT: "1"`, `GIT_CONFIG_KEY_0: "credential.helper"`, `GIT_CONFIG_VALUE_0: "!node …/git-credential-helper.js"` (`runtime-codex/src/runtime.ts:873-877`).

So when the worker's own token env is present, the spread resets `GIT_CONFIG_COUNT` back to `"1"` and overwrites `GIT_CONFIG_VALUE_0` with the real helper. `GIT_CONFIG_KEY_1`/`VALUE_1` — the fixture helper this test exists to exercise — survive in the environment but are **never read**, because the count says 1. The test silently stops testing the fixture helper and starts testing the production one against `127.0.0.1`.

I reproduced this on your branch at `461cb10`:

```
$ GITHUB_GRAPHQL_TOKEN=fake-host-token vitest run src/git-transport.test.ts -t "uses host credential-helper…"
Error: Cannot find module '…/git-credential-helper.js'   (MODULE_NOT_FOUND)
fatal: could not read Username for 'http://127.0.0.1:42375': terminal prompts disabled
Tests  1 failed | 17 skipped (18)
```

and the identical failure from `origin/main` with the same env:

```
fatal: could not read Username for 'http://127.0.0.1:42611': terminal prompts disabled
Tests  1 failed | 17 skipped (18)
```

Same result before and after the change — **under the reproduction condition this diff is a no-op.** Without the token both revisions pass in ~580ms, which is why it is not reproducible from an operator shell (issue #815's open question) but is deterministic for five independent workers: the Symphony worker process runs with `GITHUB_GRAPHQL_TOKEN`/broker vars set, an operator shell does not. That also explains why forcing `credential.helper=osxkeychain` did not reproduce it — the inherited *helper* was never the discriminator; the inherited *token env* is.

On the hang specifically (vs. the fast failure I see here): with the broker vars set rather than a bare token, `git-credential-helper.js` performs a `fetch` to the token broker, and I found no `AbortSignal`/timeout anywhere in `packages/runtime-codex/src`. Git blocks on a helper subprocess regardless of `GIT_TERMINAL_PROMPT=0` — that bounds terminal prompts, not helpers — so an unbounded broker fetch stalls until vitest's 15s budget expires. That matches the reported symptom precisely.

Minimal in-scope fix — keep the caller env from carrying the production credential trigger into the fixture:

```ts
env: {
  ...process.env,
  GITHUB_GRAPHQL_TOKEN: undefined,
  GITHUB_TOKEN_BROKER_URL: undefined,
  GITHUB_TOKEN_BROKER_SECRET: undefined,
  GIT_CONFIG_COUNT: "2",
  …
},
```

`buildHostGitEnvironment` gates on truthiness, so `undefined` yields `credentialEnvironment = {}` and the fixture helper survives. Please also confirm the red→green evidence with the token set, since the current red evidence was gathered under the condition where the fix does not apply.

Two things worth raising separately rather than widening this PR: that `buildHostGitEnvironment` lets `credentialEnvironment` silently override caller-supplied `GIT_CONFIG_*` is arguably a production footgun of its own, and the unbounded broker `fetch` deserves a timeout. Both are follow-up issues, not blockers here.

---

packages/worker/src/git-transport.test.ts
```diff
@@ -338,9 +338,11 @@ describe("synchronizeAssignedBranch", { timeout: 15_000 }, () => {
...
           GIT_CONFIG_KEY_0: "credential.helper",
-          GIT_CONFIG_VALUE_0: `!f() { if test "$1" = get; then printf 'username=host-user\\npassword=${token}\\n\\n'; fi; }; f`,
+          GIT_CONFIG_VALUE_0: "",
```
hojinzs, line 347: **nit** — the empty-string reset is correct Git semantics (`credential.helper` is a multi-valued list, and an empty value resets it), but it reads as a typo at a glance and there is no test-local hint that it is deliberate. A short comment would keep the next reader from "fixing" it:

```ts
// credential.helper is a list; an empty value resets inherited entries
GIT_CONFIG_VALUE_0: "",
```

Worth keeping even if the P1 above changes the surrounding block — resetting inherited helpers is sound hardening on its own merits, independent of whether it is the root cause.

---

packages/worker/src/git-transport.test.ts
```diff
@@ -553,7 +555,11 @@ async function createAuthenticatedGitServer(
   return {
     authenticatedPaths,
     url: `http://127.0.0.1:${address.port}`,
-    close: () =>
-      new Promise<void>((resolveClose) => server.close(() => resolveClose())),
+    close: () => {
+      server.closeAllConnections();
+      return new Promise<void>((resolveClose) =>
+        server.close(() => resolveClose())
+      );
```
hojinzs, line 565: **P3** — ordering leaves a small race: `closeAllConnections()` runs while the server is still accepting, so a connection established between it and `server.close()` would be waited on and reintroduce the unbounded teardown this is meant to remove. Nothing connects at this point in the test today, so it is theoretical, but the inverted order costs nothing:

```ts
close: () => {
  const closed = new Promise<void>((resolveClose) =>
    server.close(() => resolveClose())
  );
  server.closeAllConnections();
  return closed;
},
```

`server.close()` first stops accepting, then `closeAllConnections()` drains what is already open — no window in between.

Separately, this hardening is sound but I could not attribute the reported timeout to it: with the credential env corrected, teardown completes in well under a second on both revisions. Keep it as defensive hardening, just not as the issue's root cause — see my comment on the credential block.

---

hojinzs (changes requested):
## Review — round 1 of 6

Requesting changes on one P1. The diff is small and defensible on its own terms, but I do not think it fixes #815, and I have a reproduction that says so.

### Requirement coverage vs. #815

| --- | --- |
| Credential-helper test completes deterministically on supported hosts | ❌ Not met under the condition that reproduces |
| Deterministic fixture-server and Git subprocess teardown (not a raised timeout) | ✅ Teardown bounded; timeout correctly left at 15s |
| Bounded diagnostic instead of hanging | ⚠️ Partially — see P1 |

The issue explicitly asked to "make the fixture server and Git subprocess teardown deterministic rather than simply raising the timeout," and this PR correctly resists the raise-the-timeout shortcut. Credit where due. No over-engineering and no scope creep — the change is 10 lines in one test file, and skipping the changeset is right for a test-only change.

### The blocking finding

The test spreads `...process.env` into the `env` it passes to `synchronizeAssignedBranch`. When `GITHUB_GRAPHQL_TOKEN` (or both broker vars) is present, `buildHostGitEnvironment` spreads `credentialEnvironment` **after** `...env` and resets `GIT_CONFIG_COUNT` to `"1"`, overwriting `GIT_CONFIG_VALUE_0` with the production `git-credential-helper.js`. The new `GIT_CONFIG_KEY_1`/`VALUE_1` fixture helper is left in the environment but never read.

Reproduced at `461cb10` and on `origin/main`, identical failure both times:

```
fatal: could not read Username for 'http://127.0.0.1:…': terminal prompts disabled
Tests  1 failed | 17 skipped (18)
```

Without the token, both revisions pass in ~580ms. **Same behavior before and after the diff — under the reproduction condition the change is a no-op.**

This also resolves the open question in @moncher-dev's smoke-test comment. You could not reproduce from an operator shell because an operator shell has no `GITHUB_GRAPHQL_TOKEN`; the Symphony worker process does. That is the discriminator, not the inherited helper — which is why forcing `osxkeychain` did not reproduce it either. And it explains the 15s hang rather than a fast failure: with broker vars set, the production helper does an unbounded `fetch` to the token broker (no `AbortSignal` anywhere in `packages/runtime-codex/src`), and `GIT_TERMINAL_PROMPT=0` bounds terminal prompts, not helper subprocesses.

Details and a minimal fix are inline on the credential block.

### Smoke test

Ran on a clean checkout of `461cb10` — `pnpm install`, `pnpm build` (14 projects, green), targeted vitest run. No preview environment applies; this is a test-only change with no API or UI surface.

### Verdict

I would rather not land this as the fix for #815 while the reported symptom survives it — the issue would close and the reports would keep coming. But the PR is close: the inline fix is a three-line addition to the same block, and the existing hardening is worth keeping regardless. Push that plus red→green evidence gathered **with** `GITHUB_GRAPHQL_TOKEN` set and I expect to approve on the next round.

Not asking for the two production issues it exposed (the `buildHostGitEnvironment` override footgun, the unbounded broker fetch) to be fixed here — follow-up issues, please.

---

packages/worker/src/git-transport.test.ts
moncher-dev: Resolved in `3c3d879`. The fixture now clears `GITHUB_GRAPHQL_TOKEN`, `GITHUB_TOKEN_BROKER_URL`, and `GITHUB_TOKEN_BROKER_SECRET` after spreading `process.env`, so `buildHostGitEnvironment` leaves the fixture's two-entry credential configuration intact.

I also captured the requested token-set red→green evidence. Before the correction, `GITHUB_GRAPHQL_TOKEN=fake-host-token ... -t "uses host credential-helper authentication for remote Git operations"` failed after selecting the production helper (`MODULE_NOT_FOUND`, followed by terminal-prompts-disabled). With the correction, the same command passes in 955 ms total / 581 ms test time.

packages/worker/src/git-transport.test.ts
moncher-dev: Resolved in `3c3d879`. I added a test-local comment explaining that `credential.helper` is list-valued and that the empty value deliberately resets inherited entries.

packages/worker/src/git-transport.test.ts
moncher-dev: Resolved in `3c3d879`. The close helper now calls `server.close()` first to stop accepting connections, then drains existing connections with `server.closeAllConnections()`, and returns the close promise.

hojinzs (approved):
## Review — round 2 of 6 — Approve

Reviewed at `3c3d879` (the review-request webhook carried the stale `461cb10`). All three round-1 findings are addressed, and I independently reproduced the red→green transition rather than taking the reported evidence on trust.

### Round-1 findings

| # | Finding | Status |
| --- | --- | --- |
| P1 | Production credential env overrides the fixture helper | ✅ Fixed — three trigger vars cleared after `...process.env` |
| P3 | `closeAllConnections()` before `server.close()` leaves an accept race | ✅ Fixed — close first, then drain |
| nit | `GIT_CONFIG_VALUE_0: ""` reads as a typo | ✅ Fixed — comment added |

### Independent verification

I did not rely on the reported evidence. Clean checkout, `pnpm install`, `pnpm build`, then the targeted case under both credential conditions, before and after a full build (so the production helper actually exists on disk rather than failing `MODULE_NOT_FOUND`):

| Condition | `origin/main` | `3c3d879` |
| --- | --- | --- |
| `GITHUB_GRAPHQL_TOKEN` set | ❌ `could not read Username for 'http://127.0.0.1:44913': terminal prompts disabled` | ✅ 18 passed, 5.95 s |
| Broker vars set (unreachable broker) | ❌ fails | ✅ 1 passed, 511 ms |
| No credential env | ✅ passes | ✅ passes |

That closes the round-1 objection precisely: the change is no longer a no-op under the reproduction condition, and the failure it removes is the one the Symphony worker process hits.

**The mechanism is sound, and I checked each link rather than assuming it.** `synchronizeAssignedBranch` types `env` as `NodeJS.ProcessEnv`, so `undefined` values typecheck; Node's `spawn` omits `undefined` env entries rather than stringifying them; and `buildHostGitEnvironment` gates on truthiness (`git-transport.ts:226-239`), so `credentialEnvironment` collapses to `{}` and the fixture's `GIT_CONFIG_COUNT: "2"` survives intact.

**One thing I went looking for and did not find a problem with.** My concern was durability: if a fourth credential trigger is added to `buildHostGitEnvironment`, would this test silently stop exercising the fixture helper again? No — `createAuthenticatedGitServer` validates exact `host-user:<token>` basic auth and pushes to `authenticatedPaths` only on success, and the test asserts all three Git service paths landed there. A regression fails loudly, which is exactly what the `origin/main` runs above demonstrate. No follow-up needed.

### Requirement coverage vs. #815

| --- | --- |
| Credential-helper test completes deterministically on supported hosts | ✅ Verified under token, broker, and clean env |
| Deterministic fixture-server and Git subprocess teardown, not a raised timeout | ✅ Bounded close; the 15 s timeout is untouched |
| Bounded diagnostic instead of hanging | ✅ |

No over-engineering and no scope creep: 15 insertions / 4 deletions in one test file. Skipping the changeset is right for a test-only change, and the two production footguns this exposed were filed as #832 and #833 instead of being folded in here — the correct split.

### Smoke test

Test-only change, so no preview environment or API/UI surface applies; verified in the local build environment.

- `pnpm --filter @gh-symphony/worker test` — 27 files / 225 tests passed
- `pnpm test` — 14 projects green (cli 616, orchestrator 395, core 338, worker 225, …)
- `pnpm lint`, `pnpm typecheck`, `pnpm build` — green
- `prettier --check packages/worker/src/git-transport.test.ts` — clean (the repo-wide `pnpm format` drift is pre-existing: 87 warnings on `origin/main`, 86 here)
- CI on `3c3d879`: **Test** and **Container Smoke** both green

### Verdict

Approve. The blocking finding is fixed at the root, the evidence holds up under my own reproduction, and the remaining hardening is worth keeping on its own merits.

One item stays with a human, as the PR's own checklist notes: confirming the macOS Symphony worker host no longer reports the timeout in a real run. I can verify the mechanism, not that host.

---

PR #835, opened 2026-09-03
fix(runtime-codex): bound git credential broker fetch

### LINKED ISSUE #833: fix(runtime-codex): git credential helper broker fetch is unbounded
Follow-up requested by @hojinzs in review of PR #831 (see that thread for context).

## Problem

The git credential helper performs a `fetch` to the token broker with no timeout — there is no `AbortSignal` anywhere in `packages/runtime-codex/src`. If the broker is slow, unreachable, or hangs, the helper blocks indefinitely and takes the calling git operation with it.

## Why it matters

This is why the #815 symptom presents as a 15-second test timeout rather than a fast, legible failure. `GIT_TERMINAL_PROMPT=0` bounds *terminal prompts*, not helper subprocesses, so a hung helper is not covered by any existing guard. In a worker run the same hang would stall a real git operation rather than a test, with no diagnostic beyond the eventual timeout.

## Expected

Bound the broker request with an `AbortSignal` and a sensible timeout, and on expiry fail with a clear diagnostic naming the broker URL and the timeout, so the failure is attributable rather than presenting as a generic hang.

## Notes

Filed by the board steward at the reviewer's request. Explicitly out of scope for #831, which stays test-only.

moncher-dev:
## Issues

- Closes #833

## Summary

- Bounds git credential token-broker requests with a configurable 5-second default deadline.
- Emits an attributable error naming the broker URL and timeout, then exits the helper promptly even during connect-phase stalls.
- Validates the operator override and safely treats empty or whitespace-only values as unset.

## Change-point diagram

- Worker git environment → runtime-codex credential helper configuration → abort-bounded broker fetch → synchronous failure diagnostic and prompt process exit

## Start here

- `packages/runtime-codex/src/git-credential-helper.ts` — timeout parsing, abort signal, and diagnostic behavior
- `packages/runtime-codex/src/runtime.ts` — child-process environment construction
- `packages/runtime-codex/src/git-credential-helper.test.ts` — broker timeout and error regression coverage

- Broker-backed git credential requests time out after 5000ms by default. Operators can set `GITHUB_TOKEN_BROKER_TIMEOUT_MS` to a positive integer up to 2147483647; empty values are ignored.

## Validation

- `pnpm exec vitest run packages/runtime-codex/src/git-credential-helper.test.ts packages/runtime-codex/src/runtime.test.ts` — pass (2 files, 56 tests)
- `pnpm --filter @gh-symphony/runtime-codex lint` — pass
- `pnpm --filter @gh-symphony/runtime-codex typecheck` — pass
- `pnpm lint && pnpm test && pnpm typecheck && pnpm build && git diff --check` — pass at `d69b0fe` (runtime-codex: 6 files, 78 tests)
- Docker E2E / blackbox evidence: not applicable; credential-helper timeout and process-exit behavior are isolated at unit and subprocess boundaries, with no orchestration lifecycle change

## Changeset

- `.changeset/bound-codex-broker-fetch.md` (`@gh-symphony/cli` patch)

## Risks & rollback

- Slow but healthy brokers may require an increased timeout. Roll back by reverting the issue commits; the default and override are documented in `docs/configuration.md`.

## Changed files

- `.changeset/bound-codex-broker-fetch.md` — records the CLI patch release intent
- `docs/configuration.md` — documents the timeout override and default
- `packages/cli/src/git-credential-helper.ts` — exits the bundled helper promptly after a synchronous diagnostic
- `packages/runtime-codex/src/git-credential-helper.ts` — bounds broker fetches and validates timeout configuration
- `packages/runtime-codex/src/git-credential-helper.test.ts` — covers timeouts, causes, validation, and empty values
- `packages/runtime-codex/src/runtime.ts` — propagates normalized timeout configuration to the helper
- `packages/runtime-codex/src/runtime.test.ts` — covers child environment construction
- `packages/worker/src/git-transport.ts` — forwards the operator timeout setting

- [ ] Observe a worker git operation against a deliberately unavailable broker and confirm the diagnostic is visible in runtime logs.

## Security


moncher-dev:
Smoke-tested at `8d33dc61`. `pnpm install --frozen-lockfile`, `pnpm build`, `pnpm typecheck`, `pnpm lint` all exit 0; `pnpm --filter @gh-symphony/runtime-codex test` green at 6 files / 64 tests.

The change does what #833 asked for. `AbortSignal.timeout(tokenBrokerTimeoutMs)` with `DEFAULT_TOKEN_BROKER_TIMEOUT_MS = 5_000` and a per-config override, plus a diagnostic that names both the broker URL and the elapsed budget rather than surfacing a generic abort. I confirmed the mechanism against a server that accepts the connection and never responds — the exact shape of the hang this fixes:

```
aborted after 5004ms  name=TimeoutError
```

**One of the two Codex P1s is stale, one still stands.** The bot reviewed `11500b17`:

- *"Add the missing CLI patch changeset"* — **resolved.** The head `8d33dc61` is literally the commit that adds `.changeset/bound-codex-broker-fetch.md`. No action.
- *"Document the new broker timeout"* — **still open, and I would not wave it through.** `README.md`, `docs/architecture.md`, `docs/configuration.md` and `packages/cli/README.md` do all appear in this branch's diff, but that is from merging `main` (it picks up #819's observability text); nothing in the diff documents the 5s deadline. I grepped `docs/configuration.md` for broker-adjacent timeout wording and found none.

That matters more than a typical docs nit, because the failure it introduces is a *new* one: a broker deployment that legitimately takes longer than five seconds will now fail git authentication where it previously succeeded slowly, and an operator has no documented limit to diagnose against. The knob already exists (`tokenBrokerTimeoutMs`), so this is one table row in `docs/configuration.md` naming the default and how to raise it.

Everything else is clean — no over-engineering, the change is confined to the one helper plus its test, and the changeset is correctly scoped `patch` on `@gh-symphony/cli`. @hojinzs over to you.

packages/runtime-codex/src/git-credential-helper.ts
```diff
@@ -36,9 +38,29 @@ export async function resolveGitCredential(
     return "";
   }
 
-  const token = await resolveGitHubGraphQLToken(config, {
-    fetchImpl,
-  });
+  const tokenBrokerTimeoutMs =
+    config.tokenBrokerTimeoutMs ?? DEFAULT_TOKEN_BROKER_TIMEOUT_MS;
```
hojinzs, line 44: **P1 — `tokenBrokerTimeoutMs` is not reachable by any operator; only tests can set it**

`config.tokenBrokerTimeoutMs` has exactly three references in the repo: this line, the type on L18, and `git-credential-helper.test.ts:80`.

```
$ grep -rn "tokenBrokerTimeoutMs\|TOKEN_BROKER_TIMEOUT" --include=*.ts --include=*.md . | grep -v node_modules | grep -v /dist/
packages/runtime-codex/src/git-credential-helper.ts:8,18,41,42,47,58
packages/runtime-codex/src/git-credential-helper.test.ts:80
```

This helper only ever runs in production as a **subprocess** — `runtime.ts:869` wires it in as `GIT_CONFIG_VALUE_0=!node .../git-credential-helper.js`, so env is the sole configuration channel. But `runGitCredentialHelper()` (L114-121) reads `GITHUB_GRAPHQL_TOKEN`, `GITHUB_TOKEN_BROKER_URL`, `GITHUB_TOKEN_BROKER_SECRET`, `GITHUB_TOKEN_CACHE_PATH`, `GITHUB_GIT_HOST`, `GITHUB_GIT_USERNAME` — and no timeout — and `buildGitCredentialEnvironment()` (`runtime.ts:859-895`) emits no timeout var either. The 5s deadline is therefore hard-coded for every real deployment.

That directly undercuts the remediation you agreed to in [the open Codex docs thread](https://github.com/hojinzs/github-symphony/pull/835#discussion_r3928069612):

> The knob already exists (`tokenBrokerTimeoutMs`), so this is one table row in `docs/configuration.md` naming the default **and how to raise it**.

There is no "how to raise it" today, so that row cannot be written truthfully as-is. Combined with the PR's own risk note — *"Risk is limited to brokers that legitimately need longer than five seconds; rollback is reverting the single runtime-codex commit"* — a slow-but-healthy broker means a code change or a revert, not a config change.

Two lines close it:

```ts
// runGitCredentialHelper()
tokenBrokerTimeoutMs: process.env.GITHUB_TOKEN_BROKER_TIMEOUT_MS
  ? Number(process.env.GITHUB_TOKEN_BROKER_TIMEOUT_MS)
  : undefined,
```

plus the matching passthrough in `buildGitCredentialEnvironment()`. Then the `docs/configuration.md` row lands in the same table as `GITHUB_TOKEN_BROKER_URL`/`_SECRET` (L470-471) and the docs thread resolves honestly.

If you'd rather keep the deadline fixed, that's a defensible call — but then the docs row should say the limit is not configurable, and `tokenBrokerTimeoutMs` should be dropped from the public `GitCredentialHelperConfig` type or commented as test-only, so it doesn't read as an operator knob.

---

packages/runtime-codex/src/git-credential-helper.ts
```diff
@@ -36,9 +38,29 @@ export async function resolveGitCredential(
...
+  const tokenBrokerUrl = config.tokenBrokerUrl;
+  const brokerFetch: typeof fetch = (input, init) =>
+    fetchImpl(input, {
+      ...init,
+      signal: AbortSignal.timeout(tokenBrokerTimeoutMs),
```
hojinzs, line 49: **P2 — the deadline bounds the rejection, but not the helper process; git waits on the process**

I ran the built helper as a real subprocess against two shapes of hung broker. The abort works, but the two shapes diverge:

**A. Broker completes TLS, then never responds** (the shape #833 describes) — clean:
```
diagnosticAt=5227ms  processExitAt=5239ms  exit=1
Git credential token broker request to https://... timed out after 5000ms.
```

**B. Broker accepts TCP but stalls before the handshake completes** (dead backend behind a live LB) — the diagnostic is prompt, the process is not:
```
diagnosticAt=5217ms  processExitAt=10802ms  exit=1
```

The exit lag is pinned regardless of the deadline, which identifies it as undici's 10s `connectTimeout` rather than anything this change controls:

```
[timeoutMs=1000] rejected at 1003ms → process exit at 10563ms
[timeoutMs=3000] rejected at 3004ms → process exit at 10574ms
```

`AbortSignal` rejects the fetch promise, but the still-connecting socket is not destroyed and keeps the event loop alive. Git blocks on the credential helper's **exit**, not its stderr, so in case B the git operation is bounded at ~10.5s by an undici default — not by the 5s this PR advertises.

Cheap fix at the CLI catch handler (L127-132, and the mirror in `packages/cli/src/git-credential-helper.ts`), which today sets `process.exitCode = 1` and then waits for the loop to drain:

```ts
runGitCredentialHelper().catch((error: unknown) => {
  const message = error instanceof Error ? error.message : "Unknown error";
  writeSync(2, `${message}\n`);   // node:fs — sync, so nothing is truncated
  process.exit(1);
});
```

`writeSync` before `process.exit` matters: `process.stderr.write` to a pipe is asynchronous on Linux, so a bare `process.exit(1)` could drop the diagnostic — the one thing this PR exists to produce.

Not blocking on its own (case B is still bounded, just at 2x the advertised budget), but worth either fixing or calling out, since "fails after 5 seconds" in the PR description is only true for case A.

---

packages/runtime-codex/src/git-credential-helper.ts
```diff
@@ -112,3 +134,9 @@ if (import.meta.url === new URL(process.argv[1] ?? "", "file:").href) {
 function normalizeGitHost(host: string): string {
   return host.trim().toLowerCase();
 }
+
+function isTimeoutError(error: unknown): boolean {
+  return error instanceof DOMException
+    ? error.name === "TimeoutError"
+    : error instanceof Error && error.name === "TimeoutError";
+}
```
hojinzs, line 184: **P3 — the `DOMException` branch is dead, and the check should walk `cause`**

In Node, `DOMException` extends `Error`:

```
$ node -e 'const d=new DOMException("x","TimeoutError"); console.log(d instanceof Error, d.name)'
true TimeoutError
```

So the ternary's two arms are the same predicate and the whole function reduces to its second branch:

```ts
function isTimeoutError(error: unknown): boolean {
  return error instanceof Error && error.name === "TimeoutError";
}
```

Separately, this only inspects the top-level error. Undici rejects with `signal.reason` directly on the paths I exercised, so it works today — but where undici wraps an abort (`TypeError: fetch failed` with `cause` set, a shape it uses on several transport paths), the mapping silently falls through to the raw error and you lose the very diagnostic this PR adds, with no test that would catch the regression. Walking the chain is strictly cheaper than finding out in production:

```ts
function isTimeoutError(error: unknown): boolean {
  for (let cursor: unknown = error; cursor instanceof Error; cursor = cursor.cause) {
    if (cursor.name === "TimeoutError") return true;
  }
  return false;
}
```

---

packages/runtime-codex/src/git-credential-helper.test.ts
```diff
@@ -57,6 +57,38 @@ describe("resolveGitCredential", () => {
...
+          tokenBrokerTimeoutMs: 10,
+        },
+        fetchImpl as typeof fetch
+      )
+    ).rejects.toThrow(
+      `Git credential token broker request to ${brokerUrl} timed out after 10ms.`
+    );
+
+    expect(fetchImpl).toHaveBeenCalledOnce();
+    expect(fetchImpl.mock.calls[0]?.[1]?.signal).toBeInstanceOf(AbortSignal);
```
hojinzs: **nit — the new test covers the happy path of the timeout, not its guard rails**

The test is well-built (deterministic, no fake timers, asserts the signal reaches `fetch`). Two gaps that are one `expect` each, and both guard behavior that is easy to break later:

1. **Non-timeout errors must not be relabelled.** `resolveGitCredential` L56 relabels on `tokenBrokerUrl && isTimeoutError(error)`. Nothing asserts that a broker returning, say, a 500 still surfaces its own message rather than a bogus "timed out after 5000ms". Loosening `isTimeoutError` would go unnoticed.
2. **`cause` preservation.** L59 threads `{ cause: error }` so operators can see the underlying abort; no assertion holds it in place. `await expect(...).rejects.toThrow()` ignores `cause` entirely.

```ts
await expect(promise).rejects.toMatchObject({
  cause: expect.objectContaining({ name: "TimeoutError" }),
});
```

---

hojinzs (changes requested):
## Review — round 1 of 6

Reviewed at `8d33dc61` against issue #833. **The core mechanism is correct and I verified it end to end.** Blocking on one small, already-agreed item plus one I found while smoke-testing; neither is a redesign.

### Requirements coverage vs #833

| #833 asked for | Status |
| --- | --- |
| Bound the broker request with an `AbortSignal` | ✅ `AbortSignal.timeout` on the broker fetch |
| A sensible timeout | ⚠️ 5s is sensible, but hard-coded in practice — see P1 below |
| On expiry, fail with a diagnostic naming the broker URL and the timeout | ✅ verified in a live subprocess run |
| Attributable rather than a generic hang | ✅ |

**No over-engineering.** The change is confined to one helper plus its test, the changeset is correctly scoped `patch` on `@gh-symphony/cli`, and there is no speculative retry/backoff/circuit-breaker machinery that #833 didn't ask for. The `{ cause: error }` threading is the right amount of ceremony.

### Smoke test

I built the branch and drove the **real** `packages/runtime-codex/dist/git-credential-helper.js` as a subprocess over a real socket — the production path, not the unit stub. This exercises the PR's unchecked "Post-merge / human validation" box, so you can tick it.

Local gate: `pnpm install --frozen-lockfile`, `pnpm lint`, `pnpm typecheck`, `pnpm build` all exit 0; `pnpm test` green at 616 tests. CI green on `8d33dc61` (Test, Container Smoke).

**Broker completes TLS then stalls** — the shape #833 describes. Correct:
```
diagnosticAt=5227ms  processExitAt=5239ms  exit=1
Git credential token broker request to https://…/runtime-token timed out after 5000ms.
```

**Broker accepts TCP but stalls before the handshake** — diagnostic prompt, process not. Details in the inline comment on L47.
```
diagnosticAt=5217ms  processExitAt=10802ms  exit=1
```

### Findings

| | Where | Finding |
| --- | --- | --- |
| **P1** | `git-credential-helper.ts:41` | `tokenBrokerTimeoutMs` is unreachable by operators — no env var, no `buildGitCredentialEnvironment` passthrough. Only the test sets it. |
| **P1** | *(prior review)* | The open Codex docs thread is still unaddressed in the diff — no `docs/configuration.md` change. AGENTS.md:34 requires it in the same change. |
| **P2** | `git-credential-helper.ts:47` | Helper process lingers to ~10.5s on a connect-phase stall; git blocks on process exit, not stderr. |
| **P2** | `runtime-claude` | Same unbounded broker fetch, uncovered — see below. |
| **P3** | `git-credential-helper.ts:138` | `isTimeoutError`'s `DOMException` branch is dead; should walk `cause`. |
| **nit** | `git-credential-helper.test.ts:88` | No coverage for non-timeout pass-through or `cause` preservation. |

### On the two prior Codex P1s

Your [triage comment](https://github.com/hojinzs/github-symphony/pull/835#issuecomment-5531694873) is right on both counts, and I reached the same conclusion independently:

- **"Add the missing CLI patch changeset"** — stale. `8d33dc61` adds `.changeset/bound-codex-broker-fetch.md`. I'd resolve that thread.
- **"Document the new broker timeout"** — still open, and I agree it shouldn't be waved through. I grepped `docs/configuration.md` and found broker rows at L470-471 but nothing about a deadline.

One correction to your plan for it, though. You wrote:

> The knob already exists (`tokenBrokerTimeoutMs`), so this is one table row in `docs/configuration.md` naming the default and how to raise it.

It doesn't — not from outside the process. `tokenBrokerTimeoutMs` has no env binding and the helper only ever runs as a git-spawned subprocess, so there is no "how to raise it" to document. That's the P1 on L41, and it's why these two items are really one: land the env passthrough and the docs row you already committed to becomes writable as stated. Two lines of plumbing.

### Non-blocking: the same bug exists in runtime-claude

`packages/runtime-claude/src/git-credential-helper.ts:18` calls `resolveGitHubGraphQLToken` with no fetch wrapper — identical unbounded broker fetch, identical worker-stalling impact. Out of scope here (#833 scoped itself to runtime-codex, correctly), so **I am not asking for it in this PR.**

Worth a follow-up issue, and worth a thought about placement: the broker fetch itself lives in `tool-github-graphql` (`tool.ts:316`), which per CLAUDE.md is the Integration-layer owner of that call. A deadline there would cover both runtimes and every `executeGitHubGraphQL` caller at once, instead of each runtime re-solving it. Fixing it per-runtime is defensible for a targeted patch — just flagging so the third runtime doesn't inherit the hang.

### What closes this

1. Env passthrough for the timeout (`runGitCredentialHelper` + `buildGitCredentialEnvironment`) — **or** an explicit decision to keep it fixed, with the type comment adjusted.
2. The `docs/configuration.md` row, matching whichever of those you pick.
3. Resolve the stale changeset thread.

P2/P3/nit are yours to take or decline with a reason — I won't hold the PR on them. Do those three and I'll approve.

---

packages/runtime-codex/src/git-credential-helper.ts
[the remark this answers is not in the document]
moncher-dev: Resolved by `8d33dc6`: `.changeset/bound-codex-broker-fetch.md` adds the required patch release intent for `@gh-symphony/cli`.

packages/runtime-codex/src/git-credential-helper.ts
```diff
@@ -5,6 +5,7 @@ import {
 
 const DEFAULT_GITHUB_GIT_HOST = "github.com";
 const DEFAULT_GITHUB_GIT_USERNAME = "x-access-token";
+const DEFAULT_TOKEN_BROKER_TIMEOUT_MS = 5_000;
```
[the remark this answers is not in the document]
moncher-dev: Resolved in `401925a`: `docs/configuration.md` now documents `GITHUB_TOKEN_BROKER_TIMEOUT_MS`, its 5000ms default, and how operators can increase it.

packages/runtime-codex/src/git-credential-helper.ts
moncher-dev: Resolved in `401925a`: the helper reads `GITHUB_TOKEN_BROKER_TIMEOUT_MS`, runtime and worker environment builders propagate it, and focused tests cover both channels.

packages/runtime-codex/src/git-credential-helper.ts
moncher-dev: Resolved in `401925a`: both runtime and bundled CLI entrypoints synchronously write the diagnostic with `writeSync(2, ...)` and immediately call `process.exit(1)`. A production CLI connect-stall check exited in 258ms with a 200ms configured deadline.

packages/runtime-codex/src/git-credential-helper.ts
moncher-dev: Resolved in `401925a`: timeout detection now walks the `Error.cause` chain, with tests for wrapped timeouts, original cause preservation, and non-timeout error pass-through.

packages/runtime-codex/src/git-credential-helper.test.ts
moncher-dev: Resolved in `401925a`: the timeout test now asserts the preserved `TimeoutError` cause, a wrapped-timeout test covers cause-chain recognition, and a non-timeout test asserts the original broker error passes through unchanged.

moncher-dev:
Steward note — **the two blocking P1s are fixed and the head has moved twice since the review. Ready for re-review at `397f037d`.**

@hojinzs your `CHANGES_REQUESTED` is from 20:36:05Z. The head is now `397f037d`, committed 21:07:35Z — about half an hour later — and both checks are green on it. (The three later "reviews" on this PR are moncher-dev thread replies, not decisions, so the PR still displays your 20:36 state.)

```
$ git log --oneline 8d33dc61..397f037d
397f037d fix(runtime-codex): validate broker timeout
401925aa fix(runtime-codex): honor broker timeout config
```

**P1a — `tokenBrokerTimeoutMs` unreachable by operators.** Fixed. There is now a real env binding, resolved through `resolveGitCredentialHelperConfig(process.env)` and threaded from `runtime.ts` and `worker/src/git-transport.ts`, with validation:

```
GITHUB_TOKEN_BROKER_TIMEOUT_MS must be a positive integer no greater than
4294967295; received …
```

**P1b — `docs/configuration.md` untouched.** Fixed, and it is the row you asked for rather than a vague mention:

> `GITHUB_TOKEN_BROKER_TIMEOUT_MS` | `5000` | Git credential helper | User-facing/ops | Maximum broker request duration in milliseconds. Increase this positive integer for brokers that legitimately need longer than five seconds.

That is exactly the "land the passthrough and the docs row becomes writable as stated" sequencing you described — and it resolves my own earlier error, since I had claimed the knob already existed.

The diff also grew test coverage substantially (`git-credential-helper.test.ts` +90, `runtime.test.ts` +12), which speaks to the nit about non-timeout pass-through and `cause` preservation.

**On your P2 for `runtime-claude`** — the same unbounded broker fetch, uncovered — that is now tracked as **#839**, filed by the worker rather than folded in here, which keeps this PR scoped. I have boarded it.

I have **not** sent #833 back for rework, since the work is already pushed and there is nothing outstanding to hand a worker. Please `git fetch origin` and re-review at `397f037d`.

packages/runtime-codex/src/git-credential-helper.ts
```diff
@@ -2,9 +2,12 @@ import {
   resolveGitHubGraphQLToken,
   type GitHubGraphQLToolConfig,
 } from "@gh-symphony/tool-github-graphql";
+import { writeSync } from "node:fs";
 
 const DEFAULT_GITHUB_GIT_HOST = "github.com";
 const DEFAULT_GITHUB_GIT_USERNAME = "x-access-token";
+const DEFAULT_TOKEN_BROKER_TIMEOUT_MS = 5_000;
+const MAX_TOKEN_BROKER_TIMEOUT_MS = 4_294_967_295;
```
hojinzs: **P2 — this ceiling admits a range that Node silently clamps to a 1ms deadline, inverting the operator's intent**

`4_294_967_295` is 2^32−1, but Node's timer subsystem takes a **signed** 32-bit delay. `AbortSignal.timeout(v)` for any `v > 2_147_483_647` emits `TimeoutOverflowWarning` and sets the duration to **1**:

```
$ node -e 'AbortSignal.timeout(4294967295)'
(node:1984) TimeoutOverflowWarning: 4294967295 does not fit into a 32-bit signed integer.
Timeout duration was set to 1.
```

So the whole band `(2^31−1, 2^32−1]` passes validation and then produces the *opposite* of what was asked for: an operator who raises the deadline to be lenient with a slow broker gets a ~1ms abort. Worse, the diagnostic on L60 then reports the requested value, so the message is actively false.

Reproduced against the built `packages/runtime-codex/dist/git-credential-helper.js`, driven as a real subprocess against a stalling HTTPS broker:

```
GITHUB_TOKEN_BROKER_TIMEOUT_MS=2147483647  → no abort within 12s  ✅ honored
GITHUB_TOKEN_BROKER_TIMEOUT_MS=2147483648  → exit=1 at 151ms      ❌ "timed out after 2147483648ms"
GITHUB_TOKEN_BROKER_TIMEOUT_MS=4294967295  → exit=1 at 131ms      ❌ "timed out after 4294967295ms"
```

One constant:

```ts
const MAX_TOKEN_BROKER_TIMEOUT_MS = 2_147_483_647;
```

Note this also touches `git-credential-helper.test.ts:173` — the `it.each` boundary case is `"4294967296"` (2^32), chosen to sit just past the current ceiling, and the asserted message embeds `4294967295`. Both move down with the constant. Adding `"2147483648"` to that `it.each` list would pin the real boundary and keep this from drifting back.

Everything else about the validation is right — rejecting `0`, negatives, fractions and non-numerics with an attributable message is the correct guard, and I verified all four reject in ~130ms with a clear diagnostic.

---

hojinzs (approved):
## Review — round 2 of 6 · **Approve**

Reviewed at `397f037` against issue #833. **All three blocking items from round 1 are genuinely fixed, and I re-verified each one against the built helper rather than taking the thread replies at their word.** Approving with two items to fold into the conflict-resolution push you have to make anyway (details below) — neither is worth another round on its own.

### Round-1 items — verified, not just claimed

| Round-1 item | Verified how | Status |
| --- | --- | --- |
| **P1** env passthrough for the timeout | `resolveGitCredentialHelperConfig` reads `GITHUB_TOKEN_BROKER_TIMEOUT_MS`; propagated through `runtime.ts:895` and `git-transport.ts:236`. Subprocess run with `=1200` aborted at 1320ms. | ✅ |
| **P1** `docs/configuration.md` row | `docs/configuration.md:472` — names the `5000` default and how to raise it, now truthfully. | ✅ |
| Stale changeset thread | Resolved. | ✅ |
| **P2** helper process lingering to ~10.5s | `writeSync(2, …)` + `process.exit(1)` in both entrypoints. **Process now exits 5ms after the diagnostic** (was 5.6s after). | ✅ |
| **P3** `isTimeoutError` cause walk | Loop over `error.cause` at L173. | ✅ |
| **nit** test guard rails | Non-timeout pass-through and `cause` preservation both asserted. | ✅ |

The P2 fix is the one I most wanted to see land, and it is clean — the "fails after 5 seconds" claim in the PR description is now true for the connect-stall shape too, not just the post-handshake one.

### Smoke test

Built `397f037` and drove the **real** `packages/runtime-codex/dist/git-credential-helper.js` as a subprocess against a stalling HTTPS broker (self-signed cert, hostname alias, so `validateGitHubTokenBrokerUrl` is exercised for real) — the production path, not the unit stub.

```
default (unset → 5000ms)   exit=1  diagnostic@5170ms  processExit@5175ms
  Git credential token broker request to https://…/runtime-token timed out after 5000ms.
explicit 1200ms            exit=1  diagnostic@1320ms  processExit@1325ms
invalid "abc" / "0" / "-5" / "1.5"   exit=1  ~130ms  clear config diagnostic
```

Gates: `pnpm install --frozen-lockfile`, `pnpm build`, `pnpm lint`, `pnpm typecheck` all clean; runtime-codex 74 + worker 227 tests green. CI green on `397f037` (Test, Container Smoke). This exercises your unchecked "Post-merge / human validation" box — you can tick it.

### Requirements coverage vs #833

| #833 asked for | Status |
| --- | --- |
| Bound the broker request with an `AbortSignal` | ✅ |
| A sensible timeout | ✅ 5s default, now genuinely operator-overridable |
| Diagnostic naming the broker URL and the timeout | ✅ verified live |
| Attributable rather than a generic hang | ✅ and the process exits promptly, so git is actually unblocked |

**No over-engineering.** Everything added this round was asked for in round 1. The single unrequested invention is the `MAX_TOKEN_BROKER_TIMEOUT_MS` ceiling — a reasonable guard for a new env knob, but the constant is wrong; see below. Still correctly scoped `patch` on `@gh-symphony/cli`, still no speculative retry/backoff machinery, and the runtime-claude counterpart was correctly pushed out to #839 rather than smuggled in here.

### Two things to fold into the push you owe anyway

**1. P2 — the timeout ceiling admits values Node clamps to 1ms.** Inline on L10. `4_294_967_295` is 2^32−1, but Node's timer delay is signed 32-bit, so anything above `2_147_483_647` becomes a **1ms** deadline while the diagnostic reports the requested value. An operator raising the timeout to be lenient gets a near-instant abort and a message that lies about why. Reproduced at both `2147483648` and `4294967295`. One constant, plus the matching test boundary.

**2. The PR is not mergeable — `mergeable_state: dirty`.** `main` has moved under you (#836, #837). I reproduced and resolved the merge locally:

- `packages/worker/src/git-transport.ts` — **auto-merges correctly.** I checked the semantics rather than trusting the clean textual merge: #836 destructures `GIT_CONFIG_COUNT`/`KEY_0`/`VALUE_0` out of the credential environment and re-appends the helper after caller config; your new `tokenBrokerTimeoutMs` field lands in the `credentialVariables` rest and is spread through untouched. No interaction.
- `docs/configuration.md` — **the only real conflict.** #836 added a `GIT_CONFIG_COUNT, GIT_CONFIG_KEY_n, …` row and re-widened the table columns. Resolution is mechanical: take main's table, re-insert your `GITHUB_TOKEN_BROKER_TIMEOUT_MS` row after `GITHUB_TOKEN_BROKER_SECRET`.

I validated that resolution end to end: full `pnpm build` clean, runtime-codex 74 + worker 227 tests pass on the merged tree, and the smoke test still shows the 5000ms default and the 1200ms override honored post-merge. So the merge is safe; it just needs doing.

### Verdict

Approving. The mechanism is correct, the diagnostics are attributable, and #833 is fully satisfied — I am not holding this for another round. Land the merge with the one-constant ceiling fix alongside it and this is done; CI will re-run on that push.

---

packages/runtime-codex/src/git-credential-helper.ts
moncher-dev: Resolved in published commit `7cbe0b4`: the maximum is now Node's signed 32-bit timer ceiling, `2_147_483_647`, and the regression matrix rejects `2147483648` with the matching attributable diagnostic. The focused credential-helper suite passes with the corrected boundary.

packages/runtime-codex/src/runtime.ts
```diff
@@ -890,6 +892,13 @@ export function createGitCredentialHelperEnvironment(
           GITHUB_TOKEN_CACHE_PATH: config.githubTokenCachePath,
         }
       : {}),
+    ...(config.tokenBrokerTimeoutMs !== undefined
```
hojinzs: **P1 — an empty `GITHUB_TOKEN_BROKER_TIMEOUT_MS` is stringified to the literal `"undefined"` and breaks *all* git credential resolution**

This is mine to own — I approved in round 2 without catching it, and it landed in `401925a`, not in the new commit.

`parseGitCredentialBrokerTimeoutMs` deliberately treats `""` as unset (`git-credential-helper.ts:143`), so the helper itself handles an empty value correctly. But this guard tests `config.tokenBrokerTimeoutMs !== undefined` — and `""` passes it. The parser then returns `undefined`, and `String(undefined)` writes the seven-character string `"undefined"` into the child environment:

```
$ node -e 'import("./packages/runtime-codex/dist/index.js").then(m=>{ … })'
""      -> "undefined"     ← poisoned
"  "    -> THREW: … received "  ".
undefined -> undefined      ← correct
"7500"  -> "7500"           ← correct
```

The child then re-parses its own poisoned value and dies. Driven against the built helper:

```
$ echo -e "protocol=https\nhost=github.com\n" | \
    GITHUB_TOKEN_BROKER_TIMEOUT_MS=undefined GITHUB_GRAPHQL_TOKEN=ghs_test \
    node packages/runtime-codex/dist/git-credential-helper.js
GITHUB_TOKEN_BROKER_TIMEOUT_MS must be a positive integer no greater than 2147483647; received "undefined".
exit=1

# control, variable absent:
protocol=https / host=github.com / username=x-access-token / password=ghs_test
exit=0
```

Two things make this worse than a config-validation nit:

1. **It is reachable from the real worker path, not just the runtime helper.** `git-transport.ts:236` forwards `env.GITHUB_TOKEN_BROKER_TIMEOUT_MS` verbatim. Confirmed through `buildHostGitEnvironment` itself:
   ```
   buildHostGitEnvironment({ GITHUB_GRAPHQL_TOKEN: "ghs_test", GITHUB_TOKEN_BROKER_TIMEOUT_MS: "" })
   → GITHUB_TOKEN_BROKER_TIMEOUT_MS = "undefined"
   ```
2. **It breaks deployments that have no broker at all.** Note the repro above uses `GITHUB_GRAPHQL_TOKEN` with *no* broker configured — the static-token path that never issues a fetch and has nothing to time out. `credentialEnvironment` is built whenever `GITHUB_GRAPHQL_TOKEN || hasBroker`, so an empty variable takes down git auth for a deployment this knob does not even apply to.

Empty-valued env vars are not exotic here: `docker-compose` `- GITHUB_TOKEN_BROKER_TIMEOUT_MS=`, a k8s ConfigMap key with no value, and an unset `${{ vars.X }}` in Actions all yield `""`. And the failure lands as a git credential-helper failure with a message about a timeout the operator never set — precisely the unattributable-git-failure class #833 exists to eliminate.

Fix is to let the parser's own `undefined` decide, so both layers agree on what "unset" means:

```ts
...(() => {
  const tokenBrokerTimeoutMs = parseGitCredentialBrokerTimeoutMs(
    config.tokenBrokerTimeoutMs
  );
  return tokenBrokerTimeoutMs === undefined
    ? {}
    : { GITHUB_TOKEN_BROKER_TIMEOUT_MS: String(tokenBrokerTimeoutMs) };
})(),
```

**nit, same fix:** `"  "` currently throws eagerly out of `buildHostGitEnvironment`, failing the whole git transport rather than the one setting. A `.trim()` in the parser's empty check would fold whitespace-only into the same "unset" branch and cover both.

Worth a regression test asserting `""` produces an environment with no `GITHUB_TOKEN_BROKER_TIMEOUT_MS` key — the existing `runtime.test.ts` case pins `0` rejecting, but nothing pins empty passing through.

---

hojinzs (changes requested):
## Review — round 3 of 6 · **Changes requested (one line)**

Reviewed at `7cbe0b4` against issue #833. **Both round-2 fold-in items are correctly landed, and I re-verified each against the built helper rather than the thread replies.** I am reversing my round-2 approval for one reason only: while smoke-testing the merged tree I found a regression I missed last round, which breaks git credential resolution outright. One line, plus a test. Nothing else is outstanding.

### Round-2 fold-in items — verified

| Round-2 item | Verified how | Status |
| --- | --- | --- |
| **P2** timer ceiling clamped to 1ms | `MAX_TOKEN_BROKER_TIMEOUT_MS = 2_147_483_647` (`git-credential-helper.ts:10`). Subprocess: `2147483648` rejected in 221ms with the matching diagnostic. | ✅ |
| Test boundary moves with the constant | `it.each` now pins `"2147483648"`; the stale `"4294967296"` case is gone. | ✅ |
| Merge conflict against `main` | `main` (`a1aa06d`) is an ancestor of the head; `mergeable_state: clean`. | ✅ |
| `docs/configuration.md` merge resolution | Checked semantically, not just textually: #836's `GIT_CONFIG_COUNT, GIT_CONFIG_KEY_n, …` row **and** your `GITHUB_TOKEN_BROKER_TIMEOUT_MS` row both survive, in main's re-widened table. Nothing dropped. | ✅ |
| `git-transport.ts` auto-merge | #836's `GIT_CONFIG_*` destructure and your new field do not interact; the field rides the `credentialVariables` rest. | ✅ |

### Smoke test

Built the branch and drove the **real** `packages/runtime-codex/dist/git-credential-helper.js` as a subprocess against a stalling HTTPS broker (self-signed cert, hostname alias, so `validateGitHubTokenBrokerUrl` is exercised for real) — the production path, not the unit stub.

```
default (unset → 5000ms)      exit=1  elapsed=5233ms
  Git credential token broker request to https://broker.example:8443/runtime-token timed out after 5000ms.
override 1200ms               exit=1  elapsed=1424ms   (diagnostic names 1200ms)
connect-phase stall, 1000ms   exit=1  elapsed=1217ms   ← round-1 P2, still fixed post-merge
2147483648 (over ceiling)     exit=1  elapsed=221ms    clear config diagnostic
```

The connect-phase stall is the one I most wanted to re-check after the merge — it exits at 1217ms rather than being pinned to undici's ~10.5s `connectTimeout`, so the `writeSync` + `process.exit(1)` fix survived intact.

Gates: `pnpm install --frozen-lockfile`, `pnpm build`, `pnpm lint`, `pnpm typecheck` all exit 0; `pnpm test` green at **617 tests**. CI green on `7cbe0b4` (Test, Container Smoke).

### The one blocker

**P1 — an empty `GITHUB_TOKEN_BROKER_TIMEOUT_MS` becomes the literal string `"undefined"` in the git child environment, and every credential lookup then fails.** Inline on `runtime.ts:895`.

`parseGitCredentialBrokerTimeoutMs` deliberately maps `""` → `undefined`, so the helper handles an empty value correctly on its own. But the passthrough guards on `!== undefined`, which `""` passes, and `String(undefined)` writes `"undefined"` into the child env. The child re-parses that and exits 1.

I confirmed it end to end, including through `buildHostGitEnvironment` — so it is reachable from the worker, not only the runtime helper. The sharpest part: it fires on deployments with **no broker configured at all**, because the credential environment is built whenever `GITHUB_GRAPHQL_TOKEN || hasBroker`. A static-token deployment that never issues a broker fetch loses git auth to a timeout knob that does not apply to it, and the operator sees a diagnostic naming a value they never set.

That is the unattributable-git-failure shape #833 was filed to remove, so I do not think it should ship even though the trigger is narrow. Empty-valued env vars are ordinary: `docker-compose` `- VAR=`, an empty ConfigMap value, an unset `${{ vars.X }}`.

Details and the patch are in the inline comment.

### Requirements coverage vs #833

| #833 asked for | Status |
| --- | --- |
| Bound the broker request with an `AbortSignal` | ✅ |
| A sensible timeout | ✅ 5s default, operator-overridable, ceiling now correct |
| Diagnostic naming the broker URL and the timeout | ✅ verified live at both the default and an override |
| Attributable rather than a generic hang | ⚠️ yes for the broker path — but the P1 above introduces a *new* unattributable git failure |

**No over-engineering, and no scope drift.** The only commit since round 2 is the one-constant ceiling fix I asked for. Still correctly scoped `patch` on `@gh-symphony/cli`, still no speculative retry/backoff/circuit-breaker machinery, and the runtime-claude counterpart stayed out in #839 rather than being folded in. No unrequested inventions this round.

### What closes this

Fix the empty-string passthrough (one line as written inline), add the regression test for it, push. That is the entire remaining list — I will approve on that push and will not open a new line of review. Everything else on this PR is verified done.

---

moncher-dev:
Steward note — **re-approval needed on a new head. The conflict is resolved and this is mergeable; your approval was reset by the push.**

I set #833 to Land on your approval of `397f037d`. The worker opened the land cycle, found the head had moved to `7cbe0b4b`, correctly declined to merge without human approval on the current head, and returned the issue to `In review`. That was the right call and I am not overriding it.

**Exactly what changed since your approval**, so re-review is cheap:

```
7cbe0b4b fix(runtime-codex): bound broker timer range
7a3e53a1 Merge remote-tracking branch 'origin/main'
a1aa06df fix(worker): preserve caller git credential configuration (#836)
19c29dbb fix(worker,orchestrator): drop the implicit <cwd>/.env read … (#837)
```

Two of those four are `main` arriving through the conflict merge you already reviewed and approved elsewhere (#836, #837). The merge commit itself resolved the `docs/configuration.md` and `git-transport.ts` overlaps; I confirmed the tip still carries the `GITHUB_TOKEN_BROKER_TIMEOUT_MS` documentation row.

The only genuinely new code since your approval is `7cbe0b4b`, and it is one line plus its test:

```diff
-const MAX_TOKEN_BROKER_TIMEOUT_MS = 4_294_967_295;
+const MAX_TOKEN_BROKER_TIMEOUT_MS = 2_147_483_647;
```

That is a real correctness fix rather than churn: Node's timer range is int32, so the previous uint32 bound would have accepted values that overflow and fire immediately — turning a deliberately long broker timeout into an instant abort, which is the opposite of what an operator raising it would expect.

Current state: `mergeable=MERGEABLE`, `mergeStateStatus=CLEAN`, both checks green. Nothing outstanding but the re-approval.

Also worth noting for sequencing: **#839** (the `runtime-claude` twin of this fix) is boarded and machine-blocked by #833, deliberately, so it copies this landed contract rather than inventing a parallel one. It unblocks the moment this merges.

packages/runtime-codex/src/runtime.ts
moncher-dev: Resolved in published commit `d69b0fe`: empty and whitespace-only timeout values are normalized as unset before building the child environment, so `GITHUB_TOKEN_BROKER_TIMEOUT_MS` is omitted instead of becoming `"undefined"`. Regression tests cover both parser behavior and environment omission; the focused and full validation suites pass.

docs/configuration.md
```diff
@@ -476,6 +476,7 @@ target a non-`github.com` host.
 | ------------------------------------------------------------ | ---------------- | --------------------------------------- | ----------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------ |
 | `GITHUB_TOKEN_BROKER_URL`                                    | unset            | Worker host tools and Git transport     | User-facing/ops   | Broker endpoint for GitHub tokens. It is host-only and not inherited by agent children by default.                                                     |
 | `GITHUB_TOKEN_BROKER_SECRET`                                 | unset            | Worker host tools and Git transport     | User-facing/ops   | Shared secret sent to the GitHub token broker. It is declared as a tracker secret and removed from agent children by default.                          |
+| `GITHUB_TOKEN_BROKER_TIMEOUT_MS`                             | `5000`           | Git credential helper                   | User-facing/ops   | Maximum broker request duration in milliseconds. Increase this positive integer for brokers that legitimately need longer than five seconds.           |
```
hojinzs, line 479: **nit — the row states the floor but not the ceiling, and the ceiling fails hard**

"Increase this positive integer" reads as unbounded, but `MAX_TOKEN_BROKER_TIMEOUT_MS = 2_147_483_647` rejects anything larger — and it rejects it out of `createGitCredentialHelperEnvironment`, so the failure is not scoped to this one setting. Verified against the built runtime:

```
"2147483647" -> "2147483647"   ✅
"2147483648" -> THREW: GITHUB_TOKEN_BROKER_TIMEOUT_MS must be a positive integer
                no greater than 2147483647; received "2147483648".
```

The diagnostic is attributable, so an operator who trips it is not lost — this is just a docs completeness point. Half a clause closes it: *"…longer than five seconds, up to `2147483647`."*

While I was there, two accepted-but-surprising inputs, same nit bucket, no action needed unless you want them: `Number()` is doing the parsing, so `"1e3"` lands as `1000` and `"0x10"` as `16` despite the documented "positive integer". Both normalize through `String(tokenBrokerTimeoutMs)` before reaching the child, so nothing is poisoned — the child sees `1000`/`16`, not the raw text.

Non-blocking.

---

packages/runtime-codex/src/git-credential-helper.ts
```diff
@@ -36,9 +40,29 @@ export async function resolveGitCredential(
...
-  const token = await resolveGitHubGraphQLToken(config, {
-    fetchImpl,
-  });
+  const tokenBrokerTimeoutMs =
+    config.tokenBrokerTimeoutMs ?? DEFAULT_TOKEN_BROKER_TIMEOUT_MS;
+  const tokenBrokerUrl = config.tokenBrokerUrl;
+  const brokerFetch: typeof fetch = (input, init) =>
+    fetchImpl(input, {
+      ...init,
+      signal: AbortSignal.timeout(tokenBrokerTimeoutMs),
```
hojinzs, line 49: **nit — `signal` is assigned over `...init`, not composed with it**

`brokerFetch` spreads `init` and then overwrites `signal`, so any caller-supplied signal is silently discarded rather than combined. Today that is unreachable: `resolveGitHubGraphQLToken` (`tool-github-graphql/src/tool.ts:316-322`) issues exactly one `fetchImpl` call and passes only `method`/`headers`, no signal — I checked, and there is no retry loop, so the 5s budget really is the whole broker budget rather than 5s per attempt.

It becomes a live bug the day that function grows cancellation of its own. `AbortSignal.any` composes instead of clobbering:

```ts
signal: init?.signal
  ? AbortSignal.any([init.signal, AbortSignal.timeout(tokenBrokerTimeoutMs)])
  : AbortSignal.timeout(tokenBrokerTimeoutMs),
```

Non-blocking — noting it so the assumption is written down somewhere other than my head.

---

hojinzs (approved):
**Approving at `d69b0fe`.** Round 5. Every finding from rounds 1–4 is fixed, and I verified the fixes as a real subprocess and through real `git` rather than taking the resolution notes on faith. Two `nit`s inline, neither blocking.

## The acceptance criterion, end to end

#833 is about a hung broker taking the calling git operation with it. That is now directly testable, so I tested it rather than the unit boundary — built helper, wired the way `runtime.ts:873-875` wires it, against an HTTPS broker that completes TLS and then never answers:

```
$ GIT_TERMINAL_PROMPT=0 GIT_CONFIG_COUNT=1 \
  GIT_CONFIG_KEY_0=credential.helper GIT_CONFIG_VALUE_0="!node .../git-credential-helper.js" \
  GITHUB_TOKEN_BROKER_URL=https://broker.example:8443/token GITHUB_TOKEN_BROKER_SECRET=s \
  git ls-remote https://github.com/hojinzs/<private>.git

Git credential token broker request to https://broker.example:8443/token timed out after 5000ms.
fatal: could not read Username for 'https://github.com': terminal prompts disabled

real  0m5.435s
```

5.4s and a diagnostic naming the broker URL and the budget — versus the indefinite block the issue describes. That is the whole ask, and it is met.

## Round-2 P2 (process exit) — fixed, and the regression it was hiding is gone

That P2 was that the deadline bounded the *rejection* but not the *process*, and git waits on exit. Same three shapes, re-run at this head:

| broker shape | diagnostic | process exit | previously |
| --- | --- | --- | --- |
| TLS completes, headers never sent | 5136ms | **5141ms** | 5239ms ✅ |
| TCP accepted, TLS handshake never completes | 2143ms | **2148ms** | 10802ms ❌ |
| headers + partial body, then stalls | 5137ms | **5141ms** | *not previously tested* |

The connect-stall case is the one that mattered: it used to exit at ~10.5s pinned by undici's `connectTimeout`, so "fails after 5 seconds" was only true for one of two shapes. `writeSync(2, …)` + `process.exit(1)` closes it — exit now trails the diagnostic by ~5ms in every shape, and the diagnostic is never truncated.

I added the third row myself because no earlier round covered it: `fetch` resolves on headers, so a broker that answers and then stalls mid-body could plausibly have escaped the signal. It does not — undici carries the abort into the body stream, `response.json()` rejects, and the mapping still fires. Healthy-broker control is unaffected (`exit=0`, correct credential block, 239ms).

## Round-4 P1 (`""` → `"undefined"`) — fixed at both layers

The guard now tests the parser's own return rather than the raw config, so the two layers agree on "unset":

```
createGitCredentialHelperEnvironment  ""  -> key absent        (was "undefined")
                                      "  " -> key absent        (was THREW)
buildHostGitEnvironment (worker)      ""  -> "" passed through, child treats as unset
```

And the case that made it a P1 — an empty variable breaking deployments with **no broker at all** — is clean as a subprocess: `GITHUB_GRAPHQL_TOKEN=ghs_static` with `GITHUB_TOKEN_BROKER_TIMEOUT_MS=""` and `="  "` both exit 0 with the correct credential block. Literal `"undefined"` still rejects, which is right.

Round-3 P2 boundary holds too: `2147483647` accepted, `2147483648` rejected with the matching message.

## Requirements and scope

Against #833: `AbortSignal` ✅, sensible default ✅, diagnostic naming broker URL and timeout ✅, attributable rather than a generic hang ✅.

No over-engineering. The two things that could look like scope creep — the operator env knob and the `process.exit` change — were both asked for in review, not self-directed, and the `runtime-claude` twin was correctly pushed out to #839 instead of folded in. Changeset is present and correctly scoped `patch` on `@gh-symphony/cli`; the `docs/configuration.md` row survived the `main` merge.

## Local validation

`pnpm build`, `pnpm lint`, `pnpm typecheck` all clean; focused suites green (`git-credential-helper.test.ts` 15, `runtime.test.ts` 41, `git-transport.test.ts` 20). Both CI checks green on this head. Root `pnpm format` reports 84 files including `prettier.config.mjs` — repo-wide and pre-existing; all eight files this PR touches are Prettier-clean, so not this PR's.

Good to merge.

---

PR #836, opened 2026-09-03
fix(worker): preserve caller Git credential config

### LINKED ISSUE #832: fix(worker): buildHostGitEnvironment silently overrides caller-supplied git credential config
Follow-up requested by @hojinzs in review of PR #831 (see that thread for the full reproduction).

## Problem

`buildHostGitEnvironment` spreads `credentialEnvironment` **after** the caller's `...env` and resets `GIT_CONFIG_COUNT` to `"1"`. Any `GIT_CONFIG_KEY_0` / `GIT_CONFIG_VALUE_0` the caller supplied is silently overwritten with the production `git-credential-helper.js`, and any higher-indexed entries the caller added (`GIT_CONFIG_KEY_1`, …) remain in the environment but are never read, because the count says there is only one.

The override is silent: no warning, no error, and the caller's configuration appears present in the environment while having no effect.

## How it surfaced

This is what defeated the fix attempt in #831/#815. A test installed a fixture credential helper at index 1 and set the count accordingly; with `GITHUB_GRAPHQL_TOKEN` present, `buildHostGitEnvironment` reset the count to 1 and pointed index 0 at the production helper, so the fixture helper was never consulted and the request fell through to an authentication prompt.

## Expected

Either compose with caller-supplied `GIT_CONFIG_*` entries rather than clobbering them — appending the credential helper at the next free index and incrementing `GIT_CONFIG_COUNT` — or fail loudly when the caller has supplied git config that would be discarded. Silently discarding it is the trap.

## Notes

Filed by the board steward at the reviewer's request. Explicitly out of scope for #831, which stays test-only.

moncher-dev:
## Issues

- Closes #832

## Summary

- Preserve caller-provided `GIT_CONFIG_*` entries when host Git credentials are enabled.
- Append the Symphony credential helper at the caller's next configured index and fail clearly for malformed counts.
- Document the supported input contract, caller-first credential precedence, and host-environment trust boundary.

## Change-point diagram

- worker Git transport caller environment → existing indexed Git config remains unchanged → host credential helper is appended at `GIT_CONFIG_COUNT` → Git reads every configured entry in caller-first order

## Start here

- `packages/worker/src/git-transport.ts:223` — composes the host Git environment.
- `packages/worker/src/git-transport.test.ts:461` — verifies indexed configuration and credential variables are preserved.
- `docs/configuration.md:465` — documents the input, ordering, validation, and trust contract.

- Worker-side Git synchronization no longer silently discards caller-supplied credential helpers when a host token or token broker is configured.
- A malformed `GIT_CONFIG_COUNT` now fails host Git transport with a clear diagnostic.

## Validation

- `pnpm exec vitest run packages/worker/src/git-transport.test.ts` — pass (20 tests; Cycle 2, code unchanged during rework)
- `pnpm lint` — pass (Cycle 3)
- `pnpm test` — pass (all 14 tested workspace projects; Cycle 2, code unchanged during rework)
- `pnpm typecheck` — pass (Cycle 2, code unchanged during rework)
- `pnpm build` — pass (Cycle 2, code unchanged during rework)
- `pnpm format` — baseline failure in 86 unrelated pre-existing files; isolated `pnpm exec prettier --check docs/configuration.md` passes
- Docker E2E — not applicable: this is focused worker environment composition with no lifecycle or integration boundary change

## Changeset

- Not needed because issue #832 has no changeset label.

## Risks & rollback

- Caller-supplied Git config comes from the process and project `.env`, not a curated allowlist. Because caller entries are consulted first, an ambient `credential.helper` that returns complete credentials can shadow Symphony's tenant-scoped helper.
- Count validation rejects malformed values before Git starts, so a typo in `GIT_CONFIG_COUNT` aborts host Git transport.
- Roll back the six issue commits if caller-first helper ordering causes an unexpected transport regression.

## Changed files

- `packages/worker/src/git-transport.ts` — append host credentials without clobbering caller Git config.
- `packages/worker/src/git-transport.test.ts` — cover composition, production-token transport, broker variables, and invalid counts.
- `docs/configuration.md` — document caller configuration, helper precedence, validation, and trust boundaries.

- [ ] None required beyond CI.

## Security


moncher-dev:
Steward note — the red `Test` check looks like a CI-load flake unrelated to this change. I have re-run the failed job rather than dispatching a rework cycle for it.

The failure:

```
FAIL packages/cli/src/index.test.ts > Commander CLI entrypoint
     > forwards doctor --project-dir through the Commander entrypoint
Error: Test timed out in 5000ms.        (actual: 5653ms)
Test Files  1 failed | 134 passed (135)
```

Three reasons to read it as environmental rather than caused by this PR:

1. **This PR touches two files, both in `packages/worker`** — `git-transport.ts` and `git-transport.test.ts`. The failing test is in `packages/cli/src/index.test.ts` and exercises Commander argument forwarding for `doctor --project-dir`, which has no dependency on the worker Git transport module.
2. **The overrun is marginal** — 5653ms against a 5000ms budget, 13% over. A defect introduced by a credential-config change would be expected to hang or fail an assertion, not land just past the line.
3. **Everything else passed** — 134 of 135 files green, and `Container Smoke` succeeded.

If the re-run comes back green, this can proceed on review alone. If it fails again at the same place, it is worth a separate issue for a genuinely tight 5s budget on that test rather than treating it as this PR's problem — I did not find the same failure in other recent runs, so there is not yet evidence of a standing flake.

Leaving #832 in `In review` rather than sending it back for rework, since there is nothing in this diff to rework on the strength of that failure.

moncher-dev:
Smoke-tested at `f5671845`. `pnpm install --frozen-lockfile`, `pnpm build`, `pnpm typecheck`, `pnpm lint` all exit 0; `git-transport.test.ts` green at 20 tests. The earlier red `Test` check was the unrelated `doctor --project-dir` timeout I diagnosed above; the re-run is green.

I drove `buildHostGitEnvironment` from the built `dist` across the four cases that matter, rather than relying on the unit suite:

```
A. caller supplies index 0, token present
     GIT_CONFIG_COUNT   = 2
     GIT_CONFIG_KEY_0   = http.extraHeader        ← caller's, preserved
     GIT_CONFIG_VALUE_0 = X-Test: 1
     GIT_CONFIG_KEY_1   = credential.helper       ← appended, not overwritten
B. no caller config, token present
     GIT_CONFIG_COUNT = 1 | KEY_0 = credential.helper      ← unchanged legacy shape
C. caller config, no token
     GIT_CONFIG_COUNT = 1 | KEY_0 = http.extraHeader       ← untouched
D. GIT_CONFIG_COUNT="abc"
     throws: invalid GIT_CONFIG_COUNT for host Git transport: expected a non-negative safe integer
```

A is the #832 defect and it is fixed: the caller's entry survives and the helper lands at the next free index with the count incremented. B confirms the common path is byte-for-byte what it was, so this is not a behavior change for anyone not supplying git config. C and D show the guard is scoped and the validation is real rather than decorative.

**The Codex P1 stands and I agree with it.** This PR adds a *fatal* validation path and, in doing so, promotes project-`.env`-supplied `GIT_CONFIG_COUNT` / indexed `GIT_CONFIG_*` entries into a supported worker Git-transport input — but `docs/configuration.md` is untouched, and the diff is two files, both in `packages/worker`. Operators get no format, no ordering guarantee, and no statement that a malformed count now aborts the transport instead of being ignored. Case D is a new way for a run to fail hard on a value that previously did nothing, which is exactly the kind of change that needs a documented contract before it ships.

Concretely: state that indexed `GIT_CONFIG_*` entries from the project `.env` are honored, that the credential helper is appended after them, and that `GIT_CONFIG_COUNT` must be a non-negative safe integer or the transport fails.

No over-engineering otherwise — 2 files, the guard is minimal, and the legacy shape is preserved exactly. @hojinzs over to you.

packages/worker/src/git-transport.ts
```diff
@@ -237,10 +237,38 @@ export function buildHostGitEnvironment(
...
+  }
+
   return {
     ...env,
-    GIT_TERMINAL_PROMPT: "0",
-    ...credentialEnvironment,
+    ...credentialVariables,
+    GIT_CONFIG_COUNT: String(gitConfigCount + 1),
+    [`GIT_CONFIG_KEY_${gitConfigCount}`]: "credential.helper",
+    [`GIT_CONFIG_VALUE_${gitConfigCount}`]: credentialHelper,
```
hojinzs, line 271: **P2 — appending inverts credential precedence, and the env this reads is not curated**

This is the behavior #832 asked for (append at the next free index), so I am not asking you to change the ordering. I am asking that the trust decision it implies be stated, because it is not visible anywhere in the diff or the PR body.

Git consults `credential.helper` entries in order and stops at the first that returns a username *and* password. Before this change a host token always won: the count was reset to `1` and the Symphony helper sat at index 0. After it, **any** caller-supplied `credential.helper` outranks the Symphony helper.

The caller is not a curated allowlist. `synchronizeAssignedBranch` passes `env` straight through (`packages/worker/src/index.ts:861`, `:2027`), and that `env` is `loadLauncherEnvironment(process.env)` — `packages/runtime-codex/src/launcher.ts:98-104` merges the package `.env`, the cwd `.env`, and then `process.env`. So a `GIT_CONFIG_KEY_n=credential.helper` in a project `.env` or in the container image now silently takes precedence over the tenant-scoped credential injected by #827 (eedb876).

I confirmed the precedence flip against the built `dist`:

```
caller: GIT_CONFIG_COUNT=1, KEY_0=http.extraHeader   + GITHUB_GRAPHQL_TOKEN
  → COUNT=2, KEY_0=http.extraHeader, KEY_1=credential.helper   (caller first)
```

Two things to add:

1. The precedence rule belongs in the `docs/configuration.md` row the Codex thread is asking for — not just "entries are honored" but "caller entries are consulted **before** the Symphony helper".
2. `## Risks & rollback` currently says the risk is "limited to parsing and extending process-level Git config". It is broader than parsing: an ambient `credential.helper` can now shadow the tenant token. Please say so there.

---

packages/worker/src/git-transport.ts
```diff
@@ -237,10 +237,38 @@ export function buildHostGitEnvironment(
...
+    throw new Error(
+      "invalid GIT_CONFIG_COUNT for host Git transport: expected a non-negative safe integer"
+    );
+  }
+
   return {
     ...env,
-    GIT_TERMINAL_PROMPT: "0",
-    ...credentialEnvironment,
+    ...credentialVariables,
```
hojinzs, line 268: **P3 — `GIT_TERMINAL_PROMPT: "0"` is no longer guaranteed by this function on the credential path**

Not currently broken — I verified it. But the guarantee moved.

The old return literal set it explicitly on every path. Now the credential path sets it only because `createGitCredentialHelperEnvironment` happens to include `GIT_TERMINAL_PROMPT: "0"` (`packages/runtime-codex/src/runtime.ts:867`), and the destructure above only strips `GIT_CONFIG_COUNT` / `KEY_0` / `VALUE_0`, so it survives inside `...credentialVariables`. Checked against `dist` — a caller passing `GIT_TERMINAL_PROMPT=1` with a token still gets `"0"` out.

So this is a coupling complaint, not a bug report: the worker's "never prompt for auth" guarantee now depends on an implementation detail of a *different package*, and nothing pins it. If someone trims that factory down to just the config triple, this silently starts allowing an interactive auth prompt in the host transport, which is a hang rather than a failure.

Cheapest fix is to keep it explicit in the return literal (it costs one line and is idempotent with the spread). Failing that, add `expect(env.GIT_TERMINAL_PROMPT).toBe("0")` to the new `appends the credential helper after caller-supplied Git config` test so the coupling is at least load-bearing on a test.

---

packages/worker/src/git-transport.ts
```diff
@@ -237,10 +237,38 @@ export function buildHostGitEnvironment(
           gitUsername: env.GITHUB_GIT_USERNAME,
         })
       : {};
+  const {
+    GIT_CONFIG_COUNT: _credentialConfigCount,
+    GIT_CONFIG_KEY_0: _credentialConfigKey,
+    GIT_CONFIG_VALUE_0: credentialHelper,
+    ...credentialVariables
+  } = credentialEnvironment;
```
hojinzs, line 245: **P3 — the unenforced "the helper env has exactly one entry" assumption re-opens the #832 failure mode one index over**

This destructure discards the factory's own `GIT_CONFIG_COUNT` (`_credentialConfigCount`) and its `GIT_CONFIG_KEY_0` (`_credentialConfigKey`) without checking either, then hardcodes `"credential.helper"` again at line 270. That is correct today — `createGitCredentialHelperEnvironment` emits exactly `COUNT=1` with one entry at index 0.

The failure mode if that ever changes: a second entry from the factory (`GIT_CONFIG_KEY_1` / `VALUE_1`) is not destructured out, so it stays in `...credentialVariables` and — spread *after* `...env` at line 267-268 — **overwrites the caller's `GIT_CONFIG_KEY_1`**. Silently, with the caller's value still nowhere and the count pointing at the wrong entry. That is precisely the bug #832 was filed for, relocated from index 0 to index 1, and this diff would not catch it because both new tests use a caller count of 2 or fewer with a single-entry factory.

Two cheap options, either is fine:

- Re-emit `_credentialConfigKey` at line 270 instead of the hardcoded literal, and throw if `_credentialConfigCount !== "1"` — makes the assumption explicit and fatal rather than silent.
- Or drop the destructure and merge the factory's indexed entries in a loop over its own count, which removes the assumption entirely.

I'd take the first; the second is more machinery than the current single-entry factory earns.

---

packages/worker/src/git-transport.ts
```diff
@@ -237,10 +237,38 @@ export function buildHostGitEnvironment(
...
+    throw new Error(
+      "invalid GIT_CONFIG_COUNT for host Git transport: expected a non-negative safe integer"
+    );
+  }
+  const gitConfigCount = Number(gitConfigCountValue);
+  if (!Number.isSafeInteger(gitConfigCount)) {
+    throw new Error(
+      "invalid GIT_CONFIG_COUNT for host Git transport: expected a non-negative safe integer"
+    );
+  }
```
hojinzs, line 264: **nit — two guards, one message**

`/^\d+$/` already establishes "non-negative integer", so the second guard only ever catches overflow past `Number.MAX_SAFE_INTEGER`. Two `throw`s with a byte-identical string means the error tells you nothing about which one fired. Collapsing them reads better:

```ts
const gitConfigCount = Number(env.GIT_CONFIG_COUNT ?? "0");
if (!/^\d+$/.test(gitConfigCountValue) || !Number.isSafeInteger(gitConfigCount)) {
  throw new Error(
    "invalid GIT_CONFIG_COUNT for host Git transport: expected a non-negative safe integer"
  );
}
```

Separately, on the validation itself — I checked the accepted/rejected set against `dist` and it is tighter than Git's own parser, which is the right call but worth knowing:

```
"not-a-count" → throw     ""    → throw     "-1"  → throw
" 2"          → throw     "2k"  → throw     "007" → count=8, helper at index 7
"9007199254740993" → throw
```

`""` and `"not-a-count"` match Git (it dies with `bogus count in GIT_CONFIG_COUNT`). `" 2"` and `"2k"` are accepted by `git_parse_ulong` but rejected here. Stricter is fine for a config surface nobody hand-writes — no change requested, just don't let it surprise you later. `"007"` behaving as 7 matches Git exactly, which is the case I most expected to be wrong.

---

hojinzs (changes requested):
## Review — round 1 of 6 · `f5671845`

Requirement coverage is complete and the fix is real. Holding on one thing: the `docs/configuration.md` update that AGENTS.md makes mandatory for this change. Everything else below is P3/nit and would not block on its own.

### Requirements from #832 — all met

| #832 asked for | Status |
| --- | --- |
| Compose with caller `GIT_CONFIG_*` rather than clobbering | ✅ caller entries preserved verbatim |
| Append the helper at the next free index | ✅ appended at `GIT_CONFIG_COUNT` |
| Increment `GIT_CONFIG_COUNT` | ✅ `String(count + 1)` |
| Or fail loudly on config that would be discarded | ✅ fatal on a malformed count |

Nothing here is out of scope and nothing is over-engineered: 2 files, +70/-3, one guard, no new abstraction, no config surface invented beyond what Git already defines. I looked for scope creep specifically and did not find any.

### Verification I ran myself

I did not take the test suite's word for it. At `f5671845`, `pnpm install --frozen-lockfile` → `pnpm build` → `pnpm lint` → `pnpm typecheck` all exit 0, and `git-transport.test.ts` is 20/20 green. CI on the head commit is green on both `Test` and `Container Smoke`.

**The regression test is genuine.** I reverted `git-transport.ts` alone to base (`eedb876`) and re-ran the unmodified test file:

```
× synchronizeAssignedBranch > uses host credential-helper authentication for remote Git operations
× buildHostGitEnvironment  > appends the credential helper after caller-supplied Git config
× buildHostGitEnvironment  > rejects a malformed caller-supplied Git config count
   (17 others still pass)
```

Three fail on base, all pass on head. The transport one is the important one — it stands up a real authenticating HTTP Git server, puts a fixture helper at index 1 behind an empty-value reset at index 0, and sets a live `GITHUB_GRAPHQL_TOKEN`, so it exercises the actual #831/#815 defeat rather than asserting on a shape.

**Smoke test — driven against the built `dist`, not the unit suite:**

```
A. caller KEY_0 + host token → COUNT=2, KEY_0=http.extraHeader, KEY_1=credential.helper
B. no caller config + token  → COUNT=1, KEY_0=credential.helper        (legacy shape, byte-identical)
C. caller config, no token   → COUNT=1, KEY_0=http.extraHeader          (untouched)
D. GIT_TERMINAL_PROMPT=1 + token → GIT_TERMINAL_PROMPT=0                (still forced)
E. GIT_CONFIG_COUNT="abc"    → throws
```

A is the #832 defect, fixed. B is the one that matters for blast radius — anyone not supplying Git config gets exactly the environment they got before, so this is not a behavior change for the common path.

### Blocking

**P1 — `docs/configuration.md` is untouched.** I agree with the Codex thread and am carrying it as the one blocker.

This is not a style objection. AGENTS.md "Mandatory Post-Work Checks" item 4 requires configuration and environment variables to be documented **in the same change**, and this PR does two documentable things: it promotes caller-supplied `GIT_CONFIG_COUNT` / `GIT_CONFIG_KEY_n` / `GIT_CONFIG_VALUE_n` into a supported worker Git-transport input, and it adds a path where a bad value **aborts the transport** where it was previously ignored. The `Credential Brokers And Git Access` table (`docs/configuration.md:467-478`) has rows for `GITHUB_GIT_HOST` and `GITHUB_GIT_USERNAME` but none for any `GIT_CONFIG_*` — I grepped, there is no mention anywhere in the file.

That gap has teeth because the values are reachable from a project `.env`: `loadLauncherEnvironment` merges the package `.env` and the cwd `.env` under `process.env` (`packages/runtime-codex/src/launcher.ts:98-104`), and that merged env is what reaches `buildHostGitEnvironment`. An operator can hard-fail their own worker with a typo in a file the docs never told them was read.

Please add a row (or a short subsection) covering: indexed entries are honored; the Symphony credential helper is appended **after** them and is therefore consulted **last**; `GIT_CONFIG_COUNT` must be a non-negative safe integer or the transport fails.

### Non-blocking

- **P2** — the append inverts credential precedence, and the env it reads is not a curated allowlist. Requested behavior, so no code change asked for, but the trust decision should be written down and `## Risks & rollback` understates it. Inline at line 266.
- **P3** — `GIT_TERMINAL_PROMPT: "0"` is no longer set explicitly on the credential path; it survives only via a spread from another package. Verified not broken today. Inline at line 268.
- **P3** — the "factory emits exactly one entry" assumption is unenforced, and would reintroduce the #832 silent clobber at index 1 if it ever stopped holding. Inline at line 240.
- **nit** — two guards throwing one identical message. Inline at line 253.

### Question, not a finding

`## Changeset` says "not needed because #832 has no changeset label". That matches WORKFLOW.md:232, which gates changesets on the `changeset:*` labels, and #832 carries only `bug` — so the process was followed. But CLAUDE.md:89 says behavior-changing PRs must add one, and this PR's own `User-Visible Behavior` section describes a runtime behavior change to the published `@gh-symphony/cli`. @hojinzs — worth deciding whether #832 should get a `changeset:patch` label rather than shipping this to npm with no release note. Not something I'd hold the PR for.

### Verdict

`REQUEST_CHANGES` on the P1 alone. The code is right, the tests prove it, and the blast radius is contained — add the `docs/configuration.md` entry and I expect to approve on the next round without further rounds.

---

packages/worker/src/git-transport.ts
```diff
@@ -237,10 +237,38 @@ export function buildHostGitEnvironment(
...
+  } = credentialEnvironment;
+  if (!credentialHelper) {
+    return {
+      ...env,
+      GIT_TERMINAL_PROMPT: "0",
+    };
+  }
+
+  const gitConfigCountValue = env.GIT_CONFIG_COUNT ?? "0";
+  if (!/^\d+$/.test(gitConfigCountValue)) {
```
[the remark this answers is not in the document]
moncher-dev: Documented the indexed `GIT_CONFIG_*` input contract in `docs/configuration.md`, including the non-negative safe-integer requirement, fatal malformed-count behavior, and appended-helper ordering.

packages/worker/src/git-transport.ts
moncher-dev: Documented that caller-supplied helpers are consulted first and may shadow Symphony's tenant-scoped helper. The refreshed PR risk section now also identifies the process/project `.env` trust boundary and the precedence consequence.

packages/worker/src/git-transport.ts
moncher-dev: No code change in this rework: the requested scope was explicitly limited to documentation, and the current credential factory still guarantees `GIT_TERMINAL_PROMPT=0` on this path. The coupling remains non-blocking and is unchanged by the documentation fix.

packages/worker/src/git-transport.ts
moncher-dev: No code change in this rework: the factory currently emits exactly one indexed entry, and review classified this as a non-blocking future-proofing concern while directing the rework diff to documentation only. The documented contract matches the current implementation.

packages/worker/src/git-transport.ts
moncher-dev: Leaving the two guards unchanged because this rework is constrained to documentation and both branches already produce the intended redacted public contract without exposing caller-controlled values.

hojinzs (approved):
## Approve — cycle 2, at `a7cbb1a`

Re-reviewed at head `a7cbb1a` (the docs commit; the `review_requested` webhook carried the stale `f567184`). The rework is docs-only — `git diff f567184 a7cbb1a -- packages/` is empty — so my cycle-1 code findings apply unchanged to the code that is merging.

### Requirements vs #832

#832 asked for one of two things: compose with caller-supplied `GIT_CONFIG_*` rather than clobbering them, **or** fail loudly. This takes the first branch. I drove `buildHostGitEnvironment` from the built `dist` rather than trusting the unit suite:

```
A. caller idx 0 + host token          ← the #832 defect
     COUNT=2  KEY_0=http.extraHeader (preserved)  KEY_1=credential.helper (appended)
B. no caller config + host token
     COUNT=1  KEY_0=credential.helper             ← legacy shape, byte-for-byte
C. caller config, no host token
     COUNT=1  KEY_0=http.extraHeader              ← untouched
D. COUNT="not-a-count" + token        → throws, message carries no caller value
```

A is the defect and it is fixed. B means this is not a behavior change for anyone not supplying Git config. The regression test is real, too: the #831 test that previously set `GITHUB_GRAPHQL_TOKEN: undefined` now sets a token, so it actually exercises the path that defeated #831/#815 instead of stepping around it.

### The two blocking findings from cycle 1 are resolved

- **Codex P1 / my P2 (documentation + precedence)** — `docs/configuration.md` now carries the `GIT_CONFIG_COUNT` / `GIT_CONFIG_KEY_n` / `GIT_CONFIG_VALUE_n` row and, critically, states the precedence consequence in prose: caller helpers are consulted first and *"a helper that returns complete credentials can shadow Symphony's tenant-scoped helper."* That is the trust decision, stated plainly. The PR `## Risks & rollback` section was widened past "parsing" to name the `.env` trust boundary, which is the second half of what I asked for.

### Non-blocking declines — accepted

All three were declined as out of scope for a documentation-only rework, each with a stated reason on the thread. I verified the two that carried a factual claim:

- **`GIT_TERMINAL_PROMPT`** — confirmed still `"0"` on *both* paths at head, including when a caller passes `GIT_TERMINAL_PROMPT=1` alongside a token. The coupling to `runtime-codex` remains unpinned, which is what I actually objected to, but it is latent and not worth another cycle.
- **Single-entry factory assumption** — still correct today; the factory emits exactly one indexed entry.

Both are worth a follow-up issue rather than a third cycle here.

### Verified independently

`pnpm install --frozen-lockfile`, `pnpm build`, `pnpm lint` (exit 0), `pnpm typecheck` (exit 0); `git-transport.test.ts` green at 20/20; `prettier --check docs/configuration.md` clean. Both CI checks (`Test`, `Container Smoke`) are green on `a7cbb1a`. No Docker E2E — correct call, this is environment composition with no lifecycle or integration boundary crossed.

**No changeset is correct here.** I checked this rather than assuming: `WORKFLOW.md:231-232` gates changesets on a `changeset:major|minor|patch` label, and #832 carries only `bug`. The author's stated reason matches repo policy.

### One correction to my own cycle-1 nit

I claimed `"2k"` was *"accepted by `git_parse_ulong` but rejected here."* That is wrong — real Git rejects it too (`error: bogus count in GIT_CONFIG_COUNT`). The only genuine divergence from Git's parser is leading whitespace (`" 2"`), which Git accepts and this rejects. Stricter on one whitespace case is not worth changing; I'm noting it so the record isn't misleading.

### No over-engineering

Three files, 32 lines of production code, one guard, legacy shape preserved exactly. The count validation is the minimum needed to compute a safe append index — not scope creep.

Approving. The `009` → helper-at-index-7 behavior matches Git's own octal-insensitive parse, which was the case I most expected to be wrong and wasn't.

---

PR #837, opened 2026-09-03
fix(worker,orchestrator): isolate worker environment

### LINKED ISSUE #810: fix(worker,orchestrator): drop the implicit <cwd>/.env read and spawn workers with a run-scoped cwd
Part of #808 (root-cause chain item 5).

## Problem

The orchestrator spawns the worker with `cwd: process.cwd()` (`packages/orchestrator/src/service.ts`, worker `spawn`), i.e. the repository root in repo mode. `loadLauncherEnvironment` (`packages/runtime-codex/src/launcher.ts`) merges `<package>/.env`, `<cwd>/.env` and `process.env`. A repository-root `.env` (typically the application's own secrets) is therefore silently loaded into the worker host environment, bypassing the documented precedence (project `.env` → orchestrator env → Symphony context, `docs/configuration.md`).

## Fix outline

1. Spawn the worker with a run-scoped cwd (the run's `workspaceRuntimeDir` or the issue workspace), not the daemon cwd.
2. Remove the `<cwd>/.env` read from `loadLauncherEnvironment` (or restrict it to the explicit `SYMPHONY_PROJECT_DIR/.env`, which the orchestrator already loads), and document the single supported source.
3. `doctor` warns when a repository-root `.env` exists in repo mode and explains it is not loaded.

## Regression tests

- Worker launched with a repository-root `.env` present → its keys are absent from the worker env.
- Project `.env` still applies.

moncher-dev:
## Issues

- Closes #810

## Summary

- Spawn workers from their run-scoped runtime directory instead of the daemon repository cwd.
- Stop runtime-codex from implicitly loading `<cwd>/.env`; the orchestrator-managed project `.env` remains the supported environment source.
- Warn repo-mode operators when a repository-root `.env` exists but is intentionally not loaded.

## Change-point diagram

- orchestrator run setup → worker `cwd = workspaceRuntimeDir`
- runtime-codex launcher → package defaults + supplied environment → no ambient cwd `.env`
- CLI doctor in repo mode → repository-root `.env` detected → intentional-not-loaded warning

## Start here

- `packages/orchestrator/src/service.ts` — run-scoped worker spawn boundary
- `packages/runtime-codex/src/launcher.ts` — launcher environment-source policy
- `packages/cli/src/commands/doctor.ts` — repo-mode diagnostic

- Repository application secrets in a root `.env` are no longer silently inherited by workers.
- `doctor` explains that repository-root `.env` is ignored and directs Symphony-specific values to the managed project environment.
- `SYMPHONY_WORKER_COMMAND` runs from the run-scoped runtime directory; custom wrappers must use an absolute path or one derived from `$WORKING_DIRECTORY`.

## Validation

- `pnpm exec vitest run packages/runtime-codex/src/launcher.test.ts packages/orchestrator/src/service.test.ts packages/cli/src/commands/doctor.test.ts` — pass (3 files, 295 tests; Cycle 2)
- `pnpm lint && pnpm test && pnpm typecheck && pnpm build` — pass across all workspace packages (Cycle 2)
- `pnpm lint` — pass after docs-only rework
- `pnpm exec prettier --check docs/configuration.md` — pass after docs-only rework
- `git diff --check origin/main...HEAD` — pass at local committed head `4dde742`
- `./e2e/run-e2e.sh happy 60` — environment exception: Docker Compose v2 is unavailable, so startup exits 125; isolated rerun reproduced it and follow-up #838 records the prerequisite failure

## Changeset

- `.changeset/isolate-worker-environment.md` (`@gh-symphony/cli`, patch)

## Risks & rollback

- Risk: integrations that relied on daemon-cwd `.env` inheritance must move required values to the configured project `.env` or orchestrator environment; relative custom worker wrappers must use an explicit stable path.
- Reviewer focus: environment precedence and whether every worker launch path uses the run-scoped directory.
- Rollback: revert the issue #810 commits on this branch.

## Changed files

- `packages/orchestrator/src/service.ts` and test — use and verify the run-scoped worker cwd.
- `packages/runtime-codex/src/launcher.ts` and test — remove and regress implicit cwd `.env` loading.
- `packages/cli/src/commands/doctor.ts` and test — report ignored repository-root `.env` files in repo mode.
- `README.md`, `packages/cli/README.md`, `docs/configuration.md`, `docs/architecture.md`, and `AGENT_TEST.md` — document supported environment loading, worker-command path resolution, architecture, and regression coverage.
- `.changeset/isolate-worker-environment.md` — publish the behavior change as a CLI patch.

- [ ] Restore Docker Compose v2 availability and rerun `./e2e/run-e2e.sh happy 60` (tracked in #838).

## Security


moncher-dev:
Smoke-tested at `12a26cd8`. `pnpm install --frozen-lockfile`, `pnpm build`, `pnpm typecheck`, `pnpm lint` all exit 0; suites green across the three touched packages — cli 617, orchestrator 397, runtime-codex 64 (**1078 tests**).

Because this is epic #808's worker-environment boundary child, I exercised the new behavior against the real repo rather than the unit suite alone, with a negative control on the check it adds:

```
no repository-root .env
  PASS Repository environment file

with a repository-root .env present
  WARN Repository environment file
    Repository-root .env exists at /Users/steve/Projects/github-symphony/.env
    and is not loaded by Symphony.
    Fix: Move Symphony-specific values to the managed project .env; keep
         application secrets in the repository-root .env.
```

So the check is load-bearing rather than vacuous, and the remediation text points at the right place. (Probe file contained one inert variable, was present for a few seconds with zero active runs, and was removed.)

The core change is `cwd: process.cwd()` → `cwd: workspaceRuntimeDir` at the worker spawn, which is exactly what #810 asked for, and `readProjectEnv` now resolves `tenant.projectDir ?? store.projectDir(...)` + `/.env`.

**An operational note for whoever ships this, since it affects this host directly.** I checked whether the run-scoped cwd breaks the temporary mitigations this operator host is running, and it does not: the tenant `.env` at `.runtime/orchestrator/projects/repository/.env` (currently carrying `GITHUB_GRAPHQL_TOKEN` and `DOCKER_CONFIG`) lives in the *project* directory, which `readProjectEnv` still reads. It is the *repository-root* `.env` that stops being read. Worth stating explicitly because "drop the implicit `<cwd>/.env` read" reads like it would break both, and it does not.

**On the Codex P2 (relative `SYMPHONY_WORKER_COMMAND`) — a real behavior change, with the bot's framing overstated.** Moving cwd to the run directory does mean a relative wrapper such as `node ./scripts/worker.js` now resolves from the run dir rather than the daemon's startup dir, and would fail. But the bot cites `docs/configuration.md:561` as documenting that use, and it does not — that row says only "Shell command used to start worker processes. Useful for local E2E, debugging, or custom worker wrappers," with no relative-path guidance either way. This branch adds one line noting workers are spawned from their run-scoped runtime directory, which is a start but does not call out the consequence.

So: not a blocker in my view, but a decision worth making deliberately rather than by omission — either document that `SYMPHONY_WORKER_COMMAND` must now be absolute, or resolve it against the daemon's startup directory before spawning. Silently breaking a working relative wrapper with a worker that exits before starting would be a hard failure to diagnose.

No over-engineering; the production change is small and localized, and the docs changes are proportionate. @hojinzs over to you.

packages/orchestrator/src/service.ts
```diff
@@ -3222,7 +3222,7 @@ export class OrchestratorService {
       "bash",
       ["-lc", resolveWorkerCommand()],
       {
-        cwd: process.cwd(),
+        cwd: workspaceRuntimeDir,
```
hojinzs, line 3225: **P2 — `SYMPHONY_WORKER_COMMAND` changes its relative-path resolution base, and the docs row still doesn't say so.**

This confirms the earlier Codex thread ([r3928229314](https://github.com/hojinzs/github-symphony/pull/837#discussion_r3928229314)), which is still unresolved and unanswered.

The command runs via `bash -lc`, so its cwd is now the run directory instead of the daemon startup directory. `docs/configuration.md:561` advertises this variable for "local E2E, debugging, or custom worker wrappers", and a wrapper written the natural way — `node ./scripts/worker.js` — now dies at startup with a bare `Cannot find module` and no hint about why.

I checked the blast radius and it is narrow: every in-repo caller already uses an absolute path (`package.json:10` expands `$(pwd)`, `docker/orchestrator-e2e.Dockerfile:54` uses `/app/...`), so nothing here breaks. It is the out-of-repo operator with a relative wrapper who gets a silent failure in a patch release.

I don't think this needs a code change — resolving the command against a second base would reintroduce exactly the daemon-cwd coupling this PR removes. **A sentence in the `SYMPHONY_WORKER_COMMAND` row of `docs/configuration.md` is enough**, e.g.:

> The command is executed from the run-scoped runtime directory, so use an absolute path (or one derived from `$WORKING_DIRECTORY`); repository-relative wrapper paths are not resolved.

`docs/architecture.md` already documents the new cwd, but the configuration table is where someone configuring this variable actually looks.

---

packages/cli/src/commands/doctor.ts
```diff
@@ -579,6 +581,42 @@ function buildPathCheck(
...
+    return warnCheck(
+      "repository_environment",
+      "Repository environment file",
+      `Repository-root .env exists at ${envPath} and is not loaded by Symphony.`,
+      "Move Symphony-specific values to the managed project .env; keep application secrets in the repository-root .env.",
+      { path: envPath }
+    );
+  } catch (error) {
+    const err = error as NodeJS.ErrnoException;
+    if (err.code !== "ENOENT" && err.code !== "ENOTDIR") {
```
hojinzs, line 607: **P3 (non-blocking) — a non-ENOENT `stat` error here aborts the entire `doctor` run.**

`buildRepositoryEnvironmentCheck` is awaited directly inside `runDoctorDiagnostics`, so an `EACCES`/`EPERM` on `<repo>/.env` (root-owned file, restrictive mount, hardened CI image) propagates out and takes down every remaining diagnostic — including the checks the operator was actually running `doctor` to see.

`inspectPathState` above (~L505) has the same ENOENT/ENOTDIR narrowing but is the pre-existing convention for path checks that already gate on it. For a purely informational check like this one, degrading is strictly better than aborting: return a `warn` saying the repository-root `.env` could not be inspected, rather than rethrowing.

```ts
if (err.code !== "ENOENT" && err.code !== "ENOTDIR") {
  return warnCheck(
    "repository_environment",
    "Repository environment file",
    `Could not inspect ${envPath} (${err.code ?? "unknown error"}).`,
    "Check filesystem permissions on the repository root, then re-run 'gh-symphony doctor'.",
    { path: envPath }
  );
}
```

---

packages/cli/src/commands/doctor.ts
```diff
@@ -579,6 +581,42 @@ function buildPathCheck(
...
+      throw error;
+    }
+  }
+
+  return passCheck(
+    "repository_environment",
+    "Repository environment file",
+    `No repository-root .env exists at ${envPath}.`,
```
hojinzs, line 615: **nit** — #810 asked only for the warning; this adds a permanent `PASS Repository environment file / No repository-root .env exists at …` row to every repo-mode `doctor` run, which is a line of output that never tells the operator anything. Returning `null` in the absent case would keep the surface quiet.

I'm not pushing on this — a pass row is consistent with how the rest of the check list renders, so it's a judgement call, not a defect. Just noting it as scope slightly beyond the issue.

---

hojinzs (changes requested):
## Review — round 1

Requirements from #810 are met and I verified the behavior end-to-end rather than trusting the unit tests. One blocking item, and it's a one-line docs edit.

### Requirement coverage vs #810

| # | Requirement | Status |
|---|---|---|
| 1 | Spawn worker with run-scoped cwd, not daemon cwd | ✅ `service.ts:3225` → `workspaceRuntimeDir`; `mkdirSync(runDir)` at `:3191` runs before spawn, so the cwd always exists |
| 2 | Remove `<cwd>/.env` from `loadLauncherEnvironment`, document the single source | ✅ removed; docs updated in 5 files |
| 3 | `doctor` warns on repository-root `.env` in repo mode | ✅ new `repository_environment` check |
| RT | Repo-root `.env` keys absent from worker env | ✅ unit + black-box below |
| RT | Project `.env` still applies | ✅ existing coverage (`service.test.ts:16534`, `:16566`) + black-box below |

### Smoke test (Docker unavailable here too, so I ran the local repo-mode path)

`pnpm e2e:init`, then started the packaged CLI in repo mode against the file tracker with an adversarial `<repo>/.env` (`APP_ONLY_SECRET=leaked-value`) and a managed project `.env` (`PROJECT_ENV_MARKER=managed-project-value`), and probed the real spawned worker:

```json
{
  "cwd": ".../projects/repository/runs/repository-test-owner-test-repo-1-mtlzfwmn",
  "WORKSPACE_RUNTIME_DIR": ".../runs/repository-test-owner-test-repo-1-mtlzfwmn",
  "WORKING_DIRECTORY": ".../symphony-workspaces/test-owner_test-repo_1-484e36eec6eba867/repository",
  "APP_ONLY_SECRET": null,
  "PROJECT_ENV_MARKER": "managed-project-value"
}
```

All three properties hold: cwd is the run directory, the repository-root secret does not reach the worker, and the managed project `.env` still does. Re-running the unmodified stub worker, the happy path completes cleanly (`status=completed`, `runPhase: succeeded`) — the cwd change breaks no lifecycle step.

`gh-symphony doctor` in that repo emits the new check verbatim:

```
WARN Repository environment file
  Repository-root .env exists at .../test-repo/.env and is not loaded by Symphony.
  Fix: Move Symphony-specific values to the managed project .env; keep application secrets in the repository-root .env.
```

### Gates

`pnpm lint` ✅ · `pnpm typecheck` ✅ · `pnpm build` ✅ · targeted suites ✅ (295 tests, matching the PR body). `pnpm format` fails on 84 files, but none of them are touched by this PR (`docs/README.md`, `packages/control-plane/README.md`, `pnpm-lock.yaml`, …) — pre-existing, not yours.

### Isolation is actually complete

I swept for other leak paths rather than assuming: every remaining `.env` read is project-dir- or package-dir-scoped (`managed-project-environment.ts:16`, `service.ts:4909`, `mcp-compose.ts:100`, `launcher.ts:98`), and no CLI path loads a cwd `.env`. The cwd change is also contained — the worker has zero `process.cwd()` calls and derives everything from injected absolute env vars, and both runtime adapters pass an explicit `cwd` (`runtime-codex/src/runtime.ts:637`, `runtime-claude/src/adapter.ts:410`). `service.ts:5623`'s `process.cwd()` fallback is for `WORKING_DIRECTORY`, which is correctly unchanged.

### Over-engineering / scope

Clean. Two one-line code changes plus a diagnostic — no speculative abstraction. The removed `cwd` parameter is safe: `@gh-symphony/runtime-codex` is `private: true`, so no published API breaks. Only the `doctor` pass row (inline nit) goes slightly past the issue.

### Blocking

**P2** — the `SYMPHONY_WORKER_COMMAND` docs row doesn't mention that its relative-path resolution base moved. This is the Codex thread [r3928229314](https://github.com/hojinzs/github-symphony/pull/837#discussion_r3928229314), still unresolved and unanswered. I confirmed no in-repo caller breaks (all use absolute paths), so this is documentation, not code — see the inline comment for the suggested sentence. I'd rather not resolve the change of a documented user-facing variable with an unanswered review thread on it.

The P3 and the nit are explicitly non-blocking; take or leave them. Add the docs sentence (and either fix or reply to the Codex thread) and I'll approve — no further rounds needed from my side.

---

packages/orchestrator/src/service.ts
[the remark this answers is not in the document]
moncher-dev: Documented in `docs/configuration.md`: the command runs from the run-scoped runtime directory, so operators must use an absolute path or one derived from `$WORKING_DIRECTORY`; repository-relative wrapper paths are not resolved. No code change is needed, and the in-repo callers already use absolute paths.

moncher-dev: Documented the resolution rule in the `SYMPHONY_WORKER_COMMAND` row with the suggested operator guidance. The implementation remains decoupled from the daemon startup directory.

packages/cli/src/commands/doctor.ts
moncher-dev: Leaving this unchanged in this PR. Rethrowing non-ENOENT/ENOTDIR errors matches the existing `inspectPathState` convention for filesystem diagnostics; changing that error policy only for this informational check would widen this narrowly scoped rework.

packages/cli/src/commands/doctor.ts
moncher-dev: Keeping the PASS row. It is consistent with the rest of the doctor check list and provides a useful negative control that the repository-root `.env` condition was actually inspected.

hojinzs (approved):
## Review — round 2 (final): Approve

The round-1 blocker is resolved and the delta is exactly what I asked for. No new findings, so this is my last round.

### Round-1 items

| Item | Status |
|---|---|
| **P2 (blocking)** — `SYMPHONY_WORKER_COMMAND` docs row omits the moved resolution base | ✅ Fixed in `4dde742`. The row now carries the suggested sentence verbatim: *"The command runs from the run-scoped runtime directory, so use an absolute path or one derived from `$WORKING_DIRECTORY`; repository-relative wrapper paths are not resolved."* |
| Codex thread [r3928229314](https://github.com/hojinzs/github-symphony/pull/837#discussion_r3928229314) | ✅ Answered and resolved |
| **P3** — non-ENOENT `stat` error aborts the whole `doctor` run | ✅ Declined with reason (matches the existing `inspectPathState` convention; would widen scope). Accepted — I flagged it non-blocking and the rationale is the one I'd have given myself. |
| **nit** — permanent PASS row beyond #810's ask | ✅ Declined with reason (consistency + negative control). Accepted — it was a judgement call. |

### Delta verification

`git diff 12a26cd..4dde742` touches **only** `docs/configuration.md`. The code reviewed in round 1 is byte-identical, so the requirement coverage and isolation sweep from that review carry forward unchanged. The remaining churn in the commit is Prettier column re-padding of the table, not content.

### Gates at `4dde742`

- **CI** — `Test` ✅ · `Container Smoke` ✅ (both complete, both green)
- **Local** — `pnpm lint` ✅ · `pnpm build` ✅ · targeted suites ✅ (3 files, 295 tests, matching the PR body) · `prettier --check docs/configuration.md` ✅

### Smoke test

Re-ran the isolation property black-box against the **built** `dist` artifact (not the source), with an adversarial cwd-root `.env`:

```json
{
  "cwd": ".../scratchpad/smoke",
  "APP_ONLY_SECRET": null,
  "PROJECT_ENV_MARKER": "managed-project-value"
}
```

The repository-root secret does not reach the launcher environment, and the supplied environment still applies. This matches the round-1 end-to-end run against a real spawned worker (run-scoped cwd, no leaked secret, managed project `.env` intact, happy path `status=completed`).

### Scope

Still clean — two one-line code changes plus a diagnostic, no speculative abstraction. Nothing added in round 2 beyond the docs sentence.

Requirements from #810 (run-scoped cwd, `<cwd>/.env` removal, `doctor` warning, both regression tests) are fully met. Ship it.

---

PR #840, opened 2026-09-03
test(e2e): report unavailable Docker runtime clearly

### LINKED ISSUE #817: test(e2e): report unavailable Docker runtime clearly

moncher-dev:
## Issues

- Closes #817

## Summary

- Add a shared Docker runtime preflight to every Docker Compose E2E runner.
- Distinguish a missing Docker command, an unresolved user-level Compose plugin, and an unreachable daemon with actionable diagnostics.
- Exit unavailable environments with status 69 before cleanup traps or container startup, separating prerequisites from scenario failures and Docker's status 125.

## Change-point diagram

- Four Docker E2E runners → shared `assert_docker_runtime_is_available` → Docker command / Compose plugin / daemon probes → cleanup trap and scenario startup only when healthy

## Start here

- `e2e/lib/compose-project.sh:7` — unavailable-runtime exit contract and shared preflight
- `e2e/lib/compose-project.test.sh:55` — runtime-state and all-runner ordering regression coverage

- Docker E2E commands now stop before container startup with a precise diagnostic and status 69 when Docker prerequisites are unavailable. Missing Compose diagnostics name `$HOME/.docker/cli-plugins/` and print the effective `HOME` and `DOCKER_CONFIG` values.

## Validation

- `pnpm lint` — pass
- `pnpm test` — pass (all 14 participating workspace packages)
- `pnpm typecheck` — pass
- `pnpm build` — pass
- `bash e2e/lib/compose-project.test.sh` — pass; covers runtime states and all four runner preflight/order invariants
- `bash -n e2e/lib/compose-project.sh e2e/lib/compose-project.test.sh e2e/run-e2e.sh e2e/run-flat-tracker-keys-e2e.sh e2e/run-standalone-project-e2e.sh test/e2e/claude/run-docker-e2e.sh` — pass
- All four Docker E2E runners — expected unavailable-runtime result: actionable Compose diagnostic, exit 69, no cleanup noise, and no container startup
- `git diff --check` — pass
- Docker E2E / blackbox evidence: all four runners exercised directly against the host's unavailable Compose environment

## Changeset

- Not needed because the issue has no changeset label and changes only the repository E2E harness.

## Risks & rollback

- Risk is limited to prerequisite classification in Docker test runners. The daemon probe remains unbounded because GNU `timeout` is not portable to supported macOS hosts. Revert the three issue commits to restore direct Compose invocation.

## Changed files

- `e2e/lib/compose-project.sh` — shared Docker runtime probes, actionable diagnostics, and exit contract
- `e2e/lib/compose-project.test.sh` — healthy/unavailable runtime cases and all-runner ordering coverage
- `e2e/run-e2e.sh` — invoke preflight before TC-11
- `e2e/run-flat-tracker-keys-e2e.sh` — invoke preflight before TC-22 cleanup setup
- `e2e/run-standalone-project-e2e.sh` — invoke preflight before TC-13 project/cleanup setup
- `test/e2e/claude/run-docker-e2e.sh` — invoke preflight before TC-20
- `AGENT_TEST.md` — document the all-runner Docker preflight behavior

- [ ] None

## Security


moncher-dev:
Smoke-tested at `584f5121`, and this is the rare case where the broken environment is the ideal test fixture — Docker E2E genuinely does not work on this host, so I could exercise the real failure rather than simulate it.

**Baseline on `main`, same host, same conditions:**
```
[e2e] Scenario: happy (timeout: 30s)
[e2e] Compose project: symphony-e2e-1a11b2885a6e
Run 'docker --help' for more information
[e2e] Cleaning up...
```
That is the entire diagnostic. Four separate workers hit exactly this and reached four different wrong conclusions — "cannot start Docker", "Docker Engine not running", "Compose plugin not installed", "assigned branch context missing".

**With this PR:**
```
[e2e] Docker runtime unavailable: 'docker compose' cannot be resolved by this process.
[e2e] A user-level Compose plugin may be hidden by the current environment
      (HOME=/nonexistent, DOCKER_CONFIG=<unset>).
[e2e] Compose probe: docker: unknown command: docker compose
```

That is the correct root cause, surfaced automatically. It took me a manual `HOME=/nonexistent docker compose version` reproduction plus a read of `SAFE_RUNTIME_ENV_KEYS` to reach the same conclusion earlier in this session; naming `HOME` and `DOCKER_CONFIG` in the message is precisely the detail that collapses that investigation.

**All three cases and both entrypoints verified, with exit codes:**

| Case | Result |
|---|---|
| Compose unresolved, `e2e/run-e2e.sh` | diagnostic above, **exit 69** |
| Compose unresolved, `pnpm e2e:claude` path | identical diagnostic, **exit 69** |
| `docker` absent from PATH | `the 'docker' command is not on PATH.`, **exit 69** |
| `e2e/lib/compose-project.test.sh` | passes |

Exit 69 is the right call — it separates "this environment cannot run the test" from Docker's own 125 and from a genuine scenario failure, which is what lets callers stop conflating the two.

The diff is shell and docs only (no TypeScript), and CI is green on the head.

**This also closes #841** — filed independently asking for a `docker compose version` preflight in `test/e2e/claude/run-docker-e2e.sh` specifically. That file is in this diff and the shared `assert_docker_runtime_is_available` covers it; I verified the Claude entrypoint directly above. I have closed #841 as covered rather than opening a parallel track.

@hojinzs over to you.

e2e/lib/compose-project.sh
```diff
@@ -4,6 +4,33 @@
 # concurrent worktrees separate while allowing an explicit project to be used
 # for debugging.
 
+E2E_DOCKER_UNAVAILABLE_EXIT=69
+
+assert_docker_runtime_is_available() {
```
hojinzs, line 9: **P2 — two other Docker E2E runners still emit the opaque failure this issue exists to remove.**

`compose-project.sh` is sourced by four runners, but only two of them call the new preflight:

| Runner | Sources this lib | Calls preflight |
|---|---|---|
| `e2e/run-e2e.sh` (TC-11) | ✅ | ✅ |
| `test/e2e/claude/run-docker-e2e.sh` (TC-20) | ✅ | ✅ |
| `e2e/run-standalone-project-e2e.sh` (TC-13, `pnpm e2e:standalone-project`) | ✅ | ❌ |
| `e2e/run-flat-tracker-keys-e2e.sh` (TC-22) | ✅ | ❌ |

I reproduced this at `584f5121` with a stub `docker` that mimics an unresolvable user-level Compose plugin under `HOME=/nonexistent` — the exact condition from the issue:

```
$ pnpm e2e:standalone-project
docker: 'compose' is not a docker command.
See 'docker --help'
exit=125

$ ./e2e/run-flat-tracker-keys-e2e.sh
docker: 'compose' is not a docker command.
See 'docker --help'
exit=125
```

That is the misdiagnosis trap described in #817 — four workers, four wrong conclusions — just reached through a different door. Both files already `source` this lib, so the fix is one line each, above the first Compose invocation:

```bash
assert_docker_runtime_is_available
```

Placement note: `run-flat-tracker-keys-e2e.sh:13` and `run-standalone-project-e2e.sh:17` install `trap cleanup EXIT` before their first Compose call, so put the preflight *before* the trap — as you correctly did in the two runners in this diff — to keep the failure free of cleanup noise.

The `AGENT_TEST.md` bullet would then read "all Docker E2E runners" rather than naming two, which is also the more durable wording for the next runner someone adds.

I'd rather see this land here than as a follow-up: it's the same three-line change against a helper this PR already introduces, and splitting it leaves two documented scenarios (TC-13, TC-22) reproducing the original bug.

---

e2e/lib/compose-project.sh
```diff
@@ -4,6 +4,33 @@
...
+
+  if ! compose_version=$(docker compose version 2>&1); then
+    compose_error="$compose_version"
+    echo "[e2e] Docker runtime unavailable: 'docker compose' cannot be resolved by this process." >&2
+    echo "[e2e] A user-level Compose plugin may be hidden by the current environment (HOME=${HOME:-<unset>}, DOCKER_CONFIG=${DOCKER_CONFIG:-<unset>})." >&2
+    echo "[e2e] Compose probe: ${compose_error:-no diagnostic output}" >&2
+    return "$E2E_DOCKER_UNAVAILABLE_EXIT"
+  fi
+
+  if ! daemon_error=$(docker info 2>&1 >/dev/null); then
```
hojinzs, line 24: **P3 — the daemon probe is unbounded.**

`docker info` returns promptly when the socket is absent (I measured `0s` on this host, where the socket file does not exist). But against a socket that exists and does not answer — a daemon mid-restart, a stale socket, a remote `DOCKER_HOST` — it blocks on the connection with no client-side deadline. A preflight whose whole purpose is to fail fast with a clear diagnostic is the one place where hanging is worse than the opaque `125` it replaces, because a hang produces no diagnostic at all.

Cheap guard:

```bash
if ! daemon_error=$(timeout 30 docker info 2>&1 >/dev/null); then
```

`timeout` is coreutils and already assumed elsewhere in the harness. If you'd rather not take that on, fine to leave — flagging it as the one runtime state the three probes don't cover.

---

e2e/lib/compose-project.sh
```diff
@@ -4,6 +4,33 @@
...
+
+  if ! command -v docker >/dev/null 2>&1; then
+    echo "[e2e] Docker runtime unavailable: the 'docker' command is not on PATH." >&2
+    return "$E2E_DOCKER_UNAVAILABLE_EXIT"
+  fi
+
+  if ! compose_version=$(docker compose version 2>&1); then
+    compose_error="$compose_version"
+    echo "[e2e] Docker runtime unavailable: 'docker compose' cannot be resolved by this process." >&2
+    echo "[e2e] A user-level Compose plugin may be hidden by the current environment (HOME=${HOME:-<unset>}, DOCKER_CONFIG=${DOCKER_CONFIG:-<unset>})." >&2
```
hojinzs: **nit — name the plugin directory, not just the variables.**

#817 asked the message to point at the cause: "the Compose plugin is user-level at `$HOME/.docker/cli-plugins/`, so an isolated or run-scoped HOME will not resolve it." Naming `HOME` and `DOCKER_CONFIG` gets most of the way there, but a reader who doesn't already know *why* `HOME` matters to a Docker subcommand still has a lookup to do — and that lookup is exactly the cost this issue removes.

```bash
echo "[e2e] The Compose plugin is user-level (\$HOME/.docker/cli-plugins/), so a run-scoped HOME will not resolve it (HOME=${HOME:-<unset>}, DOCKER_CONFIG=${DOCKER_CONFIG:-<unset>})." >&2
```

Non-blocking — the current text is already a large improvement over `Run 'docker --help'`.

---

e2e/lib/compose-project.sh
```diff
@@ -4,6 +4,33 @@
...
+    echo "[e2e] Compose probe: ${compose_error:-no diagnostic output}" >&2
```
hojinzs: **nit — `compose_error` is an alias for a variable that's already in scope.**

`compose_error="$compose_version"` copies a value that is still live two lines later; the indirection also makes `compose_version` read as two different things depending on the branch. Using `${compose_version:-no diagnostic output}` directly on line 21 is one fewer name to track, and `compose_error` then comes off the `local` declaration on line 10.

---

e2e/lib/compose-project.sh
```diff
@@ -4,6 +4,33 @@
...
+    return "$E2E_DOCKER_UNAVAILABLE_EXIT"
+  fi
+
+  if ! daemon_error=$(docker info 2>&1 >/dev/null); then
+    echo "[e2e] Docker runtime unavailable: Docker Compose is installed, but the Docker daemon is not reachable." >&2
+    echo "[e2e] Daemon probe: ${daemon_error:-no diagnostic output}" >&2
+    return "$E2E_DOCKER_UNAVAILABLE_EXIT"
+  fi
+
+  echo "[e2e] Docker runtime available: ${compose_version}; daemon reachable."
```
hojinzs: **nit — success line is unconditional stdout on every healthy run.**

On a healthy host this prints on every run and nothing consumes it; the three failure branches are what carry the value. It also renders inconsistently across entrypoints: `run-e2e.sh` puts all its `[e2e]` output through a colored `log()` helper, while `run-docker-e2e.sh` has no such helper — so this line matches neither convention, and the lib can't adopt `log()` without breaking the runners that don't define it.

Either drop it, or gate it:

```bash
[ -n "${E2E_DEBUG:-}" ] && echo "[e2e] Docker runtime available: ${compose_version}; daemon reachable."
```

If you keep it as-is, note that `compose-project.test.sh:58` asserts its exact text, so the assertion moves with it.

---

hojinzs (changes requested):
## Review — round 1 of 6

Verified at head `584f5121`. The core design is right: a shared helper, three distinct runtime states, an exit code that separates "this environment cannot run the test" from "the test failed". Scope is tight (73 added lines, shell + docs only) and I found **no over-engineering and no unrequested design decisions** — every element traces to a line in #817's consolidated scope. Exit `69` is `EX_UNAVAILABLE` from `sysexits.h`, which is the right pick for "distinct non-125 status" rather than an arbitrary number.

One blocking item (P2, inline), everything else non-blocking.

### Requirement traceability vs. #817

| # | Requirement (from the consolidated scope comment) | Status |
|---|---|---|
| 1 | Cover both entrypoints — `run-e2e.sh` (TC-11) and `pnpm e2e:claude` (TC-20) | ✅ met as written — but see P2: two *sibling* runners still reproduce the bug |
| 2 | Probe for a *working* `docker compose`, not just the `docker` binary | ✅ `docker compose version` exit-0 probe |
| 3 | Handle both symptoms: `unknown command: docker compose` **and** the `unknown flag: --project-name` fallthrough | ✅ prevented at the source — both stem from the same unresolved plugin, and the probe fires before any Compose flag is passed. Verified below. |
| 4 | Name the likely cause (`HOME`, user-level plugin dir) | ⚠️ names `HOME`/`DOCKER_CONFIG`, not the `cli-plugins` path — nit inline |
| 5 | Distinct exit status from a scenario failure | ✅ `69`, documented in `AGENT_TEST.md` |

Correct call on the changeset: this touches only the repository E2E harness, not `@gh-symphony/cli` behavior, so per CLAUDE.md none is required.

### Smoke test — executed, not simulated

This host has the same broken Compose environment described in #825, plus an unreachable daemon, so I exercised the real failures rather than stubs. All three branches, both patched entrypoints:

| Case | Entrypoint | Result |
|---|---|---|
| Daemon unreachable (**real**, socket absent) | `./e2e/run-e2e.sh fail 45` | correct diagnostic + real `dial unix /var/run/docker.sock` probe text, **exit 69**, `0s`, before container startup |
| Daemon unreachable (**real**) | `bash test/e2e/claude/run-docker-e2e.sh` | identical, **exit 69** |
| Compose unresolvable (stub `docker`, `HOME=/nonexistent`) | `./e2e/run-e2e.sh fail 45` | names `HOME=/nonexistent, DOCKER_CONFIG=<unset>` + verbatim probe output, **exit 69** |
| Compose unresolvable | `bash test/e2e/claude/run-docker-e2e.sh` | identical, **exit 69** |
| `docker` absent from PATH | unit test | **exit 69**, correct message |
| `bash e2e/lib/compose-project.test.sh` | — | passes |
| `bash -n` on all four touched scripts | — | clean |

Worth calling out: the **daemon-down branch was only stub-tested in your validation**, and it holds up against a real Docker CLI — the probe's `2>&1 >/dev/null` ordering correctly captures stderr only, and the real daemon error surfaces intact.

Also confirmed the preflight sits **before** `trap cleanup EXIT` in both files, so an unavailable runtime exits clean with no `Cleaning up...` noise and without clobbering `e2e/fixtures/issues.json`. That ordering is easy to get wrong and you got it right; it's the reason the P2 comment asks for the same placement in the two remaining runners.

CI is green on the head (Test, Container Smoke).

### Verdict

**Request changes** on the P2 only. It's a three-line change (two `assert_docker_runtime_is_available` calls plus the `AGENT_TEST.md` wording) using the helper this PR already builds — I'd rather not merge a preflight that leaves two documented scenarios emitting the exact `exit 125` / `See 'docker --help'` output the issue was filed to eliminate.

Push that and I'll approve; the P3 and three nits are yours to take or decline with a reason, and none of them will hold up a second round.

---

e2e/lib/compose-project.sh
moncher-dev: Fixed in `9865b78`. Both remaining runners now call `assert_docker_runtime_is_available` before installing their cleanup traps. `AGENT_TEST.md` now documents the invariant for all Docker E2E runners, and the shell test checks all four runner files for both presence and ordering.

e2e/lib/compose-project.sh
moncher-dev: I am leaving the daemon probe unbounded in this change. The suggested `timeout` executable is GNU coreutils and is not available by default on macOS, while these repository scripts support macOS hosts. Adding a portable deadline would require a separate fallback/process-management design beyond the unavailable-Compose diagnostic scope; the daemon-unreachable path remains covered and returns promptly for the observed local-socket failure.

e2e/lib/compose-project.sh
moncher-dev: Fixed in `9865b78`. The diagnostic now names the literal `$HOME/.docker/cli-plugins/` location while continuing to print the effective `HOME` and `DOCKER_CONFIG` values. The focused test asserts the plugin path is present.

e2e/lib/compose-project.sh
moncher-dev: Fixed in `9865b78`. The helper now reports `${compose_version:-no diagnostic output}` directly and no longer declares or assigns `compose_error`.

e2e/lib/compose-project.sh
moncher-dev: Fixed in `9865b78`. Healthy preflight is now silent, and the focused test asserts empty output for a healthy Docker runtime.

e2e/lib/compose-project.sh
hojinzs: Accepted — and my premise was wrong, so thank you for checking it rather than just complying.

I claimed `timeout` was "already assumed elsewhere in the harness." It isn't. I grepped for an actual `timeout(1)` invocation across `e2e/` and `test/` and there are none; the only `timeout` tokens are the scenario-duration variable in `run-e2e.sh` and Compose's own `--timeout` flag, neither of which is the coreutils binary.

Your portability point is also confirmed by this very file: `e2e_path_hash` (`compose-project.sh:36-39`) falls back from `sha256sum` to `shasum -a 256`, a fallback that exists precisely because macOS lacks `sha256sum`. So "these scripts support macOS hosts" is this file's own established convention, not a hypothetical — and adding an unguarded GNU-only dependency to it would have been the inconsistent choice.

Agreed that a portable deadline is a separate design (a `gtimeout`/`timeout` probe with a background-PID fallback) and out of scope for an unavailable-Compose diagnostic. Not re-raising. Resolving this thread as declined-with-cause, not as fixed.

---

e2e/lib/compose-project.test.sh
```diff
@@ -33,11 +33,66 @@ stub_dir=$(mktemp -d)
...
+missing_docker_output=$(PATH="$stub_dir/empty" assert_docker_runtime_is_available 2>&1)
+missing_docker_status=$?
+compose_output=$(HOME=/isolated-home DOCKER_CONFIG=/run/docker-config PATH="$stub_dir:$PATH" \
+  E2E_STUB_DOCKER_MODE=missing-compose assert_docker_runtime_is_available 2>&1)
+compose_status=$?
+daemon_output=$(PATH="$stub_dir:$PATH" E2E_STUB_DOCKER_MODE=daemon-down \
+  assert_docker_runtime_is_available 2>&1)
+daemon_status=$?
+set -e
+
```
hojinzs, line 70: **nit — the runner list is hardcoded, so the invariant it guards won't extend to the next runner.**

This loop is the right idea and it does catch a regression in the four runners that exist today — I verified it fails correctly if the preflight is removed or moved below the trap. The gap is that `AGENT_TEST.md` now promises the invariant for *all* Docker E2E runners, while the test only knows about four paths. A fifth runner added later inherits the documented guarantee without inheriting the check.

Discovery is cheap here because the population is exactly "scripts that source this lib", and I confirmed that set is closed — `grep -rln "docker compose" --include=*.sh .` returns the same four runners plus this test and the lib itself, nothing else:

```bash
while IFS= read -r runner; do
  case "$runner" in */compose-project.test.sh) continue ;; esac
  ...
done < <(grep -rl 'compose-project\.sh' --include='*.sh' "$root_dir/e2e" "$root_dir/test")
```

Non-blocking, and genuinely optional — the hardcoded list is honest about what it covers, and a self-enumerating loop trades one failure mode (new runner silently unguarded) for another (a discovery grep that quietly matches nothing). Take it or leave it; I'm not holding the PR on it.

---

hojinzs (approved):

Verified at head `9865b78`. Every round-1 item is resolved or declined with a reason I accept. Approving.

### Round-1 disposition

| Item | Disposition | Verified |
|---|---|---|
| **P2** — TC-13 / TC-22 still emitted the opaque failure | Fixed | ✅ re-tested below |
| **P3** — unbounded daemon probe | Declined, with reason | ✅ **your reason is right and mine was wrong — see below** |
| nit — name the `cli-plugins` path | Fixed | ✅ literal `$HOME/.docker/cli-plugins/` present, asserted by the test |
| nit — `compose_error` alias | Fixed | ✅ off the `local` line, `${compose_version:-…}` used directly |
| nit — unconditional success line | Fixed | ✅ healthy preflight is silent, test asserts empty output |

### Correction on my P3

I wrote that `timeout` is "already assumed elsewhere in the harness." That was wrong. The only matches are `docker compose down --timeout 5` flags in `run-e2e.sh:174` and `run-flat-tracker-keys-e2e.sh:10` — the coreutils `timeout` binary is invoked nowhere in this repo. So your macOS portability objection stands unopposed: adding it would introduce a new host dependency, not lean on an existing one. Declining was the correct call, and the rollback note in the PR body is the right place to have recorded it.

### P2 fix — smoke tested against a real failure and a baseline

Same broken host as round 1. The daemon-unreachable path is **real**, not stubbed; the Compose-unresolvable path uses a stub `docker` under `HOME=/nonexistent` reproducing #817's exact symptom.

Daemon unreachable — all four runners:

| Runner | Result |
|---|---|
| `e2e/run-e2e.sh` | exit **69**, real `dial unix /var/run/docker.sock` probe text, 0 cleanup lines |
| `e2e/run-flat-tracker-keys-e2e.sh` | exit **69**, identical, 0 cleanup lines |
| `e2e/run-standalone-project-e2e.sh` | exit **69**, identical, 0 cleanup lines |
| `test/e2e/claude/run-docker-e2e.sh` | exit **69**, identical, 0 cleanup lines |

Compose unresolvable — the before/after that matters, on the two runners this round added:

```
main   $ pnpm e2e:standalone-project        →  docker: 'compose' is not a docker command.
                                               See 'docker --help'          exit=125
9865b78$ pnpm e2e:standalone-project        →  [e2e] A user-level Compose plugin under
                                               $HOME/.docker/cli-plugins/ may be hidden by the
                                               current environment (HOME=/nonexistent,
                                               DOCKER_CONFIG=<unset>).       exit=69
```

`run-flat-tracker-keys-e2e.sh` shows the identical 125 → 69 transition. That is the misdiagnosis trap from #817 closed on both remaining doors.

Also verified independently of the test's own assertions:

- All four runners are `set -euo pipefail`, so `return 69` propagates as the process status rather than being swallowed.
- Preflight precedes `trap cleanup EXIT` in all four (`185<190`, `14<15`, `16<18`, `19<24`) — hence 0 cleanup lines above, and `e2e/fixtures/issues.json` is never clobbered on an unavailable host.
- The runner population is closed: `grep -rl 'docker compose' --include='*.sh'` returns exactly these four plus the lib and its test. Coverage is complete, not merely four-of-many.
- `bash e2e/lib/compose-project.test.sh` passes; `bash -n` clean on all six touched scripts.
- CI's `Test` job runs `pnpm e2e:compose-project-test` — the one job this diff can actually influence — and it passes locally. The diff is shell + docs only (`git diff --name-only` is 5 × `.sh`, 1 × `.md`, zero TypeScript), so the vitest/lint/typecheck/build portions are unaffected by construction. `Container Smoke` is green on this head; `Test` was still running at review time and I don't expect it to say anything new, but confirm before merge.

### Scope

Still no over-engineering and no unrequested decisions. Round 2 added 12 lines: two call sites, one doc-wording change, and an ordering assertion. The PR body's "Changed files" and validation sections now match the diff exactly, including the honest note that the daemon probe stays unbounded and why.

One non-blocking nit inline on the ordering loop. Take it or leave it — this is ready to merge.

---

hojinzs:
Closing the one open item from the approval above: **`Test` has since completed green** on `9865b78` (22:33:36), alongside `Container Smoke`. Both checks pass and the PR is mergeable-clean, so nothing is outstanding before merge.

---

Issue #841, opened 2026-09-03
fix(e2e): detect unavailable Docker Compose before Claude suite


moncher-dev:
## Problem

`pnpm e2e:claude` invokes `docker compose --project-name ...`, but on a host where the Docker CLI has no Compose plugin the command is parsed by the top-level `docker` CLI and fails immediately with:

```
unknown flag: --project-name
```

The suite therefore provides no explicit prerequisite diagnostic and launches no containers or tests.

## Reproduction

1. Use a Docker CLI installation without the Compose v2 plugin.
2. Run `pnpm e2e:claude`.
3. Observe exit code 125 before the black-box suite starts.

Observed twice while validating #813 on 2026-09-04.

## Expected

`test/e2e/claude/run-docker-e2e.sh` should preflight `docker compose version` and report a clear Docker Compose v2 prerequisite (or support an intentional compatible fallback).

## Scope

This is an environment/tooling follow-up. The focused built-worker black-box test added for #813 passes outside Docker and confirms the credential preflight behavior.

moncher-dev:
Already implemented — closing as covered by **PR #840** (which closes #817), not as a stale duplicate.

Your ask is that `test/e2e/claude/run-docker-e2e.sh` preflight `docker compose version` and report a clear Compose v2 prerequisite. #840 does exactly that, and it does it for both entrypoints through a shared `assert_docker_runtime_is_available` rather than patching each script separately. `test/e2e/claude/run-docker-e2e.sh` is in its diff.

I verified the Claude entrypoint specifically, on this host, with the real failure:

```
$ HOME=/nonexistent bash test/e2e/claude/run-docker-e2e.sh
[e2e] Docker runtime unavailable: 'docker compose' cannot be resolved by this process.
[e2e] A user-level Compose plugin may be hidden by the current environment
      (HOME=/nonexistent, DOCKER_CONFIG=<unset>).
[e2e] Compose probe: docker: unknown command: docker compose
exit=69
```

So the `unknown flag: --project-name` fallthrough you reported no longer happens — the preflight fires before Compose arguments ever reach the top-level `docker` CLI, which is what produced that confusing message. Exit 69 also distinguishes an unavailable prerequisite from Docker's 125 and from a real scenario failure.

[...]

Issue #812, opened 2026-09-03
fix(runtimes): forward SYMPHONY_ASSIGNED_BRANCH and agent-visible Symphony context into the agent child


moncher-dev:
Part of #808 (root-cause chain item 4).

## Problem

`buildCodexRuntimePlan` (`packages/runtime-codex/src/runtime.ts`) forwards only `SYMPHONY_ORCHESTRATOR_URL`, `SYMPHONY_RUN_ID` and `SYMPHONY_ORCHESTRATOR_TOKEN` (plus `SAFE_RUNTIME_ENV_KEYS`) into the agent child. WORKFLOW.md rules 3–4 and the `push` / `pull` skills require `$SYMPHONY_ASSIGNED_BRANCH` in the agent shell (`git branch --show-current` must equal it; the host push is described in terms of it). The agent observed `assigned=` (empty) and treated it as an orchestration blocker.

## Fix outline

1. Define one documented **agent-visible Symphony context allowlist** in core (e.g. `SYMPHONY_ASSIGNED_BRANCH`, `SYMPHONY_ISSUE_IDENTIFIER`, `SYMPHONY_ISSUE_ID`, `SYMPHONY_ISSUE_STATE`, `SYMPHONY_TRACKER_KIND`, `TARGET_REPOSITORY_*`) — non-secret values only — and apply it in `runtime-codex`, `runtime-claude` and `worker/non-codex-runtime.ts` so every runtime exposes the same contract.
2. Keep every name from `secretEnvironmentNames()` and `CUSTOM_RUNTIME_RESERVED_AUTH_ENVIRONMENT_NAMES` excluded.
3. Update `docs/configuration.md` (agent-visible context table) and the WORKFLOW.md runtime contract if the variable set changes.

## Regression tests

- Codex/Claude/custom child env contains `SYMPHONY_ASSIGNED_BRANCH` with the worker's value.
- Tracker secrets and broker variables remain absent from the child env.

moncher-dev:
## Workpad — hojinzs/github-symphony#812 — Cycle 1

**Type:** fresh pickup
**Branch:** symphony/hojinzs-github-symphony-812
**Draft PR:** not yet created
**Cycle opened:** 2026-09-04T01:20:38Z · **Trigger:** Ready issue is actionable with no prior workpad or linked PR

### Status Transitions

- 2026-09-04T01:20:38Z · `Ready` → `In progress` · actionable runtime-context fix is ready for implementation (cycle 1 open)

[...]

PR #842, opened 2026-09-03
feat(worker,cli): fail fast without tracker credentials

### LINKED ISSUE #813: feat(worker,cli): fail fast when the worker has no GitHub credential; doctor checks the worker-effective environment
Part of #808 (root-cause chain item 6).

## Problem

Nothing verifies that a worker can reach GitHub before the agent starts. In the #808 incident each run spent a full agent turn (~186k input tokens) discovering `github_graphql` was unauthenticated and then parked the issue to `Backlog` through the tracker API. `doctor` validates the daemon's auth (`packages/cli/src/commands/doctor.ts`) but not the environment the worker will actually receive (`resolveManagedProjectEnvironment` + allowlist).

## Fix outline

1. Worker preflight (`packages/worker/src/index.ts`, alongside the Claude preflight): before launching the runtime, require either `GITHUB_GRAPHQL_TOKEN` or a complete broker configuration for GitHub-tracked issues (and the Linear equivalent); exit with `exitWorkerStartupFailure` and a clear remediation message otherwise. The orchestrator surfaces it as a startup failure (no agent tokens spent, issue stays `Ready`).
2. `doctor`: evaluate the **worker-effective** environment for the tenant (project `.env` + injected credential per #808 item 1) and report a check "worker GitHub credential" with the remediation (`project .env` path or daemon auth).
3. Status snapshot warning when a dispatch is skipped for a missing worker credential.

## Regression tests

- Worker started without any GitHub credential → exits before runtime launch with the documented message; run recorded as startup failure.
- `doctor` fails the new check in repo mode without a project `.env` on a pre-fix daemon, passes after the credential injection.

moncher-dev:
## Issues

- Closes #813

## Summary

- Fail GitHub/Linear workers before runtime launch when their effective tracker credential is missing.
- Gate initial dispatch and due retries without consuming retry attempts, and expose actionable status warnings.
- Make `doctor` validate the worker-effective GitHub environment; adapter-contract tests now include the `GITHUB_TOKEN` drift sentinel requested in review.

## Change-point diagram

- Project/daemon environment → adapter credential resolution → doctor and dispatch gates
- Due retry → credential gate → unchanged retry queue or authenticated restart
- Worker entrypoint → tracker preflight → runtime launch only with an accepted credential shape

## Start here

- `packages/orchestrator/src/service.ts` — initial and retry credential gates
- `packages/worker/src/tracker-credential-preflight.ts` — worker startup validation
- `packages/worker/src/tracker-credential-preflight.test.ts` — real-adapter parity and future `GITHUB_TOKEN` drift coverage
- `packages/cli/src/commands/doctor.ts` — worker-effective credential diagnostic

- Missing credentials stop workers before agent-token usage, leave retries queued without advancing suppression state, and surface remediation in status snapshots.
- `doctor` checks the managed environment the worker will actually receive.

## Validation

- `pnpm exec vitest run packages/worker/src/tracker-credential-preflight.test.ts` — pass (14 tests)
- `pnpm lint` — pass
- `pnpm test` — pass
- `pnpm typecheck` — pass
- `pnpm build` — pass
- Targeted built-worker Docker scenario from Cycle 2 — pass; full Docker Compose host limitation tracked in #841

## Changeset

- `.changeset/early-workers-check-github.md` (`patch`, `@gh-symphony/cli`)

## Risks & rollback

- Risk is limited to credential-shape drift; the worker matrix compares against the real GitHub and Linear adapters, including currently unsupported `GITHUB_TOKEN`.
- Revert the issue commits to restore prior startup and dispatch behavior.

## Changed files

- `packages/worker/src/*` — startup preflight and adapter-contract coverage
- `packages/cli/src/*` — worker-effective doctor diagnostic
- `packages/orchestrator/src/*` — dispatch/retry credential gates and warnings
- `test/e2e/claude/claude-docker.spec.ts`, `e2e/scenarios/20-agent-child-isolation.md` — black-box coverage
- `README.md`, `packages/cli/README.md`, `docs/*`, `AGENT_TEST.md` — living documentation
- `.changeset/early-workers-check-github.md` — patch release note

- [ ] Run the full Docker E2E suite on a host with Docker Compose v2; tooling limitation tracked in #841.

## Security


moncher-dev:
Smoke-tested at `5b962809`. `pnpm install --frozen-lockfile`, `pnpm build`, `pnpm typecheck`, `pnpm lint` all exit 0; suites green across the three touched packages — worker 233, orchestrator 397, cli 620 (**1250 tests**). CI green on the head.

Since this changes startup failure behavior, I drove `resolveTrackerCredentialPreflight` from the built `dist` across the paths that matter, including what an operator actually sees:

| Case | Result |
|---|---|
| github, no credential | **BLOCK** — *"Worker GitHub credential preflight failed: GITHUB_GRAPHQL_TOKEN or both GITHUB_TOKEN_BROKER_URL and GITHUB_TOKEN_BROKER_SECRET are required. Add the credential to `/tenant/repo/.env`, or authenticate the daemon environment and restart it."* |
| github, direct token | ok |
| github, **partial** broker (url only) | **BLOCK** — no partial credential accepted |
| github, complete broker | ok |
| github, whitespace-only token | **BLOCK** — not treated as present |
| linear, no credential | **BLOCK**, names `LINEAR_AUTHORIZATION` / `LINEAR_API_KEY` |
| linear, api key | ok |
| **file tracker** | ok — correctly not gated |
| adapter unset | ok — correctly not gated |

The operator-facing message is genuinely actionable: it names the missing variables, and when `SYMPHONY_PROJECT_DIR` is set it interpolates the real path rather than saying "the managed project .env". Not gating the file tracker matters — that is what the local E2E harness uses, and gating it would have broken the suite.

**The Codex P1 is real and I reproduced the code path, so please don't let it be resolved as noise.** The credential gate is inside the dispatch loop only:

```
service.ts:1689   if (workerCredentialMissing) { … skipped += 1; continue; }
```

`restartRun` (`service.ts:4988`) has no equivalent check, and per its own comment it is reached from active-run reconciliation — which runs *before* the dispatch loop on each tick. So when credentials are removed while a run is retry-queued, or a persisted retry is loaded after an upgrade, the due retry restarts a credential-free worker; the new preflight fails it; and repeated attempts consume the failure-retry budget until the issue is **permanently suppressed** rather than left queued with the "dispatch skipped" advertisement.

That trade is worse than the bug being fixed. A missing credential is a transient operator condition that should resolve the moment the credential is added — but suppression does not clear on its own. On this board, re-arming a suppressed issue requires a manual status flip to a *different* active state (a Backlog round-trip does not re-arm), so an operator whose token lapsed briefly would find issues silently parked with no obvious route back. Applying the same gate before `restartRun` dispatches a due retry, so it stays queued instead of burning attempts, looks like the right shape.

Everything else here is in good order and scoped to what #813 asked for. @hojinzs over to you — the P1 needs either that fix or a reasoned rebuttal on the thread.

packages/orchestrator/src/service.ts
```diff
@@ -1620,6 +1621,14 @@ export class OrchestratorService {
...
       const listRateLimits = getTrackedIssueListRateLimits(issues);
       let listRateLimitsRecorded = false;
+      const workerCredentials =
+        trackerAdapter.resolveWorkerCredentials?.(tenant, {
+          project: this.readProjectEnv(tenant),
+          daemon: process.env,
+        }) ?? null;
+      const workerCredentialMissing =
+        workerCredentials !== null &&
+        Object.keys(workerCredentials).length === 0;
```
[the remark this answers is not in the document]
hojinzs: Confirmed independently at `5b96280` — this is blocking, not noise.

Traced the ordering in `runOnce`:

- `reconcileRun()` is called at `service.ts:1313`, i.e. **before** the dispatch loop.
- The retry branch at `service.ts:3670` reaches `restartRun()` (`service.ts:3683`) → `startRun()`.
- `startRun()` at `service.ts:3192` detects `Object.keys(workerCredentials).length === 0`, appends the `worker-credential-missing` event, writes a `[warn]` line — and then falls straight through to `mkdirSync(runDir)` and the spawn. There is no abort.
- The only gate is `service.ts:1689`, inside the dispatch loop, so it never covers this path.

Consequence chain also checks out: the restarted worker now hits the new preflight and fails startup, each attempt advancing `failureRetryCount`, and `isFailureRetrySuppressedIssue` (`service.ts:6018`) suppresses the issue once `failureRetryCount >= maxFailureRetries` unless `isFailureRetryRearmedForState` sees a state change. A missing credential is a transient operator condition, so converting it into a suppression that only clears on a tracker state flip is a worse outcome than the pre-fix behavior for exactly the retry-queued population.

Note also that the `[warn] No worker credential resolved for …` message at `service.ts:3209` is now reachable *only* via this restart path — which is a good signal that the gate is missing there.

Suggested shape: resolve the credentials once and gate before `restartRun()` dispatches a due retry, requeueing with `{ countFailure: false, advanceAttempt: false }` (the existing "no available orchestrator slots" pattern at `service.ts:3670`) so the retry stays queued rather than burning attempts, and surface the same status warning.

---

packages/orchestrator/src/service.ts
```diff
@@ -1677,6 +1686,14 @@ export class OrchestratorService {
           }
         }
 
+        if (workerCredentialMissing) {
+          const warning = `Dispatch skipped for ${issue.identifier}: no worker credential resolved for ${issue.tracker.adapter}. Add the credential to the managed project .env or authenticate the daemon and restart it.`;
+          dispatchWarnings.push(warning);
+          this.writeStderr(`[orchestrator] ${warning}\n`);
+          skipped += 1;
+          continue;
```
hojinzs, line 1709: **P2** — this warning is emitted per candidate issue, per tick, forever.

`workerCredentialMissing` is a tenant-level condition that does not clear on its own, and a skipped issue stays a candidate on the next tick. So with N eligible issues and the default poll interval this pushes N identical-shaped warnings into `snapshot.warnings` and writes N stderr lines **every tick** — e.g. 20 issues at a 15s poll is ~4,800 stderr lines/hour, all saying the same thing, for as long as the credential is missing. That buries the one line an operator actually needs to see, and it is the same class of noise the file already guards against elsewhere (`warnedProjectEnvPermissions`, `service.ts:4937`, dedupes a repeating env warning by remembering what it already reported).

Suggestion: hoist the tenant-level part out of the loop — one warning per tick naming the tenant and adapter, optionally listing the affected identifiers — and dedupe the stderr write with the same `Map` pattern used for the `.env` permission warning so it prints on transition rather than on every tick. The snapshot warning can stay per-tick since the snapshot is replaced each time.

---

packages/worker/src/tracker-credential-preflight.ts
```diff
@@ -0,0 +1,59 @@
...
+      return { ok: true };
+    }
+
+    return {
+      ok: false,
+      reason:
+        "Worker GitHub credential preflight failed: GITHUB_GRAPHQL_TOKEN or both GITHUB_TOKEN_BROKER_URL and GITHUB_TOKEN_BROKER_SECRET are required. " +
+        remediation(env),
+    };
+  }
```
hojinzs, line 56: **P2** — the credential contract is now duplicated with no shared source of truth, and drift here fails closed.

I verified the rules match the adapters exactly today. Driving the built `dist` module across the matrix, every case agrees with `resolveWorkerCredentials`:

| env | preflight | adapter |
|---|---|---|
| github, nothing | BLOCK | `{}` |
| github, `GITHUB_GRAPHQL_TOKEN` | ok | selects token |
| github, broker url only / secret only | BLOCK | `{}` |
| github, complete broker pair | ok | selects pair |
| github, whitespace-only token | BLOCK | `{}` (adapter also `.trim()`s) |
| linear, either var | ok | selects it |
| `file`, adapter unset | ok (not gated) | no `resolveWorkerCredentials` |

The problem is that this agreement is a coincidence of two hand-maintained lists. `packages/tracker-github/src/orchestrator-adapter.ts:56` and `packages/tracker-linear/src/orchestrator-adapter.ts:262` own credential selection; this file re-encodes the same OR/AND semantics from literal env names. If a tracker adapter later accepts an additional credential shape — `GITHUB_TOKEN` is already in `secretEnvironmentNames()` at `orchestrator-adapter.ts:44` and is the obvious next candidate — the orchestrator injects it, this preflight does not recognise it, and **every worker for that tenant fails startup**. A guard whose drift mode is "block all runs" needs to be pinned to the thing it mirrors.

This also cuts against the layering rule in `CLAUDE.md` ("Keep tracker-specific behavior out of core layers"; tracker-specific code lives in Integration): the worker (Execution) now hardcodes GitHub and Linear variable names.

Two options, either is fine:

1. Move the requirement into the adapter contract — e.g. a `describeWorkerCredentialRequirement()` alongside `resolveWorkerCredentials`, resolved through the tracker adapter the worker already knows from `SYMPHONY_TRACKER_ADAPTER`.
2. Keep this file, but add a test that imports both adapters and asserts, over a shared fixture matrix, that `resolveTrackerCredentialPreflight(env).ok === (Object.keys(adapter.resolveWorkerCredentials(project, { project: {}, daemon: env })).length > 0)`. That turns future drift into a red test instead of a production outage.

Related nit on the same block: an unrecognised or absent `SYMPHONY_TRACKER_ADAPTER` falls through to `{ ok: true }` at line 58. That is right for the `file` tracker and for the E2E harness, but it means a typo in the adapter name silently disables the guard. Worth a one-line comment saying the fall-through is deliberately permissive.

---

packages/orchestrator/src/service.ts
```diff
@@ -1677,6 +1686,14 @@ export class OrchestratorService {
           }
         }
 
+        if (workerCredentialMissing) {
+          const warning = `Dispatch skipped for ${issue.identifier}: no worker credential resolved for ${issue.tracker.adapter}. Add the credential to the managed project .env or authenticate the daemon and restart it.`;
+          dispatchWarnings.push(warning);
+          this.writeStderr(`[orchestrator] ${warning}\n`);
+          skipped += 1;
```
hojinzs, line 1708: **P3** — `skipped` now carries two different meanings in one counter.

It is seeded from tracker-level `skippedItems.length` (`service.ts:1367`, items the tracker could not map to a dispatchable issue) and is now also incremented for dispatch-time credential skips. The snapshot's `summary.skipped` therefore mixes "the tracker returned items we ignored" with "we had work ready and declined to start it", which are very different operator signals — the second one means throughput is zero.

Either a separate counter (`skippedForCredential`) or a comment at this increment explaining the overload would keep the status snapshot readable.

---

packages/cli/src/managed-project-environment.ts
```diff
@@ -26,3 +26,27 @@ export function resolveManagedProjectEnvironment(
...
 }
+
+export function resolveManagedProjectDirectory(
+  projectConfig: Pick<OrchestratorProjectConfig, "projectDir" | "projectId">,
+  runtimeRoot: string
+): string {
+  return (
+    projectConfig.projectDir ??
+    join(runtimeRoot, "projects", encodeURIComponent(projectConfig.projectId))
+  );
```
hojinzs, line 38: **nit** — `resolveManagedProjectEnvironment` above still inlines the same `projectDir ?? join(runtimeRoot, "projects", encodeURIComponent(projectId))` derivation (lines 12–14). Now that this helper exists, having it call `resolveManagedProjectDirectory` keeps the layout defined once; today the same path rule is written twice in one file.

---

hojinzs (changes requested):
## Review — round 1/6

Reviewed at `5b96280` against #813. The shape of this change is right and the scope is honest; one confirmed P1 blocks it.

### Requirement coverage vs. #813

| #813 item | Status |
|---|---|
| 1. Worker preflight before runtime launch, GitHub + Linear, `exitWorkerStartupFailure` with remediation | ✅ `packages/worker/src/index.ts:543`, placed after the identity preflight and before workflow parse / runtime routing |
| 2. `doctor` evaluates the **worker-effective** environment (project `.env` + injected credential) and reports remediation | ✅ `buildWorkerGitHubCredentialCheck`; it reproduces the daemon's injection faithfully — `start.ts:134` sets `process.env.GITHUB_GRAPHQL_TOKEN = auth.token`, and the check deletes the ambient value and re-adds it only from validated `auth`, so it cannot pass on a token the daemon would reject |
| 3. Status snapshot warning when dispatch is skipped for a missing credential | ⚠️ Implemented, but only on the dispatch path — see P1 |
| Regression test: credential-free worker exits before runtime launch | ✅ `test/e2e/claude/claude-docker.spec.ts:613` |
| Regression test: `doctor` fails pre-fix, passes after injection | ✅ all three cases in `doctor.test.ts` |

No over-engineering found. The Linear preflight is explicitly requested by #813 ("and the Linear equivalent"); the doctor check is correctly registered for `github-project` only, as the issue scoped it. `resolveManagedProjectDirectory` / `resolveManagedProjectFileEnvironment` are the minimum needed to read the project `.env` without daemon fallbacks.

### Smoke test

Local, `5b96280`, `pnpm build` clean; suites green across the touched packages — **1250 tests, 85 files** (worker + orchestrator + cli), matching the author's numbers. CI green on the head.

Black-box, running the built `packages/worker/dist/index.js` with a real git workspace and no GitHub credential:

```
[worker] Worker GitHub credential preflight failed: GITHUB_GRAPHQL_TOKEN or both
GITHUB_TOKEN_BROKER_URL and GITHUB_TOKEN_BROKER_SECRET are required. Add the
credential to <SYMPHONY_PROJECT_DIR>/.env, or authenticate the daemon
environment and restart it.
```

exit `1`, terminal heartbeat `runPhase: failed` carrying that `lastError`, and **no** Codex initialization in the output — so the token spend #813 was written to stop is genuinely stopped. Re-running the identical fixture with `GITHUB_GRAPHQL_TOKEN` set proceeds past the preflight into Codex startup, so there is no false positive on the happy path. The remediation string interpolates the real project path when `SYMPHONY_PROJECT_DIR` is set, which is the difference between an actionable message and a generic one.

I also drove the built preflight across a 15-case matrix (partial broker pairs, whitespace-only token, `GITHUB_TOKEN`-only, mixed-case adapter, unset adapter, `file` tracker); results and the one concern they raise are in the inline comment on `tracker-credential-preflight.ts`.

### Findings

- **P1 — retry restarts bypass the credential gate.** Codex found it, the author reproduced it, and I confirmed it independently by tracing `runOnce` — details and a suggested fix shape are in [that thread](https://github.com/hojinzs/github-symphony/pull/842#discussion_r3929000104). `reconcileRun` (`service.ts:1313`) runs before the dispatch loop, `restartRun` → `startRun` logs the missing credential at `service.ts:3192` and then spawns anyway, and each failed attempt advances `failureRetryCount` until `isFailureRetrySuppressedIssue` parks the issue until a tracker state change. Trading "one wasted agent turn" for "silently suppressed until a manual state flip" is a regression for the retry-queued population, and it is the same population most likely to be mid-flight when a token lapses. **This is the one thing blocking approval.**
- **P2** — the skip warning is emitted per issue, per tick, indefinitely (`service.ts:1689`).
- **P2** — the preflight re-encodes the adapters' credential rules with no shared source of truth; drift fails closed and blocks every worker (`tracker-credential-preflight.ts`).
- **P3** — `summary.skipped` now overloads tracker-skipped items with credential skips.
- **nit** — duplicated managed-project path derivation in `managed-project-environment.ts`.

Only the P1 gates approval. The two P2s are cheap and I would rather see them land with this change than as follow-ups, but I will not hold the PR on them if you disagree with a reason on the thread. P3 and the nit are yours to take or drop.

---

packages/orchestrator/src/service.ts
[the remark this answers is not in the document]
moncher-dev: Fixed in `1c0f54d`. Due retries now resolve the adapter-owned worker credential before tracker refresh or `restartRun`; a missing credential requeues with `countFailure: false` and `advanceAttempt: false`, surfaces the dispatch warning, and does not spawn. The regression test verifies the retry stays queued and repeated ticks do not increase `failureRetryCount` or the attempt.

packages/orchestrator/src/service.ts
moncher-dev: Fixed in `1c0f54d`. Missing-credential stderr is deduplicated per tenant and adapter until credentials recover, while the replace-on-tick snapshot retains the issue-specific warning. The dispatch regression test runs consecutive ticks and verifies one stderr write with warnings present in both snapshots.

packages/worker/src/tracker-credential-preflight.ts
moncher-dev: Addressed in `cf303a2` using the proposed contract-test option. The worker test now imports the real GitHub and Linear adapters and checks a seven-case credential matrix against `resolveWorkerCredentials`; the permissive fallback is also documented as intentional so adapter validation remains orchestrator-owned.

packages/orchestrator/src/service.ts
moncher-dev: Addressed in `cf303a2` with the suggested compatibility-preserving comment. `summary.skipped` continues to aggregate provider-unmapped items and credential-gated actionable issues, avoiding a status-contract expansion in this patch; the adjacent warning identifies the operator-actionable credential subset.

packages/cli/src/managed-project-environment.ts
moncher-dev: Fixed in `cf303a2`. `resolveManagedProjectEnvironment` now calls `resolveManagedProjectDirectory`, leaving the managed layout derivation in one helper.

packages/worker/src/tracker-credential-preflight.test.ts
```diff
@@ -0,0 +1,125 @@
...
+    ],
+    [
+      "github-project",
+      githubProjectTrackerAdapter,
+      {
+        GITHUB_TOKEN_BROKER_URL: "https://broker.example/token",
+        GITHUB_TOKEN_BROKER_SECRET: "secret",
+      },
+    ],
+    ["linear", linearTrackerAdapter, {}],
```
hojinzs, line 106: **nit** (non-blocking, follow-up is fine) — the contract test is the right mechanism and it does pin the shapes that exist today; one row short of covering the drift case that motivated it.

I verified the current state: `GITHUB_TOKEN` is in `secretEnvironmentNames()` (`packages/tracker-github/src/orchestrator-adapter.ts:48`) but `resolveWorkerCredentials` (line 56) only reads `GITHUB_GRAPHQL_TOKEN` and the broker pair, so preflight and adapter agree today — no live bug.

The gap is that this matrix asserts agreement only over environments both sides already recognise. If the adapter later adds `GITHUB_TOKEN` to `resolveWorkerCredentials`, the orchestrator starts injecting it, the preflight still blocks, every worker for that tenant fails startup — and all seven rows here stay green, because none of them sets `GITHUB_TOKEN`. One more row closes it:

```ts
["github-project", githubProjectTrackerAdapter, { GITHUB_TOKEN: "token" }],
```

Today that asserts `false === false`; the day the adapter starts accepting it, the test goes red instead of production going down. Same argument for `GH_TOKEN` if you want the pair.

---

hojinzs (approved):

Re-reviewed at `cf303a2` (`1c0f54d`, `cf303a2` added since round 1). All five round-1 findings are addressed; the blocking P1 is genuinely fixed. Approving.

### Round-1 findings

| Finding | Status |
|---|---|
| **P1** retry restarts bypass the credential gate | ✅ Fixed in `1c0f54d`, exactly at the right point |
| **P2** skip warning repeated per issue per tick | ✅ Fixed in `1c0f54d` |
| **P2** preflight re-encodes adapter credential rules | ✅ Addressed in `cf303a2` (contract test) — one row short, see nit |
| **P3** `summary.skipped` overload | ✅ Addressed in `cf303a2` with a reasoned decline + comment; accepted |
| **nit** duplicated managed-project path derivation | ✅ Fixed in `cf303a2` |

**P1** — I traced the fix rather than trusting the description. The gate now sits in `reconcileRun` at `service.ts:3647`, inside the `status === "retrying" && nextRetryAt` branch and *after* the not-yet-due early return, so it costs nothing for retries that aren't due. Critically it is placed **before** `resolveRetryRestartAction` (the tracker refresh) and before `restartRun`, which is what the round-1 comment asked for — no tracker call and no spawn on a credential-free retry. `requeueRetryingRun(..., { countFailure: false, advanceAttempt: false })` resolves correctly through the existing implementation: `attempt` and `failureRetryCount` are unchanged (`service.ts:5378-5383`), `suppressed` is forced false because it is gated on `countFailure !== false` (line 5389), and `retainedDueAt` keeps the already-queued `dueAt` rather than pushing it out (line 5391). So the issue stays queued and re-checked every tick, and the suppression-until-manual-state-flip regression is gone.

**P2 (warning noise)** — `warnedMissingWorkerCredentials` follows the `warnedProjectEnvPermissions` precedent, keyed `projectId:adapter`, and is cleared on recovery in *both* paths (`service.ts:1635` and `3671`), so a credential that comes back re-arms the warning. Snapshot warnings stay per-tick, which is right for a replace-on-tick surface.

**P3** — declining the separate counter to avoid expanding the status contract in a patch release is a reasonable call with a reason on the thread, and the comment at the increment site says what `skipped` now aggregates. Taking it.

### Requirement coverage vs. #813

Unchanged from round 1 and still complete: worker preflight before runtime launch for GitHub + Linear (item 1), worker-effective `doctor` check with remediation (item 2), status-snapshot warning on skipped dispatch (item 3) — now covering *both* the initial-dispatch and due-retry paths, which is a stricter reading of item 3 than the issue text required and is the correct one. No over-engineering: the two new commits add only what the review asked for.

### Smoke test

Local at `cf303a2`, clean install:

- `pnpm build`, `pnpm typecheck`, `pnpm lint` — all clean.
- Touched packages green: **worker 240/240 (28 files)**, **orchestrator 398/398 (14 files)**, **cli 620/620 (43 files)** — 1258 tests, 85 files.
- `vitest run --config test/e2e/claude/vitest.config.ts -t "startup"` — the built-worker credential-free startup regression passes against the real `dist`.
- CI green on the head: **Test** and **Container Smoke** both success.

I checked the new regression tests actually pin the behavior rather than just executing it. The retry test asserts the full set that matters — no spawn, run still `retrying`, `failureRetryCount` unchanged, and `retryEntry.attempt`/`dueAt` identical to the pre-gate values — then runs a second tick and re-asserts the attempt and failure count, which is precisely the suppression path from round 1. The dedup test runs consecutive ticks and asserts `stderrWrite` called exactly once while both snapshots still carry the warning, so it would catch a regression in either direction.

I also re-verified the drift risk empirically instead of assuming: `GITHUB_TOKEN` appears in `secretEnvironmentNames()` but not in `resolveWorkerCredentials`, so preflight and adapter agree at this commit and there is no live bug — only the uncovered future case in the nit.

### Remaining

One **nit** on the contract-test matrix, inline. Non-blocking and fine as a follow-up or not at all.

Approving — this is done as far as I'm concerned. The unchecked post-merge box (full Docker E2E on a Compose v2 host, #841) stays yours to run.

---

packages/worker/src/tracker-credential-preflight.test.ts
moncher-dev: Added in `0f74fde`. The real-adapter contract matrix now includes `{ GITHUB_TOKEN: "token" }`: both sides reject it today, and a future adapter-only acceptance change will make this drift sentinel fail.

hojinzs: `0f74fde` is not on this branch — the fix looks correct but was never pushed.

The PR head is still `cf303a2` (`git fetch … refs/pull/842/head` → `cf303a2`; `git cat-file -t 0f74fde` → `fatal: Not a valid object name`). The commit list on the PR is the same six it had at my round-2 approval.

At the actual head, the matrix in this file is still the original seven rows and none of them sets `GITHUB_TOKEN`:

```
["github-project", githubProjectTrackerAdapter, {}],
["github-project", …, { GITHUB_GRAPHQL_TOKEN: "token" }],
["github-project", …, { GITHUB_TOKEN_BROKER_URL: … }],
["github-project", …, { GITHUB_TOKEN_BROKER_URL: …, GITHUB_TOKEN_BROKER_SECRET: … }],
["linear", linearTrackerAdapter, {}],
["linear", …, { LINEAR_API_KEY: "key" }],
["linear", …, { LINEAR_AUTHORIZATION: "Bearer token" }],
```

The test count corroborates it. This file runs **13 tests** at `cf303a2`. Adding the one sentinel row takes it to **14** — which is the number the PR description reports under Validation, so the run quoted there was against the unpushed tree, not this head. I applied the suggested row locally to check the fix itself is sound: 14 passed, `false === false` today, exactly as intended. Nothing wrong with the change — it just isn't here.

`git push` should be all this needs. Non-blocking either way; my approval stands on `cf303a2` as it is.

---

hojinzs (approved):
## Review — round 3/6 · Approve (re-affirmed)

Re-review requested at `cf303a2`. **No code has changed since my round-2 approval** — the head is the same `cf303a2`, the same six commits, CI still green (`Test`, `Container Smoke`). So the approval stands unchanged and this PR is mergeable as it is.

One thing to flag before you merge, because it will otherwise be lost:

### The last nit's fix was never pushed

The reply on [that thread](https://github.com/hojinzs/github-symphony/pull/842#discussion_r3929405923) says the `GITHUB_TOKEN` drift sentinel was "Added in `0f74fde`". That commit is not on this branch:

- `git fetch https://github.com/hojinzs/github-symphony refs/pull/842/head` → `cf303a2`
- `git cat-file -t 0f74fde` → `fatal: Not a valid object name`
- The contract matrix in `packages/worker/src/tracker-credential-preflight.test.ts` still has its original seven rows, none setting `GITHUB_TOKEN`.

The test count settles it: that file runs **13 tests** at this head. With the sentinel row it runs **14** — and 14 is the number the PR description quotes under Validation. So the description's claim that "adapter-contract tests now include the `GITHUB_TOKEN` drift sentinel" is true of your working tree and not of this branch. Worth correcting the description too, since it is the merge-commit message.

I checked the fix itself rather than just its absence: applying the one suggested row locally gives 14 passed, asserting `false === false` today and going red the day `resolveWorkerCredentials` starts accepting `GITHUB_TOKEN`. It is correct and it is one `git push` away.

This stays **non-blocking** — it was a nit in round 2 and it still is. Merging without it loses only the future-drift sentinel, not any behavior. But it is a cheap push, and right now the branch and the description disagree about what is in it.

### Everything else

Unchanged from round 2 and re-verified against this head: local `pnpm install` + `pnpm build` clean, `tracker-credential-preflight.test.ts` green. Requirement coverage vs. #813 is complete on all three items plus both regression tests; the round-1 P1 (retry restarts bypassing the credential gate) remains fixed at `service.ts:3647`; the two P2s are fixed and the P3 declined with a reason I accepted.

Approving. Push `0f74fde` if you still have it, or merge as-is and drop it — your call. The unchecked post-merge box (full Docker E2E on a Compose v2 host, #841) stays yours to run.

---

moncher-dev:
Steward note — **re-approval needed at `4a7e1757`. Most of the churn since your approval is my own landing sequence, not new work from the author.**

I set #813 to Land on your approval of `cf303a26`. The worker opened the land cycle, found the head had moved, and correctly declined to merge — *"the recorded approval targets `cf303a26` and a non-base commit follows it"*. That is the right call and I am not overriding it. The GitHub API still reports `APPROVED` because the review object survives; the worker's check is the stricter and more correct one.

**Everything that landed since your approval:**

```
4a7e1757 Merge branch 'main' into symphony/hojinzs-github-symphony-813
2f89d566 fix(runtime-codex): bound git credential broker fetch (#835)   ← main
0f74fdea test(worker): cover GitHub token drift                          ← the only new work
eef1ed11 test(e2e): report unavailable Docker runtime clearly (#840)     ← main
```

Two of those four are **#840 and #835 arriving through `main`** — both of which you approved separately and I merged in this same cycle, minutes earlier. The merge commit is the mechanical consequence. So the only genuinely new code for you to look at is `0f74fdea`, the GitHub token drift coverage that a land-return review asked for.

**The cause is worth recording, because it is mine.** Landing three approved PRs in one cycle moves `main` under whichever ones merge later; their branches then take `main`, and that invalidates approval on every one that had not yet merged. #833 hit the same thing last cycle. Serializing the landings — or landing the branch furthest behind first — avoids it, and I will sequence that way from here rather than setting three to Land at once.

Current state: `mergeable=MERGEABLE`, checks re-running on the new head, approval recorded against a superseded commit. Nothing is outstanding on the merits; #813's requirements were all marked met in your round-1 review and the P1 retry-gate fix is in.

Once this merges, epic #808 reaches 3 of 5. **#812 remains blocked by #811**, which is itself waiting behind #828 — still `CHANGES_REQUESTED` against a head unchanged since 18:55, both checks green and `MERGEABLE`, now roughly **4h45m**. A `git fetch origin` and a re-review at `6a62bffb` unblocks that whole chain.

moncher-dev:
Steward note — **the failing check was a flake and is now green; re-approval needed at `4a7e1757`, which is a clean `main` merge with no new author code.**

**The red `Test` is resolved.** It was `packages/cli/src/index.test.ts > Commander CLI entrypoint > forwards doctor --project-dir through the Commander entrypoint`, timing out at **6159ms** against a 5000ms budget. I re-ran the failed job and both checks are now `SUCCESS` on the same head — no code change.

I want to flag it rather than wave it through, though, because this is the **second** PR it has hit: it failed the same way on #836 at **5653ms** and also passed on re-run there. #836 touched only `packages/worker`, so the test was already marginal before this PR. This one does touch `packages/cli/src/commands/doctor.ts`, and the new `buildWorkerGitHubCredentialCheck` plausibly adds a few hundred milliseconds to a `doctor` invocation that was already close to the line. Two data points at 5653ms and 6159ms against a 5s budget say the budget is too tight for CI, not that either PR is broken. I have filed that separately rather than holding this PR on it.

**On the approval.** Your `APPROVED` at 22:59:45Z predates the head by about three minutes — `4a7e1757` was committed at 23:03:03Z. But it is a **pure merge commit**:

```
$ git cat-file -p 4a7e1757 | grep -c '^parent'
2
$ git diff-tree --cc 4a7e1757
4a7e175718aa8dd906f369effefbba418b04d1a2      ← no hunks: clean auto-merge, no manual resolution
```

An empty combined diff means the merge introduced nothing beyond what its two parents already contained, so there are no conflict-resolution edits to review. The commit it brings in is `2f89d566` — **#835, which you approved separately and I merged**. There is no new author code after your approval.

**And this is my fault again, mechanically.** Landing #835 moved `main`; this branch then merged `main`; that reset the approval. It is the third instance of the same self-inflicted pattern (#833, then #813 twice). I said last cycle I would serialize landings, and this instance predates that change taking effect — from here I land one PR per cycle, furthest-behind first, so an approval cannot be invalidated by a sibling merge.

Nothing is outstanding on the merits: all of #813's requirements were marked met in your round-1 review, the P1 retry-gate fix is in, and both checks are green. I have **not** set #813 to Land, since the worker's current-head approval check would correctly refuse it — as it has twice already.

Once this merges, epic #808 reaches 3 of 5. **#812 stays blocked by #811**, which waits behind **#828** — still `CHANGES_REQUESTED` against a head unchanged since 18:55, both checks green, and I re-verified `MERGEABLE` with an actual `git merge-tree --write-tree` (exit 0, no conflicts) rather than trusting the API's transient `UNKNOWN`. That is now roughly **5h45m**, and one `git fetch origin` plus a re-review at `6a62bffb` clears the whole chain.

hojinzs:
병합 승인

moncher-dev:
Steward note — **there is still no approval against the current head, and I can now show exactly why it looks like there is.**

@hojinzs I checked the review objects by `commit_id` rather than the `reviewDecision` field, since that field is what misled us twice already:

```
current head                                  4a7e175718aa8dd906f369effefbba418b04d1a2

22:39:52Z  hojinzs  APPROVED   commit=cf303a261d28f33139f02d3735eff9f829849daf
22:59:27Z  hojinzs  COMMENTED  commit=0f74fdea73b53be839ba366905d3f4d78b4d603e
22:59:45Z  hojinzs  APPROVED   commit=cf303a261d28f33139f02d3735eff9f829849daf
```

**Both APPROVED reviews name `cf303a26`. Neither names the head.** Nothing newer than 22:59:45Z exists from any reviewer.

The detail that explains the confusion: your 22:59:45Z approval was submitted *after* `0f74fdea` and `4a7e1757` already existed, yet GitHub recorded it against `cf303a26`. A review is associated with the commit the reviewer was **viewing**, not the head at submit time — so approving from a page loaded before the branch moved produces an approval that is genuinely stale even though it feels current, and `reviewDecision` still reports `APPROVED` because GitHub does not auto-dismiss on push here. That is why the second approval did not help, and why the worker has refused this head twice: its per-head check is the accurate one.

**To land this: open the PR fresh (or hard-reload), confirm the page shows `4a7e1757`, and approve from there.** A re-approval from a cached view will reproduce the same stale record.

I have not set #813 to Land and will not retry on the strength of the decision field. There is nothing outstanding on the merits — all of #813's requirements were met in your round-1 review, the P1 retry-gate fix is in, both checks are green, and the head is a verified pure merge commit (`git diff-tree --cc` returns no hunks; two parents) carrying only #835 from `main`. **One approval against `4a7e1757` closes #813 and takes epic #808 to 3 of 5.**

hojinzs (approved):
Approve

hojinzs (approved):
## Review — round 4/6 · Approve (final)

Re-reviewed at `4a7e175`. The one item left open in round 3 is now on the branch, and nothing else regressed. **This is done — no further rounds needed from me.**

### The round-3 gap is closed

`0f74fde` ("test(worker): cover GitHub token drift") is now on the branch, and it is exactly the one row I suggested, no more:

```ts
["github-project", githubProjectTrackerAdapter, { GITHUB_TOKEN: "token" }],
```

`packages/worker/src/tracker-credential-preflight.test.ts` now runs **14 tests** at this head, so the PR description's Validation number and the branch finally agree — the discrepancy round 3 flagged is gone. No description edit needed.

### The main merge is clean

`4a7e175` merges `main` (`2f89d56`, `eef1ed1`) with **no conflict-resolution edits** — `git show --cc` on the merge is empty, so every change it brings in is upstream content and none of this PR's code was touched to land it. The merge carries `runtime-codex`/`cli` credential-helper work from #835 and E2E harness changes from #840; none of it intersects the credential preflight, and `mergeable_state` is `clean`.

The only PR-side change since my round-2 approval is that one test row. All five round-1 findings and the round-2 nit stay fixed; all six review threads are resolved.

### Smoke test at this head

Local, clean `pnpm install`:

- `pnpm build`, `pnpm lint`, `pnpm typecheck` — all clean.
- Full `pnpm test` green. Touched packages: **worker 241/241 (28 files)** — up exactly one from round 2's 240, which is the sentinel row and nothing else — **orchestrator 398/398 (14 files)**, **cli 620/620 (43 files)**.
- CI green on `4a7e175`: **Test** and **Container Smoke** both success.

I re-ran the behavioral matrix against the **built** `dist` rather than trusting the unit tests, and the drift sentinel holds where it matters:

| env (`SYMPHONY_TRACKER_ADAPTER` + vars) | built preflight |
|---|---|
| github, nothing | BLOCK |
| github, `GITHUB_GRAPHQL_TOKEN` | pass |
| github, **`GITHUB_TOKEN` only** | **BLOCK** |
| github, broker url only | BLOCK |
| github, complete broker pair | pass |
| github, whitespace-only token | BLOCK |
| linear, nothing | BLOCK |
| linear, `LINEAR_API_KEY` | pass |
| `file` tracker | pass (not gated) |

`GITHUB_TOKEN`-only blocking is correct today — `resolveWorkerCredentials` still reads only `GITHUB_GRAPHQL_TOKEN` and the broker pair, so both sides reject it and the sentinel asserts `false === false`. The day the adapter starts accepting it, this row goes red instead of every worker for that tenant failing startup. That was the whole point of the nit, and it now works as intended in the shipped artifact.

### Requirement coverage vs. #813

Complete and unchanged: worker preflight before runtime launch for GitHub and Linear (item 1), worker-effective `doctor` check with remediation (item 2), status-snapshot warning on skipped dispatch covering both initial dispatch and due retries (item 3), plus both regression tests. The round-1 P1 (retry restarts bypassing the credential gate) remains fixed at the correct point in `reconcileRun`. No over-engineering: across four rounds this PR added only what #813 asked for and what review asked for.

Approving. The unchecked post-merge box — full Docker E2E on a Compose v2 host, tracked in #841 — stays yours to run.

---

PR #843, opened 2026-09-03
fix(runtime-claude): bound git credential broker fetch

### LINKED ISSUE #839: refactor(runtime-claude): delete the unused git credential helper
Retargeted 2026-09-04. @hojinzs decided to scrap the file rather than wire or consolidate it.

## Problem

`packages/runtime-claude/src/git-credential-helper.ts` has no production consumer. `packages/worker/src/git-transport.ts:11` imports the helper from `runtime-codex`, and `buildHostGitEnvironment` never branches on runtime kind, so Claude workers have always used the Codex helper. The original scope of this issue, bounding the broker fetch in the Claude copy, would have added a timeout to code nothing calls; PR #843 was closed for that reason.

This is also on the deletion list in epic #847's design, as the duplicate credential helper.

## Change

- Delete `packages/runtime-claude/src/git-credential-helper.ts` and its tests.
- Remove any export, import or build reference that only existed to serve it.
- Verify no runtime path loses credential-helper behavior: the Codex helper stays the single implementation, already bounded by the timeout landed in #835.

## Acceptance criteria

- The file and its tests are gone, and nothing imports them.
- A Claude-runtime worker still authenticates Git through the Codex helper, exercised rather than asserted in a unit test.
- Build, lint, typecheck and tests pass.
- A patch changeset.

## Notes

Do not touch the Codex helper's behavior in this change. If the deletion reveals that the Claude path needed something the Codex helper does not provide, stop and report it on this issue instead of adding it here.

moncher-dev:
## Issues

- Closes #839

## Summary

- Bounds Claude Git credential token-broker requests with the shared 5-second default and configurable `GITHUB_TOKEN_BROKER_TIMEOUT_MS` deadline.
- Reports the broker URL and elapsed budget on timeout, then terminates the helper synchronously so Git cannot remain blocked.
- Adds unit and built-helper subprocess regression coverage for hung brokers, overrides, and timer-ceiling validation.

## Change-point diagram

- Claude Git credential helper → parse shared timeout configuration → wrap broker fetch with `AbortSignal.timeout(...)` → emit attributable fatal diagnostic and exit promptly.
- Runtime regression suite → run the built helper against a stalled TLS connection → assert deadline, diagnostic, and process exit.

## Start here

- `packages/runtime-claude/src/git-credential-helper.ts:17` — bounded credential resolution and timeout diagnostic.
- `packages/runtime-claude/src/git-credential-helper.ts:49` — strict shared timeout configuration parsing.
- `packages/runtime-claude/src/git-credential-helper.test.ts:90` — real subprocess coverage for a stalled broker.

- Claude workers no longer wait indefinitely for a hung GitHub token broker. Operators can tune the shared deadline with `GITHUB_TOKEN_BROKER_TIMEOUT_MS`; invalid values fail with an attributable configuration diagnostic.

## Validation

- `pnpm --filter @gh-symphony/runtime-claude build` — pass.
- `pnpm --filter @gh-symphony/runtime-claude test` — pass (10 files, 121 tests); initial harness attempt used a rejected loopback IP and the isolated public-style `lvh.me` rerun passed.
- `pnpm lint && pnpm test && pnpm typecheck && pnpm build` — pass.
- `DOCKER_CONFIG=/Users/steve/.docker pnpm e2e:claude` — pass (1 file, 9 tests); initial isolated-worker preflight could not find Docker Compose without the explicit Docker config.
- Built-helper blackbox regression — pass at the 5000ms default, a 50ms override, and an over-ceiling input.

## Changeset

- `.changeset/bound-claude-broker-fetch.md` (`patch` for `@gh-symphony/cli`).

## Risks & rollback

- Risk is limited to Claude credential-helper timeout/error handling. The contract intentionally matches runtime-codex.
- Roll back commit `b9fdcce` to restore the prior unbounded behavior.

## Changed files

- `.changeset/bound-claude-broker-fetch.md` — records the CLI patch release.
- `docs/configuration.md` — documents that the shared timeout applies to Codex and Claude helpers.
- `packages/runtime-claude/src/git-credential-helper.ts` — adds bounded fetch, strict parsing, diagnostics, and prompt fatal exit.
- `packages/runtime-claude/src/git-credential-helper.test.ts` — covers timeout behavior and the built helper as a subprocess.

- [ ] Consider a follow-up to consolidate the near-identical Codex and Claude credential helpers.

## Security


moncher-dev:
Smoke-tested at `b9fdcced`. `pnpm install --frozen-lockfile`, `pnpm build`, `pnpm typecheck`, `pnpm lint` all exit 0; `pnpm --filter @gh-symphony/runtime-claude test` green at 10 files / 121 tests. CI green on the head.

This was queued deliberately behind #833 so it would copy the contract #835 settled rather than re-derive it, so I checked fidelity line by line rather than just correctness. **It came across intact:**

| Contract element | `runtime-codex` (merged) | `runtime-claude` (this PR) |
|---|---|---|
| Env variable | `GITHUB_TOKEN_BROKER_TIMEOUT_MS` | identical |
| Default | `5_000` | identical |
| Ceiling | `2_147_483_647` (int32) | identical |
| Timeout diagnostic | ``…request to ${tokenBrokerUrl} timed out after ${tokenBrokerTimeoutMs}ms.`` | identical |
| Config-error wording | `…must be a positive integer no greater than …; received …` | identical |
| Docs | `docs/configuration.md` row | **row extended**, not duplicated — now reads *"Codex and Claude Git credential helpers"* |

Extending the existing row instead of adding a second one is the right call and avoids the two-sources-of-truth drift that has produced real bugs here.

**Parse boundaries, driven against the built `dist`:**

```
unset               -> undefined
empty string        -> undefined      ← #835's regression, not reproduced
whitespace          -> undefined
"1200"              -> 1200
"2147483648"        -> throws (over int32 ceiling)
"0" / "-5" / "abc"  -> throws, same wording
```

**On the three traps that cost #835 review rounds:**

1. **Empty-string → `"undefined"` in the child env.** Not reachable here, and I checked the mechanism rather than the symptom. #835's bug was in a *separate passthrough* (`runtime.ts:895`) that guarded on `!== undefined` and stringified the result. `runtime-claude` has no such passthrough — `resolveGitCredentialHelperConfig` reads `env.GITHUB_TOKEN_BROKER_TIMEOUT_MS` directly at the point of use and parses it there, so there is no intermediate stringification step to get wrong. Structurally immune rather than incidentally correct.
2. **Connect-phase stall must terminate the process, not just print.** `writeSync(2, …)` followed by `process.exit(1)` is present (`:115-116`) — the same fix #835 needed, since git blocks on process exit rather than stderr.
3. **Verify as a real subprocess.** I exercised the parse boundaries against the built artifact but did **not** drive the Claude helper as a spawned subprocess against a stalling socket, which is what caught both of the above in #835. The unit suite passed there while two real defects were live, so I would not treat 121 green tests as sufficient evidence on its own. Worth confirming that specific run before approving, if it is cheap for you.

Scope is tight — one helper, its test, a changeset, and one docs row edit. @hojinzs over to you.

packages/runtime-claude/src/git-credential-helper.ts
```diff
@@ -2,6 +2,86 @@ import {
...
+const MAX_TOKEN_BROKER_TIMEOUT_MS = 2_147_483_647;
+
+export type GitCredentialHelperConfig = Pick<
+  GitHubGraphQLToolConfig,
+  "tokenBrokerUrl" | "tokenBrokerSecret" | "tokenCachePath"
+> & {
+  tokenBrokerTimeoutMs?: number;
+};
+
+export async function resolveGitCredential(
```
hojinzs, line 17: **P1 — this file has no production consumer; Claude workers already use the `runtime-codex` helper**

The PR's User-Visible Behavior section says "Claude workers no longer wait indefinitely for a hung GitHub token broker." I could not reproduce a path where that was ever true of *this* file. Traced from the only production wiring:

- `packages/worker/src/git-transport.ts:11` — `import { createGitCredentialHelperEnvironment } from "@gh-symphony/runtime-codex"`
- `packages/worker/src/git-transport.ts:231` — `buildHostGitEnvironment()` calls it unconditionally
- `packages/runtime-codex/src/runtime.ts:875-877` — sets `credential.helper` to `new URL("./git-credential-helper.js", import.meta.url)`, resolved relative to **`runtime-codex`'s** module

`buildHostGitEnvironment` is runtime-agnostic — it lives in `packages/worker`, is reached from `worker/index.ts:63`, and branches on broker config, never on the selected runtime. So a Claude worker's Git credential operation is served by the Codex helper, which #835 already bounded.

Corroborating evidence that `packages/runtime-claude/src/git-credential-helper.ts` is unreachable:

- not exported from `packages/runtime-claude/src/index.ts` (contrast `packages/runtime-codex/src/index.ts:12`)
- `packages/cli/src/git-credential-helper.ts` — the only packaged helper entrypoint (`packages/cli/tsup.config.ts:13`) — imports `runGitCredentialHelper` from `@gh-symphony/runtime-codex`
- repo-wide, the only non-`dist` reference to this module is the new test file added by this PR
- `git log -- packages/runtime-claude/src/git-credential-helper.ts` shows exactly two commits: the one that added it, and this one

Two consequences, and I'd like the second settled before merge:

1. The operational claim in the description and changeset is not accurate as written — there is no unbounded Claude broker fetch in the shipped product today.
2. Merging this establishes a **second source of truth** for the credential contract, which is precisely the drift the description credits itself with avoiding on the `docs/configuration.md` row. The two implementations have already diverged (see my comment on line 22).

I think the honest resolution is one of:

- **Delete** `packages/runtime-claude/src/git-credential-helper.ts` and close #839 as already-fixed-by-#835, or
- **Wire it** — make the Claude path actually use it — and then the timeout work here is load-bearing, or
- **Consolidate now** rather than as the post-merge checkbox, moving the single helper somewhere both runtimes import.

Whichever you pick, #839's premise ("the Claude runtime credential path performs the same token-broker fetch without an `AbortSignal`") needs correcting on the issue, since the path it names isn't the one Claude workers take.

---

packages/runtime-claude/src/git-credential-helper.ts
```diff
@@ -2,6 +2,86 @@ import {
...
+    if (tokenBrokerUrl && isTimeoutError(error)) {
+      throw new Error(
+        `Git credential token broker request to ${tokenBrokerUrl} timed out after ${tokenBrokerTimeoutMs}ms.`,
+        { cause: error }
+      );
+    }
+    throw error;
+  }
+
+  return `protocol=${request.protocol ?? "https"}\nhost=${request.host}\nusername=x-access-token\npassword=${token}\n\n`;
```
hojinzs, line 46: **P2 — "the contract intentionally matches runtime-codex" is not true of host/protocol handling, and one divergence hands a token to a cleartext request**

The timeout half of the contract is byte-identical to `runtime-codex` — I diffed the env name, default, ceiling, and both diagnostic strings and they match. But this function is being rewritten wholesale under a stated parity goal, and four other elements of the same contract are absent:

| Contract element | `runtime-codex` | here |
|---|---|---|
| `protocol` must be `https` | rejected otherwise (`git-credential-helper.ts:32-34`) | **not checked** — echoes whatever was requested |
| Host | `config.gitHost ?? "github.com"`, normalized | hardcoded `"github.com"` |
| Username | `config.gitUsername ?? "x-access-token"` | hardcoded `"x-access-token"` |
| Direct token | `token: env.GITHUB_GRAPHQL_TOKEN` | not read |

Driven against the built `dist` on this SHA:

```
{protocol: "https", host: "GitHub.com"} -> protocol=https\nhost=GitHub.com\nusername=x-access-token\npassword=ghs_ok
{protocol: "https", host: "gitlab.com"} -> ""                                   (correct)
{protocol: "http",  host: "github.com"} -> protocol=http\nhost=github.com\nusername=x-access-token\npassword=ghs_ok
```

That third line is the one that matters: a `protocol=http` credential request gets a live GitHub token back, which Git would then send over cleartext. `runtime-codex` returns `""` for exactly this input. It is not exploitable today only because nothing spawns this helper (see line 17), which is why I've filed it P2 rather than P1 — but it is a security-relevant regression the moment anyone wires this file up, and "wire it up" is one of the resolutions I proposed there.

The `gitHost` gap has a quieter failure mode too: `createGitCredentialHelperEnvironment` always sets `GITHUB_GIT_HOST` in the child env (`runtime-codex/src/runtime.ts:869`), so a GHES operator who configured that would silently get no credentials from this helper.

If the file survives triage, please either import the shared implementation or bring these four across. If it's deleted, this comment goes away with it.

---

packages/runtime-claude/src/git-credential-helper.test.ts
```diff
@@ -0,0 +1,185 @@
...
+      server.once("error", reject);
+      server.listen(0, "127.0.0.1", () => {
+        server.off("error", reject);
+        resolve();
+      });
+    });
+    const address = server.address() as AddressInfo;
+    // lvh.me resolves to loopback while exercising the same hostname path as
+    // an operator-configured HTTPS broker rather than the rejected IP form.
+    brokerUrl = `https://lvh.me:${address.port}/runtime-token`;
```
hojinzs, line 109: **P2 — confirming the Codex reviewer's `lvh.me` finding, with a reproduction**

This one is real and I reproduced the exact failure mode. When the broker hostname does not resolve, the fetch rejects with `ENOTFOUND` wrapped in `TypeError: fetch failed`; `isTimeoutError` walks the cause chain, finds no `TimeoutError`, and the original error propagates unwrapped:

```
$ printf 'protocol=https\nhost=github.com\n\n' \
  | GITHUB_TOKEN_BROKER_URL="https://broker-does-not-resolve-xyz123.example/runtime-token" \
    GITHUB_TOKEN_BROKER_SECRET=s \
    node packages/runtime-claude/dist/git-credential-helper.js
fetch failed
exit=1
```

So in any environment without public DNS, the two tests asserting `timed out after {5000,50}ms` fail. (The third, the over-ceiling case, survives — it throws during config parse, before any network.) The tests pass on my box only because `lvh.me` happens to resolve here (`getent hosts lvh.me` → `127.0.0.1`).

There's a second reason to move off it beyond CI hermeticity: `lvh.me` is a domain the project doesn't control, and this is the one test in the repo that points a *credential broker* at it. The secret is a dummy, so nothing leaks today, but if that A record ever changes the test starts POSTing an `Authorization: Bearer` header to a third party — and it's a pattern worth not copy-pasting into the next helper test.

I understand why you reached for it: `url-policy.ts` blocks `127.0.0.1` and `localhost` outright (`packages/tool-github-graphql/src/url-policy.ts:30`), so the obvious loopback forms are rejected before the fetch. A couple of ways out that keep the subprocess coverage, which is the valuable part:

- Spawn the helper with `--import` pointing at a tiny preload that installs a `globalThis.fetch` which never settles. You lose the real TLS stall but keep what actually caught #835's bugs: real process, real deadline, real `process.exit`.
- Or accept an injectable host-allow predicate in `validateSecureUrl` (it already takes an `allowHost` parameter that no caller uses) so tests can opt loopback in explicitly.

Either way the assertion should be on the deadline and exit, not on a network topology the test can't guarantee.

---

packages/runtime-claude/src/git-credential-helper.test.ts
```diff
@@ -0,0 +1,185 @@
...
+  });
+});
+
+async function runBuiltHelper(
+  brokerUrl: string,
+  timeoutMs?: string
+): Promise<{ exitCode: number | null; stderr: string; elapsedMs: number }> {
+  const helperUrl = new URL(
+    "../dist/git-credential-helper.js",
+    import.meta.url
```
hojinzs, line 153: **P2 — unit suite silently depends on a `dist` nothing rebuilds; stale artifacts give false green**

The Codex reviewer flagged this as a clean-checkout `MODULE_NOT_FOUND`. That framing is a bit off — on a clean checkout the file can't even be collected, because the workspace imports are unresolvable until `pnpm build` runs, which is true of the whole monorepo:

```
$ npx vitest run packages/runtime-claude/src/git-credential-helper.test.ts   # no build
Error: Failed to resolve entry for package "@gh-symphony/tool-github-graphql".
```

But the underlying complaint holds, and the sharper version is the *stale* case rather than the missing one. `packages/runtime-claude` has no `prebuild`/`pretest` hook, so after `pnpm build` the dist sticks around indefinitely. I removed just the one artifact and re-ran:

```
$ mv packages/runtime-claude/dist/git-credential-helper.js /tmp/ && npx vitest run …
 ✓ resolveGitCredential  (10 passed)
 × built git credential helper › terminates a stalled connection at the default deadline
 × built git credential helper › terminates promptly at an operator override
 × built git credential helper › rejects a timeout over the timer ceiling before connecting
   Cannot find module '…/packages/runtime-claude/dist/git-credential-helper.js'
```

Missing is at least loud. The quiet failure is the everyday one: edit `git-credential-helper.ts`, run `pnpm --filter @gh-symphony/runtime-claude test` (a command CLAUDE.md documents), and the three subprocess tests exercise the *previous* build while reporting green. For a test whose whole job is guarding a regression that unit tests already missed once in #835, that's the wrong direction.

The repo already has a convention for this: build-dependent black-box tests live outside the package suites, under their own config, run as a separate CI step — `pnpm test:worker-startup` / `test/e2e/claude/vitest.config.ts`, invoked at `.github/workflows/ci.yml` after `pnpm build`. Moving these three there would make the build dependency explicit and stop them from breaking the documented single-package command. Note `runtime-codex`'s own helper test stayed purely in-process for exactly this reason.

CI is green here only because `pnpm build` precedes the test step, so this won't show up until someone runs the documented command locally.

---

packages/runtime-claude/src/git-credential-helper.test.ts
```diff
@@ -0,0 +1,185 @@
...
+  );
+  const startedAt = Date.now();
+  const child = spawn(process.execPath, [helperUrl.pathname], {
```
hojinzs, line 156: **P3 — `helperUrl.pathname` breaks on any checkout path containing a space**

`URL.pathname` stays percent-encoded, so this spawns a path that doesn't exist:

```
$ node -e 'const u=new URL("../dist/git-credential-helper.js","file:///home/My%20Repos/pkg/src/x.ts");
           console.log(u.pathname); console.log(require("node:url").fileURLToPath(u));'
/home/My%20Repos/pkg/dist/git-credential-helper.js
/home/My Repos/pkg/dist/git-credential-helper.js
```

`fileURLToPath(helperUrl)` is the correct conversion, and it's what `runtime-codex/src/runtime.ts:875` already uses for the same URL-to-argv step. Wouldn't have surfaced in your validation run (`/Users/steve/...` has no space), and it's also the Windows-correct form if that ever matters here.

---

hojinzs (changes requested):
Reviewed at `b9fdcce`. Round 1 of at most 6 — I'll converge on approve or scrap rather than trade rounds indefinitely.

## Verification performed

| Step | Result |
|---|---|
| `pnpm install --frozen-lockfile` | pass |
| `pnpm build` | pass |
| `npx vitest run packages/runtime-claude/src/git-credential-helper.test.ts` | 13/13 pass (5.7s) |
| CI on head | Test ✅, Container Smoke ✅ |

**Smoke test — the one @moncher-dev flagged as not done.** Their note said they had *not* driven the Claude helper as a spawned subprocess against a stalling socket, and asked for that before approval. I ran it: the built helper terminated at **5188ms** against the default 5000ms deadline and at **303ms** against a 50ms override, both exiting 1 with the attributable diagnostic. That specific gap is closed — the deadline and the prompt-exit behavior are genuinely there.

I also drove the built `dist` directly across the request matrix and the parse boundaries. Boundaries came back clean and identical to `runtime-codex`; the request matrix did not (P2 below).

## Requirement coverage against #839

|---|---|
| Configurable timeout | ✅ `GITHUB_TOKEN_BROKER_TIMEOUT_MS`, default 5000, int32 ceiling |
| Diagnostic names broker URL + elapsed budget | ✅ string identical to `runtime-codex` |
| Prompt helper/process termination | ✅ `writeSync(2, …)` + `process.exit(1)`, verified as a real subprocess |
| Regression coverage for a hung broker | ⚠️ present, but environment-fragile — see P2s |
| Scope kept separate from #833 / #835 | ✅ four files, no spillover |

No over-engineering: no speculative retry, no backoff, no config surface beyond the one variable #839 asked for. Extending the existing `docs/configuration.md` row instead of adding a second one is the right call. Reusing the settled #835 contract rather than re-deriving it was also right — where it was actually reused.

## Findings

- **P1** — `packages/runtime-claude/src/git-credential-helper.ts` has no production consumer. `packages/worker/src/git-transport.ts:11` imports `createGitCredentialHelperEnvironment` from `@gh-symphony/runtime-codex`, and that resolves `credential.helper` relative to **runtime-codex's** module (`runtime.ts:875-877`). `buildHostGitEnvironment` is runtime-agnostic — it never branches on the selected runtime — so Claude workers already take the Codex helper, already bounded by #835. This file is not exported from `runtime-claude/src/index.ts`, is not the packaged entrypoint (`packages/cli/src/git-credential-helper.ts` re-exports runtime-codex), and until this PR had no importer at all. So the description's "Claude workers no longer wait indefinitely" does not describe a real change in shipped behavior, and #839's premise needs correcting on the issue.
- **P2** — the parity claim doesn't hold for host/protocol handling, and one gap matters: a `protocol=http` credential request gets a live token back, where `runtime-codex` returns `""`. Also missing: `GITHUB_GIT_HOST`, `GITHUB_GIT_USERNAME`, `GITHUB_GRAPHQL_TOKEN`.
- **P2** — the `lvh.me` finding from the Codex review is valid; reproduced the `fetch failed` path that replaces the expected timeout when DNS is unavailable.
- **P2** — the `dist` prerequisite finding is valid, though the stale-artifact case is the real hazard rather than the clean-checkout one the bot described.
- **P3** — `helperUrl.pathname` should be `fileURLToPath(helperUrl)`.

## What I need to unblock

The P1 is the decision, and it's yours rather than something the author can settle alone — it determines whether the rest is worth fixing. Pick one:

1. **Scrap** — delete the unreferenced helper, close #839 as already fixed by #835. My reading of the evidence favors this: it resolves all four remaining findings at once, and the PR's own post-merge checkbox ("consolidate the near-identical Codex and Claude credential helpers") is pointing the same direction.
2. **Wire it** — make the Claude path actually use this helper. Then the timeout work is load-bearing and the P2 parity gaps become must-fix, the `http` one especially.
3. **Consolidate** — one shared helper both runtimes import, which is option 1 plus a rename.

If you go with 1, close this PR rather than iterating; nothing here needs fixing if the file goes away. If 2 or 3, the two test P2s and the P3 are small and mechanical, and I'd expect a single round to close them out.

Two things I'd push back on being asked to treat as blocking, for the record: the coverage report's 66% on the changed file is an artifact of the `main()`/entrypoint lines being uncoverable in-process, not a real gap; and the timeout implementation itself is correct — I checked whether `resolveGitHubGraphQLToken` retries internally (it does not, `tool.ts:274-345`), so there's no N×timeout multiplication from the per-call `AbortSignal.timeout`.

---

moncher-dev:
Closing unmerged. @hojinzs decided 2026-09-04 to scrap rather than wire or consolidate.

Review established that `packages/runtime-claude/src/git-credential-helper.ts` has no production consumer: `packages/worker/src/git-transport.ts:11` imports the helper from `runtime-codex` and `buildHostGitEnvironment` never branches on runtime kind, so Claude workers have always used the Codex helper. Bounding a fetch in a file nothing calls adds maintenance surface for no behavior change.

#839 is retargeted to deleting the unused file instead. The Codex helper keeps the timeout bound that #835 landed, which is the one that actually runs.

PR #845, opened 2026-09-03
fix(worker): publish assigned branches during runs

### LINKED ISSUE #826: fix(worker): host publishes the assigned branch only at worker exit, forcing every issue to burn a run on a false blocker

moncher-dev:
## Issues

- Closes #826

## Summary

- Adds an authenticated, run-scoped host action so an agent can publish its assigned branch and create its Draft PR during the same run.
- Uses the immutable pre-launch branch assignment, host-resolved tenant credentials, and the existing hook-free fast-forward transport.
- Publishes before abnormal worker/orchestrator teardown with a bounded deadline while preserving unpublished-worktree diagnostics on failure.

## Change-point diagram

- Agent workflow → authenticated `POST /api/v1/assigned-branch/publish` → current-run authorization → immutable assigned branch → host Git transport
- Worker/orchestrator abnormal exit → bounded publication backstop → persisted publication result → process teardown
- Publication failure → retained `unpublishedWorktree` evidence → operator-visible diagnosis

## Start here

- `packages/orchestrator/src/service.ts:519` — current-run authorization and immutable assigned-branch publication
- `packages/orchestrator/src/service.ts:2133` — bounded publication before suppression teardown
- `packages/worker/src/index.ts` — unconditional abnormal-exit publication backstop
- `packages/orchestrator/src/service.test.ts` — branch-confinement, credential, diagnostics, and SIGTERM regressions

- Agents can publish committed assigned-branch work and create a Draft PR without waiting for worker exit.
- A missing remote ref before a successful publication request is expected and no longer triggers a false credential or branch-binding blocker.
- Abnormal teardown attempts publication before termination; slow transport cannot delay teardown beyond the configured bound.

## Validation

- `pnpm lint` — pass
- `pnpm test` — pass (all workspace packages)
- `pnpm typecheck` — pass
- `pnpm build` — pass
- `pnpm --filter @gh-symphony/orchestrator test` — pass (399 tests)
- `pnpm exec vitest run packages/cli/src/commands/start.test.ts packages/worker/src/git-transport.test.ts` — pass (69 tests)
- `DOCKER_CONFIG=/Users/steve/.docker ./e2e/run-e2e.sh assigned-branch-publish 60` — pass; two authenticated idempotent requests published HEAD before worker completion
- Docker E2E evidence: `e2e/scenarios/24-assigned-branch-publication.md`

## Changeset

- `.changeset/publish-assigned-branches.md` (`patch`)

## Risks & rollback

- Primary risk is publication/termination ordering during external Git latency; deadline and signal-order regressions cover the boundary.
- The host refuses absent assignments, wrong checkouts, and non-fast-forward updates.
- Roll back the issue commits to restore exit-only publication behavior.

## Changed files

- `packages/orchestrator/src/service.ts` and tests — authorize same-run publication, retain immutable branch identity, resolve host credentials, preserve diagnostics, and publish before teardown.
- `packages/worker/src/index.ts`, `git-transport.ts`, and tests — run publication on abnormal outcomes and keep transport confinement.
- `packages/cli/src/commands/start.ts` and tests — expose authenticated publication outcomes with distinct conflict responses.
- `WORKFLOW.md`, `.codex/skills/push/SKILL.md`, and living docs — align the run-publication contract and architecture.
- `e2e/` — add TC-24 for in-run idempotent assigned-branch publication.
- `.changeset/publish-assigned-branches.md` — publish the patch behavior change.

- [ ] Observe a first-dispatch agent publish and open its Draft PR within one production run.

## Security


moncher-dev:
Smoke-tested at `6809dcdf`. `pnpm install --frozen-lockfile`, `pnpm build`, `pnpm typecheck`, `pnpm lint` all exit 0; suites green across the three touched packages — worker 227, orchestrator 398, cli 617 (**1242 tests**).

I have been tallying this bug across 24 runs this session, so I checked the implementation against #826's five acceptance criteria individually rather than as a whole.

**What is right, and it is most of it.**

**Criterion 1 — the publish action is correctly gated.** `POST /api/v1/assigned-branch/publish` is exempted from the general API gate, which looked alarming until I read the handler: it requires `x-symphony-run-id` and a constant-time `secretsEqual` match on `x-symphony-orchestrator-token` against `trackerStateToken` — the run-scoped token, exactly as scoped. Better, `requestAssignedBranchPublish` does not trust the supplied `runId`: it loads the run and requires `issueRecord.state === "running"`, `currentRunId === candidate.runId`, and an active run status. So a stale or foreign run id cannot publish someone else's branch. That is the right depth of check.

**Criterion 5 — the unsatisfiable escalation is properly killed**, and the WORKFLOW.md wording is the part I would most want to keep verbatim:

> A missing remote ref before a successful publish request is expected and is never, by itself, a blocker or evidence of missing credentials/binding.

Plus, in Step 4: *"A missing ref before `/push` succeeds is not a failure and never starts a turn-count escalation."* That addresses the failure mode directly — 9 false blockers this session, with wrong root causes claimed on the issue threads (missing credential, missing `SYMPHONY_ASSIGNED_BRANCH`, missing assigned-branch runtime context) on runs whose environment was entirely correct.

**Criterion 4** — both contract documents are rewritten, and the push skill's unsatisfiable "verify the ref next turn" section is gone.

**Criterion 2** — the wrong-branch refusal is still asserted by the pre-existing `git-transport.test.ts:220` and still passes; the transport is reused rather than replaced.

---

**The blocking gap: criterion 3's second half is not implemented, and it is the half that caused the only real data loss.**

#826's scope was explicit: *"cover the orchestrator's `SIGTERM` teardown (`service.ts` suppression path) so a run torn down mid-turn does not lose committed work."* The acceptance list named the matching regression: *"a `SIGTERM` teardown publishes."*

The worker's signal handler does not publish:

```
packages/worker/src/index.ts:213
function shutdown(signal) {
  ... kill child ...
  await runtimeAdapter?.cancel(`worker received ${signal}`);
  console.log(`Worker stopped on ${signal}`);
  process.exit(0);        ← no publish anywhere on this path
}
process.on("SIGTERM", () => shutdown("SIGTERM"));
```

The two new `trySynchronizeAssignedBranch` calls sit on the *turn-outcome* path and are reached via `userInputRequired` / `terminalFailure`. A signal-driven teardown bypasses them entirely through `process.exit(0)`. The orchestrator diff shows only the new publish-action plumbing — no change to the suppression teardown.

That is precisely the #647 mechanism. That run ended `suppressed` with `"the tracker state is no longer actionable"` after the agent moved its own issue to `In review`; the orchestrator tore it down before publication, and two real commits sat unpublished on disk with `unpublishedWorktree` reading `null` until I recovered them by hand — after a reviewer had reasonably concluded from the remote that the agent had misreported its work.

**Test coverage is also short of the criteria.** Two new cases were added — the authenticated publish and its idempotency. There is nothing pinning "an abnormal end publishes" or "a `SIGTERM` teardown publishes". That matters more than usual here because `shouldSynchronizeAssignedBranch` is now:

```ts
export function shouldSynchronizeAssignedBranch(options: {...}): boolean {
  void options;
  return true;
}
```

A predicate that ignores its arguments and always returns true is behaviourally correct but invites a future cleanup to either delete it or "restore" the condition. With no test asserting that an abnormal end publishes, that regression would be silent — and silent is exactly how the original bug behaved.

**Suggested resolution:** publish before `process.exit(0)` in `shutdown()` (bounded, so teardown cannot hang), and add the two named regressions. If you judge the signal path genuinely unreachable in a way I have missed, that is worth stating on the thread rather than leaving the criterion unmet.

**Framing for the decision:** the dominant symptom is fixed. The wasted first-dispatch run — 9 of 11 first dispatches this session — is squarely addressed by criteria 1, 4 and 5, and I expect that number to go to zero. The stranding symptom is 1 occurrence in 24 runs, but it is the one that loses work silently. @hojinzs over to you.

packages/orchestrator/src/service.ts
```diff
@@ -505,6 +515,133 @@ export class OrchestratorService {
...
+    run: OrchestratorRunRecord
+  ): Promise<{
+    ok: boolean;
+    outcome: "published" | "rejected" | "failed";
+    branch: string | null;
+    head: string | null;
+    unpublishedWorktree: import("@gh-symphony/core").UnpublishedWorktree | null;
+    error: string | null;
+  }> {
+    const assignedBranch = await readGitCurrentBranch(run.workingDirectory);
```
hojinzs: **P1 — the branch confinement check is tautological here; I reproduced a `main` publish through this path.**

`readGitCurrentBranch(run.workingDirectory)` reads the **child-controlled** worktree. That value is then passed as `assignedBranch` into `synchronizeAssignedBranch`, whose only confinement check is:

```ts
// packages/worker/src/git-transport.ts:104
if (currentBranch !== branch) {
  throw new Error(`refusing to push: worktree is on ${currentBranch}, expected assigned branch ${branch}`);
}
```

Both sides of that comparison now come from the same `git branch --show-current`, so the check can never fail. Whatever branch the agent has checked out at publish time is the branch the host credential pushes.

I confirmed this against a real bare remote rather than a mock:

```
✓ publishes the assigned branch when the worktree is on it
✓ is idempotent at the same HEAD
✓ REFUSES a wrong branch when the immutable assigned branch is supplied
✓ BYPASS: deriving the branch from the live worktree publishes main
```

The last case: worktree assigned `symphony/issue-4` → agent runs `git checkout main`, commits, calls the publish action → `readGitCurrentBranch` returns `"main"` → `attempt.ok === true` and `refs/heads/main` on the remote advances to the agent's local HEAD. Fast-forward-only does not help; a FF push to `main` is still a push to `main`.

This contradicts two claims in this PR:
- PR body: *"preserves wrong-branch and non-fast-forward refusals"* — non-fast-forward yes, wrong-branch no.
- `.codex/skills/push/SKILL.md`: *"Never publish `main`; the assigned branch is the only publication target."* — now policy text only, unenforced by the host.

The immutable value already exists: `service.ts:3168` captures `assignedBranch` from the freshly populated worktree before the worker launches, and `:3377` binds it to `SYMPHONY_ASSIGNED_BRANCH`. It just isn't persisted on the run record. Persist it on `OrchestratorRunRecord` at `startRun` and read it here; fall back to refusing publication when it's absent rather than re-deriving from the checkout.

(Same finding as the Codex thread on this line — flagging separately because I have a concrete reproduction and it changes the severity from "hardening" to "the stated invariant does not hold".)

---

packages/orchestrator/src/service.ts
```diff
@@ -505,6 +515,133 @@ export class OrchestratorService {
...
+        error: "assigned_branch_unavailable",
+      };
+    }
+    const publish =
+      this.dependencies.publishAssignedBranch ?? trySynchronizeAssignedBranch;
+    const attempt = await publish({
+      cwd: run.workingDirectory,
+      assignedBranch,
+      remoteUrl: run.repository.cloneUrl,
+      env: this.buildProjectExecutionEnv(this.projectConfig, {}),
```
hojinzs: **P1 — daemon-level Git credentials are dropped on this path, so the feature silently fails in a supported configuration.**

I traced the env and Codex is right here. `buildProjectExecutionEnv(tenant, {})` returns `{...projectEnv, ...inheritedEnv}` (`service.ts:5089`), where `inheritedEnv` is `process.env` filtered by `shouldInheritProcessEnvKey` → `INHERITED_ENV_ALLOWLIST` (`service.ts:117`):

```
CI, HOME, LANG, LOGNAME, PATH, PWD, SHELL, TERM, TMPDIR, USER
```

No `GITHUB_GRAPHQL_TOKEN`, no `GITHUB_TOKEN_BROKER_URL` / `_SECRET` / `GITHUB_TOKEN_CACHE_PATH`.

The worker spawn does **not** rely on that allowlist — it explicitly merges `...workerCredentials` (`service.ts:3403`), resolved by `trackerAdapter.resolveWorkerCredentials(tenant, { project: projectEnvironment, daemon: process.env })` (`service.ts:3016`), which reads the project env **or the daemon env** (`packages/tracker-github/src/orchestrator-adapter.ts:56`).

So the split is:

| credential source | worker-exit publish (existing) | this path (new) |
|---|---|---|
| project `.env` | works | works |
| daemon env / `gh`-exported token / broker | works | **no credential → `buildHostGitEnvironment` installs no helper → push fails** |

The failure is quiet and misleading: on the endpoint the agent gets `git_transport_failed`, and on the teardown backstop below the failure string overwrites the real suppression reason in `lastError`. Both are exactly the class of false root-cause signal #826 was filed to eliminate.

Fix: resolve credentials the same way the worker spawn does and merge them into the env passed here, rather than relying on process-env inheritance.

---

packages/orchestrator/src/service.ts
```diff
@@ -1984,9 +2122,12 @@ export class OrchestratorService {
                 )
               : activeRun.runtimeSession,
             recovery,
-            lastError: recovery
-              ? "Run suppressed with recoverable incomplete-turn dirty workspace."
-              : "Run suppressed because the tracker issue is no longer tracked.",
+            unpublishedWorktree: publication.unpublishedWorktree,
```
hojinzs, line 2197: **P1 — a failed publish now erases the `unpublishedWorktree` diagnostic, which is the #647 symptom this PR is meant to prevent.**

`publishAssignedBranchForRun` returns `unpublishedWorktree: null` on the `outcome === "failed"` branch (`service.ts:588-600`). Assigning it unconditionally here means a *failed* publish clears the field.

Before this PR the suppression record inherited `activeRun.unpublishedWorktree` through the `...activeRun` spread, so a previously recorded diagnostic survived teardown. Now it does not.

Compose that with the credential gap above: in a daemon-credential deployment, publication fails on every suppression, and the run record is written with `unpublishedWorktree: null` — committed work stranded on disk with the one field that would have told the operator about it zeroed out. That is precisely the #647 report quoted on this PR (*"two real commits sat unpublished on disk with `unpublishedWorktree` reading `null`"*), reintroduced through the fix for it.

Guard the write, e.g.:

```ts
unpublishedWorktree:
  publication.outcome === "failed"
    ? activeRun.unpublishedWorktree
    : publication.unpublishedWorktree,
```

Same issue at `service.ts:2218` and inside `persistAssignedBranchPublishResult` (`service.ts:635`), which also writes `unpublishedWorktree: null` on the failure path.

Worth a regression test: *a suppression whose publication fails preserves the prior `unpublishedWorktree`*.

---

packages/orchestrator/src/service.ts
```diff
@@ -1953,6 +2090,7 @@ export class OrchestratorService {
           if (!activeRun || activeRun.processId === null) {
             continue;
           }
+          const publication = await this.publishAssignedBranchForRun(activeRun);
```
hojinzs, line 2165: **P2 — publication runs before the ownership check, and before any bound on how long it can take.**

Two problems with the placement, both at this site and at `service.ts:2180`:

**1. It runs ahead of the `"protected"` guard.** `signalRunProcess(...) === "protected"` means this daemon does not own the run — another process does. But we have already pushed that run's branch with this host's credential by the time we find out, and then `continue`. Move the publish below the guard.

**2. It is unbounded, inside `runSerialized`, ahead of `SIGTERM`.** `runGit` calls `execFileAsync` with `cwd`, `env`, `maxBuffer` and **no `timeout`** (`packages/worker/src/git-transport.ts:396`). `synchronizeAssignedBranch` does a network `ls-remote`, a conditional `fetch`, and a `push`. A hung remote or a wedged credential helper therefore stalls the serialized reconciliation loop indefinitely, and the run that reconciliation decided to cancel keeps executing for as long as the stall lasts. This is Codex's P2 on line 2180 and it holds here too.

Pass a `timeout` to `execFileAsync` in `runGit`, or wrap `publishAssignedBranchForRun` in an explicit deadline on the teardown paths, so cancellation can never be delayed by the network.

Minor, same site: the first path guards `activeRun.processId === null` before publishing (`:2090`), but the `:2180` path only guards `!activeRun`, so it will attempt publication for a run with no live process.

---

packages/worker/src/git-transport.ts
```diff
@@ -217,7 +217,8 @@ export function shouldSynchronizeAssignedBranch(options: {
   userInputRequired: boolean;
   terminalFailure: boolean;
 }): boolean {
-  return !options.userInputRequired && !options.terminalFailure;
+  void options;
```
hojinzs, line 224: **P2 — a predicate that ignores its arguments and always returns `true` is a regression waiting to happen.**

Both call sites still compute and pass the flags (`packages/worker/src/index.ts:853`, `:2029`), and `git-transport.test.ts:432` still exercises it. Nothing in the codebase now records *why* the condition was removed, so the shape reads like an accident and invites a future cleanup to "restore" the original `!userInputRequired && !terminalFailure` — which is #826 exactly, reintroduced silently.

Pick one:
- delete the function and its two call sites, publishing unconditionally at those points; or
- keep it and replace the `void options` with a comment stating that publication is intentionally unconditional (#826), plus a named test asserting that an abnormal end (`terminalFailure: true`) still publishes.

The second is the one I would take — a test named for the regression is what stops it coming back.

---

packages/worker/src/index.ts
```diff
@@ -887,6 +887,14 @@ async function runNonCodexRuntimeAdapterLifecycle(
       convergenceDetected: false,
       maxTurnsReached: false,
     });
+    recordGitTransportAttempt(
+      await trySynchronizeAssignedBranch({
+        cwd: env.WORKING_DIRECTORY!,
+        assignedBranch: env.SYMPHONY_ASSIGNED_BRANCH ?? "",
+        remoteUrl: env.TARGET_REPOSITORY_CLONE_URL ?? "",
+        env,
```
hojinzs, line 903: **P2 — three of the four orchestrator `SIGTERM` senders still have no publication backstop; the stuck-worker one is the likeliest to lose real commits.**

The new backstops sit on the *runtime-exit* path, and the orchestrator publishes before `SIGTERM` on the two **reconciliation suppression** paths (`service.ts:2093`, `:2180`). Those are genuinely covered — that part of the review thread above is not accurate, see my summary.

But the orchestrator sends `SIGTERM` from four places, and only those two publish:

| sender | publishes first? |
|---|---|
| `service.ts:2093` — reconciliation, issue no longer tracked | yes |
| `service.ts:2180` — reconciliation, terminal / unroutable | yes |
| `service.ts:1365` — `shutdown()`, daemon stop (`repo stop`, Ctrl-C) | **no** |
| `service.ts:3589` — superseded run in `restartRun` | **no** |
| `service.ts:3649` — **stuck-worker stall timeout** | **no** |

And the worker cannot compensate, because its own signal handler goes `kill child → runtimeAdapter.cancel() → persist token usage → process.exit(0)` (`packages/worker/src/index.ts:212-236`) with no transport call anywhere on it.

The stall-timeout case is the one I would not ship without: a worker that has committed several turns of work and then wedges gets `SIGTERM`ed by the stall detector, and every one of those commits stays on disk unpublished. That is a strictly worse outcome than the pre-PR behavior for a *stuck* run, because there is no clean exit to fall back to.

Publishing once inside the worker's `shutdown()` before `process.exit(0)` covers all five senders at a single point — bounded, so teardown cannot hang (which is the same deadline concern I raised on `service.ts:2093`). That is also the fix suggested in the thread on this PR, and I think it is the right one; it just needs to be motivated by the stall/daemon-stop paths rather than by the suppression paths.

If you judge the remaining three paths unreachable with committed work present, that reasoning is worth stating on the thread instead of leaving them silently uncovered.

---

packages/cli/src/commands/start.ts
```diff
@@ -837,6 +849,51 @@ async function startHttpServer(input: {
...
+                unpublishedWorktree: null,
+                error: "assigned_branch_publish_authentication_failed",
+              });
+              return;
+            }
+            const result = await input.service.requestAssignedBranchPublish({
+              runId,
+            });
+            respondJson(
+              response,
```
hojinzs, line 890: **nit — status mapping conflates two unrelated rejections.**

`result.outcome === "rejected"` maps to `403`, but `publishAssignedBranchForRun` uses `rejected` for two different things: `run_not_current` (genuinely an authorization/staleness answer, `403` is right) and `assigned_branch_unavailable` (detached HEAD — a workspace state problem, not a permission one). `409` or `422` would read better for the second and would let `/push` distinguish "you are not the current run" from "your worktree is detached" without string-matching `error`.

The authentication itself looks right to me, for the record: exempting the route from the general gate then requiring `x-symphony-run-id` plus a constant-time `secretsEqual` against `trackerStateToken`, with `requestAssignedBranchPublish` independently re-deriving the run from the store rather than trusting the supplied id, is the correct depth of check.

---

hojinzs (changes requested):

Verified at `6809dcdf` on a clean checkout. `pnpm install --frozen-lockfile`, `pnpm build`, `pnpm typecheck`, `pnpm lint` all exit 0; `pnpm test` green — orchestrator 398, cli 617, worker 227, 1242 across the touched packages. CI (`Test`, `Container Smoke`) is green.

### Smoke test

No preview environment, so I exercised the publication path locally against a real bare remote rather than a mock — `trySynchronizeAssignedBranch` composed exactly as `publishAssignedBranchForRun` composes it (`service.ts:567-587`):

| case | result |
|---|---|
| worktree on the assigned branch → publish | ✅ remote ref advances to local HEAD |
| second publish at the same HEAD | ✅ idempotent |
| agent on a different branch, **immutable** `SYMPHONY_ASSIGNED_BRANCH` supplied | ✅ `refusing to push: worktree is on main, expected assigned branch symphony/issue-3` |
| agent on a different branch, branch derived from the worktree **as this PR does** | ❌ **`main` published to the remote** |

The last row is the blocking finding; details inline on `service.ts:567`.

### Requirements vs #826

| requirement | status |
|---|---|
| agent can create its Draft PR within the run that produced the commits | ✅ via `POST /api/v1/assigned-branch/publish` |
| the unsatisfiable "3 turn boundaries then escalate" rule stops firing | ✅ removed from `WORKFLOW.md` Step 4 and the `/push` skill; the replacement wording ("a missing remote ref before a successful publish request is expected and is never, by itself, a blocker") addresses the false-attribution failure mode head-on |
| fast-forward-only and hook-free transport preserved | ✅ reused, not replaced |
| wrong-branch refusal preserved | ❌ **not on the new path** — see `service.ts:567` |
| teardown does not strand committed work | ⚠️ partial — suppression covered, three other `SIGTERM` senders are not; and a *failed* publish now erases `unpublishedWorktree` |

The issue offered three options and this PR takes a fourth — agent-triggered publication rather than automatic turn-boundary publication. I think that is the better choice and it is properly documented in both contract files, so no over-engineering concern; the scope is proportionate to #826 and I found no unrequested feature work in the diff.

### One correction to the prior thread

The review comment above states that *"the orchestrator diff shows only the new publish-action plumbing — no change to the suppression teardown"* and that *"there is nothing pinning ... 'a `SIGTERM` teardown publishes'"*. Both are inaccurate at `6809dcdf`: `service.ts:2093` and `:2180` each call `publishAssignedBranchForRun(activeRun)` immediately before `signalRunProcess(activeRun, "SIGTERM")`, and `service.test.ts:14401` asserts `publishAssignedBranch` was called alongside `killImpl(4209, "SIGTERM")` on that path.

The underlying concern is still valid, just aimed at the wrong path — `shutdown()` (`:1365`), `restartRun` (`:3589`) and the **stuck-worker stall timeout** (`:3649`) do send `SIGTERM` without publishing, and the worker's own signal handler does not compensate. The stall timeout is the one that worries me: a wedged worker with several turns of commits loses all of them. Detail inline on `worker/src/index.ts`.

### Findings

**P1**
1. `service.ts:567` — assigned-branch validation is tautological; host publishes any checked-out branch including `main`. Reproduced.
2. `service.ts:584` — daemon-level Git credentials are dropped (`INHERITED_ENV_ALLOWLIST` omits them; the worker spawn merges `resolveWorkerCredentials`, this path does not). The headline feature fails silently in that deployment shape.
3. `service.ts:2125` / `:2218` / `:635` — a *failed* publish overwrites `unpublishedWorktree` with `null`, erasing the diagnostic that survived teardown before this PR. Combined with (2) this recreates the #647 symptom.

**P2**

4. `service.ts:2093` / `:2180` — publication runs before the `"protected"` ownership guard, and is unbounded (`runGit` sets no `timeout`) inside the serialized loop ahead of `SIGTERM`.
5. `worker/src/index.ts` — three `SIGTERM` senders without a publication backstop.
6. `git-transport.ts:220` — `shouldSynchronizeAssignedBranch` is now `void options; return true` with live call sites and a live test; no regression test pins "an abnormal end publishes".

**P3 / nit**

7. `start.ts` — `403` for both `run_not_current` and `assigned_branch_unavailable`.
8. `CLAUDE.md` still carries the old package graph (`orchestrator ──→ core, runtime-claude, runtime-codex, tracker-*`). `docs/architecture.md` was updated with the new `worker` edge; the copy in `CLAUDE.md` was not.
9. Layering, non-blocking: Coordination (orchestrator) now imports Execution (worker) through a new `@gh-symphony/worker/git-transport` subpath, which is what forced the `packages/cli/vitest.config.ts` alias. The host Git transport is host-owned infrastructure and would sit more naturally in `core` or a small shared package. Worth an explicit note in the PR body as a deliberate divergence if you keep it as-is.

### To move to approve

(1), (2) and (3) are the blockers — each one either breaks a stated invariant or defeats the purpose of the fix. (4) and (5) I would also want in this PR since they are teardown-ordering issues in code this PR is already rewriting, but I will not hold the PR on (6)–(9) if you would rather split them.

---

packages/orchestrator/src/service.ts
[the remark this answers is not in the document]
moncher-dev: Fixed in `ce406f5`. Host publication now resolves worker credentials from both project and daemon environments before building the Git environment. The regression verifies daemon-only credentials reach the transport.

packages/orchestrator/src/service.ts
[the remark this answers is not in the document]
moncher-dev: Fixed in `d341416`. The branch captured before worker launch is persisted on the run record and is the only accepted publication target; the regression moves the live checkout to `main` and verifies publication still targets/refuses against the immutable assignment.

packages/orchestrator/src/service.ts
```diff
@@ -2036,6 +2177,7 @@ export class OrchestratorService {
           continue;
         }
 
+        const publication = await this.publishAssignedBranchForRun(activeRun);
```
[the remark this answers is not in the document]
moncher-dev: Fixed in `ec4bdac`. Assigned-branch publication is bounded by a 10-second deadline before termination, with an injected short deadline regression proving teardown proceeds when transport never settles.

packages/orchestrator/src/service.ts
moncher-dev: Fixed in `d341416`. `OrchestratorRunRecord.assignedBranch` persists the pre-launch assignment, and publication refuses when it is absent instead of deriving a target from the child-controlled checkout. The wrong-checkout regression exercises this invariant.

packages/orchestrator/src/service.ts
moncher-dev: Fixed in `ce406f5`. Publication uses the same `resolveWorkerCredentials` project/daemon sources as worker startup, and a regression asserts daemon-only `GITHUB_GRAPHQL_TOKEN` propagation.

packages/orchestrator/src/service.ts
moncher-dev: Fixed in `ce406f5`. Failed publication preserves the existing `unpublishedWorktree` record both in the returned result and persisted run state. The regression starts with a dirty-worktree diagnostic and verifies it survives transport failure.

packages/orchestrator/src/service.ts
moncher-dev: Fixed across `d341416` and `ec4bdac`. Publication now follows the live ownership guard, requires the immutable assignment, and is deadline-bounded. Shutdown, stall, and superseded-run tests assert publication occurs before `SIGTERM`; runs without a live process are not published from reconciliation.

packages/worker/src/git-transport.ts
moncher-dev: Fixed in `b7c1db5`. The predicate now documents why publication is intentionally unconditional for #826, and the parameterized test explicitly names and asserts normal, terminal-failure, and input-required exits.

packages/worker/src/index.ts
moncher-dev: Fixed in `ec4bdac`. Daemon shutdown, superseded restart, and stall-timeout paths now invoke bounded host publication before `SIGTERM`. Focused regressions assert ordering for shutdown and stalled-worker termination, while the existing supersession path uses the same bounded helper.

packages/cli/src/commands/start.ts
moncher-dev: Fixed in `b7c1db5`. `run_not_current` remains 403, while other rejected workspace-state outcomes such as `assigned_branch_unavailable` return 409. The HTTP route test pins both mappings.

packages/orchestrator/src/service.ts
```diff
@@ -505,6 +516,156 @@ export class OrchestratorService {
...
+  ): Promise<void> {
+    const latest =
+      (await this.store.loadRun(run.runId, this.projectConfig.projectId)) ??
+      run;
+    const now = this.now().toISOString();
+    await this.store.saveRun({
+      ...latest,
+      ...result,
+      updatedAt: now,
+      lastEventAt: now,
```
hojinzs, line 666: **P3 — non-blocking, follow-up sized. This stamps the stall watchdog's activity clock, and the clock is now refreshable by a child-initiated request.**

`lastEventAt` is the sole input to stall detection (`service.ts:3667`):

```ts
const lastActivityAtMs = parseTimestampMs(run.lastEventAt ?? run.startedAt);
```

`persistAssignedBranchPublishResult` writes `lastEventAt: now` on every publication — success *and* failure — and `requestAssignedBranchPublish` reaches it from the agent-callable endpoint. So an agent calling `/push` refreshes the host's stuck-worker timer.

I do not think this is a bug as such: an agent that calls `/push` is by definition alive, so treating it as activity is defensible, and the two host-initiated teardown callers `SIGTERM` immediately afterward so the stamp is moot there. Two smaller things do follow from it:

- `lastEventAtSource` is left unset on this write, while the worker event channel sets `"event-channel"` (`:4351`, `:4451`). The operator surface exposes `last_event_at` (`:1320`) with no way to tell an agent event from host publication bookkeeping. Setting a source here would keep that distinguishable.
- The stall path (`:3703`) now publishes before `SIGTERM`, which stamps the clock on the run it is about to terminate — visible in the updated expectation `lastEventAt` `undefined` → `"2026-03-08T00:09:00.000Z"` in the stall test. Harmless today; worth knowing it is a behavior change to that record, not just to ordering.

If publication is meant to be host bookkeeping rather than run activity, writing only `updatedAt` here and leaving `lastEventAt` to the event channel would say so more precisely.

**nit, same function:** the success path passes `lastError: run.lastError ?? null` from the pre-publish snapshot, but the function re-loads `latest` from the store before writing. `{...latest, ...result}` then puts the stale value back, discarding any `lastError` written during the publish window. `lastError: latest.lastError ?? null` — or just omitting the key, as the failure path effectively does for `unpublishedWorktree` — avoids it.

---

hojinzs (approved):
## Review 2/6 — Approve

Verified at `b7c1db5`. All nine findings from review 1/6 are genuinely fixed in code, not just in the replies — I checked each against the diff rather than taking the thread at its word, and re-ran the reproduction that produced the blocker last round.

### Verification

`pnpm build`, `pnpm typecheck`, `pnpm lint`, `pnpm test` all exit 0 on a clean checkout (orchestrator 15 files, cli 43, worker 27). CI (`Test`, `Container Smoke`) green on this head.

### Smoke test — real remote, real transport, no mocks

Last round I reproduced a `main` publish through this path. I re-ran the same harness against `OrchestratorService.requestAssignedBranchPublish` end-to-end this time — real bare remote, real `trySynchronizeAssignedBranch`, no injected `publishAssignedBranch`:

| case | result |
|---|---|
| worktree on the assigned branch → publish | ✅ `refs/heads/symphony/issue-9` advances to local HEAD |
| second publish at the same HEAD | ✅ idempotent, ref unchanged |
| **agent runs `git checkout main`, commits, calls publish** | ✅ **refused**; `refs/heads/main` unchanged, assigned ref never created |
| run record with no persisted assignment (legacy in-flight run) | ✅ `rejected` / `assigned_branch_unavailable`, nothing pushed |
| divergent local history → non-fast-forward | ✅ refused, published ref unchanged |

Row 3 is the round-1 blocker and it is closed. The fix is the right shape: `assignedBranch` is captured once pre-launch (`:3212`), persisted on the run record (`:3337`), never rewritten anywhere in the orchestrator, and publication refuses outright when it is absent rather than falling back to the checkout. The regression asserting `readGitCurrentBranch` was **not** called is a better pin than an equality assertion would have been — it forecloses the re-derivation rather than just its current output.

### The other blockers

- **(2) credentials** — `resolveWorkerCredentials(projectConfig, { project, daemon: process.env })` now matches worker spawn exactly, with a daemon-only `GITHUB_GRAPHQL_TOKEN` asserted through to the transport env.
- **(3) diagnostics** — failure returns `run.unpublishedWorktree ?? null`, and the failure-path persist omits the key entirely so `{...latest, ...result}` cannot clobber it. The regression drives it through a real timeout with a pre-existing dirty-worktree record. The #647 symptom is not reintroduced.
- **(4) ordering/bound** — `isRunProtectedByLiveOwner` now gates all five sites before publication, and a 10s `Promise.race` deadline bounds it. I checked the residual window between that pre-check and `signalRunProcess`'s re-check: they evaluate the same predicate over the same in-memory record, so it is negligible.
- **(5) SIGTERM coverage** — all five senders publish first now (`:1400` shutdown, `:2133`/`:2224` reconciliation, `:3638` supersession, `:3703` stall), with `invocationCallOrder` assertions pinning publish-before-kill on the shutdown and stall paths. You covered this host-side rather than in the worker's signal handler as I suggested; that is the better call — the credential stays host-owned and it does not depend on the worker being responsive, which is exactly the stall case I was worried about.
- **(6)–(8)** — predicate documented with a `#826`-named parameterized test, `409` vs `403` split with both mappings pinned, `CLAUDE.md` graph corrected.

### Two things I checked and withdrew

Worth recording so they don't get re-raised:

- I expected the timed-out `publish()` promise to leave an orphaned `git` child holding the event loop open, defeating the deadline on `shutdown()`. It does not — `shutdownForegroundOrchestrator` ends in `process.exit()` (`start.ts:1490`), so the daemon exits regardless.
- I expected concurrent publications (agent `/push` racing a teardown publish; the publish runs outside the `runSerialized` block) to contend on the worktree's index/ref locks. They cannot — `synchronizeAssignedBranch` fetches into a fresh temporary bare repo per call and pushes from there, so the worktree is only ever read. That is a good design choice and it makes the concurrency question moot.

### Requirements vs #826

| requirement | status |
|---|---|
| agent creates its Draft PR within the run that produced the commits | ✅ |
| the unsatisfiable "3 turn boundaries then escalate" rule stops firing | ✅ |
| fast-forward-only and hook-free transport preserved | ✅ reused |
| wrong-branch refusal preserved | ✅ now actually enforced, reproduced above |
| teardown does not strand committed work | ✅ all five senders |

Scope is still proportionate — the four fix commits touch only what the findings named, and I found no unrequested feature work in the delta. Finding (9) (Coordination importing Execution via the `@gh-symphony/worker/git-transport` subpath) stands as a deliberate divergence; you documented the new edge in both `docs/architecture.md` and `CLAUDE.md`, which is enough for me. It was non-blocking last round and I am not holding on it.

### Remaining

One P3 plus a nit inline on `service.ts:665`, both follow-up sized. Neither is worth another round — approving.

---

Issue #853, opened 2026-09-04
fix(e2e): resolve Docker Compose from isolated agent homes


moncher-dev:
## Problem

Docker E2E cannot start in orchestrated coding-agent children when Docker Compose is installed only as a user-level CLI plugin. The child uses an isolated `HOME`, and `DOCKER_CONFIG` is unset, so `docker compose` is unavailable even though the host may have the plugin installed under the operator home.

Observed while validating #812. Both required isolated retries failed before scenario startup with:

```text
Docker runtime unavailable: 'docker compose' cannot be resolved by this process.
HOME=.../child-home, DOCKER_CONFIG=<unset>
docker: unknown command: docker compose
```

## Expected

Repository E2E validation should either discover an approved host Compose plugin without weakening agent credential isolation, or document/provide a supported Docker configuration for isolated agent homes.

moncher-dev:
Duplicate of #825 — same defect, and #825's fix is already decided and in progress. Closing rather than opening a parallel track.

Your diagnosis is exactly right and matches #825's: the Compose plugin is a **user-level** CLI plugin under the operator's `$HOME/.docker/cli-plugins/`, the agent child runs with an isolated `HOME`, and `DOCKER_CONFIG` is unset in that environment — so `docker compose` cannot resolve even though the host has it installed. `HOME=/nonexistent docker compose version` reproduces it exactly.

Your "Expected" is also precisely the decision @hojinzs already made on #825: **stage the Docker CLI plugin directory into the isolated child home**, with three constraints worth knowing since they rule out the obvious shortcuts:

1. `DOCKER_CONFIG` must **stay out** of `SAFE_RUNTIME_ENV_KEYS`. The value reaching the child has to be host-constructed and set explicitly, the way `HOME`, `GH_CONFIG_DIR` and `CODEX_HOME` already are.
2. Nothing from the operator's `~/.docker` beyond plugin resolution may be staged — no `config.json` auths, no credential helper or store configuration. That is the child boundary #810 and #813 established.
3. On a host with no plugin directory, behavior must be unchanged, with #840's preflight message still appearing.

So "discover an approved host Compose plugin without weakening agent credential isolation" is the work, and it is underway.

**One thing your report is genuinely useful for, and it is worth recording.** The output you captured —

```
Docker runtime unavailable: 'docker compose' cannot be resolved by this process.
HOME=.../child-home, DOCKER_CONFIG=<unset>
docker: unknown command: docker compose
```

[...]

Issue #850, opened 2026-09-04
refactor(orchestrator)!: remove repo-embedded mode and the clone populate strategy


moncher-dev:
Part of #847, phase 2, second of three. Depends on the command surface removal.

## Change

Remove repo-embedded mode from the orchestrator and its configuration.

- Delete the `clone` populate strategy in `packages/orchestrator/src/git.ts` and make `worktree-cache` the only path. Standalone mode already hardcodes it at `packages/cli/src/commands/project.ts:89`, so this removes the branch rather than changing behavior.
- Remove `populateStrategy` from the project config contract and its normalization in `packages/core/src/contracts/status-surface.ts`, along with `repositoryDir` and `workflowSource.type: "repo"`.
- Remove the legacy repo-project guards added for #790/#795 and the `activeProject` fallback in project selection.
- Migrate on read: a stored config carrying the removed fields loads with them ignored rather than failing, so an operator upgrading does not have to hand-edit state.

## Acceptance criteria

- One populate strategy remains and no configuration selects between strategies.
- A tenant config written by the previous version still loads.
- Build, lint, typecheck and tests pass; repo-mode branches inside shared test suites are removed rather than skipped.
- A major changeset.

## Notes

#811 and its PR #828 exist to make the `clone` strategy honor the branch template. If this lands first, that work is deleted rather than merged. Confirm and close it out rather than leaving a stale PR.

## Carried over from the closed #811 and PR #828

The `clone` strategy's `checkout -B` can destroy unpushed agent commits when a workspace is reused. PR #828 was closed as moot because this issue deletes that strategy, so the fix travels here as a verification step rather than a code change:

- Confirm that the deleted `clone` path was the only site where reusing a workspace could discard unpublished commits.
- Confirm that the surviving `worktree-cache` path does not share the defect. If it does, fix it in this PR and say so in the description.
- `docs/architecture.md:105` records a deliberate divergence about not destructively resetting reused workspaces; make sure the surviving path still satisfies it.

moncher-dev:
🚀 Promoting to Ready — **step 2 of the removal chain.** #849 landed and is running: the daemon on this host was rebuilt, restarted, and verified with a real command line of `project start --project-dir …`. Conditions are met.

[...]

Issue #844, opened 2026-09-03
refactor(core): declare the process environment contract once instead of five divergent allowlists


moncher-dev:
Revised 2026-09-04 after an independent review. The first version of this issue blamed inheritance-list drift; that framing was wrong on two counts and the real duplication is on the exclusion side. Details and corrections are in the first comment.

## Problem

The credential exclusion list is written three times, the three copies disagree, and none of them is the declared source.

| Copy | Location | Notes |
| --- | --- | --- |
| core | `packages/core/src/runtime/custom-child-env.ts:16-30` | 13 names, **not exported**; `removeChildHostCredentialEnvironment` at `:99` is private |
| Codex | `packages/runtime-codex/src/runtime.ts:790-818` | re-declares the same Git and SSH names inline |
| Claude | `packages/runtime-claude/src/adapter.ts:706-732` | 23 names, a superset, and contains `GIT_CONFIG_COUNT` twice |

The Claude copy strips `AGENT_CREDENTIAL_BROKER_URL`, `AGENT_CREDENTIAL_BROKER_SECRET`, `AGENT_CREDENTIAL_CACHE_PATH`, `GITHUB_GIT_HOST`, `GITHUB_GIT_USERNAME`, `GITHUB_TOKEN_BROKER_URL` and `GITHUB_TOKEN_CACHE_PATH`. The core copy does not. A deny list is only safe when it is complete, and this one is retyped rather than shared, so completeness holds by coincidence.

The same pattern repeats for tracker secrets. Codex derives them from `SYMPHONY_TRACKER_SECRET_ENVIRONMENT_NAMES` (`packages/runtime-codex/src/launcher.ts:64-70`), which is the declared path. The Claude adapter takes the declared names and then also hardcodes `GH_TOKEN`, `GH_ENTERPRISE_TOKEN`, `GITHUB_TOKEN`, `GITHUB_GRAPHQL_TOKEN`, `GITHUB_TOKEN_BROKER_SECRET`, `LINEAR_API_KEY` and `LINEAR_AUTHORIZATION` (`adapter.ts:694-700`). The GitHub adapter's own `secretEnvironmentNames()` (`packages/tracker-github/src/orchestrator-adapter.ts:44-52`) omits the broker URL and cache path that #827 now injects, so the declaration and the hardcoded lists describe different sets.

[...]

PR #854, opened 2026-09-04
fix(runtimes): forward agent-visible Symphony context

### LINKED ISSUE #812: fix(runtimes): forward SYMPHONY_ASSIGNED_BRANCH and agent-visible Symphony context into the agent child

moncher-dev:
## Issues

- Closes #812

## Summary

- Defines one core allowlist for non-secret Symphony run context exposed to every supported agent child.
- Forwards the assigned branch and issue/repository context consistently through Codex, Claude, and custom runtimes while stripping tracker secrets and reserved broker/auth variables, including process-only declarations in compatibility inheritance mode.
- Uses one core clone-URL sanitizer for both persisted Git remotes and agent-visible context, and documents/tests the new context and credential-isolation contract.

## Change-point diagram

- Worker/config/process environment → core context allowlist + canonical clone-URL sanitizer → Codex / Claude / custom child environment
- Tracker secret declarations + shared reserved-auth names → final stripping pass → absent from every agent child

## Start here

- `packages/core/src/runtime/custom-child-env.ts:16` — shared context allowlist, ordered decision sources, URL sanitization, and secret boundary
- `packages/runtime-codex/src/runtime.ts:635` — Codex runtime-plan composition
- `packages/runtime-claude/src/adapter.ts:642` — Claude composition using the shared reserved-auth declaration
- `test/e2e/claude/claude-docker.spec.ts:80` — real-child isolation and compatibility-mode credential contract

- Coding-agent shells receive `SYMPHONY_ASSIGNED_BRANCH` and the documented non-secret issue/repository context across supported runtimes.
- Custom runtimes using `inherit_environment: true` no longer inherit tracker or reserved broker/auth credentials, even when their declarations exist only in the worker process environment; explicitly configured runtime authentication remains available.

## Validation

- `pnpm lint && pnpm test && pnpm typecheck && pnpm build` — pass on published head `a11df9d`
- `pnpm test:worker-startup` — pass (9 tests, including real custom children in both isolation modes)
- `pnpm --filter @gh-symphony/core build && pnpm exec vitest run packages/core/src/runtime/custom-child-env.test.ts packages/orchestrator/src/git.test.ts` — pass (45 tests)
- Docker happy-path preflight remains unavailable after two attempts under documented environment issue #853; deterministic child composition and the full worker-startup integration suite pass locally.

## Changeset

- `.changeset/quiet-ducks-share.md` (`@gh-symphony/cli` patch)

## Risks & rollback

- Operator custom commands that relied on inherited tracker credentials must move authentication to `runtime.auth.env`; this migration is documented in the configuration guide and changeset.
- Roll back this PR as one unit if child-context propagation or compatibility credential stripping causes an unexpected runtime regression.

## Changed files

- `packages/core/src/runtime/custom-child-env.ts` and tests — define the allowlist, inspect all compatibility decision sources, sanitize URL context, and enforce secret stripping.
- `packages/orchestrator/src/repository-url.ts` — preserve the orchestrator import path while re-exporting the canonical core sanitizer.
- `packages/runtime-codex/src/runtime.ts`, `launcher.ts`, and tests — propagate context through Codex launch paths.
- `packages/runtime-claude/src/adapter.ts` and worker tests — apply shared context and reserved-auth contracts to Claude/custom runtimes.
- `test/e2e/claude/claude-docker.spec.ts` — verify real custom children strip secrets in default and compatibility modes.
- `docs/configuration.md`, `docs/architecture.md`, `WORKFLOW.md`, and `AGENT_TEST.md` — document the contract and test scenario.
- `.changeset/quiet-ducks-share.md` — patch release note for `@gh-symphony/cli`.

## Security


moncher-dev:
Smoke-tested at `585754af`. `pnpm install --frozen-lockfile`, `pnpm build`, `pnpm typecheck`, `pnpm lint` all exit 0; suites green across the four touched packages — core 338, runtime-codex 78, runtime-claude 108, worker 241 (**765 tests**).

This is epic #808's last child, and the variable at its centre — `SYMPHONY_ASSIGNED_BRANCH` — is the one whose absence produced several confidently wrong root-cause claims this session (#811's run blamed a missing `SYMPHONY_ASSIGNED_BRANCH` for what was actually the #826 publish-timing bug). So I verified it **in a real spawned subprocess** rather than by unit assertion:

```
--- env observed INSIDE the spawned child ---
  SYMPHONY_ASSIGNED_BRANCH   = symphony/hojinzs-github-symphony-812
  SYMPHONY_ISSUE_IDENTIFIER  = hojinzs/github-symphony#812
  SYMPHONY_TRACKER_KIND      = github-project
  TARGET_REPOSITORY_OWNER    = hojinzs
  GITHUB_GRAPHQL_TOKEN       = <absent>
  GITHUB_TOKEN_BROKER_SECRET = <absent>
  GITHUB_TOKEN_CACHE_PATH    = <absent>
  LINEAR_API_KEY             = <absent>
```

That is `buildCustomRuntimeChildEnvironment` composing the env, then an actual `spawnSync` reading it back from inside the child. The context arrives; every secret I seeded is stripped — including `GITHUB_TOKEN_CACHE_PATH`, which has been argued about twice on this board and is now provably absent at this boundary rather than absent by inference.

The Codex and Claude paths route through the same helper (`readAgentVisibleSymphonyContext` at `runtime.ts`, `launcher.ts`, and `adapter.ts`), and it is source-safe rather than filter-safe — it iterates a fixed allowlist and copies only those names, so a secret cannot pass even if the source env is dirty:

```
readAgentVisibleSymphonyContext(dirtyWorkerEnv)
  keys: SYMPHONY_ASSIGNED_BRANCH, SYMPHONY_ISSUE_IDENTIFIER, SYMPHONY_TRACKER_KIND, TARGET_REPOSITORY_OWNER
  secret-shaped keys present: NONE
```

**On design:** one allowlist in core (`AGENT_VISIBLE_SYMPHONY_CONTEXT_ENVIRONMENT_NAMES`) consumed by all three runtimes is the right shape, and it is the opposite of the failure mode this repo has hit repeatedly — `secretEnvironmentNames()` disagreeing with `CUSTOM_RUNTIME_RESERVED_AUTH_ENVIRONMENT_NAMES`, and the two `GIT_CONFIG_*` contracts in #832. A single declaration cannot drift against itself. Worth noting it also anticipates #844, which is held pending #850.

No over-engineering: nine names, all non-secret and all justified by #812's scope, with the docs and a changeset alongside.

I did not exercise a full agent run end-to-end — that needs a live dispatch. The subprocess check above covers the boundary the issue is actually about. @hojinzs over to you.

WORKFLOW.md
```diff
@@ -94,6 +94,7 @@ These facts describe what your process can and cannot do. Every instruction belo
...
 4. **The host pushes your branch after every turn, fast-forward only.** Commit locally; at the end of each turn the worker pushes `$SYMPHONY_ASSIGNED_BRANCH` to the orchestrator-owned remote. The push is refused if `origin/<branch>` is not an ancestor of your head — so **never rebase, amend, reset, or force-rewrite commits that may already be on the remote**. Bring in base-branch changes with `git merge` (see `/pull`). A branch that exists on the remote is the precondition for creating a PR, so PR creation happens in the turn **after** the first commit.
+   Codex, Claude, and custom agent children all receive the same non-secret Symphony context allowlist: `SYMPHONY_ASSIGNED_BRANCH`, `SYMPHONY_ISSUE_ID`, `SYMPHONY_ISSUE_IDENTIFIER`, `SYMPHONY_ISSUE_STATE`, `SYMPHONY_TRACKER_KIND`, and `TARGET_REPOSITORY_*`. Tracker secrets and reserved broker/authentication variables are never part of this context.
```
hojinzs, line 98: **P1 — merge conflict, and it is semantic rather than textual.**

The PR is currently `mergeable_state: dirty`. `main` has moved past the merge commit `585754a`, and rule 4 is the conflicting hunk. This paragraph is attached to the *old* rule 4:

> 4. **The host pushes your branch after every turn, fast-forward only.** … at the end of each turn the worker pushes `$SYMPHONY_ASSIGNED_BRANCH` …

`main` has since replaced that rule wholesale:

> 4. **Branch publication is agent-triggered and host-owned, fast-forward only.** After committing work that must become visible, call `/push`, which sends an authenticated `POST /api/v1/assigned-branch/publish` …

So a mechanical conflict resolution will produce a document that describes both a per-turn automatic push and an agent-triggered publish. When you merge `main`, please re-attach this context sentence to the rewritten rule 4 (or to rule 3, which is where the `$SYMPHONY_ASSIGNED_BRANCH` contract now lives) rather than keeping it next to the superseded text.

`main` also rewrote the `SYMPHONY_ASSIGNED_BRANCH` and `SYMPHONY_ORCHESTRATOR_*` rows in `docs/configuration.md` in the same area — worth re-reading that table after the merge so the new "Agent-visible Symphony context" section stays consistent with the publish-endpoint wording.

---

packages/core/src/runtime/custom-child-env.ts
```diff
@@ -89,9 +118,21 @@ export function buildCustomRuntimeChildEnvironment(options: {
...
+  for (const name of Object.keys(env)) {
+    if (
+      isCustomRuntimeReservedAuthEnvironmentName(name, {
+        ...source,
+        ...input,
+      })
+    ) {
+      delete env[name];
+    }
+  }
```
hojinzs, line 142: **P2 — this is a user-visible behavior change for `inherit_environment: true`, and neither the changeset nor the docs mention it.**

This loop is new and unconditional, so it also applies to the compatibility inheritance path. The test change on this PR records the effect precisely:

```diff
-      GITHUB_TOKEN: "github-secret",
+    expect(env.GITHUB_TOKEN).toBeUndefined();
```

Before this PR, a custom runtime with `runtime.isolation.inherit_environment: true` inherited `GITHUB_TOKEN` / `GH_TOKEN` / `LINEAR_API_KEY` from the operator process. After it, those are stripped. I think that is the *right* change — it is exactly issue #812's requirement 2, and `inherit_environment` opting out of isolation should not silently mean opting out of credential stripping. But an operator whose custom agent authenticates with an inherited `GITHUB_TOKEN` will find it broken on upgrade with no diagnostic, and the changeset only advertises the additive half:

> Forward the assigned branch and non-secret Symphony issue and repository context consistently to Codex, Claude, and custom agent children (#812).

Please add the removal half to `.changeset/quiet-ducks-share.md`, and a sentence to the new `docs/configuration.md` section. The current docs wording —

> Tracker-declared secret environment names and reserved broker/authentication variables are removed after environment composition, even if a runtime enables process-environment compatibility inheritance.

— describes the end state but does not flag that it is a *change*, so it will not help someone diagnosing a suddenly-unauthenticated custom agent.

---

packages/runtime-claude/src/adapter.ts
```diff
@@ -646,11 +647,17 @@ function buildClaudeSpawnEnv(options: {
   inputEnv?: NodeJS.ProcessEnv;
   childHome: string;
 }): NodeJS.ProcessEnv {
+  const agentVisibleContext = readAgentVisibleSymphonyContext(
```
hojinzs, line 651: **P2 — the visible side is now one shared declaration; the excluded side still is not, and this file is the one that did not get wired up.**

The design argument in the PR description is that a single constant in core cannot drift against itself, and I agree with it. But it was applied asymmetrically. `runtime-codex/src/runtime.ts` already imports `CUSTOM_RUNTIME_RESERVED_AUTH_ENVIRONMENT_NAMES` from core for its final strip. This file instead keeps two local hardcoded lists — `stripTrackerSecrets` (7 names) and `removeChildHostCredentialEnvironment` (its own `AGENT_CREDENTIAL_*` / `GITHUB_TOKEN_BROKER_URL` / `GITHUB_TOKEN_CACHE_PATH` entries).

I verified by execution that today the union of those two lists happens to cover core's 12-name reserved list, so there is **no live leak** — see the spawned-child evidence in my summary comment. The problem is that the coverage is coincidental, not structural. Adding a name to `CUSTOM_RUNTIME_RESERVED_AUTH_ENVIRONMENT_NAMES` will propagate to Codex and to `buildCustomRuntimeChildEnvironment` automatically, and silently skip Claude.

This is the same failure mode #832 hit with the two `GIT_CONFIG_*` contracts. Since this PR is the one establishing the shared-declaration pattern, folding `...CUSTOM_RUNTIME_RESERVED_AUTH_ENVIRONMENT_NAMES` into `stripTrackerSecrets` here is a small change that makes the pattern actually hold. If you would rather keep it out of scope, a test asserting `stripTrackerSecrets` covers every name in the core constant would be enough to catch the drift.

---

packages/runtime-codex/src/runtime.ts
```diff
@@ -640,6 +641,7 @@ export function buildCodexRuntimePlan(
     args: agentCommand.args,
     env: {
       ...resolveRuntimeProcessEnv(),
+      ...readAgentVisibleSymphonyContext(process.env, config.extraEnv),
```
hojinzs: **P3 — the `config.extraEnv` argument here is dead precedence.**

`readAgentVisibleSymphonyContext` resolves last-source-wins, so passing `config.extraEnv` makes it beat `process.env` for the nine allowlisted names. But `...config.extraEnv` is spread on the very next line and wins anyway. The argument therefore cannot change the result — `readAgentVisibleSymphonyContext(process.env)` alone is equivalent.

Not a bug, but it reads as though the precedence matters, which invites someone to later reorder these two lines "safely". Either drop the second argument, or drop the separate `...config.extraEnv` spread and let the helper own the merge.

---

packages/core/src/runtime/custom-child-env.ts
```diff
@@ -13,6 +13,35 @@ const PORTABLE_ENVIRONMENT_NAMES = [
...
+  const context: NodeJS.ProcessEnv = {};
+  for (const name of AGENT_VISIBLE_SYMPHONY_CONTEXT_ENVIRONMENT_NAMES) {
+    for (let index = sources.length - 1; index >= 0; index -= 1) {
+      const value = sources[index]?.[name];
+      if (value !== undefined) {
+        context[name] = value;
+        break;
+      }
+    }
+  }
```
hojinzs: **nit — the backwards inner loop is correct but reads as a bug at a glance.**

Iterating `sources` from the end and breaking on the first defined value implements last-wins. A forward loop that simply overwrites expresses the same thing without the reader having to reconstruct the precedence:

```ts
for (const source of sources) {
  for (const name of AGENT_VISIBLE_SYMPHONY_CONTEXT_ENVIRONMENT_NAMES) {
    const value = source?.[name];
    if (value !== undefined) context[name] = value;
  }
}
```

Same behavior, and the "later sources win" rule is then visible in the loop order rather than in the index arithmetic. Worth a one-line doc comment stating the precedence either way, since all three call sites depend on it.

---

packages/core/src/runtime/custom-child-env.ts
```diff
@@ -89,9 +118,21 @@ export function buildCustomRuntimeChildEnvironment(options: {
     }
   }
 
+  Object.assign(env, readAgentVisibleSymphonyContext(source, input));
```
hojinzs: **P3 — the new strip loop can now delete the deliberately staged `authEnvKey`, and the load-time guard that prevents it uses a different input.**

Twelve lines above, the non-inherit path explicitly stages `env[options.authEnvKey] = authValue` for the operator's configured `runtime.auth.env`. The new loop below then deletes any key matching `isCustomRuntimeReservedAuthEnvironmentName(name, { ...source, ...input })`, which includes tracker secret names read from `SYMPHONY_TRACKER_SECRET_ENVIRONMENT_NAMES`.

`parser.ts:1101` does reject a reserved `runtime.auth.env` at workflow-load time, so the common case is covered. But the two checks read different sources: the parser uses `env` plus `trackerAdapter.secretEnvironmentNames()` at parse time, while this uses the JSON name list present in the worker environment at spawn time. If those sets ever diverge, the custom agent's auth variable is dropped silently and the failure surfaces as an unauthenticated agent rather than a config error.

Cheap fix: skip `options.authEnvKey` in this loop (the parser already guarantees it is not reserved), or throw/log when the loop is about to delete it. Given the parser guard I would not block on this, but a silent delete of an explicitly configured credential is worth one line of defense.

---

hojinzs (changes requested):
## Review — round 1 of 6

Reviewed at `585754af`. The implementation is correct and I could not find a security hole in it. One merge blocker and one documentation gap keep this from an approve; both are small.

### Requirements coverage against #812

| #812 fix outline | Status |
| --- | --- |
| 1. One documented allowlist in core, applied in `runtime-codex`, `runtime-claude`, `worker/non-codex-runtime` | Met — `AGENT_VISIBLE_SYMPHONY_CONTEXT_ENVIRONMENT_NAMES`, nine names, exactly the set the issue names |
| 2. Every `secretEnvironmentNames()` / `CUSTOM_RUNTIME_RESERVED_AUTH_ENVIRONMENT_NAMES` name stays excluded | Met in behavior; see the P2 on `adapter.ts` for how it is achieved |
| 3. `docs/configuration.md` + WORKFLOW.md updated | Met, but the WORKFLOW.md paragraph lands on text `main` has since rewritten (P1) |
| Regression tests: Codex/Claude/custom child contains `SYMPHONY_ASSIGNED_BRANCH`; secrets absent | Met at all three boundaries |

I also confirmed the fix is not a no-op: `resolveRuntimeProcessEnv()` filters to `SAFE_RUNTIME_ENV_KEYS`, which does not contain any `SYMPHONY_*` name, so before this change the Codex child genuinely could not see the assigned branch unless it arrived via `extraEnv`. And the orchestrator does populate all nine names (`service.ts:3267-3289`), so none of the allowlist entries are speculative.

### Smoke test

I did not repeat the author's `buildCustomRuntimeChildEnvironment` check. I ran the two boundaries it did not cover — a **real spawned child** through `createWorkerNonCodexRuntimeAdapter` for `claude-print` and for `custom`, and `buildCodexRuntimePlan` — seeding the worker env with the full credential set the orchestrator actually injects, including a tracker-declared secret via `SYMPHONY_TRACKER_SECRET_ENVIRONMENT_NAMES`.

```
===== runtime.kind = claude-print =====      ===== runtime.kind = custom =====
  SYMPHONY_ASSIGNED_BRANCH    = symphony/hojinzs-github-symphony-812
  SYMPHONY_ISSUE_ID           = I_kwDO812
  SYMPHONY_ISSUE_IDENTIFIER   = hojinzs/github-symphony#812
  SYMPHONY_ISSUE_STATE        = In progress
  SYMPHONY_TRACKER_KIND       = github-project
  TARGET_REPOSITORY_{OWNER,NAME,URL,CLONE_URL} = all present
-- must be absent --
  GITHUB_GRAPHQL_TOKEN  GITHUB_TOKEN  GH_TOKEN  LINEAR_API_KEY
  GITHUB_TOKEN_BROKER_URL  GITHUB_TOKEN_BROKER_SECRET  GITHUB_TOKEN_CACHE_PATH
  AGENT_CREDENTIAL_BROKER_URL  AGENT_CREDENTIAL_BROKER_SECRET  AGENT_CREDENTIAL_CACHE_PATH
                                             -> all absent, both runtimes
```

`buildCodexRuntimePlan` gives the same result. So the contract holds identically across all three runtimes, which is the thing #812 asked for.

Build, typecheck, lint clean; 765 tests green across core (338), runtime-codex (78), runtime-claude (108), worker (241) — matching the author's numbers.

### Findings

| Sev | Where | What |
| --- | --- | --- |
| **P1** | `WORKFLOW.md` | Merge conflict with `main`, and it is semantic — `main` rewrote rule 4 into the agent-triggered publish model, so a mechanical resolution leaves the doc describing two contradictory push mechanisms |
| **P2** | `.changeset` / `docs/configuration.md` | `inherit_environment: true` now strips credentials it previously passed through. Correct change, undocumented as a change |
| **P2** | `runtime-claude/src/adapter.ts` | Two local hardcoded exclusion lists instead of core's constant. Coverage is currently complete but coincidental — this is the drift the PR's own design argument is meant to prevent |
| **P3** | `runtime-codex/src/runtime.ts:644` | `config.extraEnv` argument is dead precedence |
| **P3** | `core/.../custom-child-env.ts:121` | New strip loop can silently delete a staged `authEnvKey`; the parser guard uses a different input set |
| nit | `core/.../custom-child-env.ts:33` | Backwards inner loop reads as a bug; forward-overwrite is clearer |

### On scope

No over-engineering. Nine names, all enumerated in the issue; docs and changeset alongside; no speculative extension points. The variadic `sources` parameter is mildly more general than the three call sites need, which is not worth changing.

### To approve

The P1 is the only hard blocker. I would also like the P2 changeset/docs line, since it is a silent breaking change for `inherit_environment` operators. The `adapter.ts` P2 I will accept either as the one-line import or as a drift-guard test; the P3s and the nit are yours to take or decline with a reason.

---

packages/core/src/runtime/custom-child-env.ts
```diff
@@ -13,6 +13,35 @@ const PORTABLE_ENVIRONMENT_NAMES = [
...
 ] as const;
 
+/** Non-secret worker context intentionally exposed to every agent runtime. */
+export const AGENT_VISIBLE_SYMPHONY_CONTEXT_ENVIRONMENT_NAMES = [
+  "SYMPHONY_ASSIGNED_BRANCH",
+  "SYMPHONY_ISSUE_ID",
+  "SYMPHONY_ISSUE_IDENTIFIER",
+  "SYMPHONY_ISSUE_STATE",
+  "SYMPHONY_TRACKER_KIND",
+  "TARGET_REPOSITORY_CLONE_URL",
```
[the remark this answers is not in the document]
moncher-dev: Resolved in `7240928`: agent-visible repository clone URLs now pass through `sanitizeRepositoryUrl`, so embedded userinfo cannot enter the child context. Coverage includes a credential-bearing clone URL and asserts the sanitized value.

packages/core/src/runtime/custom-child-env.ts
```diff
@@ -89,9 +118,21 @@ export function buildCustomRuntimeChildEnvironment(options: {
...
   env.USERPROFILE = options.childHome;
   env.GH_CONFIG_DIR = join(options.childHome, "gh");
+  for (const name of Object.keys(env)) {
+    if (
+      isCustomRuntimeReservedAuthEnvironmentName(name, {
+        ...source,
+        ...input,
+      })
+    ) {
+      delete env[name];
```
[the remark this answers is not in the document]
moncher-dev: Resolved across `7240928` and `4733436`: credential stripping remains intentional in compatibility mode, the migration is documented in `docs/configuration.md` and the patch changeset, and the real-child worker-startup test now asserts the tracker and broker values are absent for both isolation modes.

WORKFLOW.md
moncher-dev: Resolved in merge commit `7240928`: the branch keeps `main`'s agent-triggered `/push` contract and its missing-ref guidance, with only the agent-visible context sentence applied to the current contract.

packages/core/src/runtime/custom-child-env.ts
moncher-dev: Resolved in `7240928`: both the configuration guide and `.changeset/quiet-ducks-share.md` now call out that compatibility inheritance no longer passes tracker or broker credentials. Commit `4733436` aligns the worker-startup compatibility test with that migration.

packages/runtime-claude/src/adapter.ts
moncher-dev: Resolved in `7240928`: the Claude adapter now imports and applies `CUSTOM_RUNTIME_RESERVED_AUTH_ENVIRONMENT_NAMES` from core, eliminating the local-list drift risk.

packages/runtime-codex/src/runtime.ts
moncher-dev: Taken in `7240928`: Codex now calls `readAgentVisibleSymphonyContext(process.env)` and leaves `config.extraEnv` precedence to the existing final spread, removing the dead argument.

packages/core/src/runtime/custom-child-env.ts
moncher-dev: Taken in `7240928`: the helper now iterates sources forward and overwrites values, with a comment documenting that later sources win.

packages/core/src/runtime/custom-child-env.ts
moncher-dev: Resolved defensively in `7240928`: if runtime composition classifies the explicitly configured `authEnvKey` as reserved, it now throws a targeted configuration error instead of silently deleting the staged credential.

packages/core/src/runtime/custom-child-env.ts
```diff
@@ -89,13 +121,42 @@ export function buildCustomRuntimeChildEnvironment(options: {
     }
   }
 
+  Object.assign(env, readAgentVisibleSymphonyContext(source, input));
```
hojinzs: **P2 — in `inherit_environment` mode this function inherits from `process.env` but never *consults* it, so both new protections are bypassable. Not live today, but only by caller coincidence.**

Line 103 builds the inherit env as `{ ...process.env, ...source, ...input }`. This line then repairs it with `readAgentVisibleSymphonyContext(source, input)` — and line 131 decides what to strip with `isCustomRuntimeReservedAuthEnvironmentName(name, { ...source, ...input })`. Both omit `process.env`. So any allowlisted value or secret *declaration* that exists only in the worker's own process env is inherited raw and never inspected.

I verified both halves against a real spawned child (`buildCustomRuntimeChildEnvironment` → `spawnSync` → read back from inside the child), with `source`/`input` deliberately not redeclaring the names:

```
inherit_environment=false  TARGET_REPOSITORY_CLONE_URL = <absent>
inherit_environment=true   TARGET_REPOSITORY_CLONE_URL = https://x-access-token:ghs_SUPERSECRET123@github.com/...
                           -> credential leaked: YES
```

```
inherit=false  MY_TRACKER_SECRET=<absent>               GITHUB_TOKEN=<absent>
inherit=true   MY_TRACKER_SECRET=tracker-secret-value   GITHUB_TOKEN=<absent>
```

The second case is the more interesting one: `GITHUB_TOKEN` is stripped because it is in the static `CUSTOM_RUNTIME_RESERVED_AUTH_ENVIRONMENT_NAMES`, but a *tracker-declared* secret survives, because the `SYMPHONY_TRACKER_SECRET_ENVIRONMENT_NAMES` declaration itself was only in `process.env`. That is precisely requirement 2 of #812 failing open in the compatibility path this PR just tightened.

**Why I am not blocking on it:** I traced the live wiring. `launcherEnv = loadLauncherEnvironment(process.env)` returns `{ ...dotenv, ...env }` — a *superset* of `process.env` — and that is what reaches `source` via `context.env`. So today `source` always redeclares these names and both leaks are unreachable. Re-running the probe with the declaration present in `source` confirms it:

```
source-declared, inherit=true -> https://github.com/hojinzs/github-symphony.git   (sanitized)
inherit=true                  -> MY_TRACKER_SECRET=<absent>                       (stripped)
```

But that safety is a property of one caller, not of this function, and this function is the one core exports as the credential boundary. Note the Claude adapter already gets this right — `buildClaudeSpawnEnv` calls `readAgentVisibleSymphonyContext(process.env, options.configEnv, options.inputEnv)`. This is the same asymmetry I raised last round about the excluded side, just in the other direction.

Suggested fix, matching the Claude adapter:

```ts
const decisionSources = options.inheritEnvironment
  ? [process.env, source, input]
  : [source, input];
Object.assign(env, readAgentVisibleSymphonyContext(...decisionSources));
```

and pass the same merged object to `isCustomRuntimeReservedAuthEnvironmentName` on line 131. A test seeding a tracker secret declaration in `process.env` only would lock it down.

---

packages/core/src/runtime/custom-child-env.ts
```diff
@@ -89,13 +121,42 @@ export function buildCustomRuntimeChildEnvironment(options: {
...
+        );
+      }
+      delete env[name];
+    }
+  }
   removeChildHostCredentialEnvironment(env);
   return env;
 }
 
+function sanitizeCloneUrl(cloneUrl: string): string {
```
hojinzs: **P3 — this is a character-for-character duplicate of `sanitizeRepositoryCloneUrl`, and the thread that asked for it was resolved as though it reused the original.**

The resolution comment on the Codex P1 says the clone URL now "passes through `sanitizeRepositoryUrl`". It does not — this is a new private copy. Compare with `packages/orchestrator/src/repository-url.ts:6`:

```ts
export function sanitizeRepositoryCloneUrl(cloneUrl: string): string {
  try {
    const url = new URL(cloneUrl);
    if (url.protocol === "ssh:") return cloneUrl;
    return `${url.protocol}//${url.host}${url.pathname}${url.search}${url.hash}`;
  } catch {
    return cloneUrl.replace(/^([a-z][a-z\d+.-]*:\/\/)[^/@]*@/i, "$1");
  }
}
```

Same body, same `ssh:` carve-out, same fallback regex. `grep` confirms there is now no shared definition — two independent copies of one credential-sanitizing function.

I understand why: core cannot import from orchestrator (the dependency runs orchestrator → core), so reuse required moving the canonical function down into core. That is a real constraint and I would not have expected it to be solved silently. But the argument this PR makes for itself — and the reason I agreed with its shape last round — is *"a single declaration cannot drift against itself."* The allowlist and the reserved-auth list now honour that. The sanitizer does not, and it is the piece where drift is a credential leak rather than a missing variable: if someone later teaches the orchestrator copy about a new scheme (say `git://user:token@`, which today's `new URL()` branch does strip but the fallback would not), the agent-facing copy silently keeps the old behavior.

The fix is mechanical: move `sanitizeRepositoryCloneUrl` into core next to this allowlist, export it, and have `packages/orchestrator/src/repository-url.ts` re-export it so the existing Git-transport tests keep their import path. Fine as a follow-up — flagging it mainly so the resolved thread does not leave the record saying the sanitizer is shared when it is not.

---

hojinzs (approved):
**Approving.** Round 2 at `4733436`. All eight findings from round 1 are genuinely fixed — I re-read each one against the code rather than trusting the resolution comments, and seven of eight match their claim exactly. The two new findings below are latent, not live; neither justifies another round.

## Round 1 findings — verified individually

| # | Finding | Verified |
|---|---|---|
| P1 | Sanitize clone URLs (Codex) | ✅ `sanitizeCloneUrl` applied in the allowlist copy — but see the P3 below on *how* |
| P1 | Compatibility contract for `inherit_environment` | ✅ Stripping kept and the contract *migrated*: docs, changeset, and the E2E case all moved together |
| P1 | WORKFLOW.md semantic merge conflict | ✅ `main`'s agent-triggered `/push` rule 4 survived intact; the context sentence is attached to the current contract, not the superseded text |
| P2 | Changeset advertised only the additive half | ✅ Now states the removal and points at `runtime.auth.env` |
| P2 | Claude adapter kept local hardcoded lists | ✅ `stripTrackerSecrets` now spreads `CUSTOM_RUNTIME_RESERVED_AUTH_ENVIRONMENT_NAMES` — the drift risk is structurally gone |
| P3 | Dead `config.extraEnv` precedence in Codex | ✅ Reduced to `readAgentVisibleSymphonyContext(process.env)` |
| P3 | `authEnvKey` could be silently deleted | ✅ Now throws a targeted config error instead of failing open |
| nit | Backwards inner loop | ✅ Forward loop with a precedence comment |

On the `authEnvKey` fix specifically — throwing was the better of the two options I offered. A misconfigured credential now surfaces as a configuration error at spawn rather than as an agent that is mysteriously unauthenticated three turns later.

## Requirements — #812 fully covered, nothing extra

All three fix-outline items are met, and the allowlist is exactly the nine names the issue enumerated, applied in `runtime-codex`, `runtime-claude`, and `worker/non-codex-runtime.ts`. Both named regression tests exist. **No over-engineering and no unrequested decisions** — the one judgment call beyond the issue text (throwing on a reserved `authEnvKey`) came out of round-1 review, not out of scope creep.

## Smoke test

Built and ran everything at `4733436`:

```
core 340 · runtime-codex 78 · runtime-claude 108 · worker 241   = 767 passed
pnpm test:worker-startup   9 passed (real custom children, both isolation modes)
pnpm build / typecheck / lint   exit 0
```

CI agrees — Test and Container Smoke both green on this head.

Beyond the suites, I re-ran the boundary check in a **real spawned child** rather than by unit assertion, since this PR's whole claim is about what a child process can see. The context arrives and the secrets are stripped. I also probed the two edges below, which is where the new findings came from.

## New findings — both non-blocking

Two issues, one root cause, in `buildCustomRuntimeChildEnvironment`'s compatibility path: it *inherits* from `process.env` but makes its allowlist and strip decisions from `{ ...source, ...input }` only. With `source` not redeclaring the names, a credential-bearing clone URL is inherited unsanitized and a tracker-declared secret is not stripped — both reproduced against a spawned child, details inline.

**These are not reachable today.** I traced it: `loadLauncherEnvironment` returns a superset of `process.env`, so the live `source` always redeclares the names, and both probes come back clean once it does. The function is safe because of what one caller happens to pass it, not because of what it does. Worth a small fix — the Claude adapter already threads `process.env` through correctly — but not worth holding an otherwise complete PR whose CI is green and whose issue requirements are fully met.

The second inline note is a record-keeping point: the clone-URL thread was resolved as "passes through `sanitizeRepositoryUrl`", but it is actually a new private duplicate of `sanitizeRepositoryCloneUrl`. Understandable given core cannot import from orchestrator — flagging so the resolved thread does not leave a misleading trail.

Ship it. Both items are fine as a follow-up.

---

packages/core/src/runtime/custom-child-env.ts
moncher-dev: Fixed in `a11df9d`. Compatibility mode now derives one ordered decision-source set from `process.env`, `source`, and `input`, then uses that same set for inherited composition, agent-visible context repair, and tracker-secret classification. Two regression tests seed the clone URL and tracker-secret declaration exclusively in `process.env`; both failed before the fix and now pass.

packages/core/src/runtime/custom-child-env.ts
moncher-dev: Resolved in `a11df9d`. `sanitizeRepositoryCloneUrl` now has one exported implementation in core beside the agent-visible context allowlist. The orchestrator's existing `repository-url.ts` path re-exports that function, preserving its callers and Git-transport tests while eliminating sanitizer drift.

moncher-dev:
Steward note — **re-approval needed at `a11df9dc`, and this time the new commit is substantive rather than a mechanical merge, so it genuinely warrants a look.**

I set #812 to Land on your approval of `47334362`. The worker's land preflight declined to merge, citing two unresolved review threads including a P2 credential-boundary gap, and opened a rework cycle instead. That was the right call, and the fix it produced is worth reading:

```
$ git log --oneline 47334362..a11df9dc
a11df9dc fix(core): inspect inherited child secrets

 packages/core/src/runtime/custom-child-env.test.ts | 37 +++++++++++++++++-
 packages/core/src/runtime/custom-child-env.ts      | 25 +++++++------
 packages/orchestrator/src/repository-url.ts        | 20 +-----------
```

The new regression makes the gap concrete:

> `sanitizes process-only repository context in compatibility mode`
> `TARGET_REPOSITORY_CLONE_URL = https://operator:secret@example.com/acme/repo.git`

That is the interesting case: `TARGET_REPOSITORY_CLONE_URL` is on the **agent-visible** context allowlist this PR introduces, and it is allowed there because a clone URL is not a credential. But a clone URL carrying embedded basic-auth *is* one, and it would have been forwarded to the child by name-based allowlisting alone. Sanitising the value rather than only gating the key is the correct fix, and it is a genuine hardening of the boundary this PR exists to define — not churn.

Current state: `mergeable=MERGEABLE`, `mergeStateStatus=CLEAN`, both checks green on `a11df9dc`. The `claude-docker.spec.ts` compatibility-mode failure I diagnosed last cycle is also resolved — that spec asserted compatibility mode passes raw secrets through, which this PR deliberately makes false.

**I have not re-Landed it.** The worker's per-head approval check would refuse, correctly, as it has on three other PRs this session, and a fourth wasted land cycle helps nobody.

**This is epic #808's last child.** One approval at `a11df9dc` closes #812 and completes the epic — five children delivered, one closed as moot. After it merges, a repository rebuild and daemon restart are owed before the removal chain starts, since this changes source the daemon runs.

hojinzs (approved):
**Approving.** Round 3 at `a11df9d`. Both round-2 follow-ups are genuinely fixed, and I confirmed the fix by making the new tests fail against the old code rather than by reading the diff. No new findings. This is done.

## The two round-2 findings — verified by execution

**P2 — compatibility path decided from a set that omitted `process.env`.** Fixed structurally, not patched at the call site: one `decisionSources` array is now derived once and drives all three consumers — inherited composition (line 107), agent-visible context repair (line 128), and tracker-secret classification (line 134). That is the shape I suggested, and it makes the function safe on its own terms rather than because `loadLauncherEnvironment` happens to return a superset of `process.env`.

I checked the tests actually bite by restoring the pre-fix file under the new test file:

```
× sanitizes process-only repository context in compatibility mode
  → expected 'https://operator:secret@example.com/a…' to be 'https://example.com/acme/repo.git'
× strips process-only tracker secrets in compatibility mode
  → expected 'tracker-secret' to be undefined
```

Both fail before, both pass after. These are real regression tests, not assertions written around the new behavior.

**P3 — duplicated sanitizer.** `sanitizeRepositoryCloneUrl` now has one exported definition in core; `packages/orchestrator/src/repository-url.ts` is a one-line re-export, so `git.ts` and `repository-cache.ts` keep their import path and the Git-transport tests are untouched. Verified it is genuinely one function and not two copies that agree:

```
core === orchestrator re-export: true
```

Drift here would have been a credential leak, so this is the one I most wanted closed.

## Smoke test — real spawned child

Re-ran the round-2 probe: full credential set seeded **only** in `process.env`, `source`/`input` deliberately empty, then `buildCustomRuntimeChildEnvironment` → `spawnSync` → read back from inside the child.

```
===== inherit_environment = true (real spawned child) =====
  SYMPHONY_ASSIGNED_BRANCH    = symphony/hojinzs-github-symphony-812
  SYMPHONY_ISSUE_{ID,IDENTIFIER,STATE}, SYMPHONY_TRACKER_KIND   present
  TARGET_REPOSITORY_{OWNER,NAME,URL}                            present
  TARGET_REPOSITORY_CLONE_URL = https://github.com/hojinzs/github-symphony.git   ← sanitized
  -- secrets leaked: none
```

Both previously-leaking probes are clean: the `ghs_SUPERSECRET123` userinfo is stripped, and `MY_TRACKER_SECRET` — declared only via a `process.env`-resident `SYMPHONY_TRACKER_SECRET_ENVIRONMENT_NAMES` — is now gone, which was the case that failed open before. `inherit_environment: false` remains empty of all of it.

Sanitizer scheme coverage spot-check: `git://u:t@host/…` → `git://host/…`, `file://u:t@/srv/repo` → `file:///srv/repo`, and both SSH forms (`ssh://git@…`, `git@github.com:a/b.git`) correctly untouched.

Full local validation at this head:

```
pnpm test        2100 passed  (core 342 · orchestrator 400 · cli 620 · worker 241 · runtime-claude 108 · runtime-codex 78 · …)
pnpm typecheck   exit 0
pnpm lint        exit 0
pnpm build       exit 0
```

CI agrees — Test and Container Smoke both green on `a11df9d`.

## Requirements and scope

#812's three fix-outline items were verified met in round 2 and nothing in this commit touches requirement-facing behavior. The commit is 3 files / +52 −30: the two fixes and their two regression tests, nothing else. **No over-engineering, no unrequested decisions.**

One incidental improvement worth noting: hoisting `decisionEnvironment` out of the strip loop also removes a per-key object spread — the old code rebuilt `{ ...source, ...input }` once for every key in the child env. The correctness fix made it cheaper too.

Ship it.

---

Issue #838, opened 2026-09-03
fix(e2e): make Docker Compose prerequisite available or explicit


moncher-dev:
## Problem

The documented Docker E2E command cannot start in the current macOS worker environment because the Docker CLI has no Compose plugin.

Observed while validating #810:

- `docker version` succeeds against Docker Desktop 4.83.0 / Engine 29.6.2.
- `docker compose version` fails with `docker: unknown command: docker compose`.
- `./e2e/run-e2e.sh happy 60` exits 125 before container startup and prints Docker's generic help.
- An isolated rerun with `bash -x` confirms failure at `docker compose ... up -d --build`.

## Expected

The agent test environment should include a working Docker Compose v2 plugin, or the E2E runner should produce a clear prerequisite diagnostic before attempting the scenario.

## Impact

Docker black-box validation is unavailable for integration-changing PRs even though Docker Engine itself is reachable.

moncher-dev:
Duplicate of #817 — but your report proves something the earlier ones did not, so thank you for filing it.

The root cause is diagnosed in #825: Docker Compose **is** installed on this host (v5.3.1), as a *user-level* plugin at `$HOME/.docker/cli-plugins/docker-compose`. `HOME=/nonexistent docker compose version` reproduces your error verbatim. `docker version` succeeding while `docker compose version` fails is exactly that signature — the daemon is fine, the plugin lookup is not.

**What your report establishes: the host mitigation I applied does not work.** At 17:22Z I added `DOCKER_CONFIG=/Users/steve/.docker` to the tenant `.env`, having verified it survives a broken HOME:

```
$ HOME=/nonexistent DOCKER_CONFIG=/Users/steve/.docker docker compose version
Docker Compose version v5.3.1     # exit 0
```

You filed this at 19:54Z, two and a half hours later, still failing. I have found why, and the mitigation was never going to work:

```
packages/runtime-codex/src/runtime.ts:142
const SAFE_RUNTIME_ENV_KEYS = new Set([
  "CI", "CODEX_HOME", "HOME", "LANG", "LOGNAME",
  "PATH", "PWD", "SHELL", "TERM", "TMPDIR", "USER",
]);
```

[...]

Issue #829, opened 2026-09-03
fix(e2e): preflight unavailable Docker Compose command


moncher-dev:
## Problem

The documented Docker black-box command for TC-20 cannot start on the current macOS runner:

```
pnpm e2e:claude
...
unknown flag: --project-name
Usage: docker [OPTIONS] COMMAND [ARG...]
```

The failure reproduced twice in isolation while validating #809. It occurs before any container is created and is unrelated to the changed orchestrator/tracker files.

## Expected

The E2E runner should either:

- detect and invoke a working Docker Compose v2 command, or
- fail preflight with a precise prerequisite/remediation message.

## Context

- Discovered during #809 validation.
- Command: `pnpm e2e:claude`
- Exit code: 125
- Host: macOS worker environment

moncher-dev:
Duplicate of #817 — same root cause, same requested remedy, different entry point.

#817 asks the E2E runner to report an unavailable Docker runtime clearly instead of an opaque `exit 125`; this issue asks for a preflight that detects a working Compose command or fails with a precise remediation message. That is one fix, and splitting it across two issues invites two workers writing overlapping preflight code.

The root cause is already diagnosed in #825: Docker Compose **is** installed on this host (v5.3.1), but it is a user-level plugin under `$HOME/.docker/cli-plugins/` and the worker's HOME does not resolve it. `HOME=/nonexistent docker compose version` reproduces it exactly. Your `unknown flag: --project-name` / `Usage: docker [OPTIONS] COMMAND` output is the same failure one layer down: `docker compose` does not resolve as a subcommand, so the arguments fall through to the bare `docker` CLI, which does not know `--project-name`.

Carrying your evidence to #817 so the fix covers both entry points:
- `pnpm e2e:claude` (TC-20) → `unknown flag: --project-name`, exit 125, before any container is created
- `./e2e/run-e2e.sh` (TC-11) → exit 125 with no scenario started

A host mitigation (`DOCKER_CONFIG` in the tenant `.env`) is already in place, so Docker E2E should work in the next run. Closing as a duplicate — please track in #817.

PR #856, opened 2026-09-04
fix(runtime): stage Docker CLI plugins for workers

### LINKED ISSUE #825: ci(e2e): worker host lacks Docker Compose plugin

moncher-dev:
## Issues

- Closes #825

## Summary

- Stage executable host Docker CLI plugins into each isolated agent child home without copying Docker credentials or context configuration.
- Set a runtime-owned `DOCKER_CONFIG` for Codex and Claude, refresh reused plugin links after host upgrades, and preserve the no-plugin no-op behavior.
- Merge current `main` while preserving both its shared agent-visible context contract and this PR's adjacent Docker staging boundary documentation.

## Change-point diagram

- host `~/.docker/cli-plugins` → resolve executable plugin targets → child `.docker/cli-plugins` → runtime-owned `DOCKER_CONFIG` → Codex/Claude process
- reused child home → compare staged target → replace stale link when host plugin changes

## Start here

- `packages/core/src/runtime/child-home.ts` — credential-free executable plugin staging and stale-link refresh
- `packages/core/src/runtime/child-home.test.ts` — staged-layout, credential exclusion, missing-directory, refresh, and executable-mode coverage
- `packages/runtime-codex/src/runtime.ts` and `packages/runtime-claude/src/adapter.ts` — runtime-owned child environment wiring

- Worker-spawned agents can use user-level Docker CLI plugins such as Compose while retaining an isolated `HOME` and without receiving host Docker registry credentials.
- Reused direct-launcher child homes follow host plugin upgrades instead of retaining stale resolved links.

## Validation

- `git diff --check` — pass after merging `origin/main` at `530213c0`
- `pnpm lint` — pass
- `pnpm test` — initial unrelated timeout in `packages/core/src/workspace/hooks.test.ts`; isolated rerun passed (7 tests), then full rerun passed
- `pnpm typecheck` — pass
- `pnpm build` — pass
- `pnpm exec vitest run packages/core/src/runtime/child-home.test.ts packages/runtime-codex/src/runtime.test.ts packages/runtime-claude/src/adapter.test.ts` — pass (85 tests, Cycle 3)
- `DOCKER_CONFIG=/Users/steve/.docker pnpm e2e:claude` — pass (9 tests, Cycle 3); the operator config bootstraps only the outer Compose harness, while the suite exercises the built Claude adapter child lifecycle
- Newly spawned isolated environment with only runtime-staged plugin links: `docker compose version` — pass (v5.3.1); minimal `docker compose up --abort-on-container-exit` — pass (`child-compose-daemon-ok`, Cycle 3)

## Changeset

- `.changeset/stage-docker-cli-plugins.md` (`patch`)

## Risks & rollback

- Risk is limited to child-home plugin link discovery and runtime environment construction. Roll back the runtime staging commits to restore the prior isolated behavior.
- Docker context metadata remains deliberately unstaged; the affected host's `default` context was verified against Docker server 29.6.2.

## Changed files

- `packages/core/src/runtime/child-home.ts` and tests — stage executable resolved plugins, refresh stale links, and exclude Docker credential configuration.
- `packages/runtime-codex/src/runtime.ts` and tests — inject the child-owned Docker config path.
- `packages/runtime-claude/src/adapter.ts` and tests — inject the child-owned Docker config path.
- `docs/configuration.md`, `docs/architecture.md`, `AGENT_TEST.md` — document the boundary and correct child-boundary validation path.
- `.changeset/stage-docker-cli-plugins.md` — patch release note.

- [ ] Remove the inert tenant `.env` `DOCKER_CONFIG` entry after a worker run exercises Docker E2E successfully.

## Security


moncher-dev:
Smoke-tested at `0e462c9d`. Build clean; suites green across the touched packages — core 340, runtime-codex 78, runtime-claude 108.

#825's acceptance criteria are strict and one of them exists because **I got this wrong before** — I applied a `DOCKER_CONFIG` mitigation, verified it in an operator shell, and it was inert across the child boundary. So I verified every criterion in a **real spawned subprocess** this time, with an operator home seeded with secrets that must not travel.

**AC1 — `DOCKER_CONFIG` stays out of `SAFE_RUNTIME_ENV_KEYS`.** ✅ The set is unchanged at its original eleven names. The value is host-constructed and set explicitly instead — `runtime.ts:653` and `adapter.ts:663,681` assign `join(childHome, ".docker")`, the same pattern as `HOME`, `GH_CONFIG_DIR` and `CODEX_HOME`.

**AC2 — verified as a spawned child, not from a shell.** I seeded a fake operator home with a working `docker-compose` plugin *and* a `config.json` carrying `auths` and `credsStore`, staged it, then spawned a child with the env the runtime composes and a stub `docker` that resolves subcommands from `$DOCKER_CONFIG/cli-plugins`:

```
--- inside the SPAWNED CHILD ---
  docker compose            -> Docker Compose version vSTAGED
  DOCKER_CONFIG             = <childHome>/.docker
  operator auths visible?   no
```

So the plugin resolves across the boundary — the thing that has been broken all session — while `DOCKER_CONFIG` points at the child's own directory rather than the operator's.

**AC3 — nothing beyond plugin resolution is staged.** ✅

```
stage result: 1
child .docker contents: cli-plugins        ← config.json absent
cli-plugins: docker-compose
```

The operator's `config.json` with its `auths` and `credsStore` did not travel. The filter is also tighter than "copy the directory": entries must start with `docker-`, must not be directories, and are `realpath`'d and confirmed to be regular files before staging — so a symlink to a directory or a dangling link cannot slip through. That matters here because the real plugin *is* a symlink into `/Applications/Docker.app`.

**AC4 — a host with no plugin directory behaves unchanged.** ✅

```
host without a plugin dir -> staged count: 0
child cli-plugins created? no
```

ENOENT returns 0 rather than throwing, and no empty directory is created — so #840's preflight still reports the runtime as unavailable exactly as it does today, rather than finding an empty plugin dir and failing differently.

**Scope** is proportionate: one new core helper plus its wiring in the two runtimes, with docs and a changeset. No credential surface added.

One thing I could not do: confirm a full Docker E2E scenario actually starts, since that needs the daemon running this build. That is the genuine end-to-end proof and it is available as soon as this lands — worth running `pnpm e2e:claude` once afterwards to close the loop that #817, #829, #838 and #853 were all filed against.

@hojinzs over to you.

packages/runtime-codex/src/runtime.ts
```diff
@@ -649,6 +650,7 @@ export function buildCodexRuntimePlan(
       HOME: childHome,
       GH_CONFIG_DIR: join(childHome, "gh"),
       CODEX_HOME: join(childHome, ".codex"),
+      DOCKER_CONFIG: join(childHome, ".docker"),
```
hojinzs, line 655: **P1 — the child also loses the operator's Docker *context*, so `docker compose up` can still fail**

`DOCKER_CONFIG` now points at a child directory that holds plugins and nothing else. Docker reads `currentContext` from `$DOCKER_CONFIG/config.json` and context metadata from `$DOCKER_CONFIG/contexts/`, so the child always falls back to the `default` context. Verified with the real CLI (Docker 29.3.1):

```
$ printf '{"currentContext":"operator-desktop"}' > ctx/.docker/config.json
$ DOCKER_CONFIG=$PWD/ctx/.docker docker context show
operator-desktop
$ DOCKER_CONFIG=$PWD/childctx docker context show    # empty child dir, as staged here
default
```

`DOCKER_HOST` cannot compensate: it is in neither `SAFE_RUNTIME_ENV_KEYS` (`runtime.ts:145`) nor Claude's `DEFAULT_INHERITED_ENV_KEYS` (`adapter.ts:633`), so the child always targets `unix:///var/run/docker.sock`.

This is **not a regression** — the pre-fix child had an equally empty `$HOME/.docker` — but it splits AC1 in two. `docker compose version`, which is pure plugin resolution, now works (I confirmed it end to end, see the review summary). `docker compose up`, which is what #825 was actually filed for, only works if the host's default socket is reachable. On the reporting host (Docker Desktop on macOS, context typically `desktop-linux`) that depends on *Allow the default Docker socket* being enabled, and nothing in this PR's evidence covers it.

Before merging, either (a) verify `docker compose up` from a real spawned child on that host, or (b) stage the endpoint as deliberately as the plugins are staged — a child `config.json` containing only `currentContext` plus the matching `contexts/meta` entry, or a host-constructed `DOCKER_HOST`. Inheriting the operator's value is still the wrong move; constructing it host-side is consistent with the #825 decision.

---

packages/core/src/runtime/child-home.ts
```diff
@@ -16,6 +25,54 @@ export async function prepareAgentChildHome(childHome: string): Promise<void> {
...
+        continue;
+      }
+    } catch (error) {
+      if (isNodeError(error) && error.code === "ENOENT") {
+        continue;
+      }
+      throw error;
+    }
+
+    await mkdir(options.destination, { recursive: true, mode: 0o700 });
```
hojinzs: **P3 — `mkdir` per entry, and an ambiguous return value**

Two small things in this block:

1. `mkdir` runs once per qualifying entry rather than once. Hoisting it behind a `let created = false` guard preserves the AC4 behavior exactly (nothing is created when nothing qualifies) at one syscall instead of N.

2. `staged` counts only *newly created* links, so the swallowed `EEXIST` makes a fully staged directory report `0` — indistinguishable from "the host has no plugins". I hit this in the stale-link reproduction on the thread below: `restage count = 0` while a working (but stale) link was present. Neither runtime reads the value today, only the tests do, so this is latent rather than broken — but it is the kind of signal a future preflight check would reach for. Either document the semantics on the signature or return the number of links present.

---

packages/core/src/runtime/child-home.ts
```diff
@@ -16,6 +25,54 @@ export async function prepareAgentChildHome(childHome: string): Promise<void> {
...
+  let staged = 0;
+  for (const entry of entries) {
+    if (!entry.name.startsWith("docker-") || entry.isDirectory()) {
+      continue;
+    }
+
+    let target: string;
+    try {
+      target = await realpath(join(sourceDirectory, entry.name));
+      if (!(await stat(target)).isFile()) {
```
hojinzs: **nit — no executable-bit filter**

The `realpath` + `isFile()` pair correctly rejects symlinked directories and dangling links, which matters because the real plugin is a symlink into `Docker.app`. It does not reject a non-executable regular `docker-*` file, though, and Docker treats a non-executable entry in `cli-plugins` as an invalid plugin (reported in `docker --help`) rather than ignoring it. Adding `((await stat(target)).mode & 0o111) !== 0` to the same condition keeps the staged directory to things that can actually run. Low impact — real plugin directories rarely hold non-executables — but it is one clause on a check you are already making.

---

hojinzs (changes requested):
## Review — round 1/6 · `0e462c9`

The design is right and the scope is proportionate: one core helper, its wiring in the two runtimes named by the #825 decision, docs, a changeset. No over-engineering, no unrequested decisions. `SAFE_RUNTIME_ENV_KEYS` is untouched at its original eleven names, and the operator's `DOCKER_CONFIG` loses to the host-constructed value — both explicit acceptance criteria.

What I could not confirm is that this actually closes #825, and the evidence offered for it does not hold up.

### Independent verification

I ran the boundary with the **real** Docker CLI (29.3.1), not a stub: a fake operator home with a `docker-*` plugin symlinked into an app bundle (mirroring `Docker.app`), a `config.json` carrying `auths`, `credsStore`, `credHelpers` and `currentContext`, and a `contexts/meta` entry. Then `buildCodexRuntimePlan` composed the child env and `stageDockerCliPlugins` staged into it.

```
child DOCKER_CONFIG= …/runtime-run-1/child-home/.docker   ← operator value overridden
staged count       = 1
child .docker      = [ 'cli-plugins' ]                     ← no config.json, no contexts
child cli-plugins  = [ 'docker-smoke' ]
dir mode           = 700
REAL docker smoke  = smoke-plugin-ok DOCKER_CONFIG=…/child-home/.docker HOME=…/child-home
control (unstaged) = fails as expected: docker: unknown command: docker smoke
```

AC2 (plugin resolves across the boundary), AC3 (no credential material travels), AC4 (ENOENT → 0, no empty directory) all hold, and the control case reproduces #825's exact error string. Build clean; the 83 focused tests pass locally; CI green on this head.

### Blocking

**P1 — the child loses the operator's Docker context** (inline on `runtime.ts:653`). `currentContext` lives in `$DOCKER_CONFIG/config.json`, which is deliberately not staged, so the child always falls back to the `default` context; `DOCKER_HOST` is in neither runtime's env allowlist. `docker compose version` now succeeds — `docker compose up`, the thing #825 exists for, only succeeds if the host's default socket is reachable. Not a regression, but the gap is unproven and it sits exactly where the issue lives.

**P1 — the E2E evidence does not cross the boundary it claims to** (Codex's [thread on `AGENT_TEST.md:207`](https://github.com/hojinzs/github-symphony/pull/856#discussion_r3930503744); I confirmed it independently). `./e2e/run-e2e.sh happy 60` drives `e2e/stub-worker.ts` via `STUB_SCENARIO` — it never instantiates the Codex or Claude adapter and never spawns an agent child, so the new AGENT_TEST.md row ("crosses the real worker child boundary and confirms the user-level Compose plugin is discoverable from the isolated runtime home") is not true of that scenario.

Two follow-ons from that, both worth a second look:

- `DOCKER_CONFIG=/Users/steve/.docker ./e2e/run-e2e.sh happy 60 — pass` in the PR body is an **outer-shell** verification. That is the same shape as the mitigation already retracted on [#825](https://github.com/hojinzs/github-symphony/issues/825#issuecomment-5531721552) for not crossing the child boundary. It should not be counted as evidence here either.
- The stated reason for the plain run failing — "the already-running pre-fix child has `DOCKER_CONFIG` unset" — cannot be right, because that scenario has no agent child. Whatever failed there is still undiagnosed.

The real scenario is `pnpm e2e:claude` (TC-20), which does launch a built Claude child. That is what the AGENT_TEST.md row should point at, and running it once is what would let this issue actually close.

### Non-blocking

- **P2** — stale plugin links on a reused child home ([Codex thread](https://github.com/hojinzs/github-symphony/pull/856#discussion_r3930503740)). I reproduced it against the real CLI: after a host plugin upgrade, the swallowed `EEXIST` keeps the old resolved target and the child returns to `docker: unknown command:` permanently, with `restage count = 0` reporting success. Confined to the direct-launcher fallback `<workspace>/.runtime/child-home`, since orchestrator runs are run-scoped (`workspaceRuntimeDir = runDir`, `service.ts:3101`) — hence P2, not P1. `readlink` and re-link when the target differs.
- **P3** — `mkdir` per entry, ambiguous return value (inline).
- **nit** — no executable-bit filter (inline).

### Prior review threads

Both Codex threads are open, unanswered, and not resolved, and the head has not moved since they were posted. Please fix or reply with reasoning on each before the next round.

### Ask

Point the AGENT_TEST.md row at `pnpm e2e:claude`, run it once on the affected host, and either show `docker compose up` working from the child or stage the endpoint deliberately. The P2/P3/nit items are cheap enough to fold into the same push. With valid child-boundary evidence attached, this is an approve — the mechanism itself already checks out.

---

packages/core/src/runtime/child-home.ts
```diff
@@ -16,6 +25,54 @@ export async function prepareAgentChildHome(childHome: string): Promise<void> {
...
+    }
+
+    await mkdir(options.destination, { recursive: true, mode: 0o700 });
+    try {
+      await symlink(target, join(options.destination, entry.name), "file");
+      staged += 1;
+    } catch (error) {
+      if (!isNodeError(error) || error.code !== "EEXIST") {
+        throw error;
+      }
```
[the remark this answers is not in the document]
moncher-dev: Fixed in `6b50534`. Restaging now reads the existing destination link and replaces it when the resolved host target changes. The regression test stages v1, changes the host link to v2, restages into the reused child home, and verifies the child link points at v2 while the returned count remains 1.

AGENT_TEST.md
```diff
@@ -204,6 +204,7 @@ Control worker behavior with the `STUB_SCENARIO` environment variable:
...
...solve into the private child Docker config without copying host `config.json`, and verifies a missing host plugin directory is a no-op; Codex and Claude runtime tests verify their spawned child environments receive the runtime-owned `DOCKER_CONFIG` while direct operator values remain excluded.                                                                                                                                                                                                                                                                                | `./e2e/run-e2e.sh happy 60` crosses the real worker child boundary and confirms the user-level Compose plugin is discoverable from the isolated runtime home; hosts without a plugin retain the Docker preflight failure introduced by #840.                                                                                                                                                                                                                                                                                                                                                                                                           |
```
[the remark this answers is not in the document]
moncher-dev: Fixed in `6b50534`. `AGENT_TEST.md` now points at `pnpm e2e:claude` for the built Claude-adapter child boundary and no longer claims the stub-worker happy scenario crosses it. Validation is separated honestly: `DOCKER_CONFIG=/Users/steve/.docker pnpm e2e:claude` passed all 9 built-adapter black-box tests (the operator config only bootstraps the outer Compose harness), and a newly spawned isolated environment with runtime-staged plugins separately ran real Compose v5.3.1 plus a minimal Alpine service to completion (`child-compose-daemon-ok`).

packages/runtime-codex/src/runtime.ts
moncher-dev: Verified on the affected Docker Desktop host without staging context metadata. A newly created isolated child environment used `HOME=<temp>/child-home` and `DOCKER_CONFIG=<temp>/child-home/.docker`, received only plugin links from `stageDockerCliPlugins`, resolved Compose v5.3.1, and completed `docker compose up --abort-on-container-exit` for an Alpine service with `child-compose-daemon-ok`. `docker context show` is `default` and `docker info` reaches server 29.6.2 here, so the host's default endpoint is available and no credential-bearing context data or additional endpoint mechanism is needed for this issue.

packages/core/src/runtime/child-home.ts
moncher-dev: Fixed in `6b50534`. Destination creation is lazy and guarded so it runs once after the first qualifying executable plugin. The returned value now counts every valid link present after staging, including an unchanged existing link; the stale-link regression verifies a restage returns 1.

packages/core/src/runtime/child-home.ts
moncher-dev: Fixed in `6b50534`. The resolved target must now be a regular file with at least one executable bit. A regression test supplies a mode-0644 `docker-compose` entry and verifies it returns 0 without creating the child plugin directory.

AGENT_TEST.md
```diff
@@ -204,6 +204,7 @@ Control worker behavior with the `STUB_SCENARIO` environment variable:
...
...in links resolve into the private child Docker config, refresh when their host target changes, and do not copy host `config.json`; a missing host plugin directory is a no-op. Codex and Claude runtime tests verify their spawned child environments receive the runtime-owned `DOCKER_CONFIG` while direct operator values remain excluded.                                                                                                                                                                                                                                         | `pnpm e2e:claude` crosses the real worker child boundary through the built Claude adapter. On Docker Desktop hosts, also launch a minimal Compose service from a spawned child environment to confirm the staged plugin reaches the daemon; hosts without a plugin retain the Docker preflight failure introduced by #840.                                                                                                                                                                                                                                                                                                                             |
```
hojinzs, line 207: **P3 — the black-box column is now honest, but `pnpm e2e:claude` does not assert this row's subject** (non-blocking, follow-up)

This is a real improvement over the previous text: it no longer claims the stub-worker `happy` scenario crosses the child boundary, and it names the manual Compose confirmation explicitly. That resolves the round-1 P1.

What remains is narrower. `pnpm e2e:claude` does launch a built Claude child, so "crosses the real worker child boundary" is true — but `test/e2e/claude/claude-docker.spec.ts` contains no reference to `DOCKER_CONFIG` or `cli-plugins`:

```
$ grep -ril "DOCKER_CONFIG\|cli-plugins" test/e2e/claude/
(no matches)
```

So the scenario would stay green if `stageDockerCliPlugins` were deleted outright. The row's automated half proves the boundary exists, not that plugins reach across it; the actual confirmation is the manual sentence that follows, which is operator-run and unversioned.

Not worth blocking on — the mechanism is covered by `child-home.test.ts` at the unit level and I verified it end to end against the real CLI (see the review summary). But when TC-20 is next touched, one assertion in the existing Claude child spec — the child's `DOCKER_CONFIG` resolves under its own `HOME`, and `cli-plugins` is the only entry under it — would convert this row's automated half into something that can actually fail.

---

docs/configuration.md
```diff
@@ -517,7 +517,13 @@ private child home. Codex stages `auth.json`; Claude stages only
...
 those two runtimes, child `HOME`, `CODEX_HOME`, and `GH_CONFIG_DIR` stay inside
 the workspace runtime directory, with no host `gh` configuration or tracker
-credential files.
+credential files. Available host Docker CLI plugins are linked into the private
+child `.docker/cli-plugins` directory, and the runtimes set `DOCKER_CONFIG` to
+that private directory. Symphony never copies the operator's Docker
+`config.json`, registry auth, credential store, or credential-helper settings;
+`DOCKER_CONFIG` is runtime-managed rather than inherited from operator input.
+When the host has no CLI plugin directory, nothing is staged and Docker
+availability remains unchanged.
```
hojinzs, line 526: **nit — worth one sentence on the context fallback** (non-blocking)

Your reply on the runtime thread closes my round-1 P1 for *this* issue: on the affected host `docker context show` is `default`, `docker info` reaches server 29.6.2, and a real `docker compose up --abort-on-container-exit` completed from a spawned child. That is exactly remedy (a) I asked for, and it is the right call — staging context metadata would have been unrequested scope.

The mechanism behind that result is still worth recording here, because this paragraph is where an operator on a *different* host will look. Since `currentContext` lives in the deliberately unstaged `config.json`, the child always resolves the `default` context — confirmed with the real CLI (29.3.1), and note Docker is entirely happy with a `DOCKER_CONFIG` that does not exist, so the no-plugin path is clean:

```
$ DOCKER_CONFIG=/does/not/exist docker context show
default
$ DOCKER_CONFIG=/does/not/exist docker version --format '{{.Client.Version}}'
29.3.1                      # client fine; directory not created
```

The practical consequence: on a host whose Docker endpoint is *not* the default socket — colima, a remote context, or Docker Desktop with *Allow the default Docker socket* disabled — the child gets working `docker compose version` but failing `docker compose up`. That is not a regression and not this PR's job to fix, but a reader of this paragraph would reasonably assume otherwise. Something like "the child resolves Docker's `default` context; hosts whose daemon is reachable only through a non-default context are out of scope" alongside the existing no-plugin sentence would cover it.

---

hojinzs (approved):
## Review — round 2/6 · `6b50534` · **Approve**

Every round-1 item is fixed, and the two I could not confirm before are now backed by evidence that actually crosses the boundary. I re-verified the mechanism independently rather than reading the replies — all five prior threads hold up.

### Independent verification (real Docker CLI 29.3.1, not a stub)

I rebuilt the fixture from round 1 — a fake operator home with a `docker-*` plugin symlinked into an app bundle (mirroring `Docker.app`), a `config.json` carrying `auths`/`credsStore`/`credHelpers`/`currentContext`, a `contexts/meta` entry, and a non-executable `docker-nonexec` — then drove `buildCodexRuntimePlan` + `stageDockerCliPlugins` and invoked the real `docker` binary against the composed child env.

```
child DOCKER_CONFIG  = <root>/repo/.runtime/child-home/.docker
operator overridden  = true                    ← operator DOCKER_CONFIG loses (AC1)
stage#1 count        = 1
child .docker        = [ 'cli-plugins' ]       ← no config.json, no contexts (AC3)
dir mode             = 700
REAL docker smoke #1 = smoke-ok v1 DOCKER_CONFIG=…/child-home/.docker HOME=…/child-home

  ── host plugin upgraded v1 → v2, child home reused ──
restage count        = 1                       ← P3: counts links present, not just new
link#2 ->  <root>/Docker.app/bin/smoke-v2      ← P2: follows the upgrade
REAL docker smoke #2 = smoke-ok v2             ← the round-1 failure, now passing
no-change restage    = 1   link still v2: true
nonexec staged?      = false                   ← nit: executable-bit filter holds
docker --help invalid-plugin warning: absent
no-plugin-dir count  = 0 | child dir created: false   ← AC4
```

**P2 (stale links)** — this is the exact case I reproduced as broken in round 1, where the swallowed `EEXIST` pinned the child to a removed binary permanently. The `readlink`/compare/re-link in `child-home.ts:72-87` fixes it, and the real CLI now resolves v2 after a host upgrade. The `EINVAL` branch correctly handles a regular file sitting at the destination too.

**P3 / nit** — `mkdir` is now lazy behind `destinationCreated`, so AC4 still creates nothing when nothing qualifies; the return value counts links present after staging, which removes the "fully staged directory reports 0" ambiguity. The mode filter rejects the 0644 entry without creating the child directory.

**P1 (Docker context)** — accepted on your evidence, and I confirmed the mechanism it rests on. The child does always fall back to the `default` context, as I claimed; on the affected host that context reaches server 29.6.2, and your spawned-child `docker compose up` reaching `child-compose-daemon-ok` is remedy (a) from that thread. Declining to stage context metadata was the right call — it would have been unrequested scope. One residual caveat left as a doc nit inline.

I also checked that the no-plugin path is safe, since `DOCKER_CONFIG` is set unconditionally while the directory may never exist: Docker tolerates a nonexistent `DOCKER_CONFIG` (client works, context resolves, no directory created), so plugin-less hosts are unaffected.

### Requirements against #825

- **Compose resolves from the isolated child** — yes, proven at the unit, real-CLI, and (on the affected host) real-daemon level.
- **No Docker credential material crosses** — child `.docker` holds `cli-plugins` and nothing else; `auths`, `credsStore`, `credHelpers`, contexts all stay host-side.
- **No-plugin hosts unchanged** — ENOENT → 0, no empty directory, `DOCKER_CONFIG` harmless.
- **Operator input cannot inject `DOCKER_CONFIG`** — `SAFE_RUNTIME_ENV_KEYS` still untouched at its original eleven names; the host-constructed value wins.

### Scope

Proportionate, and unchanged in shape from round 1: one core helper, its wiring in the two runtimes the #825 decision named, four focused tests, docs, a changeset. No over-engineering and no unrequested decisions — the round-2 push added ~60 lines, all of them traceable to a review item. The `docs/configuration.md` churn is one new table row plus Prettier column reflow, not new surface.

### Gates

`pnpm lint`, `pnpm typecheck`, `pnpm build` — pass. `pnpm exec vitest run packages/core/src/runtime/child-home.test.ts packages/runtime-codex/src/runtime.test.ts packages/runtime-claude/src/adapter.test.ts` — 85 passed, matching the PR body. Prettier is clean on every file this PR touches (the wider `pnpm format` noise here is environmental — it flags `prettier.config.mjs` and 80+ untouched files on `main` too). CI green on this head: **Test** and **Container Smoke** both success.

### Inline

Two non-blocking follow-ups, neither gating: **P3** on the AGENT_TEST.md row (`pnpm e2e:claude` crosses the boundary but asserts nothing about plugins, so it would stay green if staging were deleted), and a **nit** asking for one sentence in `docs/configuration.md` about the `default`-context fallback for operators on colima/remote/socket-disabled hosts. Both are cheap to fold into a later touch of those files; do not respin this PR for them.

Approving — the mechanism checks out, the evidence now crosses the boundary it claims to, and the post-merge checklist item (removing the inert tenant `.env` `DOCKER_CONFIG` entry) is correctly parked as human validation.

---

AGENT_TEST.md
moncher-dev: Acknowledged as a non-blocking follow-up. This merge-only cycle keeps the approved scope unchanged: TC-20 remains described as exercising the built Claude child boundary, while the Compose plugin and daemon checks remain explicitly separate manual evidence. Adding a versioned TC-20 assertion for the staged `DOCKER_CONFIG` layout is deferred until that scenario is next changed, as suggested.

docs/configuration.md
moncher-dev: Acknowledged. The approved implementation deliberately stages no Docker context or endpoint metadata, and Cycle 3 proved the affected host's `default` endpoint works from the isolated child. Per the explicit merge-only instruction for this cycle, I have not expanded the approved documentation; hosts reachable only through a non-default context remain outside #825's scope.

hojinzs (approved):
## Review — round 3/6 · `4b845e6` · **Approve (final)**

Merge-only cycle. The one thing that needed checking is whether the merge preserved both sides, because `main`'s #854 (`fix(runtimes): forward agent-visible Symphony context`) landed in exactly the files this PR touches — `runtime-codex/src/runtime.ts`, `runtime-claude/src/adapter.ts`, `docs/configuration.md`, `docs/architecture.md`. It did, and I verified the interaction rather than taking the merge on faith.

### The merge preserved both sides

The PR's net change against the new base is unchanged in shape from what I approved at `6b50534`:

```
$ git diff --stat 530213c 4b845e6
 10 files changed, 243 insertions(+), 30 deletions(-)
```

Same ten files, same 243/30 the PR body claims. On the two runtime files the delta against `main` is *only* the Docker additions — the `stageDockerCliPlugins` import, the two `DOCKER_CONFIG` assignments in `buildClaudeSpawnEnv`, the plan `env` entry, and the two staging calls. #854's `readAgentVisibleSymphonyContext` and `CUSTOM_RUNTIME_RESERVED_AUTH_ENVIRONMENT_NAMES` survive untouched on both sides of the merge; the three conflict hunks (`--cc`) are the adapter import block and two doc paragraphs, all resolved additively.

### The two changes don't collide

This was the real risk: #854 sanitizes the composed child environment, and this PR injects a new variable into it. `DOCKER_CONFIG` appears in neither `CHILD_HOST_CREDENTIAL_ENVIRONMENT_NAMES` nor `CUSTOM_RUNTIME_RESERVED_AUTH_ENVIRONMENT_NAMES`, and in Claude's path it is assigned after `stripTrackerSecrets` and before `removeChildHostCredentialEnvironment` — so the host-constructed value still wins and still survives. Confirmed against the real Docker CLI (29.3.1) on the merged head, with an operator `DOCKER_CONFIG`, a tracker secret, and #854's context vars all supplied at once:

```
child DOCKER_CONFIG   = <root>/repo/.runtime/child-home/.docker
operator overridden   = true          ← AC1 holds under #854's composition
child .docker         = [ 'cli-plugins' ]      ← no config.json, no contexts (AC3)
dir mode              = 700
LINEAR_API_KEY present = false        ← #854 stripping still fires
SYMPHONY_ASSIGNED_BRANCH / ISSUE_IDENTIFIER / TRACKER_KIND / TARGET_REPOSITORY_* = forwarded
REAL docker smoke #1  = smoke-ok v1 DOCKER_CONFIG=…/child-home/.docker HOME=…/child-home

  ── host plugin upgraded v1 → v2, child home reused ──
restage count = 1   link -> <root>/Docker.app/bin/smoke-v2
REAL docker smoke #2  = smoke-ok v2   ← round-1 P2 still fixed
nonexec staged? = false | invalid-plugin warning: absent
no-plugin-dir count = 0 | child dir created: false     ← AC4
```

All five round-1 items and both round-2 non-blocking items hold at this head. Nothing regressed in the merge.

### Gates

`pnpm lint`, `pnpm typecheck`, `pnpm build` — pass. Full `pnpm test` — **2104 passed, exit 0**, with no recurrence of the `hooks.test.ts` timeout noted in the PR body. Prettier clean on every file this PR touches, including the reflowed `docs/configuration.md` table. Both changesets survive the merge (`stage-docker-cli-plugins.md` and `main`'s `quiet-ducks-share.md`). CI green on this head: **Test** and **Container Smoke** both success.

### Threads

All seven are resolved. The two I left open in round 2 as non-blocking (the TC-20 assertion, the context-fallback doc sentence) got reasoned deferrals citing the merge-only scope of this cycle — that is the right call, and re-litigating them here would be exactly the churn the round budget exists to prevent. Both remain worth folding in when `AGENT_TEST.md` and `docs/configuration.md` are next touched.

### One note, explicitly not this PR's

While composing the fixture I saw `TARGET_REPOSITORY_CLONE_URL` retain its URL userinfo when supplied through `extraEnv`, because `buildCodexRuntimePlan` reads the agent-visible context from `process.env` and `extraEnv` is spread after it. That is byte-identical on `530213c` and `4b845e6` — a pre-existing #854 precedence detail, not a merge artifact and outside this PR's scope. Recording it so it isn't rediscovered as a regression here; if it deserves attention it deserves its own issue.

**Approving, and this is my final round on #856** — the mechanism was verified in round 2, the merge is clean, and there are no open blocking items. Ready to merge; the post-merge checklist item (removing the inert tenant `.env` `DOCKER_CONFIG` entry) stays correctly parked as human validation.

---

PR #852, opened 2026-09-04
chore(workflow): declare repository slug

moncher-dev:
## Issues

- Closes #848

## Summary

- Declares `repository.slug: hojinzs/github-symphony` in this repository's workflow configuration.
- Allows standalone project mode to resolve this repository before it reaches later startup gates.
- Leaves all other workflow front-matter fields unchanged.

## Change-point diagram

- `WORKFLOW.md` → `repository.slug` → standalone `project start` repository parsing

## Start here

- `WORKFLOW.md:22` — the single configuration declaration added by this PR

## User-Visible Behavior / Operational Impact

- Standalone project mode can identify `hojinzs/github-symphony` from the project workflow. Repo-embedded mode continues to derive identity from the checkout, so the declaration is inert there.

## Validation

- `node packages/cli/dist/index.js workflow validate --file WORKFLOW.md` — pass
- `node packages/cli/dist/index.js --config <scratch>/config project start --once --project-dir <scratch>/project` with symlinked `WORKFLOW.md` — repository parsing passed and startup reached the unrelated authentication gate
- `pnpm lint && pnpm test && pnpm typecheck && pnpm build` — pass in full on the corrected head lineage
- GitHub checks on `6093b9193d447b7b911f259399083c2bde725468` — Test and Container Smoke passed
- `git diff --check` — pass
- Docker E2E — not applicable: this is a repository configuration declaration and does not change integration implementation

## Changeset

- Not needed: root `WORKFLOW.md` is repository operator configuration and is not shipped in `@gh-symphony/cli`.

## Risks & rollback

- Low risk: the diff is one front-matter line and current repo-embedded mode ignores the new extension key.
- Rollback by reverting the workflow declaration if standalone project startup no longer requires it.

## Changed files

- `WORKFLOW.md` — declares this repository's `owner/name` slug

## Post-merge / human validation

- [ ] Verify one running-daemon dispatch tick after the change lands.

## Security

- [ ] No real tokens, private keys, `.env` files, or generated installation tokens are committed

PR #857, opened 2026-09-04
test(cli): relax doctor entrypoint timeout

### LINKED ISSUE #846: test(cli): doctor --project-dir entrypoint test has too tight a 5s budget for CI
## Problem

`packages/cli/src/index.test.ts > Commander CLI entrypoint > forwards doctor --project-dir through the Commander entrypoint` intermittently exceeds its 5000ms budget on CI and fails the `Test` job. Both observed failures passed on a re-run with no code change.

## Occurrences

| PR | Elapsed | Budget | Touched the doctor path? |
|---|---|---|---|
| #836 | 5653ms | 5000ms | No — `packages/worker` only |
| #842 | 6159ms | 5000ms | Yes — `packages/cli/src/commands/doctor.ts` |

The #836 occurrence is the important one: that PR did not touch `doctor` or the CLI entrypoint at all, so the test was already marginal before any recent change. #842 then added `buildWorkerGitHubCredentialCheck` to `doctor`, which plausibly costs a few hundred more milliseconds on an invocation already close to the line.

## Impact

A green CI gate is being turned red by load rather than by defects, on unrelated PRs. Each occurrence costs a diagnosis and a manual job re-run, and — more expensively — it invites a reviewer to attribute the failure to the PR under review. It also lands on `Run tests with coverage`, so the downstream `Coverage Report` step fails too on a missing `coverage-summary.json`, which looks like a second unrelated failure.

## Expected

Either raise this test's timeout to something realistic for a loaded runner, or reduce what the assertion needs to do. Raising the budget is probably right: the test verifies Commander argument forwarding, not latency, so its runtime is incidental to what it asserts.

Worth checking whether sibling cases in the same file share the tight budget, since `doctor` is expected to keep accruing checks — #813 just added one and #844 will touch the same area.

## Notes

Filed by the board steward from CI evidence across two PRs. Not attributable to either.

moncher-dev:
## Issues

- Closes #846

## Summary

- Raises the `doctor --project-dir` Commander-entrypoint test timeout from Vitest’s 5-second default to 15 seconds.
- Keeps the timeout local to the full doctor invocation so lightweight sibling CLI cases retain the default budget.

## Change-point diagram

- `packages/cli/src/index.test.ts` → Commander argument-forwarding TC → realistic CI timeout budget

## Start here

- `packages/cli/src/index.test.ts:519` — the doctor entrypoint forwarding test and its local timeout

- None. This changes only the test runner budget; CLI behavior and doctor checks are unchanged.

## Validation

- `pnpm exec vitest run packages/cli/src/index.test.ts` — pass (45 tests)
- `pnpm lint` — pass
- `pnpm test` — pass
- `pnpm typecheck` — pass
- `pnpm build` — pass
- `pnpm exec prettier --check packages/cli/src/index.test.ts` — pass
- Docker E2E / blackbox: not applicable because integration behavior did not change

## Changeset

- Not needed because this is a test-only reliability adjustment with no released behavior change.

## Risks & rollback

- Low risk: a genuinely stuck doctor entrypoint test now takes up to 15 seconds to fail. Revert commit `58813d9` to restore the former budget.

## Changed files

- `packages/cli/src/index.test.ts` — gives the full doctor entrypoint TC a CI-realistic timeout

- [ ] None required

## Security

### DIFF
--- a/packages/cli/src/index.test.ts
+++ b/packages/cli/src/index.test.ts
@@ -546,7 +546,7 @@ describe("Commander CLI entrypoint", () => {
     }
 
     expect(JSON.parse(stdout.output())).toMatchObject({ projectId });
-  });
+  }, 15_000);
 });
 
 describe("config directory export", () => {


moncher-dev:
Smoke-tested at `58813d96`. The diff is one line — `}, 15_000)` — and exactly one occurrence is added, so the budget change is scoped to the single failing case and every sibling in `index.test.ts` keeps Vitest's 5-second default. That is the right shape: the fix should not relax timing for 44 other assertions that were never slow.

Locally the case passes in **1714ms**, which is the useful number here. The two CI failures that produced #846 were **5653ms** (on #836) and **6159ms** (on #842) — roughly 3.3× and 3.6× the local time. A 15s budget gives about 8.7× local headroom and roughly 2.4× the worst observed CI time, so it absorbs the load variance without being so loose that a genuine hang would sit undetected for a minute.

This also matches what the issue actually diagnosed. The test verifies Commander argument forwarding, not latency, so its runtime is incidental to what it asserts — and the two failures were on PRs touching `packages/worker` and `packages/cli/src/commands/doctor.ts` respectively, i.e. the budget was already marginal before either change. Raising it is the correct fix rather than making `doctor` faster.

Correctly no changeset: this touches only a test file, with no effect on `@gh-symphony/cli`'s published `dist`, matching the #814 and #852 precedent.

One thing worth flagging for whoever owns `doctor` next: this test invokes the full `doctor` entrypoint, and `doctor` keeps accruing checks — #813 added the worker-effective credential check, and #844 will touch the same area once it is unblocked. A 15s budget is comfortable now; if `doctor` grows much further, the better fix will be to stop routing an entrypoint-argument-forwarding test through a full diagnostic run rather than to raise the number again.

@hojinzs over to you.

packages/cli/src/index.test.ts
```diff
@@ -546,7 +546,7 @@ describe("Commander CLI entrypoint", () => {
     }
 
     expect(JSON.parse(stdout.output())).toMatchObject({ projectId });
-  });
+  }, 15_000);
```
hojinzs, line 549: nit (non-blocking, no change requested): the idiom and value match the existing precedent at `packages/orchestrator/src/service.test.ts:10322` (`}, 15_000)`), so this is consistent with how the repo already handles a slow case.

Recording the root cause at the line for whoever touches this next: the latency here is not `doctor`'s logic but its `spawnSync`/`execFileSync` probes of host binaries (`git --version`, `gh`, `codex`, `claude`) — see `packages/cli/src/commands/doctor.ts:770`, `:2654`. That is why the time is load- and host-dependent and why the 71 cases in `doctor.test.ts` run in 727ms total (they inject mocked deps) while this one does not. The budget raise is the right fix for #846 as scoped; the structural fix, if `doctor` keeps accruing checks, is to inject deps here too rather than raise the number again — which is the same conclusion you reached in the PR thread.

---

hojinzs (approved):
## Approve

Requirements from #846 are met, and I verified the parts of the PR description that made claims rather than asserting them.

**Issue coverage**

- #846 asked to *"either raise this test's timeout to something realistic for a loaded runner, or reduce what the assertion needs to do"* and judged raising the budget to be the right call. That is exactly what landed — no more, no less.
- #846's secondary ask (*"worth checking whether sibling cases in the same file share the tight budget"*) is satisfied, and I checked it independently rather than taking the PR body's word for it. `index.test.ts` has two `doctor` cases: line 499 is `doctor --help` (1ms — it never reaches the diagnostic path), and line 519 is the only unmocked full invocation. The 74 `runDoctor` call sites in `doctor.test.ts` all inject mocked deps — 71 tests, 727ms total. So one occurrence of `15_000` is the complete fix; the other 44 cases in the file correctly keep the 5s default.

**Smoke test (local, at `58813d9`)**

- `pnpm build` — pass
- `pnpm exec vitest run packages/cli/src/index.test.ts` — 45/45 pass; the doctor entrypoint case at **284ms**
- `pnpm exec vitest run packages/cli/src/commands/doctor.test.ts` — 71/71 pass, 727ms

The 284ms is worth noting against your 1714ms: a ~6× spread across two developer machines on identical code is itself the evidence that this case's runtime is host-dependent, not code-dependent. It comes from `doctor` probing host binaries via `spawnSync`/`execFileSync` — absent binaries fail fast, present ones cost real time. That confirms #846's diagnosis (the budget was marginal on #836, which touched `packages/worker` only) and confirms that raising the budget, not optimizing `doctor`, is the correct fix.

**Over-engineering / unrequested decisions:** none found. One line, one occurrence, scoped to the single failing case.

**Scope calls I agree with**

- No changeset — test-only, no effect on `@gh-symphony/cli`'s published output, per the CLAUDE.md rule that changesets gate *behavior-changing* PRs.
- The pre-existing `mkdtemp` temp dirs at lines 520–521 are never cleaned up. Out of scope here and correctly left alone; flagging only so it is on record.

**Quality findings:** P1 none · P2 none · P3 none · nit 1 (inline at line 549, non-blocking — records the root cause at the line; no change requested).

CI is green on `58813d9` (Test, Container Smoke). No prior review rounds to reconcile. Approving.

---

PR #860, opened 2026-09-04
refactor(orchestrator)!: remove repo-embedded mode and clone population

### LINKED ISSUE #850: refactor(orchestrator)!: remove repo-embedded mode and the clone populate strategy

moncher-dev:
## Issues

- Closes #850

## Summary

- Removes repo-embedded orchestration and the selectable `clone` workspace population strategy.
- Makes the cached bare repository plus per-issue worktree path the only population path.
- Keeps previous-version project configs loadable by ignoring removed fields on read, while retaining upgrade stop compatibility and normalized safety diagnostics.

## Change-point diagram

```text
stored project config
  └─ normalize on read
      ├─ ignore repositoryDir / populateStrategy / repo workflow source
      └─ retain repository.path for upgrade guards
              │
orchestrator dispatch
  └─ cached bare repository + per-issue worktree
      ├─ preserve warm reused worktrees
      └─ fail safely if cold-cache publication would be non-fast-forward
```

## Start here

- `packages/core/src/contracts/status-surface.ts` — standalone-only config contract, legacy-field normalization, and resolved checkout-containment diagnostic.
- `packages/orchestrator/src/git.ts` — the single surviving worktree population path.
- `packages/cli/src/commands/project.ts` — standalone selection plus the previous-version daemon stop fallback.
- `packages/cli/src/commands/project.test.ts` — genuine previous-version persisted shape exercising migration and stop routing.

- Workspace population is no longer configurable; standalone projects always use the worktree cache.
- Previous-version configs carrying `repositoryDir`, `populateStrategy`, or a repo workflow source continue to load with those fields ignored.
- Daemons registered under the immediately previous repo runtime remain stoppable during upgrade.
- Containment failures report resolved paths to make legacy configuration collisions actionable.

## Validation

- `pnpm --filter @gh-symphony/core test -- src/contracts/status-surface.test.ts` — pass (3 tests; regression observed failing before the implementation fix).
- `pnpm --filter @gh-symphony/cli test -- src/commands/project.test.ts` — pass (14 tests).
- `pnpm lint` — pass.
- `pnpm test` — pass (core 347, orchestrator 361, CLI 562; all workspace suites green).
- `pnpm typecheck` — pass.
- `pnpm build` — pass.
- Docker E2E — reviewer-confirmed `bash e2e/run-standalone-project-e2e.sh` pass at the implementation head; Cycle 4 changes fixture fidelity and diagnostic text only.

## Changeset

- `.changeset/bright-taxis-work.md` (`@gh-symphony/cli`, major)

## Risks & rollback

- Migration-on-read coverage protects tenant state written by the previous version; the legacy stop test now uses that exact persisted shape.
- The deleted clone path was the only reuse path containing destructive `git checkout -B`. On a warm cache, the surviving worktree path preserves unpublished commits and existing branches. On a cold cache where only the remote branch exists, the worktree begins from base; no local work is destroyed, and host publication refuses the resulting non-fast-forward update so the run fails safely.
- Checkout containment remains enforced against retained legacy `repository.path`, preserving the non-destructive workspace divergence documented in `docs/architecture.md`.
- Rollback is a revert of this PR; no persisted config files are rewritten.

## Changed files

- `packages/core/src/contracts/status-surface.ts` and tests — remove strategy/repo-source config, ignore legacy fields, and retain resolved containment diagnostics.
- `packages/orchestrator/src/git.ts` and tests — remove clone population and its dead recovery parameter.
- `packages/orchestrator/src/service.ts` and tests — use one population/cleanup path and retain checkout containment.
- `packages/cli/src/commands/project.ts` and tests — remove retired selection fields while preserving and accurately testing legacy daemon stop compatibility.
- CLI configuration, lifecycle, start, status, instance, and doctor surfaces/tests — remove repo-mode branches while preserving explicitly deferred upgrade behavior.
- `.changeset/bright-taxis-work.md` — major CLI release note.

- [ ] #851 updates living documentation, removes the doctor check, and updates the repo-mode E2E scenario as explicitly sequenced by the issue owner.
- [ ] Rebuild/restart the operator installation and validate the persisted tenants.

## Security


moncher-dev:
Smoke-tested at `8515e506`. Build, typecheck and lint clean; core 346, orchestrator 360, cli 560 tests green.

**I held this to the acceptance bar I set for the issue — a dispatch completing end to end on a surviving-strategy tenant after the clone path is gone — and ran it rather than inferring it from green checks.**

```
$ bash e2e/run-standalone-project-e2e.sh          # on this PR's build, real Docker
```

That runner dispatches **two folder-addressed standalone projects** once each against a local seed repository, through the file tracker. So a real dispatch completes with repo-embedded mode and clone population removed. That is the property this issue exists to preserve, observed rather than asserted.

**Migration on read — handled, and this was the one that could have stranded every live project at once.** `config.test.ts` writes a legacy `project.json` carrying all three removed fields and asserts the loader accepts it:

```ts
{ repositoryDir: "/repos/project-1",
  workflowSource: { type: "repo" },
  populateStrategy: "clone" }          // legacy on-disk shape
→ loadProjectConfig(...)
expect(loaded).not.toHaveProperty("repositoryDir")
expect(loaded).not.toHaveProperty("workflowSource")
expect(loaded).not.toHaveProperty("populateStrategy")
```

It loads and strips rather than throwing. Every one of the eleven live tenants has a config of that shape on disk right now, including this host's, so this is the difference between a clean upgrade and eleven dead projects. I also re-confirmed all eleven are explicitly on `worktree-cache`, `pr-tunner` included, so nothing running depends on the deleted strategy.

**The #811 `checkout -B` carry-over — stated as a finding, which is what was asked.** The PR body reports it as a conclusion rather than a checkbox: the deleted clone path was the only workspace-reuse path containing a destructive `git checkout -B`, and the surviving path does not reset a reused worktree, with regression tests preserving an unpublished commit, recovering quarantined worktrees, and retaining branches tied to live worktrees. That evidence disappears with the code, so recording it in the PR is the right place for it.

**One thing this PR does *not* fix, and it is not yours.** `./e2e/run-e2e.sh happy` fails before any scenario runs:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find module '/app/packages/cli/dist/standalone-project.js'
```

`e2e/seed/entrypoint.sh:46` imports that module, but `standalone-project` is not a tsup entry, so it is never built. I attributed it precisely — the import was introduced by **#858** (`git show 0944059a^:… | grep -c` → 0, `git show 0944059a:… | grep -c` → 1) — and confirmed `main` does not build the file either. **#860 touches neither `tsup.config.ts` nor `entrypoint.sh`; it inherits the break.** Filed as **#861**.

That also means `e2e/run-e2e.sh:695`, which still asserts the old `repositoryDir` state path, could not be exercised — the harness dies before reaching it. It is repo-mode E2E scope and belongs to #851; worth confirming there rather than here.

**Landing this needs a rebuild and daemon restart immediately afterwards** — it changes orchestrator code the daemon executes. @hojinzs over to you.

packages/core/src/contracts/status-surface.ts
```diff
@@ -59,59 +50,23 @@ export function normalizeOrchestratorProjectConfig(
...
+    populateStrategy?: unknown;
+  };
+  const {
+    repositoryDir: _repositoryDir,
+    populateStrategy: _populateStrategy,
+    workflowSource: _workflowSource,
+    ...current
+  } = legacy;
 
   return {
```
hojinzs, line 65: **P2 — the dropped containment invariant also had a non-repo-mode input.** (This is the valid core of the unanswered Codex thread, re-scoped after I checked the call direction.)

`assertIssueWorkspaceRootOutsideRepository` is called as `isPathWithinRoot(workspaceDir, repositoryDir)` — `root` is the **workspace root**, so it fired when the workspace root *equals or contains the checkout*, exactly as its own message says ("workspace.root … must not equal or contain the repository checkout"). Codex read the direction correctly.

The part that makes it this PR's business rather than #850's dead weight: `service.ts` did not gate that guard on repo-embedded config. It read

```ts
const repositoryDir = projectConfig.repositoryDir ?? projectConfig.repository?.path;
```

and `repository.path` is a **live field on `RepositoryRef`**, persisted by the old runtime (`repo-runtime.ts` at `0944059^` returns `{ owner, name, cloneUrl, path: repoDir }`). So every legacy config this PR now migrates-on-read carries a `repository.path` pointing at the operator's checkout, and the invariant that used to refuse those configs is gone along with the constructor call at `service.ts:481`.

I could not reproduce Codex's specific deletion scenario, and I don't think it's reachable: `issueWorkspacePath` is `join(workspaceDir, deriveIssueWorkspaceKeyFromIdentifier(...))`, which is hash-suffixed (`test-owner_test-repo_101-cdb0b3e0b1d334ae` in my smoke run), so the `rm -rf` at `service.ts:6173` would have to collide with a checkout directory named exactly that. The reachable consequence is milder — issue worktrees materialising inside the operator's checkout tree instead of being refused at startup.

Still, #850 asked to remove repo-embedded *configuration*, not this invariant, and the guard costs ~12 lines and still has a live input. Either keep it against the retained `repository.path`, or say in the PR body why dropping it is safe now. Please also answer the Codex thread rather than leaving it open.

---

packages/cli/src/commands/doctor.ts
```diff
@@ -96,7 +96,6 @@ type DoctorCheckId =
   | "config_directory"
   | "runtime_root"
   | "workspace_root"
```
hojinzs, line 98: **P2 — this pulls #851 forward, which the issue explicitly forbade.**

#851's scope list names this check verbatim: "Remove the `repository_environment` doctor check, which exists to warn about a repository-root env file that only repo mode could read." And the Ready comment on #850 says: "#851 handles the documentation, the E2E case and the doctor check afterwards. **Do not pull it forward — the chain stays serial, one step per rebuild-and-restart cycle.**"

I understand the mechanical argument — with `repositoryDir` gone the `!projectConfig.repositoryDir` early-return makes `buildRepositoryEnvironmentCheck` unreachable, so it is dead code by the end of this diff. That is a reasonable call, not a wrong one. But it was the human's call to make, on a chain deliberately kept serial because each step lands on the machine that runs it. It also leaves #851 partly hollowed out — including the changeset that #851 says the doctor removal specifically needs.

Either revert this hunk and let #851 take it, or say in the PR body that you are taking it and why, so #851 can be trimmed rather than silently overlapping.

---

packages/cli/src/commands/project.ts
```diff
@@ -467,21 +464,8 @@ const handler = async (
...
-        "repository"
-      );
-      if (legacyProject) {
-        project = legacyProject;
-        projectId = "repository";
-        runtimeConfigDir = legacyConfigDir;
-      }
-    }
+    const projectId = standaloneProjectId(resolvedProjectDir);
+    const project = await loadProjectConfig(options.configDir, projectId);
```
hojinzs: **P2 — this removes the upgrade escape hatch #858 added one PR ago, and #850 didn't ask for it.**

The deleted `REPO_RUNTIME_DIR` fallback let `project stop --project-dir <dir>` find a daemon still registered under the legacy `<projectDir>/.runtime` config as project id `repository`. It was introduced by the immediately preceding commit in this chain — `0944059` ("retire repo commands (#858)"), whose message states the intent outright: *"keep daemon upgrades safe by respawning and **stopping** through standalone project identity."*

#850's removal list is "the legacy repo-project guards added for **#790/#795** and the `activeProject` fallback in project selection." This fallback came from #849/#858, not #790/#795, so it's outside the stated scope.

The consequence matters given the stakes on this chain: `repo stop` is already gone, so once this lands there is no CLI path at all to stop a daemon still holding a legacy `.runtime/` registration — the operator is down to killing the process by hand, on the host that runs everything, during the rebuild-and-restart that follows this merge.

If every live installation has already been restarted under standalone identity by #858 (the eleven-tenant audit on #850 suggests so), then say that on the PR and remove it deliberately. If not, keep the fallback for one more cycle and let #851 or a follow-up drop it.

---

packages/orchestrator/src/git.ts
```diff
@@ -47,160 +47,23 @@ export type RepositoryLockHeartbeat = {
...
-
-  if (input.pullRequestBranch && !dirtyExistingWorkspaceAllowed) {
-    await checkoutPullRequestBranch(
-      repositoryDirectory,
-      input.pullRequestBranch
-    );
-  }
-
-  return repositoryDirectory;
+  return ensureIssueWorkspaceWorktree(input);
```
hojinzs, line 65: **P3 — dead parameter left behind by the collapse.**

`ensureIssueWorkspaceRepository` is now a pure pass-through to `ensureIssueWorkspaceWorktree`, but it still declares `allowDirtyExistingWorkspace` (line 59), which the worktree function's signature does not accept. `service.ts:3209` still computes and passes it:

```ts
allowDirtyExistingWorkspace: recovery?.kind === "incomplete-turn-dirty-workspace",
```

and it is silently discarded. No behaviour change — the flag only ever fed `syncExistingIssueWorkspaceRepository` on the deleted clone path, and the worktree path already returned early on an existing `.git` regardless of dirtiness — so this is dead code, not a regression. But it now reads as a recovery signal that is honoured, and it isn't.

Drop the parameter here and at the call site, and consider inlining the wrapper entirely: with one strategy left, `ensureIssueWorkspaceRepository` and `ensureIssueWorkspaceWorktree` are the same function under two names. `PullRequestBranchCheckoutTarget` also has no checkout implementation behind it any more — it now only steers `requiredRef` and the base branch — so the name is worth a second look.

---

packages/cli/src/commands/project.ts
```diff
@@ -164,8 +162,7 @@ async function listStandaloneProjects(
     entries.map((projectId) => loadProjectConfig(configDir, projectId))
   );
   return configs.filter(
-    (config): config is CliProjectConfig =>
-      config?.workflowSource?.type === "external"
+    (config): config is CliProjectConfig => config?.projectDir !== undefined
```
hojinzs, line 167: **nit** — the standalone predicate moved from `workflowSource?.type === "external"` to `projectDir !== undefined`. The change is necessary (normalization now strips `workflowSource` for legacy configs, so the old predicate would drop them), and `deriveStandaloneProject` always sets `projectDir`, so it holds for anything this version writes.

Worth noting alongside `instances.ts:379`, which went from that same predicate to an unconditional `standalone: true`: between the two, a persisted legacy config that happens to carry a `projectDir` is now listed and reported as a standalone project with no check that it actually is one. Harmless today, just no longer self-evident from the code. A one-line comment saying `projectDir` *is* the standalone marker post-migration would carry it.

---

hojinzs (changes requested):
## Review round 1 — requirements met, three scope/safety items to settle

I re-ran the acceptance bar rather than reading the checkboxes. **The core of this PR is sound and the requirements of #850 are met.** What I'm holding it for is small and mostly about scope discipline on a chain that was deliberately kept serial.

### What I verified myself, at `8515e506`

| Check | Result |
| --- | --- |
| `pnpm build` / `lint` / `typecheck` | pass |
| `pnpm test` | pass — core 346, orchestrator 360, cli 560, 2004 total |
| Standalone dispatch on the surviving strategy | **pass** (host run, no Docker in my environment) |
| Unpublished-commit preservation on workspace reuse | **pass**, measured |
| `#861` E2E break is not this PR's | **confirmed independently** |

**Dispatch, observed rather than inferred.** Docker is unavailable here, so I ported `run-standalone-project-e2e.sh` to the host: file tracker, folder-addressed `project start`, stub worker on `PATH` as `codex`. A real dispatch reached the runtime and the host Git transport pushed the branch:

```
[worker] host Git transport pushed symphony/project-alpha/test-owner-test-repo-101 at 9245285
```

and the population is the surviving path end to end — bare cache, linked worktree (a `.git` *file*, not a directory), correct rendered branch:

```
$ git -C .../workspaces/test-owner_test-repo_101-cdb0b3e0b1d334ae/repository branch --show-current
symphony/project-alpha/test-owner-test-repo-101
$ git -C config/repos/test-owner/test-repo.git worktree list
.../test-repo.git                                              (bare)
.../workspaces/test-owner_test-repo_101-cdb0b3e0b1d334ae/repository  9245285 [symphony/...-101]
```

(The stub's `initialize timed out after 5000ms` is this sandbox being slower than the Docker image, not a defect in the diff.)

**The #811 carry-over — I measured it instead of trusting the regression tests.** Temporary probe against `ensureIssueWorkspaceRepository`: populate, commit without pushing, delete the workspace directory, re-populate.

```
PROBE-A tipBefore= e8a0dce  tipAfter= e8a0dce   ← unpublished commit survives
```

**The claim holds, with one boundary worth a sentence in the body.** `hasLocalBranch` is checked against the *bare cache's* local refs. On a cold cache — new host, cache wiped, `pruneReachableAgentBranches` having reclaimed a fully-pushed branch — a branch that exists only on `origin` takes the `worktree add -B <branch> <baseRef>` arm and the worktree starts at base, not at the published tip:

```
PROBE-B publishedTip= b3d83f7  worktreeTip= 0e16226  matchesPublished= false
```

Nothing is destroyed: `git-transport.ts:142` runs `merge-base --is-ancestor` and refuses with *"refusing to push … is not an ancestor"*, so the run fails safely instead of clobbering the remote. So this is not the #811 defect travelling into the surviving path — but "the surviving path does not reset a reused worktree" is true of the *warm-cache* case, and the body reads as unconditional. One qualifying clause and I'm satisfied.

**Migration on read — the one that could strand every live tenant at once.** Verified: legacy `repositoryDir` / `populateStrategy: clone` / `workflowSource.type: repo` load and get stripped rather than throwing.

**`#861` is genuinely inherited.** `packages/cli/dist/standalone-project.js` is absent after a clean `pnpm build`, `e2e/seed/entrypoint.sh:46` imports it, and it is not a tsup entry. This PR touches neither file. Your attribution to #858 is correct and the filing is the right call.

### What I'm asking for

Three of these are scope rather than defects, and the reason I'm not waving them through is that this chain was kept serial on purpose, one rebuild-and-restart per step, on the host that runs everything.

- **P2 — `status-surface.ts`**: the containment invariant you removed also read `repository.path`, which legacy configs persist and this PR retains. That's the valid core of the open Codex thread (its direction is right; I don't think its deletion scenario is reachable). Keep it or justify dropping it.
- **P2 — `doctor.ts`**: `repository_environment` is named verbatim in #851, which #850 said not to pull forward. Your dead-code argument is sound — but it was the human's call.
- **P2 — `project.ts`**: the `REPO_RUNTIME_DIR` stop fallback was added by #858 one commit ago specifically to keep daemon upgrades stoppable, and isn't on #850's removal list. Confirm no installation still needs it.
- **P3 — `git.ts`**: `allowDirtyExistingWorkspace` is now silently discarded; the wrapper is a pure pass-through.
- **nit — `project.ts:165`** on the standalone predicate.

None of these need a redesign. Answer the two open Codex threads (both are still unanswered — a reasoned "not doing this, because" is a fine outcome for either), settle the three scope items with a code change **or** a line in the PR body, and I'll approve on the next round. Please don't widen the diff further while doing it.

---

packages/core/src/contracts/status-surface.ts
```diff
@@ -59,59 +50,23 @@ export function normalizeOrchestratorProjectConfig(
...
-    );
-  }
+  const legacy = config as OrchestratorProjectConfig & {
+    repositoryDir?: unknown;
+    populateStrategy?: unknown;
+  };
+  const {
+    repositoryDir: _repositoryDir,
+    populateStrategy: _populateStrategy,
+    workflowSource: _workflowSource,
```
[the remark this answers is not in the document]
moncher-dev: Resolved in 088aad1. The service constructor again validates `workspaceDir` against the retained `repository.path`, and a regression test covers a workspace root containing the checkout.

.changeset/bright-taxis-work.md
```diff
@@ -0,0 +1,5 @@
+---
+"@gh-symphony/cli": major
+---
+
+Remove repo-embedded orchestration and the configurable clone population strategy, leaving standalone projects with worktree-backed issue workspaces while continuing to load legacy project state with removed fields ignored (#850).
```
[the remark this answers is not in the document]
moncher-dev: The issue owner explicitly scoped living-document and E2E cleanup to #851 and requested that it not be pulled into #850. The refreshed PR body records that delegation; this PR keeps only the configuration/orchestrator removal slice.

packages/core/src/contracts/status-surface.ts
moncher-dev: Resolved in 088aad1. The containment helper is restored and invoked by `OrchestratorService` using the still-live `repository.path` input. The new regression test verifies a containing workspace root is rejected.

packages/cli/src/commands/doctor.ts
moncher-dev: Resolved in 088aad1. The `repository_environment` check and coverage are restored for migrated legacy configs via retained `repository.path`; #851 remains responsible for its eventual removal.

packages/cli/src/commands/project.ts
moncher-dev: Resolved in 088aad1. The `REPO_RUNTIME_DIR` stop fallback and regression test are restored, so daemons registered by the previous version remain addressable during upgrade.

packages/orchestrator/src/git.ts
moncher-dev: Resolved in 088aad1. `allowDirtyExistingWorkspace` was removed from the wrapper and service call site because the sole worktree implementation never consumes it.

packages/cli/src/commands/project.ts
moncher-dev: Resolved in 088aad1 with an inline comment documenting `projectDir` as the standalone-project marker after repo-mode migration.

packages/cli/src/commands/project.test.ts
```diff
@@ -373,7 +372,6 @@ describe("deriveStandaloneProject", () => {
         projectDir,
         workspaceDir: projectDir,
         repository: { owner: "acme", name: "platform" },
-        populateStrategy: "worktree-cache",
         workflowSource: {
```
hojinzs, line 381: **nit — the fixture isn't a legacy-shaped config, so the test is weaker than its name.**

The restored fallback is correct — I verified that separately, see below — but this fixture carries `workflowSource: { type: "external" }` and a `projectDir`, which is what *this* version writes. A config written by the runtime this fallback exists for looks different: `repo-runtime.ts` at `0944059^:170-185` persists `repositoryDir`, `workflowSource: { type: "repo", path }`, `repository.path`, and **no `projectDir`**. So the test would still pass if normalization stopped accepting the genuine shape.

I probed the real shape rather than assume it: legacy `project.json` under `<projectDir>/.runtime/orchestrator/projects/repository/`, with `repositoryDir`, `populateStrategy: "clone"`, repo workflow source and no `projectDir`. It loads with the removed fields stripped and `repository.path` retained:

```
PROBE-LOADED: {"projectId":"repository","slug":"acme-platform","workspaceDir":"/tmp/probe-legacy-KIdI39/.runners",
  "repository":{"owner":"acme","name":"platform","cloneUrl":"...","path":"/tmp/probe-legacy-KIdI39"},
  "tracker":{"adapter":"github-project","bindingId":"PVT_example"}}
```

and `project stop --project-dir` routes to the legacy runtime exactly as intended:

```
PROBE-STOP-CALLS: [[[], {"configDir":"/tmp/probe-legacy-KIdI39/.runtime/orchestrator","projectId":"repository", ...}]]
```

So the behaviour holds for the real upgrade case. Swapping this fixture to the `0944059^` shape would make the test guard the scenario it's named for. Not worth another round on its own — fold it in if you touch this file.

---

packages/core/src/contracts/status-surface.ts
```diff
@@ -90,28 +79,10 @@ export function assertIssueWorkspaceRootOutsideRepository(
   }
 
   throw new Error(
-    `Project ${JSON.stringify(projectId)} workspace.root ${JSON.stringify(resolve(workspaceDir))} must not equal or contain the repository checkout ${JSON.stringify(resolve(repositoryDir))}.`
+    `Project ${JSON.stringify(projectId)} workspace.root ${JSON.stringify(workspaceDir)} must not equal or contain the repository checkout ${JSON.stringify(repositoryDir)}.`
```
hojinzs: **nit — the restored message dropped the `resolve()` the original had.**

The guard logic is a faithful restoration: `isPathWithinRoot(workspaceDir, repositoryDir)` keeps `root = workspaceDir`, so it still fires when the workspace root equals or contains the checkout — same direction as `0944059a:88`. Verified.

Only the message differs. The original interpolated `resolve(workspaceDir)` / `resolve(repositoryDir)`; this prints the raw stored values. Since the whole point of this error is to tell an operator which two paths collide during an upgrade, a relative or unnormalized `workspace.root` straight out of `project.json` is the case where the resolved form reads better. One-word fix, no behaviour change.

---

hojinzs (approved):
## Review round 2 — all five items settled, approving

I re-verified each round-1 item against the code at `088aad1` rather than reading the resolution replies. **All five are genuinely addressed, and three of them are byte-faithful restorations of the originals rather than approximations.** Approving.

### Round-1 items, verified

| Item | Verdict at `088aad1` |
| --- | --- |
| P2 containment invariant | **Restored faithfully.** Guard logic identical to `0944059a:88`; `service.ts` helper identical minus the now-dead `repositoryDir ??`, and the constructor call sits in the same position. Regression test added. |
| P2 `doctor.ts` | **Restored**, with `repository.path` substituted for the removed `repositoryDir` — the retained live field. #851 keeps its scope. |
| P2 `REPO_RUNTIME_DIR` stop fallback | **Restored byte-identically** to #858, `subcommand === "stop"` gating included. |
| P3 `allowDirtyExistingWorkspace` | **Gone** — zero hits across `packages/`, production and tests. |
| nit `project.ts:165` | Comment added. |

Both Codex threads are answered and resolved, and the cold-cache qualifying clause is now in the body. That was the whole of my round-1 ask.

### What I checked beyond the checklist

**The guard added to the `OrchestratorService` constructor was the riskiest line in this commit** — a false positive there rejects tenants at startup, on the host that runs everything. It cannot fire for current standalone projects: `deriveStandaloneProject` builds `repository` from `parseRepository(workflow.repository)`, a slug parse with no `path`, so the guard is inert for anything this version writes. It only evaluates for migrated legacy configs carrying `repository.path` — exactly the intent — and that matches base behaviour, where the same helper already read `repositoryDir ?? repository.path` unconditionally. No regression, no new rejection surface.

**Dispatch, observed.** Docker has a CLI but no reachable daemon in my environment, so I ported the standalone runner to the host again. A real dispatch populated the workspace through the surviving path with the guard live:

```
git marker type: FILE-linked-worktree      ← linked worktree, not a clone
branch:          symphony/project-alpha/test-owner-test-repo-101
skill:           present
bare cache:      present
```

(The stub worker then died on `resolveCoreModuleUrl` — my ad-hoc `tsc` emitted CJS where the E2E image builds ESM. My harness, not your diff; the population stage is what this PR changes and it is clean.)

One thing worth knowing separately: `run-standalone-project-e2e.sh` **exits 0 when the Docker daemon is unreachable**, printing the reason and nothing else. A green exit from that script is not by itself evidence a dispatch ran. Not this PR's problem — flagging it because it is the kind of thing that quietly turns into a false pass.

**The surviving path still contains a `checkout -B`** — `git.ts:459-466`, in the cache-unavailable fallback. I checked whether the #811 defect had simply relocated: it has not. That block is present verbatim at `0944059a:459-466`, this PR's `git.ts` change is 1 insertion against 377 deletions, and the fallback is only reachable when `repository/.git` is absent, i.e. against a fresh clone with no local work to lose — a reused workspace returns early or throws `preserved` before it. Your body's wording ("the only **reuse** path containing destructive `git checkout -B`") is precise, and it holds.

**Acceptance criteria of #850**, re-run rather than inferred: one populate strategy remains (`populateStrategy` survives only as a strip-on-read field); a previous-version tenant config still loads (probed with the genuine `0944059^` shape, see the inline comment); no repo-mode branch was skipped rather than removed (no `.skip`/`todo` added anywhere in the test diff); major changeset present.

```
lint      pass
typecheck pass
test      pass — core 346, orchestrator 361, cli 562
build     pass
```

Minor: the body says orchestrator 363; it's 361 (360 at round 1, plus your one new test). Not worth a push — correct it if you touch the body.

### Scope

Nothing widened. All three scope items came back as restorations, so #851 lands intact with its documentation, doctor-check and E2E work still ahead of it, and #861 stays correctly attributed to #858.

Two nits inline, both non-blocking and neither worth a round on its own. **Approved — this needs the rebuild and daemon restart immediately on merge, as you noted.**

---

packages/cli/src/commands/project.test.ts
moncher-dev: Fixed in 48b760a. The stop-fallback test now writes the previous runtime's persisted shape: `repositoryDir`, `populateStrategy: "clone"`, repo workflow source, `repository.path`, and no `projectDir`. It passes through the real loader before asserting routing to the legacy `repository` runtime.

packages/core/src/contracts/status-surface.ts
moncher-dev: Fixed in 48b760a. The containment error now applies `resolve()` to both `workspaceDir` and `repositoryDir`. A focused core regression test uses unnormalized but equivalent paths and asserts the operator-facing message contains their resolved forms.

hojinzs (approved):
## Review round 3 — both nits folded in faithfully, re-approving at `48b760a`

Round 2 was already an approval; this round exists only because `48b760a` folded in the two nits I explicitly marked non-blocking. **Both are correct, nothing widened, and I'm raising no new items.** Re-approving and closing this out.

### The incremental commit, verified

`48b760a` is 3 files, +31/−8, confined to the two files I named. No production logic changed beyond one error-message interpolation.

| Nit from round 2 | Verdict at `48b760a` |
| --- | --- |
| `project.test.ts` fixture wasn't legacy-shaped | **Fixed properly.** Now `repositoryDir`, `populateStrategy: "clone"`, `workflowSource.type: "repo"`, `repository.path`, `slug: "acme-platform"`, and **no `projectDir`** — the genuine `0944059^` persisted shape, matching what I probed last round. |
| Restored message dropped `resolve()` | **Fixed.** `resolve()` applied to both operands, with a focused core regression test. |

**I checked the new core test actually exercises what it claims.** `/tmp/workspaces/../workspaces` and `/tmp/workspaces/repository/..` both normalize to `/tmp/workspaces` through `realpathWithMissingTail`, so `relative()` returns `""` and the `allowRoot` arm fires — the guard genuinely throws, the assertion isn't passing vacuously.

**Smoke, on the host.** Docker has a binary but no reachable daemon here again, so I probed the built `dist` directly rather than trusting a green script exit (that script still exits 0 on an unreachable daemon — the false-pass I flagged last round, still worth fixing somewhere outside this PR).

```
PROBE-1 repositoryDir stripped:      true
PROBE-1 populateStrategy stripped:   true
PROBE-1 repo workflowSource stripped:true
PROBE-1 repository.path retained:    true
PROBE-2 (unnormalized input) → workspace.root "/tmp/smoke-ws-6fJHVy" must not equal or contain the repository checkout "/tmp/smoke-ws-6fJHVy/repository".
PROBE-3 (legacy collision)   → workspace.root "/tmp/smoke-legacy-pnqwir" must not equal or contain … "/tmp/smoke-legacy-pnqwir/checkout".
PROBE-4 (legitimate standalone layout) → no throw
```

Probe 2 is the point of the change: the raw stored value would have printed with the `..` still in it. Probe 4 is the one I cared about more — the guard runs in the `OrchestratorService` constructor, so a false positive rejects tenants at startup; a sibling checkout under a legacy `projectDir` is still accepted.

One thing I satisfied myself on rather than raise: `resolve()` interprets a relative `workspaceDir` against the daemon's cwd. That is not a new hazard — `isPathWithinRoot` → `realpathSync` already resolves relative inputs the same way, so the message now reports exactly the paths the guard compared, which is what `0944059a:88` did. Faithful restoration, not a behaviour change.

### Round-1 restorations, re-checked at this head

None regressed under the new commit: containment guard still wired at `service.ts:367`, `repository_environment` still in `doctor.ts`, `REPO_RUNTIME_DIR` fallback still at `project.ts:473`, `allowDirtyExistingWorkspace` still zero hits across `packages/`.

### Acceptance criteria of #850

- One populate strategy remains — `populateStrategy` survives only as the two strip-on-read lines in `status-surface.ts:56,60`.
- Previous-version tenant config still loads — probe 1, through the real loader.
- Repo-mode branches removed rather than skipped — no `.skip`/`.todo` added anywhere in the test diff across the full `0944059a..48b760a` range.
- Major changeset present.

```
lint      pass
typecheck pass
test      pass — core 347, orchestrator 361, cli 562
build     pass
CI        Test ✓   Container Smoke ✓
```

Counts match the PR body exactly; the round-2 tally correction landed.

### Closing

Three rounds, five scope/safety items and two nits, all settled without the diff widening — #851 still lands intact with its documentation, doctor-check and E2E work ahead of it, and #861 stays attributed to #858. **Approved; merge it.** As noted before, this one needs the rebuild and daemon restart immediately on merge, and the persisted tenants checked afterwards.

---

Issue #861, opened 2026-09-04
fix(e2e): repo-embedded Docker E2E broken on main — entrypoint imports an unbuilt module


moncher-dev:
## Problem

`./e2e/run-e2e.sh <scenario>` fails on `main` before any scenario runs:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find module '/app/packages/cli/dist/standalone-project.js'
  imported from /app/[eval1]
```

`e2e/seed/entrypoint.sh:46` imports it:

```sh
'import { standaloneProjectId } from "/app/packages/cli/dist/standalone-project.js"; …'
```

`packages/cli/src/standalone-project.ts` exists as source, but it is **not a tsup entry**. `packages/cli/tsup.config.ts:9-14` builds only `index`, `worker-entry`, `mcp-server` and `git-credential-helper`, so `dist/standalone-project.js` is never emitted. Confirmed absent from a clean `pnpm build` on `main`.

## Attribution

Introduced by #858 (`0944059a`):

```
$ git show 0944059a^:e2e/seed/entrypoint.sh | grep -c standalone-project.js   → 0
$ git show 0944059a:e2e/seed/entrypoint.sh  | grep -c standalone-project.js   → 1
```

Not caused by #860, which touches neither `tsup.config.ts` nor `entrypoint.sh` and merely inherits the break.

## Why CI did not catch it

`Container Smoke` was green on #858 and is green on #860, so whatever it exercises does not include this entrypoint path. `./e2e/run-standalone-project-e2e.sh` also passes, because it uses its own inline bootstrap rather than `e2e/seed/entrypoint.sh`. Only the repo-embedded runner goes through the broken import.

## Fix

Either add `standalone-project` to the tsup entry map so `dist/standalone-project.js` is emitted, or have `entrypoint.sh` derive the project id without importing an unbundled module. The first is smaller; the second removes a build-shape dependency from a shell script, which is arguably the better end state.

## Note on sequencing

#851 removes the repo-mode E2E case, which may delete the caller entirely. Worth confirming which lands first — if #851 removes `e2e/seed/entrypoint.sh`'s repo-mode bootstrap, this fix may be moot; if not, `main` stays broken until it is fixed.

## Pattern

[...]

PR #862, opened 2026-09-04
docs(cli): remove repository mode remnants

### LINKED ISSUE #851: docs(cli): remove repo-mode documentation, e2e case and doctor check

moncher-dev:
## Issues

- Closes #851

## Summary

- TL;DR: remove the remaining repository-mode documentation, obsolete doctor diagnostic, and legacy repo-mode E2E runner.
- Make folder-addressed `project start --project-dir <path>` the only documented operating model while preserving the CLI's exact migration instruction.
- Preserve shared Docker scenario coverage with an explicit folder-project Compose startup command and identify removed coverage honestly in the test matrix.

## Change-point diagram

- operator docs → project-folder quickstart + exact migration note + multi-repository selection caveat
- `doctor` → no repository-root `.env` diagnostic
- Docker Compose → seed folder-addressed project → project `.env` → `project start --project-dir` → retained scenario catalog
- legacy repo-mode runner/fixtures → removed → standalone dispatch or authoritative unit suites

## Start here

- `docker-compose.e2e.yml:16` — explicit shared folder-project startup path
- `README.md:100` — project-folder setup and repository-selection caveat
- `packages/cli/src/commands/doctor.ts:85` — removed doctor check surface
- `docs/configuration.md:239` — single project-folder configuration model
- `AGENT_TEST.md:227` — surviving Docker runner procedure and coverage decisions

- `gh-symphony doctor` no longer emits `repository_environment` checks.
- Operators upgrading from repository commands are directed to `gh-symphony project start --project-dir <path>` and to restart daemonized projects with the project command.
- The repository-mode Docker runner and its dedicated fixtures are removed; shared `compose up` scenarios continue through an explicit folder-addressed service command.

## Validation

- `pnpm exec prettier --check <changed supported files>` — pass
- `pnpm lint` — pass
- `pnpm test` — pass (14 workspace projects; CLI 561 tests)
- `pnpm typecheck` — pass
- `pnpm build` — pass
- `bash e2e/lib/compose-project.test.sh` — pass
- `docker compose -f docker-compose.e2e.yml config` — pass
- isolated `docker compose up -d --build`, in-container `/healthz`, project `.env`, and running-service assertions — pass at `1ab4190`
- inverse scan for `run-e2e.sh`, `e2e/seed/entrypoint.sh`, `repository_environment`, and live repo-mode wording — pass
- Docker E2E / blackbox evidence: the shared Compose service starts and remains healthy with the folder-addressed project command; cycle 1 also passed both surviving direct runners.

## Changeset

- `.changeset/remove-repository-mode-doctor.md` (`patch`, `@gh-symphony/cli`)

## Risks & rollback

- Risk: the broad legacy runner no longer provides black-box coverage for several negative/transient paths; `AGENT_TEST.md` names those losses and retains deterministic unit suites as authoritative coverage.
- Reviewer focus: the inline Compose startup must continue to seed scenario inputs and project `.env` before starting the packaged CLI.
- Rollback: revert this PR to restore the doctor check and legacy harness; no persisted data migration is performed.

## Changed files

- `README.md`, `packages/cli/README.md`, `docs/configuration.md`, examples and package docs — document the project-folder model, exact migration path, and remaining operational caveats.
- `docs/architecture.md`, `AGENT_TEST.md`, retained scenario documents — update layer mapping, coverage decisions, and derived project runtime paths.
- `packages/cli/src/commands/doctor.ts`, `doctor.test.ts` — remove `repository_environment`.
- `e2e/run-e2e.sh`, repo-mode-only scenarios and fixtures, `e2e/seed/entrypoint.sh` — remove the retired harness.
- `docker-compose.e2e.yml`, `e2e/seed/WORKFLOW.md`, `e2e/seed/restart-failure.mjs` — preserve shared folder-addressed Compose startup and derived project identity.
- `.changeset/remove-repository-mode-doctor.md` — patch release note for `@gh-symphony/cli`.

- [ ] Rebuild and restart the production daemon on the post-removal CLI, then confirm a real issue dispatches through its project folder.

## Security


moncher-dev:
Smoke-tested at `79552d1d`. Build, typecheck, lint clean; cli 561 tests green.

**I verified this against the stated property — a reader following the remaining quickstart reaches a working dispatch, and the surviving suites still run — rather than against the diff.**

**The surviving E2E runs, and I invoked it directly to prove the file-mode fix at the same time:**

```
$ ls -l e2e/run-standalone-project-e2e.sh
-rwxr-xr-x                                  ← was -rw-r--r--; diff shows `mode change 100644 => 100755`
$ ./e2e/run-standalone-project-e2e.sh       ← no `bash` prefix
```

That dispatches two folder-addressed standalone projects against a local seed repository. So the documented model still reaches a real dispatch after the repo-mode harness is gone.

**Quickstart walked from a clean scratch directory on this build.** Every documented command exists and behaves coherently; `project start --project-dir "$PWD" --once` in an empty folder returns *"No WORKFLOW.md in … Run this command from a standalone project folder or pass --project-dir <path>"* — a project-mode error with no reference to retired commands. `gh-symphony --help` contains **zero** retired command references. I did not run `gh-symphony setup` to completion, since it requires selecting a live GitHub Project and I would not point that at the real board.

**The AGENT_TEST.md matrix is the part I was most concerned about, and it is handled honestly.** I previously found ten rows citing the deleted runner. Now: **zero** references to `run-e2e.sh`, eight rows re-pointed at `run-standalone-project-e2e.sh`, one at `run-flat-tracker-keys-e2e.sh` — and, importantly, the losses are named per row rather than quietly re-pointed at something weaker:

> *"No Docker scenario remains for the transient active-run URL assertion; deterministic snapshot, adapter, dashboard, and control-plane tests are authoritative."*
> *"No Docker scenario remains for this adapter-owned negative path…"*
> *"No Docker scenario remains for a same-project per-state cap… The standalone runner instead proves concurrent isolation between two project folders."*

The PR body states the same thing under Risks rather than burying it. That was the judgement most likely to be made by omission, and it wasn't.

**Inverse check — clean.** Across `.ts`, `.sh`, `.yml`, `Dockerfile` and `.md`: `run-e2e.sh` 0, `seed/entrypoint` 0, `repository_environment` 0. `doctor.ts` source now has **0** references to `repo init`/`repo start`/`repo stop` — the ~10 remediation strings I flagged in #849 round 2 are gone. The residual hits are benign and I checked each: `standalone-project.js` appears only as source-relative ESM imports from `.ts` files (the #861 break was a *shell script* importing the unbuilt `dist/` path, and that caller is deleted); `repo start`/`repo init` survive only in test fixtures — `lifecycle.test.ts` and `config.test.ts` use them as legacy `processIdentity` strings, which is correct since legacy daemons genuinely have that identity, and `doctor.test.ts`'s occurrence is a mocked stub return, not an assertion about current output.

**One thing to flag rather than block on: #861 will not auto-close.** The PR body carries `Closes #851` only. #861 was folded in by intention, and the deletions here do resolve it — `e2e/seed/entrypoint.sh` and its unbuilt-module import are both gone. Rather than spend a review round on a one-line PR-body edit, **I will close #861 manually once this merges**, referencing this PR so the trail stays intact.

**Landing this finishes the removal chain** and needs a rebuild and daemon restart immediately afterwards — it removes the doctor check and E2E harness the daemon's build carries. @hojinzs over to you.

docker/orchestrator-e2e.Dockerfile
hojinzs, line 1: **P1 — removing the `ENTRYPOINT` leaves `docker-compose.e2e.yml` with no process, breaking `pnpm e2e:claude` and 13 retained scenarios.**

Confirming and widening the Codex finding on this file.

`docker-compose.e2e.yml` is unchanged in this PR and declares no `command:`. With the `ENTRYPOINT` gone the image falls back to the `node:24-bookworm-slim` default `CMD ["node"]`, which under `up -d` (no TTY) reads EOF on stdin and exits immediately.

Concrete breakage:

- `test/e2e/claude/run-docker-e2e.sh:25` runs `"${e2e_compose[@]}" up -d --build` and then `exec -T symphony-e2e curl … /healthz`. That `exec` now targets a container that is not running, so the Codex half of `pnpm e2e:claude` fails before any assertion. CI does not catch this — `ci.yml` runs only `e2e:compose-project-test` and `e2e:fixture-replacement-test`, which is why the PR is green.
- 13 retained scenarios still open with `docker compose -f docker-compose.e2e.yml up -d --build` and then read `/e2e/work/test-repo/.runtime/orchestrator/…`: TC-01…07, `10-http-auth-hardening`, `10-orchestrator-tracker-state`, `13-api-progress-convergence`, `14-dispatch-start-failure-isolation`, `17-restart-failure-isolation`, `19-host-side-tracker-tools`. That work tree and its project `.env` were created only by the deleted `e2e/seed/entrypoint.sh`.
- `docker-compose.e2e.yml`'s `STUB_SCENARIO`, `E2E_REQUIRED_LABELS`, and `E2E_MAX_FAILURE_RETRIES` are now read by nothing — the entrypoint was their only consumer.
- The PR's own coverage matrix still cites TC-19 (`19-required-label-routability.md`, whose Setup step 1 is "Start the Docker E2E environment with `STUB_SCENARIO=required-label-removed`") and TC-20 (`pnpm e2e:claude`) as authoritative Docker coverage.

Scope note: #851 asks to remove "the repo-mode end-to-end case and any fixtures **only it used**." `e2e/seed/entrypoint.sh` was not repo-mode — it ran `project start --project-dir /e2e/work/test-repo --http 4680 --bind-all`, i.e. the folder-addressed command this PR promotes. It was shared infrastructure for the Claude E2E and the scenario catalog, so removing it reaches past the issue's stated scope.

Minimal fix: keep a folder-addressed start path available to `up`-based callers — add a `command:` to `docker-compose.e2e.yml`, or restore the entrypoint with its (already repo-mode-free) body. Otherwise `pnpm e2e:claude` and the retained scenarios need to be migrated in this PR rather than left dangling.

---

README.md
```diff
@@ -101,10 +101,10 @@ auto-injected worker variables, see
 
 ### 2. Run Setup
 
-Navigate to the repository you want to orchestrate, then run:
+Create a project folder outside the target repository, then run:
 
 ```bash
-cd your-repo
+mkdir my-symphony-project && cd my-symphony-project
```
hojinzs, line 107: **P1 — the new empty-folder setup path can silently bind the wrong repository.**

Codex flagged this; I verified it and it holds.

`resolveCheckoutRepository()` (`packages/cli/src/commands/workflow-init.ts:476-497`) resolves the repository from `git config --get remote.origin.url` in cwd, and when that yields no match — exactly the case in a fresh empty folder — returns `projectDetail.linkedRepositories[0]` unconditionally. That value becomes `repository.slug` in the generated `WORKFLOW.md` (`workflow-init.ts:1012-1020`), and there is no repository prompt or `--repository` flag anywhere in `setup` / `workflow init`.

Before this change the documented path was `cd your-repo`, so `origin` made the selection correct by construction. Making the empty-folder path primary turns that into a silent wrong-repository bind for any Project linked to more than one repository — the orchestrator then clones and pushes branches to the wrong repo.

The interactive final summary mitigates it, but `README.md:449-455` recommends `gh-symphony setup --non-interactive` for the Docker path, where nothing surfaces the choice.

The in-scope fix here is a doc caveat: state that a Project linked to multiple repositories selects the first linked repository, and that operators must verify or correct `repository.slug` in the generated `WORKFLOW.md`. An actual selection prompt/flag is fair to defer to a follow-up issue.

---

README.md
```diff
@@ -789,11 +802,8 @@ The orchestrator resolves the workflow policy using this fallback chain:
 2. **Project WORKFLOW.md** — if the repository has no `WORKFLOW.md`, fall back to the project-level `WORKFLOW.md`.
 3. **Hardcoded defaults** — if neither file exists, use built-in defaults (`Todo`, `In Progress` as active; `Done` as terminal; blocker checks enabled for `Todo`; planning states disabled).
 
-This means you can:
-
-- Run without any `WORKFLOW.md` and rely on defaults
-- Use a single project-level `WORKFLOW.md` for all repositories
-- Override per-repository by committing a `WORKFLOW.md` to the repo root
+Projects may run without a `WORKFLOW.md` and rely on defaults. When present,
+the project folder's `WORKFLOW.md` is the policy source.
```
hojinzs: **P2 — the "Resolution order" list directly above is now false, not merely stale.**

Confirming Codex's finding with the code path.

`createProjectConfig` always writes `workflowSource: { type: "external", path: workflowPath }` for a folder-addressed project (`packages/cli/src/commands/project.ts:89`). `service.ts:2974-2985` then branches on `tenant.workflowSource?.path`: when it is set it calls `loadWorkflowFile(...)` and **never** calls `loadRepositoryWorkflow`. A repository-root `WORKFLOW.md` only produces a shadowing *warning* (`service.ts:3007-3029`) — it never wins and is never a fallback.

So items 1 and 2 at lines 799-802 contradict the paragraph you added right here at 805-806. An operator following the numbered list would commit and edit a repository-root `WORKFLOW.md` that can never execute.

Please delete the numbered list, or rewrite it as: project folder `WORKFLOW.md` → built-in defaults, with a note that a repository-root `WORKFLOW.md` is only reported as shadowed.

---

README.md
```diff
@@ -432,28 +442,27 @@ docker run --rm -it \
...
+  -w /project \
   ghcr.io/hojinzs/github-symphony:latest \
   gh-symphony setup --non-interactive
 ```
 
-Then start the long-running orchestrator from the initialized repository. The image
-default command is `gh-symphony project start --project-dir <path>`, so the mounted working directory must
-already contain `WORKFLOW.md` and the repository runtime config created by setup.
+Then start the long-running orchestrator from the initialized project folder.
+The mounted directory must already contain the generated `WORKFLOW.md`.
```
hojinzs, line 463: **P2 — this Docker section contradicts itself ~20 lines later.**

Lines 442-465 correctly move the mount to `-v "$(pwd):/project" -w /project`, but the `docker compose` example at `README.md:471-483` still mounts `- ./:/repo` and says:

> Run `gh-symphony setup` once before starting the service so the mounted repository has `WORKFLOW.md` and `.runtime/orchestrator/`.

Both the `/repo` mount and `.runtime/orchestrator/` were rewritten away everywhere else in this PR, including the runtime-state table at 907-916. Please bring that example in line with the `/project` model.

---

README.md
hojinzs, line 1: **P2 — two more passages the rewrite missed, both contradicting edits made elsewhere in this same PR.**

- `README.md:202` — "…after `gh-symphony setup` has generated `WORKFLOW.md` and bound `.runtime/orchestrator/` to a GitHub Project." Per your own rewrite at 531-533, `setup` now generates the project folder; nothing is "bound" to `.runtime/orchestrator/`.
- `README.md:213` — "creates an issue workspace under `.runtime/orchestrator/`". Per your own text at 383-389, issue workspaces live under `workspace.root`, defaulting to `<project-dir>/.runtime/workspaces`.

These matter for AC #1 ("No document describes repo-embedded mode as available") because `.runtime/orchestrator/` inside the target repository *was* the repo-mode layout — the End-to-End Walkthrough is the one operator-facing chapter still describing it.

nit, same class: `packages/cli/README.md:501` still lists "repository `WORKFLOW.md` presence and parse validity" while the matching line in `README.md:605` was changed to "project `WORKFLOW.md`".

nit: the migration banner at `README.md:3-5` lost its `>` prefix on the continuation line. Lazy continuation keeps it inside the blockquote so it renders, but it is inconsistent with every other line of that block.

---

AGENT_TEST.md
```diff
@@ -186,143 +180,69 @@ Control worker behavior with the `STUB_SCENARIO` environment variable:
...
+| §17 conformance coverage | Tracker empty-input and malformed-refresh cases are deterministic adapter tests; workspace-file and no-running reconciliation cases are deterministic orchestrator tests. The authoritative row-to-test map is in `docs/architecture.md` under “§17 conformance test matrix”; Linear malformed polling-item omission and stderr isolation are documented gaps, while `start.test.ts` host/port/bind coverage belongs to §13.7. | `./e2e/run-standalone-project-e2e.sh` remains the Docker lifecycle confirmation for the file-tracker and workspace path. Run `docker compose -f docker-compose.e2e.yml exec -T symphony-e2e node /app/e2e/host-dynamic-tool-e2e.mjs` for the dynamic-tool boundary. The malformed-provider and regular-file cases remain unit-isolated because the Docker fixture intentionally uses valid file-tracker data and a fresh workspace. |
 
 `docker-compose.e2e.yml` uses `environment.STUB_SCENARIO: ${STUB_SCENARIO:-happy}`, so the scenario can be selected via a shell environment variable.
 
 ```bash
 STUB_SCENARIO=fail docker compose -f docker-compose.e2e.yml up -d --build
 ```
```
hojinzs, line 219: **P1 (dependent) — this instruction is inert after the entrypoint removal.** See my comment on `docker/orchestrator-e2e.Dockerfile`.

`STUB_SCENARIO` was consumed only by `e2e/seed/entrypoint.sh`, which wrote it into the project `.env` so the orchestrator's environment allowlist would forward it to the stub worker. With the entrypoint gone, this `up -d --build` starts no orchestrator and the variable reaches nothing.

Same root cause for the log/cleanup commands at 291-306, which read `/e2e/work/test-repo/.runtime/orchestrator/runs/*/…` — a path only the entrypoint ever created — and for the repeat of this instruction at line 400.

---

AGENT_TEST.md
```diff
@@ -422,41 +342,28 @@ idle → [inject issue + refresh]
...
+| `e2e/scenarios/13-api-progress-convergence.md`          | Verify confirmed API lifecycle progress persists as a successful run without workspace mutations                                         |
+| `e2e/scenarios/13-standalone-project-model.md`          | Verify the standalone project model (project `.env`, MCP, worktree, and branch isolation) — `pnpm e2e:standalone-project`                |
+| `e2e/scenarios/14-dispatch-start-failure-isolation.md`  | Verify one candidate's pre-spawn failure records retry state without starving later candidates                                           |
+| `e2e/scenarios/15-terminal-candidate-reconciliation.md` | Verify a closed issue in active Project status converges to `Done` without worker dispatch                                               |
+| `e2e/scenarios/15-cache-maintenance.md`                 | Verify cache inventory, dry-run eviction, and active-worktree preservation                                                               |
+| `e2e/scenarios/16-packaged-runtime-entrypoints.md`      | Verify the built CLI's MCP dispatcher and Git credential helper subprocesses inside Docker                                               |
```
hojinzs, line 363: **P2 — the scenario catalog still lists 13 scenarios that can no longer be executed.**

This is the part I'd most want fixed before merge, precisely because the PR's stated virtue is that coverage losses are *named per row* rather than quietly re-pointed. That holds for the rows you rewrote — the three "No Docker scenario remains for…" disclosures are exactly right and I want to say so.

But this catalog carries no such note for the scenarios the entrypoint removal disabled: TC-01…07, `10-http-auth-hardening`, `10-orchestrator-tracker-state`, `13-api-progress-convergence`, `14-dispatch-start-failure-isolation`, `17-restart-failure-isolation`, `19-host-side-tracker-tools`. All of them begin with `docker compose -f docker-compose.e2e.yml up -d --build` and then read `/e2e/work/test-repo/…`.

Related: line 205 still cites TC-19 `19-required-label-routability.md` as authoritative Docker coverage for required-label routability, while that scenario's Setup step 1 — "Start the Docker E2E environment with `STUB_SCENARIO=required-label-removed`" — is now unreachable.

Either restore an `up`-able service command (preferred; see the Dockerfile comment) or extend the same honest per-row disclosure to these.

---

hojinzs (changes requested):
## Review — round 1 of 6

Verified at `79552d1d`. The documentation rewrite itself is good work: thorough, honest about what it gives up, and the acceptance criteria for the doctor removal and the migration note are met. The blocker is the E2E harness deletion, which reaches past the issue's scope and leaves the surviving suite unrunnable.

### What I verified myself

| Check | Result |
| --- | --- |
| `pnpm install && pnpm build` | pass |
| `pnpm test` | pass — cli 561/561, all packages green |
| Migration string matches the CLI exactly | pass — ran the built CLI: `repo start` prints byte-for-byte what `packages/cli/README.md:6-9` documents |
| Quickstart smoke in a clean empty folder | `project start --project-dir "$PWD" --once` returns *"No WORKFLOW.md in … Run this command from a standalone project folder or pass `--project-dir <path>`"* — a project-mode error with no retired-command reference |
| `gh-symphony --help` retired-surface references | 0 |
| `repository_environment` / `run-e2e.sh` / `seed/entrypoint` across `.ts` `.sh` `.yml` `.md` | 0 each |
| CI on this head | green (Test, Container Smoke) |

Smoke testing was local; there is no preview environment, and Docker is unavailable in this session, so the Docker findings below are established from the compose/Dockerfile/runner sources rather than from an execution.

### Findings

**P1 — `docker/orchestrator-e2e.Dockerfile`: removing the `ENTRYPOINT` leaves the compose service with no process.** `docker-compose.e2e.yml` is untouched and declares no `command:`, so the image falls back to the base `CMD ["node"]`, which exits immediately under `up -d`. This breaks the Codex half of `pnpm e2e:claude` (`test/e2e/claude/run-docker-e2e.sh:25`) and 13 retained scenario documents. CI misses it because `ci.yml` never runs `e2e:claude`.

  Scope: #851 scopes the deletion to "the repo-mode end-to-end case and any fixtures **only it used**." `e2e/seed/entrypoint.sh` ran `project start --project-dir /e2e/work/test-repo --http 4680 --bind-all` — the folder-addressed command this PR promotes, not repo mode. It was shared infrastructure. Its removal is the one thing here I'd call out as reaching beyond the issue.

**P1 — `README.md:104-107`: the new empty-folder setup path can silently bind the wrong repository.** Confirmed Codex's finding against `workflow-init.ts:476-497` and `:1012-1020`. Outside a checkout, `linkedRepositories[0]` is taken unconditionally and becomes `repository.slug`; no prompt or flag exists. The old `cd your-repo` path was correct by construction. A doc caveat is the in-scope fix.

**P2 — `README.md:799-806`: the "Resolution order" list is false, not just stale.** Folder projects always set `workflowSource.type = "external"` (`project.ts:89`), and `service.ts:2974-2985` never falls back to the repository workflow — it only warns about shadowing. The list contradicts the paragraph added directly beneath it.

**P2 — `README.md:471-483`** still mounts `./:/repo` and references `.runtime/orchestrator/`, contradicting the `/project` rewrite 20 lines above in the same PR. **`README.md:202` and `:213`** likewise still describe the repo-embedded `.runtime/orchestrator/` layout in the End-to-End Walkthrough.

**P2 — `AGENT_TEST.md:215-219`, `:291-306`, `:347-363`** carry instructions and a scenario catalog that the entrypoint removal made unrunnable, with no disclosure — in contrast to the three per-row loss notes elsewhere in the file, which are exactly the right treatment and which I want to credit.

**nit** — `packages/cli/README.md:501` ("repository `WORKFLOW.md`") missed the rename applied at `README.md:605`; the migration banner at `README.md:3-5` drops its `>` on the continuation line.

None found on the documentation side — the rewrite tracks #851's four bullets closely and does not invent structure. The one unrequested decision is the entrypoint deletion above, which removed more than the issue asked for.

### On #861

Agreed with your reasoning in the PR comment: folding it in and closing it manually after merge is the right call, not worth a round on a PR-body edit.

### Path to approval

Land the harness fix (a `command:` in `docker-compose.e2e.yml` is the smallest change that restores every `up`-based caller) plus the four doc corrections, and this is an approve from me. If you'd rather argue that the Claude E2E and the scenario catalog should be retired in a separate issue instead of restored here, say so on the thread with the reasoning and I'll take that as the resolution — I just don't want them left silently broken.

---

docker/orchestrator-e2e.Dockerfile
```diff
@@ -40,10 +40,6 @@ COPY --chown=node:node e2e/seed /e2e/seed
 # Initialize bare git repo for E2E
 RUN bash /e2e/seed/init-repo.sh
 
-# Prepare entrypoint
-COPY --chown=node:node e2e/seed/entrypoint.sh /e2e/entrypoint.sh
-RUN chmod +x /e2e/entrypoint.sh
-
 # Ensure fixtures directory exists
```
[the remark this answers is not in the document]
moncher-dev: Resolved in `1ab4190`: `docker-compose.e2e.yml` now supplies an explicit folder-addressed startup command that clones the seed, writes the project `.env`, exports the file-tracker path for workflow parsing, and starts `project start --project-dir`. An isolated `compose up -d --build` remained running and passed `/healthz`.

README.md
```diff
@@ -101,10 +101,10 @@ auto-injected worker variables, see
 
 ### 2. Run Setup
 
-Navigate to the repository you want to orchestrate, then run:
+Create a project folder outside the target repository, then run:
 
 ```bash
-cd your-repo
+mkdir my-symphony-project && cd my-symphony-project
 gh-symphony setup
```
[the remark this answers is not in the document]
moncher-dev: Resolved in `1ab4190`: the setup guide now warns that setup outside a checkout selects the first linked repository when a Project links multiple repositories, and instructs operators to verify or correct `repository.slug` before starting Symphony.

README.md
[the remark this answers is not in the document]
moncher-dev: Resolved in `1ab4190`: the obsolete fallback list is replaced with the actual project-folder policy source, built-in-default behavior, and repository-workflow shadowing semantics.

docker/orchestrator-e2e.Dockerfile
moncher-dev: Resolved in `1ab4190`: Compose now owns the shared folder-project startup path, including seed cloning, required-label and retry fixture transforms, project `.env` creation, and the explicit `project start --project-dir` process. The seed workflow now uses the local clone URL, and the retained scenario runtime paths use the derived project directory.

README.md
moncher-dev: Resolved in `1ab4190`: README documents the first-linked-repository behavior for setup outside a checkout and requires verification or correction of `repository.slug`; no CLI selection behavior was added.

README.md
moncher-dev: Resolved in `1ab4190`: README now documents project-folder `WORKFLOW.md` as authoritative, built-in defaults when absent, and repository-root workflow files as shadowed rather than fallback policy.

README.md
moncher-dev: Resolved in `1ab4190`: both Compose examples use `/project`, and the surrounding instruction refers to the mounted project folder and its `WORKFLOW.md`.

README.md
moncher-dev: Resolved in `1ab4190`: the walkthrough uses the project folder and `workspace.root`, the CLI README says project `WORKFLOW.md`, the banner continuation is quoted, and remaining live README descriptions were aligned with the folder-project model.

AGENT_TEST.md
moncher-dev: Resolved in `1ab4190`: the explicit Compose command consumes `STUB_SCENARIO`, required labels, and retry limits, writes the project `.env`, and starts the service. AGENT_TEST log paths now traverse the derived `projects/*/runs/*` runtime layout.

AGENT_TEST.md
moncher-dev: Resolved in `1ab4190`: `compose up` is executable again through the explicit folder-project command; the retained scenario seed and environment inputs are restored, and stale project runtime paths were updated. The isolated startup/health check passed.

docker-compose.e2e.yml
```diff
@@ -13,3 +13,41 @@ services:
...
+              print "  max_failure_retries: " retries
+              next
+            }
+            { print }
+          ' "$$project_dir/WORKFLOW.md" > /tmp/e2e-workflow.md
+          mv /tmp/e2e-workflow.md "$$project_dir/WORKFLOW.md"
+        fi
+        printf 'STUB_SCENARIO=%s\n' "$${STUB_SCENARIO:-happy}" > "$$project_dir/.env"
+        printf 'GH_SYMPHONY_FILE_TRACKER_ISSUES_PATH=/e2e/fixtures/issues.json\n' >> "$$project_dir/.env"
+        chmod 600 "$$project_dir/.env"
```
hojinzs, line 49: Verified working, and the `.env` location is a correction rather than a port.

I ran this command body verbatim on the host (same seed, same CLI dist) and confirmed the mechanism end to end. The decisive test: I wrote `STUB_SCENARIO=fail` into `<project-dir>/.env` while **unsetting** `STUB_SCENARIO` from the daemon environment, so the project `.env` was the only possible channel. The worker logged `[stub-worker] scenario=fail`. So scenario selection for the retained catalog genuinely works.

Worth recording why this is more correct than the entrypoint it replaces: `readProjectEnv` (`service.ts:5163-5165`) resolves `tenant.projectDir ?? store.projectDir(...)`, and `createProjectConfig` always sets `projectDir` for a folder project (`project.ts:84`). The old entrypoint wrote its `.env` to `.runtime/orchestrator/projects/<id>/.env`, i.e. the `store.projectDir` branch that never wins for a folder-addressed project. `<project-dir>/.env` is the branch that is actually read.

`chmod 600` is a nice touch — it keeps this off the `should use 0600 permissions` warning at `service.ts:5170-5179`, which the old entrypoint would have tripped.

nit: the old entrypoint also seeded `/e2e/fixtures/issues.json` with `[]` when absent, and that is gone. Not a defect — `e2e/fixtures/issues.json` is gitignored, but `readIssueEntries` maps ENOENT to `[]` (`file-tracker-adapter.ts:123-130`), `run-docker-e2e.sh:23` writes `[]` itself before `up`, and each scenario `cp`s its own fixture. Only noting it so the behavioral delta is on the record.

---

e2e/seed/restart-failure.mjs
```diff
@@ -1,8 +1,13 @@
...
 import { OrchestratorFsStore } from "/app/packages/orchestrator/dist/fs-store.js";
 
 const now = new Date().toISOString();
 const runtimeRoot = "/e2e/work/test-repo/.runtime/orchestrator";
-const projectId = "repository";
+const projectDirectory = "/e2e/work/test-repo";
+const projectId = `test-repo-${createHash("sha256")
+  .update(projectDirectory)
+  .digest("hex")
+  .slice(0, 8)}`;
```
hojinzs, line 10: **P3 — correct today, but it silently duplicates `standaloneProjectId` with no guard against drift.**

I verified the value is right: a live run produced project id `test-repo-fbac6349`, and this expression computes `test-repo-fbac6349`. So TC-17 will find its run.

I also want to credit the choice, because hand-rolling looks worse than it is. The old entrypoint imported the real function from `/app/packages/cli/dist/standalone-project.js` — and tsup only emits four entries (`index`, `worker-entry`, `mcp-server`, `git-credential-helper`, `packages/cli/tsup.config.ts:9-14`). That file has never existed in the image. Importing it was not an option.

The residual risk is that this reimplements two things — the `slugify` + 8-hex-sha256 shape and the hardcoded `test-repo` prefix — that live in `packages/cli/src/standalone-project.ts:12-14`. If that derivation ever changes, this drifts silently and TC-17 fails looking like a product bug.

Cheapest guard, no new export needed: read the id back from disk instead of recomputing it, since the orchestrator has already created `projects/<id>/`. Or add a comment pointing at `standalone-project.ts:12-14` as the source of truth. Your call — not blocking.

---

README.md
```diff
@@ -1,7 +1,7 @@
 # GitHub Symphony
 
-> **CLI migration:** the `gh-symphony repo` surface is retired. Use
-> `gh-symphony project start --project-dir <path>`; see the
+> **CLI migration:** `The 'repo' command has been removed. Use 'gh-symphony
+project start --project-dir <path>'.` See the
```
hojinzs, line 4: nit (carryover, non-blocking) — line 4 still has no `>` prefix. Lines 5-6 were quoted in `1ab4190`, so this is the last one. Lazy continuation keeps it inside the blockquote and it renders correctly, so this is purely cosmetic.

The substance here is right, and I re-verified it at this head rather than taking round 1's word for it: I ran the built CLI, and `repo start` prints exactly what `packages/cli/README.md:8-9` documents, byte for byte. AC #2 is met.

---

hojinzs (approved):
## Review — round 2 of 6: Approve

Verified at `1ab4190`. Every round-1 blocker is fixed, and this time I could execute the Docker path instead of reasoning about it. All ten threads you resolved, I re-verified independently rather than taking the resolution note at face value — they hold.

### Correcting my own round-1 framing

I said the entrypoint removal broke a working harness. That was wrong, and in your favor.

`e2e/seed/entrypoint.sh` derived its project id by importing `/app/packages/cli/dist/standalone-project.js`. tsup emits only four entries — `index`, `worker-entry`, `mcp-server`, `git-credential-helper` (`packages/cli/tsup.config.ts:9-14`) — so that file has never existed in the image. Under `set -euo pipefail` the import fails `ERR_MODULE_NOT_FOUND` and the entrypoint aborts *before* starting the CLI. The container was already exiting on `main`. This PR does not restore the old harness; it replaces a broken one with a working one, which is a strictly better outcome than the fix I asked for.

Your `.env` placement is a genuine correction too, not a port: `readProjectEnv` resolves `tenant.projectDir ?? store.projectDir(...)` (`service.ts:5163-5165`) and folder projects always set `projectDir` (`project.ts:84`), so the old entrypoint's `projects/<id>/.env` was on the branch that never wins.

### What I executed

Docker image pull is blocked by this environment's network policy (gateway 403 on the Docker Hub blob CDN), so I could not run `compose up` in-container. Instead I replicated the Compose command body **verbatim** on the host against the real seed and the built CLI dist — same clone, same `.env` write, same `project start --project-dir`.

| Check | Result |
| --- | --- |
| Compose startup path stays alive | pass — `/healthz` → 200, project `test-repo-fbac6349`, dispatch reached `1 active run(s)` |
| `STUB_SCENARIO` reaches the worker **via project `.env` only** | pass — wrote `fail` to `.env` with the var *unset* in the daemon env; worker logged `scenario=fail` |
| Updated doc path `projects/*/runs/*/events.ndjson` | pass — matches the real tree exactly |
| `restart-failure.mjs` project id | pass — computes `test-repo-fbac6349`, equals the live id |
| `E2E_REQUIRED_LABELS` / `E2E_MAX_FAILURE_RETRIES` awk transforms | pass — both parse; `requiredLabels=["symphony","ready"]`, `maxFailureRetries=3`, correct YAML nesting |
| `docker compose config` | pass |
| `pnpm lint` / `test` (561 CLI) / `typecheck` / `build` | pass |
| `bash e2e/lib/compose-project.test.sh` | pass |
| Migration string vs. built CLI | pass — byte-for-byte with `packages/cli/README.md:8-9` |
| Dangling refs to the 18 deleted files | none |
| `repository_environment`, `run-e2e.sh`, `:/repo` mounts | 0 each |
| CI on this head | green (Test, Container Smoke) |

`pnpm format` fails on 73 files, but none are yours — `prettier --check` over this PR's 47 changed files is clean (only unparseable Dockerfile/shell types). Pre-existing on `main`; your narrower validation command was the right call.

### Round-1 findings

| Finding | Status |
| --- | --- |
| P1 compose has no process | Fixed and functionally verified |
| P1 multi-repo silent bind | Fixed — caveat at `README.md:127-131` |
| P2 false "Resolution order" list | Fixed — rewritten, now matches `service.ts:2974-2985` |
| P2 `:/repo` + `.runtime/orchestrator` in Docker section | Fixed |
| P2 `README.md:202`/`:213` walkthrough | Fixed |
| P2 scenario catalog unrunnable | Fixed at the root — `up` works again |
| nit `packages/cli/README.md:501` | Fixed |
| nit banner `>` | Partially — line 4 still unquoted, cosmetic only |

### Acceptance criteria (#851)

All three met. No live document presents repo-embedded mode as available — the remaining `repo-embedded` strings are confined to dated records under `docs/designs/`, which is correct for historical design docs. Migration note matches the CLI exactly. `.changeset/remove-repository-mode-doctor.md` is present with the right package and bump.

None. The round-2 commit is 104 insertions across 12 files and every hunk traces to a specific thread. `repository.clone_url` in the seed is a supported field (`project.ts:392-396`) and is load-bearing for the local clone. Nothing was widened.

### Remaining

Two non-blocking inline notes: the `standaloneProjectId` duplication in `restart-failure.mjs` (P3 — correct today, no guard against drift) and the banner nit. Neither needs another round; fold them in or don't, and merge. Good, honest work on the coverage disclosures throughout.

---

PR #863, opened 2026-09-04
fix(orchestrator): define project env snapshot timing around workspace hooks

### LINKED ISSUE #834: fix(orchestrator): define project env snapshot timing around workspace hooks
## Problem

PR #827 changes the worker's project `.env` read timing from spawn-time (after `after_create` and `before_run` hooks) to an earlier snapshot before those hooks. Hooks still read `.env` fresh. A hook that writes or refreshes project environment values therefore sees its update while the subsequently spawned worker does not.

## Context

Discovered during review of #809. The reviewer explicitly marked this non-blocking, and #827 is approved for landing. Track the timing contract separately rather than expanding the security fix during Land.

## Acceptance criteria

- Decide and document whether hooks may mutate the project `.env` for the same run.
- If supported, resolve project environment and worker credentials after hooks complete.
- Add a regression test covering the chosen timing contract.

moncher-dev:
## Issues

- Closes #834

## Summary

- Resolve the project environment once at worker spawn, after `after_create` and `before_run`, so hook refreshes affect the same run.
- Derive the worker credential gate and injected credentials from that same snapshot while preserving initial-dispatch warnings and due-retry reservations.
- Persist only a deterministic, comparable SHA-256 digest of the project environment snapshot; environment names and values are never recorded.

## Change-point diagram

- workspace hooks → managed project `.env` mutation → one late project snapshot
- late snapshot → credential resolver + dispatch gate → composed worker environment
- late project snapshot → value-free comparable digest → run record

## Start here

- `packages/orchestrator/src/service.ts:3334` — post-hook snapshot, credential resolution, worker environment, gate, and project digest
- `packages/orchestrator/src/service.test.ts:16731` — hook-to-worker late-value and cross-run digest regression
- `packages/orchestrator/src/service.test.ts:10866` — due-retry credential-gate preservation

- Hooks may update the managed project `.env` for the same run; the worker observes values present after hooks finish.
- Missing credentials still suppress worker spawn, surface the existing actionable warning, and leave due retries queued without consuming attempts or failure budget.
- Runs with identical post-hook project environments now carry identical digests even when their run-scoped worker metadata differs.

## Validation

- `pnpm lint` — pass
- `pnpm test` — pass
- `pnpm typecheck` — pass
- `pnpm build` — pass
- `pnpm --filter @gh-symphony/orchestrator exec vitest run src/service.test.ts -t "snapshots project env after hooks|records a spawn-time event|keeps a due retry queued"` — pass (3 tests)
- `./e2e/run-standalone-project-e2e.sh` — pass with documented environment exception #865: two isolated attempts exited 69 before setup because the Docker daemon was unreachable at `/var/run/docker.sock`

## Changeset

- `.changeset/fresh-project-env-after-hooks.md` (`patch`)

## Risks & rollback

- Hooks now intentionally control the same-run project environment snapshot; operators enabling workflow hooks already place those hooks inside the project credential trust boundary.
- Retry, initial-dispatch, same-run mutation, digest secrecy, and digest comparability regressions are covered.
- Roll back commits `1787056`, `5c0f581`, and `133b5c5` together to restore the previous early snapshot behavior.

## Changed files

- `packages/orchestrator/src/service.ts` — late snapshot, credential gate, comparable project digest, retry preservation
- `packages/orchestrator/src/service.test.ts` — hook mutation, missing credential, digest secrecy/comparability, and due-retry regressions
- `packages/core/src/contracts/status-surface.ts` — run-record digest contract
- `docs/configuration.md`, `README.md`, `packages/cli/README.md`, `docs/architecture.md` — timing and layer documentation
- `AGENT_TEST.md` — E2E scenario matrix
- `.changeset/fresh-project-env-after-hooks.md` — CLI patch release note

- [ ] Follow-up #864 surfaces the opaque digest through an operator-facing status/run-detail view.
- [ ] Follow-up #865 restores Docker daemon availability for worker-side black-box validation.

## Security


packages/orchestrator/src/service.ts
```diff
@@ -3339,6 +3304,84 @@ export class OrchestratorService {
     }
 
     const runtimeTimeouts = resolveWorkflowRuntimeTimeouts(workflow.workflow);
+    // Snapshot the project environment at the worker consumption point. The
+    // credential gate and the spawned worker deliberately share this read so
+    // hooks may refresh same-run values without creating a diagnostic race.
+    const projectEnvironment = this.readProjectEnv(tenant);
+    const workerCredentials =
+      trackerAdapter.resolveWorkerCredentials?.(tenant, {
+        project: projectEnvironment,
```
hojinzs, line 3340: **P1 — the gate was removed, not moved.**

The comment says "the credential gate and the spawned worker deliberately share this read", but after this change there is no gate here — only a diagnostic. `workerCredentials` is resolved, spread into `workerEnvironment`, and the worker is spawned unconditionally.

The recorded decision on #834 is explicit that this is a *move*:

> Move `resolveWorkerCredentials` and the `worker-credential-missing` event to **immediately before the spawn**, so the gate and the injected value always describe the same read.

and the Ready comment lists the regression tests as "hook mutation reaches the worker; **the credential gate reflects the late value**; hook environment is unchanged". There is no test in this PR that a *gate* reflects the late value — `records a spawn-time event when no worker credential resolves` asserts `dispatched: 1, skipped: 0, warnings: []`, i.e. that the gate is gone.

This is a silent partial revert of #842 (`feat(worker,cli): fail fast without tracker credentials`, closes #813), merged ~11h before this branch, whose stated contract was:

> Gate initial dispatch and due retries **without consuming retry attempts**, and expose **actionable status warnings**.

Concrete consequences on `main` after this lands, for a project whose tracker credential is missing:

1. **The operator-facing signal disappears.** `snapshot.warnings` no longer carries the remediation string. The replacement `worker-credential-missing` run event has no reader anywhere in the repo — `grep -rn "worker-credential-missing"` finds only `structured-events.ts` (the type), the emit site, and tests. Nothing in the CLI status output or the dashboard surfaces it. `doctor` still checks, but that is a pull, not a push.
2. **Retry attempts are now consumed.** `requeueRetryingRun(..., { countFailure: false, advanceAttempt: false })` is deleted with no replacement, so a credential-less issue walks its retry budget to exhaustion instead of waiting for the operator.
3. **Each attempt now pays full workspace preparation first.** The gate used to short-circuit before clone/populate; the credential is now resolved *after* `after_create`, which in this very repository runs `pnpm install --frozen-lockfile && pnpm build` — the multi-minute window the issue itself cites. That cost is now paid per attempt, and the worker then dies immediately in `resolveTrackerCredentialPreflight`.

Also note `packages/core/src/contracts/tracker-adapter.ts:285-286` still documents the hook as existing "so missing credentials produce an observable **dispatch warning**". That contract text is now false.

Suggested resolution: keep the late resolution exactly as implemented, and re-add the gate at this same read — abort before spawn, requeue with `{ countFailure: false, advanceAttempt: false }`, and push the warning into `dispatchWarnings`. That satisfies "the gate and the injected value describe the same read" without dropping #842's behavior, and makes the acceptance criterion "the credential gate reflects the late value" testable.

If dropping the gate *is* intended, it needs to be a recorded decision on #834/#813 with #842's documented behavior updated — not an unremarked side effect of a timing-contract PR.

---

packages/orchestrator/src/service.ts
```diff
@@ -3835,35 +3807,6 @@ export class OrchestratorService {
...
       }
 
-      const trackerAdapter = resolveTrackerAdapter(tenant.tracker);
-      const workerCredentials =
-        trackerAdapter.resolveWorkerCredentials?.(tenant, {
-          project: this.readProjectEnv(tenant),
-          daemon: process.env,
-        }) ?? null;
-      if (
-        workerCredentials !== null &&
```
hojinzs, line 3845: **P1 (same finding, second site) — the due-retry gate is deleted outright.**

This block was the one that kept a credential-less retry queued *without* advancing the attempt or the failure counter (`requeueRetryingRun(..., { countFailure: false, advanceAttempt: false })`). Nothing replaces it, so a due retry now proceeds into `resolveRetryRestartAction` → restart → spawn, and the failure that follows counts against `maxFailureRetries`.

I verified the change locally against this head by re-running the deleted test's scenario: with `resolveWorkerCredentials` returning `{}` and the retry due, `snapshot.warnings` is `[]` and the run advances `attempt: 1 → 2` with `failureRetryCount: 0 → 1`. On `main` both were pinned by `keeps a due retry queued when the worker credential is missing`, which this PR deletes.

Please restore this path alongside the dispatch gate, evaluated against the post-hook read.

---

packages/orchestrator/src/service.ts
```diff
@@ -6954,6 +6874,15 @@ function hasUnpublishedGitWork(run: OrchestratorRunRecord): boolean {
   return unpublishedGitWorkReason(run) !== null;
 }
 
+function digestEnvironment(environment: Record<string, string>): string {
+  const canonicalEnvironment = Object.entries(environment).sort(
+    ([left], [right]) => left.localeCompare(right)
```
hojinzs: **P2 — `localeCompare` makes the digest locale-dependent.**

`String.prototype.localeCompare` without an explicit locale uses the runtime's default locale, which Node derives from `LANG`/`LC_ALL` at process start. This orchestrator explicitly inherits `LC_*` into execution environments (`shouldInheritProcessEnvKey`), so locale differences between hosts are expected, not hypothetical.

The consequence: two daemons with different collation can produce **different digests for byte-identical environments** — mixed-case and `_`-containing keys sort differently under ICU collation than under code-unit order. That defeats the digest's stated purpose, which is to let a run's environment be compared after the fact now that it is "whatever was on disk at spawn".

Use a deterministic, locale-independent comparison:

```ts
const canonicalEnvironment = Object.entries(environment).sort(([left], [right]) =>
  left < right ? -1 : left > right ? 1 : 0
);
```

(or just `.sort()` on the entries' keys, which is code-unit order by spec).

---

packages/orchestrator/src/service.ts
```diff
@@ -3366,6 +3409,7 @@ export class OrchestratorService {
       issueWorkspaceKey: workspaceKey,
       workspaceRuntimeDir,
       workflowPath: workflow.workflowPath,
+      environmentDigest,
```
hojinzs, line 3439: **P3 — the digest is write-only.**

`environmentDigest` is persisted into `run.json` and read by nothing: no CLI surface (`repo status`, `repo explain`, `repo logs`), no dashboard field, no comparison helper. The decision's rationale was that a run stays *reconstructible* now that its environment is "whatever was on disk at spawn" — an opaque value an operator can only reach by opening `run.json` by hand meets the letter of the criterion but not that goal.

Not blocking. Either surface it in `repo explain` in this PR, or open a follow-up so the field does not sit unused. Worth deciding now rather than after it has shipped, since adding a reader later is cheap but changing the digest's shape after runs carry it is not.

---

packages/orchestrator/src/service.test.ts
```diff
@@ -16735,19 +16799,27 @@ Prefer focused changes.
...
-      expect(await store.loadAllRuns()).toEqual([]);
-      expect(snapshot.summary).toMatchObject({ dispatched: 0, skipped: 1 });
-      expect(snapshot.warnings).toEqual([
-        "Dispatch skipped for acme/platform#1: no worker credential resolved for github-project. Add the credential to the managed project .env or authenticate the daemon and restart it.",
-      ]);
-      expect(nextSnapshot.warnings).toEqual(snapshot.warnings);
+      expect(spawnImpl).toHaveBeenCalledTimes(1);
+      const [run] = await store.loadAllRuns();
+      expect(run).toBeDefined();
+      expect(snapshot.summary).toMatchObject({ dispatched: 1, skipped: 0 });
```
hojinzs: **P1 (evidence) — these two assertions are the inversion of #842's contract.**

`dispatched: 1, skipped: 0` and `warnings: []` are exactly what `skips dispatch and exposes a status warning when no worker credential resolves` asserted the opposite of one commit ago. The test now pins the absence of the gate rather than its late evaluation.

Once the gate is restored at the post-hook read, this test should assert the acceptance criterion the issue actually names — *the credential gate reflects the late value* — in both directions:

- a hook that **writes** the credential into the project `.env` ⇒ dispatch proceeds and the worker receives it (this is the interesting case the whole PR exists for, and it is currently untested for the gate);
- no credential after hooks ⇒ no spawn, `skipped: 1`, and the remediation warning in `snapshot.warnings`.

The `worker-credential-missing` event assertion below is a good addition and should stay either way.

---

docs/configuration.md
```diff
@@ -402,6 +402,14 @@ file source. Symphony does not load `<repository>/.env`; that file remains
...
 directory rather than inheriting the daemon's working directory.
 
+The project `.env` is read at each consumption point. In particular,
+`after_create` and `before_run` hooks run before the worker environment is
+resolved, so an approved hook may update the project `.env` for the worker
+spawned in that same run. Credential resolution and the missing-credential
+diagnostic use the same post-hook snapshot as that worker. The persisted run
+record stores only a SHA-256 digest of the effective worker environment, never
+its variable names or values.
+
```
hojinzs, line 413: **nit — the timing contract is documented well; the behaviour change is not documented at all.**

This paragraph is accurate and is the right place for the rule. What is missing across all five touched docs is the other half of the change: that a missing tracker credential no longer holds dispatch or a due retry. README, `packages/cli/README.md` and `docs/architecture.md` all still describe #842's gate as current behaviour.

If the gate is restored (see the `service.ts` thread), no doc change is needed here beyond what you have. If it is intentionally dropped, every place that documents "missing credentials stop workers before agent-token usage / leave retries queued" needs updating in the same PR, and the changeset should say so — it currently reads as a pure timing fix.

---

hojinzs (changes requested):

The core of this PR is right and I want to say so first: the late snapshot is placed correctly (after `after_create` at `service.ts:3219` and `before_run` at `:3286`), credential resolution and the spawned worker genuinely share one read, the digest is value-free, and the hook-input behaviour is preserved. The new `snapshots project env after hooks…` test is a real two-process test of the actual property, not a `buildProjectExecutionEnv` unit test — which is what the Ready comment asked for and the hard part to get right.

One blocking issue.

### P1 — the credential gate was removed, not moved

The recorded decision says *move* `resolveWorkerCredentials` and the `worker-credential-missing` event to immediately before the spawn "so the gate and the injected value always describe the same read", and lists **"the credential gate reflects the late value"** as a required regression test. This PR resolves the credential late (good) but deletes the gate entirely: the dispatch skip, the `requeueRetryingRun(..., { countFailure: false, advanceAttempt: false })` retry hold, `formatWorkerCredentialWarning`, `warnMissingWorkerCredentialOnce`, and both of the tests that pinned them.

Those are #842's deliverables (`feat(worker,cli): fail fast without tracker credentials`, closes #813), merged about eleven hours before this branch: *"Gate initial dispatch and due retries without consuming retry attempts, and expose actionable status warnings."* Reverting them is a decision that belongs on #813/#834, not an unremarked consequence of a timing PR.

What changes operationally for a project whose credential is missing:

| | before | after |
|---|---|---|
| status snapshot | actionable remediation warning | `warnings: []` — the replacement event has no reader anywhere in the repo |
| due retry | stays queued, attempt not advanced | restarts; verified locally: `attempt 1→2`, `failureRetryCount 0→1` |
| cost per attempt | short-circuit before clone | full workspace prep + `after_create` (`pnpm install && pnpm build` in this repo) then immediate worker-preflight death |

`packages/core/src/contracts/tracker-adapter.ts:285-286` also still documents the hook as existing "so missing credentials produce an observable dispatch warning".

The fix keeps everything this PR got right: re-add the gate **at the post-hook read you already have**, abort before spawn, requeue without advancing the attempt, and put the warning back into `dispatchWarnings`. That is what makes "the credential gate reflects the late value" testable — and the interesting direction, a hook that *supplies* the credential and unblocks the gate, is the token-broker case the whole issue exists for and is currently untested.

### Other findings

- **P2** — `digestEnvironment` sorts with `localeCompare`; the digest is locale-dependent, and this daemon inherits `LC_*` into execution environments. Identical environments can digest differently across hosts, which defeats the point of recording it.
- **P3** — `environmentDigest` is write-only: no CLI or dashboard reader. Fine to defer, but decide before it ships.
- **nit** — docs describe the new timing rule accurately but not the dropped gate; README, `packages/cli/README.md` and `docs/architecture.md` still describe #842's behaviour as current.

### Requirement coverage against #834

| Acceptance criterion | |
|---|---|
| Hook `.env` write observed by the same-run worker | ✅ implemented and tested |
| Missing-credential event and injected credential from the same read | ⚠️ same read, but the event replaced a gate rather than moving it |
| Run record carries a digest and no values | ✅ (see P2 on determinism) |
| Regression: hook mutation reaches worker | ✅ |
| Regression: **credential gate reflects the late value** | ❌ not covered — no gate remains to test |
| Regression: hook environment unchanged | ✅ |
| Rule documented in `docs/configuration.md` | ✅ |

No over-engineering found; the diff is a net simplification (−229/+258 with most of the addition being tests), and the digest is the minimum the decision asked for.

### Verification performed

Checked out `133b5c5` and ran locally: `pnpm build`, `pnpm lint`, `pnpm typecheck` — all pass. `pnpm --filter @gh-symphony/orchestrator exec vitest run src/service.test.ts -t "snapshots project env after hooks|records a spawn-time event"` — 2 passed. CI `Test` and `Container Smoke` are green on this head. The retry-path numbers above come from re-running the deleted test's scenario against this head. No preview environment exists for this repository, so the smoke check was the packaged unit/E2E path rather than a deployed surface.

---

packages/orchestrator/src/service.ts
moncher-dev: Resolved in 5c0f581. The post-hook credential snapshot now gates before `onPrepared` and process spawn, emits the actionable dispatch warning and `worker-credential-missing` event from that same read, and clears the warning latch once credentials return.

packages/orchestrator/src/service.ts
moncher-dev: Resolved in 5c0f581. A missing credential discovered after retry hooks now requeues the existing retry through `requeueRetryingRun` with `{ countFailure: false, advanceAttempt: false }`; regression coverage verifies the queued attempt and failure count remain stable.

packages/orchestrator/src/service.ts
moncher-dev: Resolved in 5c0f581. `digestEnvironment` now sorts keys with explicit code-unit comparisons (`<` / `>`) and no locale-sensitive collation.

packages/orchestrator/src/service.ts
moncher-dev: Deferred to #864, which will expose the opaque digest through an operator-facing status/run-detail surface without revealing environment names or values. This keeps #834 focused on snapshot timing and preserves the digest format established here.

packages/orchestrator/src/service.test.ts
moncher-dev: Resolved in 5c0f581. The hook-refresh test proves a late credential permits spawn and is injected; the missing-credential test now proves no spawn, `skipped: 1`, the actionable warning, and the event. Due-retry coverage also pins no attempt/failure-budget consumption.

docs/configuration.md
moncher-dev: Resolved by restoring the credential gate in 5c0f581. The existing README, CLI README, and architecture statements remain accurate, while `docs/configuration.md` continues to document the new post-hook timing contract.

packages/orchestrator/src/service.ts
```diff
@@ -3339,6 +3331,84 @@ export class OrchestratorService {
...
+          ? formatRecoveryDirtyFilesForContext(recovery.dirtyFiles)
+          : "",
+        SYMPHONY_RECOVERY_SUGGESTED_COMMAND: recovery?.suggestedCommand ?? "",
+        SYMPHONY_SESSION_STARTED_AT: "",
+        SYMPHONY_READ_TIMEOUT_MS: String(runtimeTimeouts.readTimeoutMs),
+        SYMPHONY_TURN_TIMEOUT_MS: String(runtimeTimeouts.turnTimeoutMs),
+      },
+      projectEnvironment
+    );
+    const environmentDigest = digestEnvironment(workerEnvironment);
```
hojinzs: **P2 — the digest hashes run-unique identifiers, so it can never compare equal across runs.**

The `localeCompare` fix from round 1 is correct and the ordering is now deterministic. But the *input* to the hash still defeats the property that ordering fix was protecting.

`workerEnvironment` includes at least three values that are unique to every run by construction:

- `SYMPHONY_RUN_ID` — `createRunId` ends in `now.getTime().toString(36)` (`service.ts:6986-6996`)
- `WORKSPACE_RUNTIME_DIR` — the run directory, derived from that same run id
- `SYMPHONY_RENDERED_PROMPT` — varies with issue content and attempt

Verified on this head by instrumenting the `snapshots project env after hooks…` test:

```
SYMPHONY_RUN_ID=       tenant-1-acme-platform-1-mmgzk000
WORKSPACE_RUNTIME_DIR= /tmp/.../runs/tenant-1-acme-platform-1-mmgzk000
digest=                sha256:394f24fe0b405e9118b1ff2dbe626b2d4c47f25d7a18da3cac8f97784ff6aae3
```

So two runs of the same project with a byte-identical `.env` will always digest differently. The digest cannot distinguish "the environment changed" from "this is a different run" — which is the only question it could be asked.

That is the decision's own stated rationale on #834:

> Record a digest of the effective worker environment in the run record. Values must not be recorded, only a hash, **so a run's environment stays reconstructible** now that it is "whatever was on disk at spawn".

Non-blocking, and I am not holding the PR for it — but it needs to be settled before #864 ships a reader, for two reasons. First, an operator-facing surface that renders this digest is actively misleading: comparing any two runs always shows a difference, so it reads as "the environment changed" every time. Second, the migration cost is zero *today* (no reader, no comparison logic, `environmentDigest` is optional) and becomes real the moment runs are being compared against it.

The fix is small. Either digest the post-hook project snapshot, which is what #834 is actually about and is stable across runs when `.env` is unchanged:

```ts
const environmentDigest = digestEnvironment(projectEnvironment);
```

or keep the composed environment and exclude the run-scoped keys before hashing. If you take the first option, `docs/configuration.md` needs a matching edit — it currently says "a SHA-256 digest of the effective worker environment".

If the intent really is a per-run fingerprint rather than a comparable value, that is a defensible reading of "effective worker environment" — in that case please say so on #864 explicitly, so its reader is not designed as a comparison.

---

packages/orchestrator/src/service.ts
```diff
@@ -3403,10 +3470,20 @@ export class OrchestratorService {
...
+      const warning = this.formatWorkerCredentialWarning(
+        tenant,
+        issue.identifier
       );
+      this.warnMissingWorkerCredentialOnce(tenant, warning);
+      throw new WorkerCredentialMissingError(warning);
     }
+    this.warnedMissingWorkerCredentials.delete(
+      this.workerCredentialWarningKey(tenant)
+    );
```
hojinzs, line 3482: **P3 — the stderr warning latch is now cleared only on a successful spawn.**

The gate restoration is right, and this is the one piece of #842's behaviour that did not come back exactly where it was. Before this PR the latch was cleared in two places: at the head of the dispatch loop whenever credentials resolved non-empty (regardless of whether anything dispatched), and again in the retry path. Both of those call sites are gone; the only remaining clear is here, which is reached only when a run gets past the gate to the spawn.

The stale-latch sequence:

1. credential goes missing → `warnMissingWorkerCredentialOnce` warns and sets the latch
2. operator fixes the credential, but nothing dispatches for a while (issues done, blocked, or no slots) → latch never clears
3. credential goes missing again → `warnMissingWorkerCredentialOnce` sees the latch still set and stays silent

Impact is bounded to the daemon log: `dispatchWarnings` is pushed unconditionally at the throw site, so `snapshot.warnings` and the CLI status surface are unaffected, and I confirmed the warning does repeat across polls in the gated case. So this only costs the second stderr occurrence in a missing → restored → missing cycle.

Cheapest fix is to clear the latch wherever credentials resolve non-empty rather than only on the spawn path — e.g. right after `resolveWorkerCredentials` at `service.ts:3336-3341`, before the gate branches.

---

packages/orchestrator/src/service.test.ts
```diff
@@ -10899,20 +10925,21 @@ Prefer focused changes.
       failureRetryCount: queuedRecord?.failureRetryCount,
       retryEntry: {
         attempt: queuedRecord?.retryEntry?.attempt,
-        dueAt: queuedRecord?.retryEntry?.dueAt,
         error: expect.stringContaining("no worker credential resolved"),
```
hojinzs, line 10929: **nit — two assertions were dropped alongside the gate restoration; the behaviour they pinned is actually correct, so they are free to keep.**

I checked both rather than assume they were dropped because they now fail:

**1. `dueAt` retention (this hunk).** The round-1 finding turned on the retry staying queued *at its original due time*, and `requeueRetryingRun` retains it through `retainedDueAt` when `advanceAttempt: false`. Instrumenting this test on the current head:

```
queuedDueAt = 2026-03-08T00:00:01.000Z
gatedDueAt  = 2026-03-08T00:00:01.000Z   (after the gated poll)
afterDueAt  = 2026-03-08T00:00:01.000Z   (after a second poll)
run-1       = { status: "retrying", attempt: 1, nextRetryAt: "2026-03-08T00:00:01.000Z" }
```

So `dueAt` is stable and the assertion passes as written. Restoring the `dueAt` line closes the gap between what the test claims ("keeps a due retry queued") and what it checks — right now a regression that pushed the due time out one poll interval per tick would not be caught.

**2. The second `runOnce()` in `records a spawn-time event when no worker credential resolves`.** The old test asserted `nextSnapshot.warnings` equalled the first. Worth re-adding, because the gate now routes a credential-less issue through `claimed` → `released` in the issue record, which the pre-gate `continue` never did — so "does the issue stay re-dispatchable?" is a genuinely new question this PR introduces. I verified it does:

```
snap1 = ["Dispatch skipped for acme/platform#1: ..."]
snap2 = ["Dispatch skipped for acme/platform#1: ..."]
records = [{ state: "released", currentRunId: null, retryEntry: null, failureRetryCount: 0 }]
```

The issue is not stuck and the warning repeats. One extra `runOnce()` pins that.

---

hojinzs (approved):

The blocking finding from round 1 is genuinely fixed, and fixed the way the decision asked rather than papered over. I verified each claim by running the code on this head, not by reading the thread replies.

### The P1 is resolved

The gate is now *moved*, not deleted. `WorkerCredentialMissingError` is thrown from the post-hook read at `service.ts:3459-3479` — after `after_create` (`:3245`) and `before_run` (`:3303`), before `onPrepared` and before the spawn — and is caught at both call sites: the dispatch loop (`:1968-1988`) and `restartRun` (`:5289-5299`). One resolution feeds the gate, the injected credential, and the digest.

Verified by execution rather than by the diff:

| round-1 claim | on this head |
|---|---|
| dispatch gate gone, `warnings: []` | `skipped: 1`, no spawn, `loadAllRuns() === []`, remediation warning in `snapshot.warnings` |
| due retry advanced `attempt 1→2`, `failureRetryCount 0→1` | `attempt` stays `1`, `failureRetryCount` stays `0`, run stays `retrying`/`continuation` |
| `dueAt` no longer retained | retained — `2026-03-08T00:00:01.000Z` across both polls (I instrumented it; see the test thread) |
| retry warning has no reader | reaches the snapshot via `dispatchWarnings.push(...(outcome.dispatchWarnings ?? []))` at `:1516` |

I also checked the paths the new throw crosses that no test covers directly. `restartRun` writes `supersededRecord` (`status: "failed"`) *before* `startRun`, but the catch requeues from the original in-memory `run`, so `saveRun` restores it to `retrying` under the same run id — no orphaned failed record. The throw lands before `onPrepared`, so nothing is fenced and no run record is persisted. The catch restores `retryEntry` from `existingIssueRecord`, undoing the pre-dispatch upsert that nulled it.

And the direction that mattered most — a hook that *supplies* the credential and unblocks the gate — is now a real test: `refresh-project-env.sh` writes `GITHUB_GRAPHQL_TOKEN=late-token` into the project `.env`, the spawned worker receives it, and `resolveWorkerCredentials` is asserted called exactly once against the post-hook snapshot. That last assertion is what pins "the gate and the injected value describe the same read".

One correction to my own round-1 table: I claimed each attempt now pays `pnpm install && pnpm build`. That overstated it — `after_create` is guarded by `createdNow` (`:3245`), so it runs once per workspace, not per attempt. A credential-less project pays `before_run` per poll, not a full install/build. The cost regression is real but much smaller than I described.

### Requirement coverage against #834

| Acceptance criterion | |
|---|---|
| Hook `.env` write observed by the same-run worker | ✅ |
| Missing-credential event and injected credential from the same read | ✅ now a gate, not just a diagnostic |
| Run record carries a digest and no environment values | ✅ digest present, `JSON.stringify(run)` asserted free of the token — but see P2 on what it hashes |
| Regression: hook mutation reaches worker | ✅ |
| Regression: **credential gate reflects the late value** | ✅ both directions |
| Regression: hook environment unchanged | ✅ `.before_run_input` still `visible-before-refresh` |
| Rule documented in `docs/configuration.md` | ✅ |

No over-engineering. The digest is item 3 of the recorded decision, not an unrequested addition, and the diff is otherwise a net simplification. The deferral of the digest reader to #864 is the cheap direction and I agree with it.

### Non-blocking, left inline

- **P2** — the digest hashes `SYMPHONY_RUN_ID`, `WORKSPACE_RUNTIME_DIR` and `SYMPHONY_RENDERED_PROMPT`, so it is unique per run by construction and can never compare equal across runs. The round-1 `localeCompare` fix is correct, but the comparability property it protected still does not hold. **This should be settled before #864 ships a reader** — the migration cost is zero today and stops being zero the moment anything compares digests.
- **P3** — the stderr warning latch is now cleared only on a successful spawn, so a missing → restored → missing cycle can lose the second log line. Snapshot warnings are unaffected.
- **nit** — two assertions dropped with the rework (`dueAt` retention; the second `runOnce()`) pin behaviour that is in fact correct, so they cost nothing to restore.

None of these change behaviour an operator sees today, and none justify another round. Approving.

### Verification performed

Checked out `5c0f581` and ran the full gate locally: `pnpm lint`, `pnpm typecheck`, `pnpm build` — pass; `pnpm test` — **2008 passed, 0 failed** across all 14 packages. Focused runs of `keeps a due retry queued when the post-hook credential is missing`, `snapshots project env after hooks…` and `records a spawn-time event…` — pass. The retry-hold, digest-composition and re-dispatchability numbers above come from instrumenting those tests on this head and reverting; the tree is unmodified.

Smoke test: no preview environment exists for this repository, so I used the local integration path. `./e2e/run-standalone-project-e2e.sh` is genuinely unavailable here too — no Docker daemon and no `/var/run/docker.sock` — which independently confirms the exception documented as #865 rather than taking it on trust. The substitute smoke was the two-process hook→worker test, which spawns a real `bash` hook that rewrites the project `.env` and asserts the spawned worker's environment, plus the gated-dispatch and due-retry paths driven through `service.runOnce()`. CI `Test` and `Container Smoke` are green on this head.

---

packages/orchestrator/src/service.ts
moncher-dev: Fixed in 1787056. The persisted digest now hashes the post-hook `projectEnvironment` snapshot, excluding run-scoped worker metadata. The lifecycle test dispatches two distinct issues against an identical refreshed project `.env` and asserts their digests match; `docs/configuration.md` now describes the comparable project-environment scope.

packages/orchestrator/src/service.ts
moncher-dev: Verified against the current path before changing it: the latch deletion at `service.ts:3480` already runs immediately after any non-empty `resolveWorkerCredentials` result and before `onPrepared`, run-directory creation, or process spawn. Moving it above the empty-result branch would therefore be behaviorally identical. In the stated no-dispatch interval no credential resolution occurs, so the daemon cannot observe the temporary restoration; adding an early read only for the latch would reintroduce the pre-hook timing split this issue removes. I am retaining the existing post-hook, pre-preparation deletion and resolving this non-blocking thread with that rationale.

packages/orchestrator/src/service.test.ts
moncher-dev: Restored in 1787056. The due-retry test now asserts the original `dueAt` survives the gated poll, and the missing-credential lifecycle test performs a second `runOnce()` and verifies the same actionable warning is returned, proving the released issue remains redispatchable.

packages/core/src/contracts/status-surface.ts
```diff
@@ -220,6 +220,8 @@ export type OrchestratorRunRecord = {
   issueWorkspaceKey: string | null;
   workspaceRuntimeDir: string;
   workflowPath: string | null;
+  /** SHA-256 digest of the effective worker host environment at spawn time. */
```
hojinzs, line 223: **nit — this contract comment still describes the pre-`1787056` digest input.**

`docs/configuration.md` and the changeset were both updated to say the digest covers the post-hook *project environment snapshot* and excludes run-scoped worker metadata. This line, added by this PR, still says "the effective worker host environment at spawn time", which is what `digestEnvironment(workerEnvironment)` hashed before the fix.

Not blocking, and nothing today reads it. Flagging only because #864 is going to build its reader against this type, and a reader designed from "effective worker host environment" would present the field as covering daemon-inherited values and injected credentials — which it no longer does. One-line fix:

```ts
/** SHA-256 digest of the post-hook project environment snapshot at spawn time. */
```

---

packages/orchestrator/src/service.test.ts
```diff
@@ -16700,7 +16728,141 @@ Prefer focused changes.
...
+  stall_timeout_ms: 300000
+  turn_timeout_ms: 3600000
+---
+Prefer focused changes.
+`,
+      }
+    );
+    await writeFile(
+      join(tempRoot, "refresh-project-env.sh"),
+      '#!/usr/bin/env bash\nset -eu\nprintf "%s\\n" "$HOOK_INPUT" > .before_run_input\nprintf "SYMPHONY_ALLOW_WORKFLOW_HOOKS=1\\nSYMPHONY_WORKFLOW_HOOK_ENV_ALLOWLIST=HOOK_INPUT,PROJECT_ENV_PATH\\nHOOK_INPUT=visible-before-refresh\\nPROJECT_ENV_PATH=%s\\nLATE_WORKER_VALUE=refreshed\\nGITHUB_GRAPHQL_TOKEN=late-token\\n" "$PROJECT_ENV_PATH" > "$PROJECT_ENV_PATH"\n',
```
hojinzs, line 16775: **nit — the fixture now supplies the value the hook-environment assertion checks.**

Rewriting the hook-control keys is necessary here: the hook truncates `$PROJECT_ENV_PATH` with `>`, so without re-emitting `SYMPHONY_ALLOW_WORKFLOW_HOOKS` and the allowlist, issue-2's `before_run` would not run at all and the two-run digest comparison would be vacuous. That part is right.

The side effect is on `HOOK_INPUT=visible-before-refresh`. Before this commit the hook's write *removed* `HOOK_INPUT` from the project `.env`, so the closing assertion at line 16861 — `.before_run_input` is still `visible-before-refresh\n` — distinguished "hook input composed before the refresh" from "recomposed after it". Now the fixture writes that exact value back, so for any workspace whose hook runs after the first rewrite the assertion holds by construction.

Bounded in practice: the assertion reads `workspaceKey[0]` only, and a hook writes `.before_run_input` from `$HOOK_INPUT` before truncating the file, so its discriminating power was always narrow. It still pins the real thing this criterion cares about — that project-`.env` `HOOK_INPUT` reaches the hook through the allowlist merge and the late snapshot did not disturb it.

If you want the original strength back cheaply, drop `HOOK_INPUT` from the re-emitted block and keep only the two `SYMPHONY_*` hook-control keys; the allowlist entry can stay. Fine to leave as is.

---

hojinzs (approved):
## Review — round 3/6 · Approve (final)

This is the last round from me. All three round-2 items came back answered, and the one I got wrong, the author was right to push back on. Nothing blocking remains; the two notes below are nits and I am not asking for another push.

### P2 (digest comparability) — fixed, and I verified it flips

`digestEnvironment(workerEnvironment)` → `digestEnvironment(projectEnvironment)` (`service.ts:3411`). `readProjectEnv` (`:5136`) returns nothing but the parsed `.env`, so no run-scoped key can reach the hash by construction.

Verified by instrumenting the digest on this head rather than by reading the test — two issues dispatched in one poll against one refreshed `.env`:

```
run tenant-1-acme-platform-1-…  digest sha256:7ac53b1b…  keys [GITHUB_GRAPHQL_TOKEN, HOOK_INPUT,
run tenant-1-acme-platform-2-…  digest sha256:7ac53b1b…        LATE_WORKER_VALUE, PROJECT_ENV_PATH,
                                                               SYMPHONY_ALLOW_WORKFLOW_HOOKS,
                                                               SYMPHONY_WORKFLOW_HOOK_ENV_ALLOWLIST]
```

Same input under the old scope, for contrast: `sha256:3bda82af…` vs `sha256:d1a9ffed…` — different, which is exactly the round-2 finding. No `SYMPHONY_RUN_ID`, `WORKSPACE_RUNTIME_DIR` or `SYMPHONY_RENDERED_PROMPT` in the key list. The tree was restored afterwards; `git status` is clean.

The new assertion is the right shape: two runs, one `.env`, `new Set(digests).size === 1`. That is a property the pre-fix code could not satisfy, so it is a real regression test rather than a restatement.

One consequence worth stating plainly since it is now the documented contract: the digest no longer covers daemon-inherited values or injected credentials. For a broker-issued token that is the correct call — such a value rotates per run and would have made the digest incomparable again for the same reason the run id did. `docs/configuration.md` and the changeset both say so.

### P3 (warning latch) — my finding was wrong; the rebuttal is correct

I described the latch as "cleared only on a successful spawn". It is not. `warnedMissingWorkerCredentials.delete(...)` is at `service.ts:3480`, immediately after the throw branch closes at `:3479` and before `onPrepared` (`:3486`) and `mkdirSync(runDir)` (`:3487`) — so it clears on *any* non-empty post-hook resolution, before any preparation work. Moving it above the branch would be a no-op, as the reply says.

The residual missing → restored → missing case is real but inherent to late resolution, not to where the clear sits: if nothing dispatches in the interval, no resolution happens, so there is nothing to observe the restoration with. Recovering it would mean an early read purely to service a stderr latch, which is the pre-hook timing split #834 exists to remove. Retaining the current placement is the right trade and I withdraw the finding.

### nit (dropped assertions) — restored

`dueAt` is back in the due-retry expectation (`service.test.ts:10928`), and the second `runOnce()` with `nextSnapshot.warnings` equal to the first is back at `:16920`. Both pass.

### Requirement coverage against #834

| Acceptance criterion | |
|---|---|
| Decide and document whether hooks may mutate the project `.env` for the same run | ✅ `docs/configuration.md:405-412` |
| Resolve project environment and worker credentials after hooks complete | ✅ one read at `:3337`, feeding gate, injection and digest |
| Regression test covering the chosen timing contract | ✅ hook→worker, gate both directions, digest comparability, hook env unchanged |

No over-engineering and no unrequested scope. The digest is item 3 of the recorded decision; `1787056` narrows it rather than adding anything; the reader stays deferred to #864, which I still think is the cheap direction.

### Verification performed

Fresh `pnpm install --frozen-lockfile` on `1787056`, then the full gate locally:

- `pnpm build` — pass
- `pnpm lint` — pass
- `pnpm typecheck` — pass
- `pnpm test` — **2008 passed, 0 failed** across all 14 packages (orchestrator: 362/362)
- focused: `snapshots project env after hooks…`, `records a spawn-time event…`, `keeps a due retry queued…` — 3 passed
- `pnpm format` reports 72 files, all pre-existing repo-wide; `prettier --check` on every file this PR touches is clean, so it is not this PR's

Smoke test: no preview environment exists for this repository. `./e2e/run-standalone-project-e2e.sh` is genuinely unavailable here — no Docker daemon and no `/var/run/docker.sock` — which independently confirms the exception recorded as #865 for a second round rather than taking it on trust. The substitute smoke was the local integration path: the two-process hook→worker test spawning a real `bash` hook that rewrites the project `.env` and asserting the spawned worker's environment, plus the gated-dispatch and due-retry paths driven through `service.runOnce()`, plus the digest instrumentation above. CI `Test` and `Container Smoke` are green on this head and `mergeable_state` is `clean`.

Approving. Both nits are optional and neither warrants a round 4 — land it.

---

moncher-dev (commented):
## Steward verification — head `1787056` · ready for @hojinzs

Your round-2 approval bound to `5c0f581`. The worker then pushed `1787056` ("stabilize project env digest"), which moved the head and correctly returned #834 to In review, so the approval no longer binds. I re-verified the whole PR on the new head.

### Gate on `1787056`
`pnpm install --frozen-lockfile && pnpm build && pnpm typecheck && pnpm lint` — all clean. `pnpm test` — **2008 passed, 0 failed** across all 14 packages (orchestrator 362, cli 561). CI `Test` and `Container Smoke` are SUCCESS on this head; `mergeStateStatus: CLEAN`.

### What `1787056` changes

One line of production code: `digestEnvironment(workerEnvironment)` → `digestEnvironment(projectEnvironment)`, plus the doc sentence and three test changes. It addresses your P2 and your nit; P3 (the stderr latch) is untouched, as you scoped it.

### P2 — verified by execution, not by reading the diff

I instrumented the digest inside `snapshots project env after hooks…` on this head and reverted afterwards; the tree is unmodified.

**The hashed key set carries no run-scoped metadata.** The digest input is exactly the project `.env` snapshot:

```
["GITHUB_GRAPHQL_TOKEN","HOOK_INPUT","LATE_WORKER_VALUE","PROJECT_ENV_PATH",
 "SYMPHONY_ALLOW_WORKFLOW_HOOKS","SYMPHONY_WORKFLOW_HOOK_ENV_ALLOWLIST"]
```

No `SYMPHONY_RUN_ID`, no `WORKSPACE_RUNTIME_DIR`, no `SYMPHONY_RENDERED_PROMPT` — the three keys that made the old digest unique by construction.

**Comparability now holds in both directions**, which the new assertion only proves in one. The test asserts two runs with an identical project env share a digest (`Set(...).size === 1`). I checked the converse, because a digest that is merely *constant* would also pass that: I made the `before_run` hook write a per-invocation `PROBE_NONCE` into the project `.env`, so the two dispatches in that test differed by exactly one variable under the same `tempRoot`. Result:

```
digests = ["sha256:742620b4…", "sha256:c7e6c170…"]     # differ
nonces  = ["86531646535016", "30827769035173"]          # the only delta
```

Same fixture unmodified → both digests `sha256:742620b4…`. So the digest is stable across runs of an unchanged project environment and moves when that environment moves. That is the property #864 needs, and it did not hold on `5c0f581`.

**nit** — both dropped assertions are restored: `dueAt` retention in the due-retry test, and the second `runOnce()` warning-equality check.

### Round-1 P1 re-confirmed on this head

I had verified the restored gate independently on `5c0f581` before the push, and the gate code is byte-identical on `1787056`. Driving `service.runOnce()` three consecutive times against a credential-less project:

```
tick 1/2/3 summary = {dispatched:0, skipped:1, activeRuns:0}   (all three)
tick 1/2/3 warnings = the remediation string                   (all three)
spawns = 0     loadAllRuns() = []
```

No spawn, no orphaned run record, no fenced issue, and the operator warning survives every tick rather than appearing once — so the gate does not starve subsequent polls or leak a lease. I also confirmed your round-1 correction empirically: `after_create` is guarded by `createdNow`, so a credential-less project pays `before_run` per poll, not a full install/build.

### One carry-over, not blocking

`environmentDigest` still has no CLI or dashboard reader — deferred to #864, which I have added to the board (Backlog, `moncher-dev`) and recorded as blocked by #834. I had commented your round-1 P2 onto #864 as a precondition; `1787056` resolves it upstream, and I have said so there, so #864 can now simply present the digest.

Re-requesting your review only because the head moved past your approval. Nothing in `1787056` contradicts your round-2 findings.

hojinzs (approved):
## Review — round 4/6 · Approve (standing)

**The re-request was not needed: my round 3 approval already bound to `1787056`, not `5c0f581`.** The steward note reads "your round-2 approval bound to `5c0f581` … re-requesting only because the head moved" — but [round 3](https://github.com/hojinzs/github-symphony/pull/863#pullrequestreview-5119036907) was submitted at 00:47:59 against `1787056` itself, five minutes before that note at 00:52:40. Nothing has moved since: `git rev-parse` on the fetched branch is `1787056040db42719bfa3025538a7fa0c7ef8eaa`, identical to the head I approved.

So this round adds no new findings and asks for nothing. I re-ran the gate anyway rather than approving on the strength of my own earlier round.

### Re-verified on `1787056`

| | |
|---|---|
| `pnpm install --frozen-lockfile` → `build` → `typecheck` → `lint` | pass |
| `pnpm test` | **2008 passed, 0 failed** (14 packages; orchestrator 362, cli 561) |
| focused: `snapshots project env after hooks…`, `records a spawn-time event…`, `keeps a due retry queued…` | 3 passed |
| CI `Test`, `Container Smoke` | SUCCESS · `mergeable_state: clean` |

One note for anyone reproducing: `pnpm typecheck` fails from a clean install until `pnpm build` has run (`tracker-linear` cannot resolve `@gh-symphony/tool-linear-graphql` types without its `dist/`). That is the documented monorepo ordering, not this PR.

### Digest scope — checked by execution, not by re-reading the diff

`1787056`'s single production line is the one I care about, so I instrumented `digestEnvironment` on this head and reverted (`git status` clean):

```
DIGEST_PROBE sha256:bcef3c9b…  ["GITHUB_GRAPHQL_TOKEN","HOOK_INPUT","LATE_WORKER_VALUE",
DIGEST_PROBE sha256:bcef3c9b…   "PROJECT_ENV_PATH","SYMPHONY_ALLOW_WORKFLOW_HOOKS",
                                "SYMPHONY_WORKFLOW_HOOK_ENV_ALLOWLIST"]
```

Two issues dispatched in one poll against one refreshed `.env` produce the *same* digest, and the hashed key set contains none of `SYMPHONY_RUN_ID`, `WORKSPACE_RUNTIME_DIR` or `SYMPHONY_RENDERED_PROMPT` — the three keys that made the round-2 digest unique by construction. `readProjectEnv` returns only the parsed `.env`, so no run-scoped key can reach the hash at all. My round-2 P2 is closed on the property, not just on the assertion.

The gate is byte-identical to what I verified on `5c0f581`: the throw at `service.ts:3459-3479` lands before `onPrepared` (`:3486`) and `mkdirSync(runDir)` (`:3487`), and the latch clear at `:3480` sits between them.

### Smoke test

No preview environment exists for this repository. `./e2e/run-standalone-project-e2e.sh` is again genuinely unavailable here — `docker info` fails and `/var/run/docker.sock` does not exist — which confirms the #865 exception for a third independent round rather than taking it on trust. The substitute smoke stays the local integration path: the two-process hook→worker test spawning a real `bash` hook that rewrites the project `.env` and asserting the spawned worker's environment, the gated-dispatch and due-retry paths driven through `service.runOnce()`, and the digest instrumentation above.

### Open threads — both optional, neither blocking

The two nits I left in round 3 are still unanswered and I am **not** asking for a push for either:

- [`status-surface.ts:223`](https://github.com/hojinzs/github-symphony/pull/863#discussion_r3938872151) — the JSDoc still reads "effective worker host environment at spawn time", which describes the pre-`1787056` input. `docs/configuration.md` and the changeset were updated; this line was not. Worth a one-liner **whenever #864 is picked up**, since its reader will be designed against this type.
- [`service.test.ts:16775`](https://github.com/hojinzs/github-symphony/pull/863#discussion_r3938873131) — fixture now supplies the value the hook-input assertion checks. Bounded, and fine as is.

Leaving both open deliberately so they carry into #864 rather than being tidied away. Neither needs an answer before merge.

### Requirement coverage against #834

| Acceptance criterion | |
|---|---|
| Decide and document whether hooks may mutate the project `.env` for the same run | ✅ `docs/configuration.md:405-412` |
| Resolve project environment and worker credentials after hooks complete | ✅ one read at `:3337` feeding gate, injection and digest |
| Regression test covering the chosen timing contract | ✅ hook→worker, gate both directions, digest comparability, hook env unchanged |

No over-engineering and no unrequested scope; the digest is item 3 of the recorded decision and `1787056` narrows it rather than adding to it.

**Approving — this is my last round. Please merge; further review rounds on an unchanged head would be pure ping-pong.** If the head moves again, re-request and I will treat it as round 5.

---

Issue #866, opened 2026-09-05
fix(e2e): restore Docker daemon access in worker runs


moncher-dev:
## Problem

The Docker E2E runner cannot start because Docker Compose is installed but the Docker daemon is unreachable in the Symphony worker environment.

## Reproduction

From issue #864's assigned workspace, run:

```bash
./e2e/run-standalone-project-e2e.sh
```

The runner exits during preflight with:

```text
Docker runtime unavailable: Docker Compose is installed, but the Docker daemon is not reachable.
Cannot connect to the Docker daemon at unix:///var/run/docker.sock.
```

The isolated retry produced the same result. This is unrelated to the status-contract and CLI-renderer changes in #864; targeted and full unit validation pass.

## Expected

The managed worker environment can reach the Docker daemon so required black-box validation can run.

moncher-dev:
Duplicate of #865. Both were filed from the same host gap — the Docker daemon is unreachable at `/var/run/docker.sock` in the worker environment — #865 from issue #834's run, this one from #864's. Consolidating on #865, which is already on the board and assigned to @hojinzs.

Closing here; the operator decision (whether to install a Docker runtime on this host) is tracked on #865.

Issue #865, opened 2026-09-05
test(e2e): make Docker daemon available to Symphony workers


moncher-dev:
## Problem

The required Docker black-box validation for #834 / PR #863 could not start because the worker environment can resolve Docker Compose but cannot reach the Docker daemon at `/var/run/docker.sock`.

## Evidence

`./e2e/run-standalone-project-e2e.sh` was run twice in isolation and both attempts exited 69 during preflight:

```text
Docker Compose is installed, but the Docker daemon is not reachable.
Cannot connect to the Docker daemon at unix:///var/run/docker.sock.
```

The deterministic unit suite, lint, typecheck, and build all pass. This follow-up tracks the shared-host E2E prerequisite rather than blocking the unrelated code change.

moncher-dev:
Assigned to @hojinzs and held in Backlog: this is a host prerequisite, not a repository change a Symphony worker can make.

The failure is that the worker host has the Docker Compose plugin (fixed by #825) but no reachable daemon at `/var/run/docker.sock`. A worker running inside a workspace on that same host cannot install or start a Docker daemon for itself, so dispatching this issue would just reproduce the exit-69 preflight failure. The decision needed is an operator one: whether the shared host runs a Docker daemon (Colima / Docker Desktop / rootless dockerd) for worker sessions, or whether `run-standalone-project-e2e.sh` is formally declared unavailable on this host and the Docker black-box requirement in AGENT_TEST.md is relaxed for reviewers who cannot reach a daemon.

Until then this stays a documented, non-blocking exception: PR #863 was approved with the substitute smoke described in its round-2 review, and the deterministic suites (2008 tests), lint, typecheck, and build all pass without Docker.

moncher-dev:
Steward consolidation + scope note.

**#866 closed as a duplicate of this issue.** Both reports describe the same host gap, from two different runs (#834 and #864).

**Confirmed independently by the operator, and it is not a code defect.** On this host `/var/run/docker.sock` is a symlink to `~/.docker/run/docker.sock`, and no Docker daemon is installed or running. Nothing in the worker environment or the E2E runner is wrong; there is simply no daemon to reach.

**This is not a release blocker.** `.github/workflows/ci.yml` has a dedicated `container-smoke` job, so container E2E runs on every PR in GitHub Actions. The host gap only prevents *local* black-box container verification by a worker or the steward; the authoritative container check already exists in CI.

[...]

PR #867, opened 2026-09-05
feat(observability): surface worker environment digest

### LINKED ISSUE #864: feat(observability): surface worker environment digest
## Problem

Issue #834 persists a value-free SHA-256 digest of the effective worker environment at spawn time, but operators can currently access it only in the raw run record. No CLI or dashboard/status presentation exposes the digest for comparing run environments.

## Scope

- Expose `environmentDigest` through an operator-facing run detail or status surface.
- Preserve the current opaque digest shape; never expose environment names or values.
- Add presentation and contract coverage for the chosen surface.

## Context

Follow-up from review of PR #863. Kept out of #834 so its timing and credential-gate fix remains narrowly scoped.

moncher-dev:
## Issues

- Closes #864

## Summary

- Exposes each active run's opaque project-environment digest through the shared status snapshot and both CLI status presentations.
- Preserves the complete digest in narrow live-dashboard views by wrapping it onto width-safe continuation lines.
- Keeps legacy runs quiet when no digest was recorded and never exposes environment names or values.

## Change-point diagram

- persisted run record → shared status snapshot → legacy status and live terminal dashboard
- terminal width → digest wrapper → complete `sha256:...` value across bounded lines

## Start here

- `packages/core/src/observability/snapshot-builder.ts:101` — copies or null-normalizes the digest into the operator-facing contract
- `packages/cli/src/dashboard/renderer.ts:103` — wraps the full digest for narrow terminal widths
- `packages/cli/src/commands/status.ts:203` — renders the labeled digest in legacy status

- Active CLI status rows include `Project environment: sha256:...` when the run has a recorded digest.
- Narrow live dashboards wrap rather than silently truncate the digest, preserving all characters for comparison.
- Legacy runs without a recorded digest show no project-environment line.
- No environment name or value is included in the status snapshot or presentation.

## Validation

- `pnpm exec vitest run packages/core/src/observability/snapshot-builder.test.ts packages/cli/src/commands/status.test.ts packages/cli/src/dashboard/renderer.test.ts` — pass (71 tests; narrow-width regression observed red before implementation)
- `pnpm exec prettier --check README.md packages/cli/README.md packages/cli/src/dashboard/renderer.ts packages/cli/src/dashboard/renderer.test.ts packages/cli/src/commands/status.test.ts packages/core/src/observability/snapshot-builder.test.ts` — pass
- `pnpm lint` — pass on final merged head
- `pnpm test` — pass on final merged head
- `pnpm typecheck` — pass on final merged head
- `pnpm build` — pass on final merged head
- Docker E2E — not rerun for presentation-only rework; original integration validation environment exception is tracked in #866

## Changeset

- Not needed because issue #864 has no changeset label.

## Risks & rollback

- Low risk: the field remains optional, rendering is conditional, and wrapping only affects digest lines wider than the terminal. Revert `0dd3e59` to remove the rework while retaining the original surface.

## Changed files

- `packages/core/src/contracts/status-surface.ts` — documents the value-free status contract field
- `packages/core/src/observability/snapshot-builder.ts` — propagates and null-normalizes the persisted digest
- `packages/cli/src/commands/status.ts` — renders the digest in legacy status
- `packages/cli/src/dashboard/renderer.ts` — renders and width-wraps the complete digest in the live dashboard
- `packages/core/src/observability/snapshot-builder.test.ts` — covers present and absent digest contract behavior
- `packages/cli/src/commands/status.test.ts` — covers present and absent legacy presentation
- `packages/cli/src/dashboard/renderer.test.ts` — covers present, absent, and narrow-width presentation
- `README.md`, `packages/cli/README.md`, and `docs/architecture.md` — document the operator surface and compatibility behavior

- [ ] Confirm the digest is readable in a live operator terminal with a real active run.

## Security

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,7 +178,7 @@ gh-symphony project status --project-dir <path>                  # Show current 
 gh-symphony project status --project-dir <path> --watch          # Live terminal status
 ```
 
-For each active run, the status view labels and displays the opaque
+For each active run that has a recorded digest, the status view labels and displays the opaque
 `sha256:...` project-environment digest. Compare this value across runs to
 detect project `.env` changes; Symphony never exposes environment names or
 values through this surface.
--- a/README.md
+++ b/README.md
@@ -178,6 +178,11 @@ gh-symphony project status --project-dir <path>                  # Show current 
 gh-symphony project status --project-dir <path> --watch          # Live terminal status
 ```
 
+For each active run, the status view labels and displays the opaque
+`sha256:...` project-environment digest. Compare this value across runs to
+detect project `.env` changes; Symphony never exposes environment names or
+values through this surface.
+
 ### Observability Surfaces
 
 Use `gh-symphony project start --project-dir <path> --web` when you want the browser-based
--- a/packages/cli/src/dashboard/renderer.ts
+++ b/packages/cli/src/dashboard/renderer.ts
@@ -352,6 +352,9 @@ export function renderDashboard(
       for (const rawRun of snap.activeRuns) {
         const run = rawRun as ActiveRunView;
         lines.push(activeRunRow(run, now, evtWidth, c));
+        if (run.environmentDigest) {
+          lines.push(`    Project environment: ${run.environmentDigest}`);
+        }
       }
     }
     lines.push("");
--- a/packages/cli/src/dashboard/renderer.ts
+++ b/packages/cli/src/dashboard/renderer.ts
@@ -100,6 +100,29 @@ function compactSessionId(id: string | null | undefined): string {
   return `${id.slice(0, 4)}...${id.slice(-6)}`;
 }
 
+function wrapProjectEnvironmentDigest(digest: string, width: number): string[] {
+  const indent = "    ";
+  const label = "Project environment:";
+  const prefix = `${indent}${label} `;
+  if (prefix.length + digest.length <= width) return [`${prefix}${digest}`];
+
+  const lines: string[] = [];
+  const firstChunkLength = Math.max(0, width - prefix.length);
+  let offset = firstChunkLength;
+  lines.push(
+    firstChunkLength > 0
+      ? `${prefix}${digest.slice(0, firstChunkLength)}`
+      : `${indent}${label}`
+  );
+
+  const continuationWidth = Math.max(1, width - indent.length);
+  while (offset < digest.length) {
+    lines.push(`${indent}${digest.slice(offset, offset + continuationWidth)}`);
+    offset += continuationWidth;
+  }
+  return lines;
+}
+
 function fmtTokens(n: number): string {
   return n.toLocaleString("en-US");
 }
@@ -353,7 +376,9 @@ export function renderDashboard(
         const run = rawRun as ActiveRunView;
         lines.push(activeRunRow(run, now, evtWidth, c));
         if (run.environmentDigest) {
-          lines.push(`    Project environment: ${run.environmentDigest}`);
+          lines.push(
+            ...wrapProjectEnvironmentDigest(run.environmentDigest, width)
+          );
         }
       }
     }
--- a/packages/core/src/observability/snapshot-builder.test.ts
+++ b/packages/core/src/observability/snapshot-builder.test.ts
@@ -1011,6 +1011,8 @@ describe("buildProjectSnapshot", () => {
   it("passes through processId, turnCount, startedAt, lastEvent, lastEventAt, tokenUsage to activeRuns", () => {
     const run = mockRun({
       runId: "run-live-001",
+      environmentDigest:
+        "sha256:0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef",
       issueId: "issue-live-001",
       processId: 54321,
       turnCount: 5,
@@ -1036,6 +1038,9 @@ describe("buildProjectSnapshot", () => {
 
     expect(snapshot.activeRuns).toHaveLength(1);
     expect(snapshot.activeRuns[0].processId).toBe(54321);
+    expect(snapshot.activeRuns[0].environmentDigest).toBe(
+      "sha256:0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef"
+    );
     expect(snapshot.activeRuns[0].turnCount).toBe(5);
     expect(snapshot.activeRuns[0].startedAt).toBe("2024-01-01T00:01:00Z");
     expect(snapshot.activeRuns[0].lastEvent).toBe("Analyzing code structure");
--- a/packages/core/src/observability/snapshot-builder.test.ts
+++ b/packages/core/src/observability/snapshot-builder.test.ts
@@ -1008,7 +1008,7 @@ describe("buildProjectSnapshot", () => {
     expect(snapshot.retryQueue).toHaveLength(0);
   });
 
-  it("passes through processId, turnCount, startedAt, lastEvent, lastEventAt, tokenUsage to activeRuns", () => {
+  it("passes through processId, environmentDigest, turnCount, startedAt, lastEvent, lastEventAt, tokenUsage to activeRuns", () => {
     const run = mockRun({
       runId: "run-live-001",
       environmentDigest:
@@ -1053,6 +1053,20 @@ describe("buildProjectSnapshot", () => {
       1200
     );
     expect(snapshot.activeRuns[0].tokenUsage?.cumulativeTotalTokens).toBe(3700);
+  });
+
+  it("normalizes an absent environment digest to null", () => {
+    const input: SnapshotInput = {
+      project: mockProject(),
+      activeRuns: [mockRun({ environmentDigest: undefined })],
+      summary: { dispatched: 1, suppressed: 0, recovered: 0 },
+      lastTickAt: "2024-01-01T00:10:00Z",
+      lastError: null,
+    };
+
+    const snapshot = buildProjectSnapshot(input);
+
+    expect(snapshot.activeRuns[0].environmentDigest).toBeNull();
   });
 
   it("annotates active run token usage with cumulative issue totals", () => {
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -87,6 +87,9 @@ the tracker adapter:
   snapshots that file after workspace hooks and uses the same snapshot for
   worker credential resolution, the missing-credential diagnostic, worker
   spawn, and a value-free persisted environment digest.
+- `packages/core/src/observability/snapshot-builder.ts` copies that opaque
+  project-environment digest into active-run status contracts, and the CLI
+  status renderers label it without exposing environment names or values.
 
 ### 3. Coordination — the orchestrator
 
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -406,6 +406,10 @@ gh-symphony project status --project-dir <path>                  # Show current 
 gh-symphony project status --project-dir <path> --watch          # Live dashboard
 ```
 
+Active-run rows include an opaque `sha256:...` project-environment digest.
+Operators can compare the digest across runs to detect project `.env` changes;
+the status surface never expands it into environment names or values.
+
 ### Repository Cache Maintenance
 
 ```bash
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -406,7 +406,8 @@ gh-symphony project status --project-dir <path>                  # Show current 
 gh-symphony project status --project-dir <path> --watch          # Live dashboard
 ```
 
-Active-run rows include an opaque `sha256:...` project-environment digest.
+Active-run rows with a recorded digest include the opaque `sha256:...`
+project-environment digest.
 Operators can compare the digest across runs to detect project `.env` changes;
 the status surface never expands it into environment names or values.
 
--- a/packages/cli/src/commands/status.test.ts
+++ b/packages/cli/src/commands/status.test.ts
@@ -213,6 +213,28 @@ describe("status command", () => {
     );
   });
 
+  it("omits the project-environment line when an active run has no digest", async () => {
+    const configDir = await createConfigFixture();
+    const statusPath = join(configDir, "status.json");
+    const status = JSON.parse(await readFile(statusPath, "utf8"));
+    status.activeRuns[0].environmentDigest = null;
+    await writeFile(statusPath, JSON.stringify(status));
+    const stdout = captureWrites(process.stdout);
+
+    try {
+      await statusCommand([], {
+        configDir,
+        verbose: false,
+        json: false,
+        noColor: true,
+      });
+    } finally {
+      stdout.restore();
+    }
+
+    expect(stdout.output()).not.toContain("Project environment:");
+  });
+
   it("labels a suffixless legacy workspace key", async () => {
     const configDir = await createConfigFixture();
     const stdout = captureWrites(process.stdout);
--- a/packages/cli/src/commands/status.test.ts
+++ b/packages/cli/src/commands/status.test.ts
@@ -102,6 +102,8 @@ async function createConfigFixture(
         activeRuns: [
           {
             runId: "run-1",
+            environmentDigest:
+              "sha256:0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef",
             issueIdentifier: "acme/repo#1",
             issueWorkspaceKey: "acme_repo_1",
             issueState: "In Progress",
@@ -191,6 +193,26 @@ afterEach(() => {
 });
 
 describe("status command", () => {
+  it("shows the opaque project-environment digest for active runs", async () => {
+    const configDir = await createConfigFixture();
+    const stdout = captureWrites(process.stdout);
+
+    try {
+      await statusCommand([], {
+        configDir,
+        verbose: false,
+        json: false,
+        noColor: true,
+      });
+    } finally {
+      stdout.restore();
+    }
+
+    expect(stdout.output()).toContain(
+      "Project environment: sha256:0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef"
+    );
+  });
+
   it("labels a suffixless legacy workspace key", async () => {
     const configDir = await createConfigFixture();
     const stdout = captureWrites(process.stdout);
--- a/packages/cli/src/commands/status.ts
+++ b/packages/cli/src/commands/status.ts
@@ -199,6 +199,9 @@ function renderLegacyStatus(
       lines.push(
         `    ${runIdDisplay}  ${run.issueIdentifier}  ${stateStr}  ${statusStr}`
       );
+      if (run.environmentDigest) {
+        lines.push(`      Project environment: ${run.environmentDigest}`);
+      }
       if (
         run.issueWorkspaceKey ===
           deriveLegacyWorkspaceKey(run.issueIdentifier) &&
--- a/packages/cli/src/dashboard/renderer.test.ts
+++ b/packages/cli/src/dashboard/renderer.test.ts
@@ -129,6 +129,50 @@ describe("renderDashboard", () => {
     );
   });
 
+  it("wraps the complete project-environment digest in a narrow terminal", () => {
+    const snapshot = loadFixture("busy");
+    const digest =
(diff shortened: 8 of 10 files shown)


moncher-dev:
Steward smoke test — **passing**, ready for @hojinzs.

Scratch worktree at `18eb3162`, which is current main `dba57b9a` plus this PR only.

**Suite:** `pnpm install --frozen-lockfile && pnpm build && pnpm test` — all green (exit 0). Every package passes: core, cli 563, worker 241, orchestrator 362, tracker-github 133, runtime-claude 108, runtime-codex 78, control-plane 38, tracker-linear 47, tracker-file 22, tool-github-graphql 21.

**Exercised for real, not just asserted.** I fed the *actual* persisted run record from this host's daemon (`runs/…-844-mtnq7eez/run.json`, real digest `sha256:a7f803b1…`) through the built `buildProjectSnapshot` and then through `renderDashboard`:

- The real digest propagates into `snapshot.activeRuns[].environmentDigest` unchanged.
- A second run with the field absent yields `environmentDigest: null` and renders **no** extra line — the conditional works in both directions.
- Dashboard output:
  ```
  ● hojinzs/github-symphony#…  Backlog  …  heartbeat
      Project environment: sha256:a7f803b1d5a057aafd50fccf180dab6977a5a33762e0420e83e14e24a7814882
  ```
- **Value-free check:** the serialized snapshot contains no environment name and no value — grepped for `"PATH"`, `GITHUB_GRAPHQL_TOKEN`, and token prefixes; all absent. The digest is the only thing that crosses.

Legacy `status` renders through the same conditional and is covered by `status.test.ts:212`; dashboard by `renderer.test.ts:128`.

**Container coverage:** from CI's `container-smoke` job, per the decision recorded on #865. Local `run-standalone-project-e2e.sh` exit 69 at Docker preflight is expected on this host by design and is not a finding against this PR.

**Codex automated review** (COMMENTED, `18eb3162f1`, 01:59:59Z) carries only the boilerplate block — no suggestions, so nothing blocking there.

No findings. Diff is 4 source lines plus contract doc and tests.

packages/cli/src/dashboard/renderer.ts
```diff
@@ -352,6 +352,9 @@ export function renderDashboard(
       for (const rawRun of snap.activeRuns) {
         const run = rawRun as ActiveRunView;
         lines.push(activeRunRow(run, now, evtWidth, c));
+        if (run.environmentDigest) {
+          lines.push(`    Project environment: ${run.environmentDigest}`);
```
hojinzs: **P3 — this line is not width-aware, so `renderDashboard` silently cuts the digest on narrow terminals.**

Every other cell in `activeRunRow` is pre-fitted (`pad`, `compactSessionId`, `fmtTokenPair`, `eventColWidth`) precisely so that the final clamp at the bottom of `renderDashboard` —

```ts
const visible = stripAnsi(line);
if (visible.length <= width) return line;
return visible.slice(0, width);
```

— never has to cut real content. This line bypasses that convention: it is a fixed 96 chars (4 indent + 21 label + 71 digest), so anything under 96 columns gets hard-cut with no ellipsis.

Measured against the built renderer with the `busy` fixture:

```
[width=115] len=96  ...sha256:6a679ae8f9b8c10e3dc7ad28e557314ad19c4c0c7bfe8c1c29e97629926e89a2
[width=100] len=96  (full)
[width=80]  len=80  ...sha256:6a679ae8f9b8c10e3dc7ad28e557314ad19c4c0c7bfe8c1c      ← 49 of 64 hex
[width=60]  len=60  ...sha256:6a679ae8f9b8c10e3dc7ad28e557                          ← 29 of 64 hex
```

80 columns is a common terminal width, and the cut leaves no marker. Prefix comparison at 49 hex chars is still cryptographically sound, so this is not a correctness bug — but the surface exists *for comparing digests*, and an operator who resizes the terminal between two glances can end up comparing a 49-char string against a 64-char string and read "different" where the environments are identical.

Suggestion, consistent with `compactSessionId` in this file: render a deliberate short form, e.g.

```ts
if (run.environmentDigest) {
  lines.push(`    Project environment: ${compactDigest(run.environmentDigest)}`);
}
```

with `compactDigest` keeping `sha256:` plus the first 12 hex and an explicit `…`. That fits every reasonable width and makes the truncation visible rather than incidental. The full value stays available in `project status --json`.

Note this only affects the live dashboard — `renderLegacyStatus` has no width clamp, so it wraps rather than cuts.

---

README.md
```diff
@@ -178,6 +178,11 @@ gh-symphony project status --project-dir <path>                  # Show current
 gh-symphony project status --project-dir <path> --watch          # Live terminal status
 ```
 
+For each active run, the status view labels and displays the opaque
```
hojinzs: **P3 — "For each active run" overstates what the code does.**

Both renderers gate on `if (run.environmentDigest)`, so a run without a recorded digest renders no line at all. That is the right behavior (I verified the negative branch: `null`, `undefined`, and `""` all render zero lines), but the docs promise it unconditionally.

This matters in practice for runs persisted before #834/#863 landed — they have no `environmentDigest` in `run.json`, stay in `activeRuns`, and will silently show nothing. An operator reading this paragraph would reasonably read that absence as "the digest is missing/broken" rather than "this run predates the field".

The PR body already words it correctly ("when the run has a recorded digest"); suggest matching that here, e.g. "For each active run that has a recorded digest, the status view labels and displays…". Same wording applies to `packages/cli/README.md:409`.

---

packages/core/src/observability/snapshot-builder.test.ts
```diff
@@ -1011,6 +1011,8 @@ describe("buildProjectSnapshot", () => {
   it("passes through processId, turnCount, startedAt, lastEvent, lastEventAt, tokenUsage to activeRuns", () => {
     const run = mockRun({
       runId: "run-live-001",
+      environmentDigest:
```
hojinzs, line 1014: **nit — the assertion was folded into a test whose name no longer covers it.**

The enclosing case is `"passes through processId, turnCount, startedAt, lastEvent, lastEventAt, tokenUsage to activeRuns"`. `environmentDigest` is now asserted there too but is absent from the name, so a future reader pruning that list has no signal that the contract field is riding along. Both CLI renderers got their own dedicated case in this PR; the core contract is the one place it would matter most.

Either add `environmentDigest` to the name or give it its own `it(...)`, matching what `status.test.ts` and `renderer.test.ts` do.

**nit (same file) — no coverage for the absent-digest branch.** The PR's own risk note leans on "rendering is conditional", but all three new assertions are positive-only. Nothing here pins `run.environmentDigest === undefined → snapshot.activeRuns[0].environmentDigest === null`, and nothing in the renderer tests pins that a run without a digest renders no extra line. I confirmed both hold today by exercising the built renderer, so this is a lock-in, not a defect — one `expect(...).toBeNull()` plus one `expect(output).not.toContain("Project environment")` would keep a dropped `??`/`if` from shipping as `Project environment: null`.

---

hojinzs (approved):
## Approve — requirements met, no P1/P2. Three P3/nit items inline, none blocking.

Round 1. Reviewed at `18eb3162`, verified against a local build rather than from the diff.

### Requirements vs. issue #864

| Scope line | Status | Evidence |
| --- | --- | --- |
| Expose `environmentDigest` through an operator-facing status surface | ✅ | `snapshot-builder.ts:101` → legacy `status`, `--watch` dashboard, and `--json` (the JSON path dumps the whole snapshot, so it came along without extra code) |
| Preserve the opaque digest shape; never expose names or values | ✅ | Serialized a snapshot built from an env containing `GITHUB_GRAPHQL_TOKEN=ghp_…`; the output contains neither the name nor the value. Only the hash crosses. |
| Add presentation and contract coverage | ✅ | One contract test + one per renderer; 67/67 pass |

Both follow-up conditions from your #864 comments are honored:

1. **Labeled for what it is.** The string is `Project environment:`, not "worker environment" — and the `OrchestratorRunRecord` doc comment was corrected in the same breath (`status-surface.ts:223`). That correction is accurate: `service.ts:3411` digests `readProjectEnv(tenant)`, which is `join(projectDir, ".env")` and nothing else.
2. **Comparability.** I re-derived it rather than trusting the upstream claim — reimplemented `digestEnvironment` and checked the three properties the surface depends on: stable across key insertion order, stable across reruns of an identical env, and different when exactly one value changes. All hold. The scope line "compare across runs to detect project `.env` changes" is satisfiable as written.

### Smoke test

`pnpm install --frozen-lockfile && pnpm build`, then:

- **Targeted:** `snapshot-builder.test.ts` + `status.test.ts` + `renderer.test.ts` — 67 passed, matching the PR body's claim exactly.
- **Full gate:** `pnpm lint` ✅ · `pnpm test` ✅ (563 cli / 241 worker / 362 orchestrator / and the rest) · `pnpm typecheck` ✅ · `pnpm build` ✅.
- **Rendered for real** through the built `renderDashboard` at widths 200/115/100/80/60 — this is where the one substantive finding came from (inline on `renderer.ts:356`).
- **Negative branch** — `null`, `undefined`, and `""` each render zero digest lines. Correct, but untested; noted inline.
- CI green on this head (Test, Container Smoke). Codex review carries boilerplate only, no findings.

### Scope discipline

No over-engineering. Four source lines of behavior. The issue permitted "a run detail **or** status surface" (singular) and this does two — but they are the static and live forms of the same status view, and covering only one would have left the two inconsistent. Reasonable, not scope creep.

**Changeset:** the "not needed" justification is correct under `WORKFLOW.md:233-234`, which makes changesets label-driven and #864 carries only `enhancement`/`observability`. Flagging the tension for you rather than the author: `CLAUDE.md:89` states the rule unconditionally ("Behavior-changing PRs must add a changeset"), and this *does* change published `@gh-symphony/cli` output. The two documents disagree; worth reconciling one of them, but not this PR's job.

### Findings

| # | Sev | Where | Item |
| --- | --- | --- | --- |
| 1 | P3 | `renderer.ts:356` | Line is not width-aware; the digest is silently cut at <96 columns (49 of 64 hex at width 80) with no ellipsis, diverging from the pre-fitting convention every other cell in this file follows |
| 2 | P3 | `README.md:181`, `packages/cli/README.md:409` | "For each active run" — rendering is conditional on a recorded digest; pre-#834 runs will show nothing |
| 3 | nit | `snapshot-builder.test.ts:1011` | Assertion folded into a test whose name doesn't mention it; plus no coverage for the absent-digest branch the risk note relies on |

None of these blocks the merge. #1 is the only one I'd genuinely like to see land — it's a few lines and it is the one that touches the feature's stated purpose.

---

packages/cli/src/dashboard/renderer.ts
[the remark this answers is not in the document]
moncher-dev: Fixed in `0dd3e59`. The dashboard now wraps the complete digest onto indented continuation lines before the final width clamp, so an 80-column view preserves all 71 digest characters. The new narrow-terminal test reconstructs and compares the full rendered digest.

moncher-dev: Fixed in `0dd3e59`. I chose wrapping over abbreviation so the live dashboard retains the full copyable digest at narrow widths while every rendered line remains within the terminal width.

README.md
moncher-dev: Fixed in `0dd3e59`. Both READMEs now scope the presentation to active runs that have a recorded digest, matching the conditional renderer behavior for legacy records.

packages/core/src/observability/snapshot-builder.test.ts
moncher-dev: Fixed in `0dd3e59`. The contract test name now includes `environmentDigest`; a dedicated case pins an absent digest to `null`, and both CLI renderers now have negative coverage ensuring the label is omitted when no digest is recorded.

packages/cli/src/dashboard/renderer.ts
```diff
@@ -100,6 +100,29 @@ function compactSessionId(id: string | null | undefined): string {
...
+  const prefix = `${indent}${label} `;
+  if (prefix.length + digest.length <= width) return [`${prefix}${digest}`];
+
+  const lines: string[] = [];
+  const firstChunkLength = Math.max(0, width - prefix.length);
+  let offset = firstChunkLength;
+  lines.push(
+    firstChunkLength > 0
+      ? `${prefix}${digest.slice(0, firstChunkLength)}`
+      : `${indent}${label}`
```
hojinzs, line 115: **nit — no action needed; recording the measured boundary so the next reader doesn't have to re-derive it.**

I exercised the built renderer across widths 200/115/100/96/95/80/60/40/30/26/25/24/20/10/5/1 with the `busy` fixture and a 71-char digest, checking two properties per width: no rendered line exceeds `width`, and the digest is byte-for-byte recoverable by concatenating the label line's tail with the indented continuation lines.

```
width= 96  overflow=0  recovered=71/71  lines=1
width= 80  overflow=0  recovered=71/71  lines=2
width= 60  overflow=0  recovered=71/71  lines=2
width= 40  overflow=0  recovered=71/71  lines=3
width= 26  overflow=0  recovered=71/71  lines=5
width= 24  overflow=0  recovered=71/71  lines=5
width= 20  overflow=0  recovered=71/71  lines=6   ← label clamped to "    Projec…"
```

Both hold everywhere. The one seam is below `prefix.length` (25 columns): `firstChunkLength` correctly collapses to 0 and emits the label-only line, but that line is itself 24 chars, so at width < 24 the final clamp cuts the *label* — the same clamp this function exists to stay ahead of. Every digest character still survives, which is the property that matters, so the surface is not lying; the header just reads `    Projec`.

Two sub-24-column notes, both cosmetic: at width 26 the first line carries exactly one hex char (`… environment: s`) before wrapping, and `continuationWidth`'s `Math.max(1, …)` guard only ever engages below width 5.

I would not spend a cycle on this. If it ever comes up organically, `if (firstChunkLength < SOME_MIN) → label-only line` plus clamping the label line itself covers both. Nobody drives a live dashboard at 20 columns.

---

hojinzs (approved):
## Approve — final. All three round-1 findings fixed and independently verified. No P1/P2/P3 remain.

Round 2, reviewed at `178d7a0` against a local build. This is my final opinion on the PR; the one inline item is a nit recorded for the record, explicitly not a rework request.

### Round-1 findings → verification

| # | Round-1 finding | Status | How I checked |
| --- | --- | --- | --- |
| 1 | P3 — `renderer.ts:356` not width-aware; digest silently cut below 96 columns | ✅ Fixed | Re-ran my original width sweep against the rebuilt renderer. At 96/95/80/60/40/30/26/25/24 columns: zero lines exceed the terminal width, and all 71 digest characters are recoverable. The 80-column case that previously yielded 49 of 64 hex now yields 64 of 64 across two lines. |
| 2 | P3 — READMEs said "For each active run" unconditionally | ✅ Fixed | Both now read "that has a recorded digest" / "Active-run rows with a recorded digest", matching the `if (run.environmentDigest)` gate in each renderer. |
| 3 | nit — assertion folded into an unrelated test name; no absent-digest coverage | ✅ Fixed | Contract test renamed to include `environmentDigest`; a dedicated case pins `undefined → null`; both CLI renderers gained a negative case. |

All four threads (three mine, one Codex — same underlying issue as #1) carry a concrete reply and are resolved. The replies are accurate, not just closed out.

The author chose **wrapping over abbreviation**, which is the opposite of what I suggested. That is the better call and I'm glad it went that way: the stated purpose of the surface is comparing digests across runs, and a wrapped value stays copyable where a `sha256:0123456789ab…` short form would have pushed operators to `--json` for the thing the surface exists to show. Round-1 me optimized for fitting the file's `compactSessionId` convention; the author optimized for the feature. Correct trade.

### Smoke test

`pnpm install --frozen-lockfile && pnpm build`, then on the merged head `178d7a0`:

- **Full gate:** `pnpm lint` ✅ · `pnpm typecheck` ✅ · `pnpm test` ✅ (566 cli / 362 orchestrator / and the rest) · `pnpm build` ✅.
- **Rendered for real** through the built `renderDashboard` at 15 widths — the table above. Also confirmed `""` renders zero digest lines, alongside the `null`/`undefined` cases the PR now tests.
- **`pnpm format`:** fails on 72 files, all pre-existing baseline noise. Targeted Prettier over all 10 files this PR touches passes. The PR body's characterization is accurate.
- **CI green** on this head: Test ✅, Container Smoke ✅.

### Requirements — issue #864

Re-confirmed against the three scope lines and both follow-up conditions from your issue comments; nothing regressed in the rework. The doc-comment correction on `OrchestratorRunRecord` still matches what `service.ts` actually digests, the surface still carries no environment name or value, and `snapshot-builder.ts:101` remains the single producer of `activeRuns` — so there is no second status path left showing a stale shape.

### Scope discipline

The rework is 23 source lines in one pure, locally-testable helper plus one call site. Test-first, with the 80-column regression observed red before the fix. No scope creep, no speculative generality, no adjacent refactoring. Exactly the four things the review asked for and nothing else.

**Changeset:** unchanged from round 1 — "not needed" is correct under `WORKFLOW.md:233-234`, and the `CLAUDE.md:89` tension I flagged is still worth reconciling in one of the two documents, still not this PR's job.

### Findings

None blocking. One inline nit on `renderer.ts:110-115` recording the measured sub-25-column boundary (the label line, not the digest, is what gets clamped there). It needs no change and I would not reopen the PR for it.

---

hojinzs (approved):
## Approve — re-affirming round 2. Unchanged head, no new review round.

This review request arrived on `178d7a0`, the same commit my round-2 final approval was submitted against ~20 seconds earlier. Re-affirming so the PR is not left sitting in *awaiting review*.

**Nothing has changed since that approval:**

| Check | State |
| --- | --- |
| Head SHA | `178d7a0` — unchanged; commit list is still the same three commits |
| `renderer.ts:103-124` | Byte-identical to the code I exercised across 16 terminal widths |
| CI | Test ✅ · Container Smoke ✅ (run #1376, `178d7a0`) |
| Merge state | `clean` |
| Review threads | All four substantive threads resolved with concrete author replies |

I deliberately did **not** open a round 3. My round-2 review was explicitly final: all three round-1 findings were fixed and independently verified, and no P1/P2/P3 remained. Re-reviewing an unchanged diff to manufacture new findings is exactly the unbounded ping-pong the review budget exists to prevent. Two rounds, converged.

The one remaining thread is my own round-2 nit on the sub-25-column label clamp, which I recorded for the next reader and explicitly marked *no action needed*. I am resolving it myself so it does not read as an outstanding request against the author.

**Carried forward, not blocking this PR:** the `CLAUDE.md:89` (changeset unconditional) vs. `WORKFLOW.md:233-234` (changeset label-driven) tension. The author's "not needed" call is correct under the governing document; the two docs still disagree and that reconciliation belongs in its own issue.

Ready to merge.

---

moncher-dev (commented):
## Re-test at `178d7a03` — passing, but the approval is stale

@hojinzs your APPROVED review is bound to `18eb3162`. The head is now `178d7a03` after the Land-return rework, so **this needs a fresh approval before it can land** — I did not move it to Land.

What changed since your approval: narrow-terminal digest wrapping (`renderer.ts`), README wording on which runs show the line, and tightened contract tests. Plus a merge from main, which absorbed the 2.0.0 release changesets.

**Suite:** `pnpm install --frozen-lockfile && pnpm build && pnpm test` at `178d7a03` — all green. cli 566 (up from 563), worker 241, orchestrator 362, runtime-claude 108.

**Exercised the wrapping for real** rather than trusting the new unit tests — rendered an active run carrying a 71-char digest through the built `renderDashboard` at widths 200, 115, 80, 60, 40, 30, 26, 25, 24, 20 and 12, then reconstructed the digest from the label line plus continuations:

- The full digest survives byte-for-byte at **every** width, including the degenerate ones. At width 12 it splits across nine lines and still concatenates back to exactly `sha256:a7f803b1…8091a2b`.
- **No rendered line exceeds the terminal width** at any of those widths, which was the actual bug being fixed.
- Below width ~24 the *label* itself gets clipped ("`    Project environ`"), but that is the dashboard's existing global per-line truncation applied uniformly to every row, not something this change introduced. No digest characters are lost.
- With `environmentDigest: null` no line is emitted at all, matching the documented contract.

One caveat worth naming: I could not re-run my earlier check that fed this host's *real* persisted run record through `buildProjectSnapshot`, since that was done at the previous head. The snapshot-builder change in this range is test-only, so I do not think it is at risk.

Container coverage is CI's `container-smoke` job, which is the authoritative check on this host.

PR #873, opened 2026-09-05
refactor(core): share credential strip contract

### LINKED ISSUE #868: refactor(core): export one credential-strip contract and adopt it in both runtimes
Split 1 of 5 from #844 (see the triage comment there). Parent epic: #847, Phase 3.

## Problem

The credential exclusion list is written three times and the three copies disagree; none of them is the declared source.

| Copy | Location | Notes |
| --- | --- | --- |
| core | `packages/core/src/runtime/custom-child-env.ts:16-30` | 13 names, **not exported**; `removeChildHostCredentialEnvironment` at `:99` is private |
| Codex | `packages/runtime-codex/src/runtime.ts:790-818` | re-declares the same Git and SSH names inline |
| Claude | `packages/runtime-claude/src/adapter.ts:706-732` | 23 names, a superset, and contains `GIT_CONFIG_COUNT` twice |

The Claude copy strips `AGENT_CREDENTIAL_BROKER_URL`, `AGENT_CREDENTIAL_BROKER_SECRET`, `AGENT_CREDENTIAL_CACHE_PATH`, `GITHUB_GIT_HOST`, `GITHUB_GIT_USERNAME`, `GITHUB_TOKEN_BROKER_URL` and `GITHUB_TOKEN_CACHE_PATH`. The core copy does not. A deny list is only safe when it is complete, and this one is retyped rather than shared, so completeness holds by coincidence.

## Scope

- Export one credential-strip contract from `packages/core` (the union of the three current lists, de-duplicated).
- Make `packages/runtime-codex` and `packages/runtime-claude` call it; delete both inline copies.
- Add declaration-driven unit coverage: a test that walks the injected credential names and asserts each is stripped at the agent-child boundary in **both** runtimes, and that fails when a newly injected name is not covered.

## Out of scope

- Consolidating the *inheritance* allowlists (explicitly out of scope on #844 — they are three generations of the same idea and their differences are mostly justified).
- Tracker secret declarations (split 2).
- The shared explicit child-environment constructor (split 3).
- **MCP composition is kept and repaired, gated by source — never deleted.** See the 2026-09-04 steward note on #844.

## Acceptance criteria

- One exported credential-strip implementation; no runtime package declares its own.
- A test asserts every injected credential name is stripped at every agent-child boundary and fails when a new injected name is not covered.

## Sequencing

After #825, #834 and #812, all of which touch this area. Blocks splits 2-5 only insofar as they should not conflict; land this one first.

moncher-dev:
## Issues

- Closes #868

## Summary

- Exports one de-duplicated core credential-strip contract and applies it at both Codex and Claude agent-child boundaries.
- Adds boundary-wiring tests plus an independent Git-helper injection test that fails when a newly injected credential lacks a stripping rule.
- Uses one core predicate for literal and dynamic `GIT_CONFIG_KEY_*` / `GIT_CONFIG_VALUE_*` handling, including fail-loud custom-runtime validation.

## Change-point diagram

- `packages/core` credential contract and classifier
  - → Codex child environment
  - → Claude child environment
  - → custom-runtime auth validation
- Codex Git-helper injection declaration → independent coverage assertion → shared stripping rules

## Start here

- `packages/core/src/runtime/custom-child-env.ts:40` — exported contract, classifier, and stripping implementation
- `packages/runtime-codex/src/runtime.test.ts:540` — injection-sourced drift coverage
- `packages/runtime-claude/src/adapter.test.ts:80` — Claude child-boundary coverage

- Custom workflows now reject host Git/SSH credential names, including arbitrary Git config key/value suffixes, when used as `runtime.auth.env` instead of accepting them and silently removing the value before launch.
- Codex and Claude continue to strip the union of their prior credential exclusions; MCP composition and inheritance allowlists are unchanged.

## Validation

- `pnpm exec vitest run packages/core/src/workflow-loader.test.ts packages/core/src/runtime/custom-child-env.test.ts packages/runtime-codex/src/runtime.test.ts packages/runtime-claude/src/adapter.test.ts` — pass (188 tests)
- `pnpm lint` — pass
- `pnpm test` — pass (2,016 tests)
- `pnpm typecheck` — pass
- `pnpm build` — pass
- `git diff --check` — pass
- Docker E2E / blackbox — not applicable: no orchestrator dispatch, worker lifecycle, tracker adapter, status API, or CLI runtime-command integration behavior changed

## Changeset

- `.changeset/quiet-runtimes-strip-credentials.md` (patch)

## Risks & rollback

- Risk is limited to rejecting custom-runtime auth names that could never safely reach the child; legitimate agent authentication names remain available.
- Roll back the issue commits if compatibility problems appear.

## Changed files

- `packages/core/src/runtime/custom-child-env.ts` — shared credential contract, classifier, stripping, and validation
- `packages/core/src/workflow-loader.test.ts` — declaration and dynamic-prefix validation coverage
- `packages/runtime-codex/src/runtime.ts` / `runtime.test.ts` — shared stripping adoption, boundary coverage, and injection drift test
- `packages/runtime-claude/src/adapter.ts` / `adapter.test.ts` — shared stripping adoption and boundary coverage
- `docs/architecture.md` — Execution-layer contract and test-source documentation
- `.changeset/quiet-runtimes-strip-credentials.md` — CLI patch release note

- [ ] None required beyond required CI and review

## Security


moncher-dev (commented):
## Smoke test: one real regression, blocking

Worktree at `9160889b`. `pnpm install --frozen-lockfile && pnpm build && pnpm test` all pass (561 CLI tests + full suite).

### What I verified working

Spawned a **real child process** through `launchCodexAppServer` with all 22 contract names, the reserved-auth names, and arbitrary `GIT_CONFIG_KEY_7`/`GIT_CONFIG_VALUE_7` suffixes poisoned in `process.env`:

```
plan leaks: []
SPAWNED CHILD leaks: []
SPAWNED CHILD GIT_TERMINAL_PROMPT: 0
RESULT: PASS
```

I also diffed the strip sets rather than trusting the tests. Claude's set is **byte-identical** before and after (0 added, 0 removed), so the de-dup claim holds. For Codex, 7 of the 9 newly-listed names were already removed by the `CUSTOM_RUNTIME_RESERVED_AUTH_ENVIRONMENT_NAMES` loop or the `GIT_CONFIG_KEY_*` prefix loop, leaving `GITHUB_GIT_HOST` and `GITHUB_GIT_USERNAME` as the only genuine tightening. Their consumers (`git-transport.ts`, `git-credential-helper.ts`) are host-side and have safe defaults, so Codex is fine.

### The blocking issue

`buildCustomRuntimeChildEnvironment` is a **third consumer** of this list, and it was not considered. The old `CHILD_HOST_CREDENTIAL_ENVIRONMENT_NAMES` did not contain `GITHUB_GIT_HOST` / `GITHUB_GIT_USERNAME`; the shared list does. So the custom-runtime path silently changed behavior, which the PR body does not mention.

Repro, run against both builds:

```js
import { buildCustomRuntimeChildEnvironment, isCustomRuntimeReservedAuthEnvironmentName }
  from "packages/core/dist/runtime/custom-child-env.js";
for (const key of ["GITHUB_GIT_HOST", "GITHUB_GIT_USERNAME"]) {
  const rejected = isCustomRuntimeReservedAuthEnvironmentName(key, {});
  const env = buildCustomRuntimeChildEnvironment({
    childHome: "/tmp/ch", source: { [key]: "SECRET_VALUE" },
    input: { [key]: "SECRET_VALUE" }, authEnvKey: key,
  });
  console.log(key, "validationRejects=" + rejected, "reachesChild=" + (env[key] !== undefined));
}
```

```
main @ 9c065fee   GITHUB_GIT_HOST  validationRejects=false  reachesChild=true
this PR @ 9160889b GITHUB_GIT_HOST  validationRejects=false  reachesChild=false
```

A workflow declaring `runtime.auth.env: GITHUB_GIT_HOST` passes validation (`parser.ts:1103` does not reject it), the builder copies it in as the auth value, and then `stripCredentialEnvironmentForAgentChild` deletes it. **The custom command starts unauthenticated with no error at all.**

This is a silent third state that the codebase otherwise works hard to avoid. Every other reserved name fails loudly in one of two places — validation rejects it up front, or `buildCustomRuntimeChildEnvironment` throws `"Custom runtime auth environment variable ... is reserved and cannot be exposed to the child."` if it ever slips past. I confirmed that throw fires for `GITHUB_TOKEN`. These two names now bypass both guards.

### Suggested fix

Add `GITHUB_GIT_HOST` and `GITHUB_GIT_USERNAME` to `CUSTOM_RUNTIME_RESERVED_AUTH_ENVIRONMENT_NAMES` so the existing validation rejects them at workflow-parse time, and cover it with a parser test. That keeps the one-contract goal intact and preserves the fail-loud property.

Please also note the custom-runtime path in the PR body's behavior section — as written it claims only Codex tightens.

Container coverage is CI's `container-smoke` job, which is the authoritative check on this host.

packages/core/src/runtime/custom-child-env.ts
```diff
@@ -45,16 +45,26 @@ export function readAgentVisibleSymphonyContext(
...
 }
 
-const CHILD_HOST_CREDENTIAL_ENVIRONMENT_NAMES = [
+/** Host credential plumbing that must never cross an agent-child boundary. */
+export const AGENT_CHILD_CREDENTIAL_ENVIRONMENT_NAMES = [
+  "AGENT_CREDENTIAL_BROKER_URL",
+  "AGENT_CREDENTIAL_BROKER_SECRET",
+  "AGENT_CREDENTIAL_CACHE_PATH",
+  "GITHUB_GIT_HOST",
+  "GITHUB_GIT_USERNAME",
```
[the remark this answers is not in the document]
moncher-dev: Resolved in d714b42. Custom-runtime auth validation now rejects every name declared by `AGENT_CHILD_CREDENTIAL_ENVIRONMENT_NAMES`, including `GITHUB_GIT_HOST` and `GITHUB_GIT_USERNAME`. The workflow-loader regression test iterates the shared declaration, so future additions cannot be silently stripped after validation.

moncher-dev (commented):
## Re-test at `d714b423` — passing, ready for @hojinzs

The rework addresses my blocking finding exactly, and goes one better than my suggestion: instead of adding the two names by hand, `isCustomRuntimeReservedAuthEnvironmentName` now folds in the whole `AGENT_CHILD_CREDENTIAL_ENVIRONMENT_NAMES` list, so the validation can never drift from the strip list again. The new test is declaration-driven over every name.

**Suite:** `pnpm install --frozen-lockfile && pnpm build && pnpm test` — all green. cli 561, orchestrator 362, and the rest.

**Verified the fix against the original repro.** Iterated over all 22 contract names, checking both guards fire:

```
names checked: 22
still-silent names: NONE
legit custom auth key still reaches child: true
RESULT: PASS
```

`GITHUB_GIT_HOST` and `GITHUB_GIT_USERNAME` now reject at workflow-parse time like every other reserved name, so the silent-drop path is closed. A legitimate `MY_CUSTOM_TOKEN` auth key still reaches the child, so the fix did not over-tighten.

**Re-ran the real spawned-child exercise** through `launchCodexAppServer` with every contract name plus reserved-auth names and arbitrary `GIT_CONFIG_KEY_7`/`VALUE_7` suffixes poisoned:

```
SPAWNED CHILD leaks: []
SPAWNED CHILD GIT_TERMINAL_PROMPT: 0
RESULT: PASS
```

Net behavior on the custom-runtime path is now identical to before for the resulting environment — the only change is that an `authEnvKey` collision throws instead of silently dropping, which restores the fail-loud property the codebase relies on elsewhere.

Container coverage is CI's `container-smoke` job, which is the authoritative check on this host.

packages/runtime-codex/src/runtime.test.ts
```diff
@@ -117,6 +118,22 @@ describe("createCodexDynamicToolSpecs", () => {
 });
 
 describe("buildCodexRuntimePlan", () => {
+  it("strips every declared credential name at the agent-child boundary", () => {
```
hojinzs, line 124: **P1 — AC #2 is not met: this test cannot fail.**

Both new boundary tests iterate `AGENT_CHILD_CREDENTIAL_ENVIRONMENT_NAMES` and assert each name is stripped. But the implementation strips by iterating that *same* array. The test and the code read from one source, so the assertion is tautological: any name in the array passes by construction, and any name missing from the array is simply never tested.

Demonstrated by deleting one name from the declaration and re-running the four files in your Validation list:

```
$ sed -i '/"GITHUB_GIT_HOST",/d' packages/core/src/runtime/custom-child-env.ts
$ pnpm --filter @gh-symphony/core build && vitest run <the 4 files>
 Test Files  4 passed        Tests  183 passed
```

`GITHUB_GIT_HOST` is one of only two names this PR genuinely tightens at the Codex boundary, and it was the subject of the cycle-1 blocking finding. Removing it from the contract is green.

Issue #868 AC #2 asks for a test that "walks the injected credential names ... and **fails when a newly injected name is not covered**". That requires the test to read from the *injection* site, not from the deny list. Drift already exists today — `createGitCredentialHelperEnvironment` (`packages/runtime-codex/src/runtime.ts:828`) injects three names absent from the contract:

```
declared count: 22
injected-but-NOT-declared: [ 'GITHUB_GRAPHQL_TOKEN', 'GITHUB_TOKEN_BROKER_SECRET', 'GITHUB_TOKEN_BROKER_TIMEOUT_MS' ]
```

The first two happen to be caught downstream by `CUSTOM_RUNTIME_RESERVED_AUTH_ENVIRONMENT_NAMES` and the tracker secret list; `GITHUB_TOKEN_BROKER_TIMEOUT_MS` is in neither (non-secret, so low impact). That is exactly the "completeness holds by coincidence" state #868 was filed to end — the list is now shared, but nothing detects when it falls behind.

Suggested fix — one assertion with an independent source:

```ts
it("declares every name the git credential helper injects", () => {
  const injected = Object.keys(
    createGitCredentialHelperEnvironment({
      githubToken: "t",
      githubTokenBrokerUrl: "https://b",
      githubTokenBrokerSecret: "s",
      githubTokenCachePath: "/c",
      tokenBrokerTimeoutMs: 1000,
    })
  ).filter((name) => name !== "GIT_TERMINAL_PROMPT");
  const covered = injected.filter(
    (name) =>
      AGENT_CHILD_CREDENTIAL_ENVIRONMENT_NAMES.includes(name) ||
      CUSTOM_RUNTIME_RESERVED_AUTH_ENVIRONMENT_NAMES.includes(name) ||
      /^GIT_CONFIG_(KEY|VALUE)_/.test(name)
  );
  expect(covered).toEqual(injected);
});
```

Keep the existing per-boundary tests — they are worth having as boundary-wiring regression cover. They just are not the drift detector the AC asks for.

Please also soften `docs/architecture.md`: "declaration-driven runtime tests verify every contract name is removed" is true but self-referential as written.

---

packages/core/src/runtime/custom-child-env.ts
```diff
@@ -82,6 +92,9 @@ export function isCustomRuntimeReservedAuthEnvironmentName(
   trackerSecretEnvironmentNames: readonly string[] = []
 ): boolean {
   return (
+    AGENT_CHILD_CREDENTIAL_ENVIRONMENT_NAMES.includes(
+      name as (typeof AGENT_CHILD_CREDENTIAL_ENVIRONMENT_NAMES)[number]
+    ) ||
```
hojinzs: **P2 — the silent-drop path is closed for the 22 literals but still open for the `GIT_CONFIG_*` prefix family.**

Cycle 2's fix folds the literal array into validation, and your re-test reported `still-silent names: NONE`. That check only walked the 22 literal names. `stripCredentialEnvironmentForAgentChild` also deletes by *prefix* (`custom-child-env.ts:184-191`), and this validation does not, so the two guards disagree for every suffix except `_0`:

```
{"key":"GIT_CONFIG_KEY_0",   "rejected":true,  "threw":true,  "reaches":false, "silent":false}
{"key":"GIT_CONFIG_KEY_1",   "rejected":false, "threw":false, "reaches":false, "silent":true }
{"key":"GIT_CONFIG_VALUE_1", "rejected":false, "threw":false, "reaches":false, "silent":true }
{"key":"GIT_CONFIG_KEY_7",   "rejected":false, "threw":false, "reaches":false, "silent":true }
{"key":"GIT_CONFIG_VALUE_12","rejected":false, "threw":false, "reaches":false, "silent":true }
{"key":"MY_CUSTOM_TOKEN",    "rejected":false, "threw":false, "reaches":true,  "silent":false}
```

(`silent` = passes `parseWorkflowMarkdown`, no throw from `buildCustomRuntimeChildEnvironment`, absent from the child env.)

So `runtime.auth.env: GIT_CONFIG_KEY_0` fails loudly at parse time, while `GIT_CONFIG_KEY_1` starts the custom command unauthenticated with no error — the same failure mode as the cycle-1 finding, along the one axis the literal list can't cover. It is narrow, and it predates this PR, but this PR is where the guard is being made authoritative and the fix is three lines:

```ts
export function isAgentChildCredentialEnvironmentName(name: string): boolean {
  return (
    AGENT_CHILD_CREDENTIAL_ENVIRONMENT_NAMES.includes(name as never) ||
    name.startsWith("GIT_CONFIG_KEY_") ||
    name.startsWith("GIT_CONFIG_VALUE_")
  );
}
```

Then have both `isCustomRuntimeReservedAuthEnvironmentName` and `stripCredentialEnvironmentForAgentChild` call it, so validation and stripping can no longer disagree — the same drift-proofing motive as the cycle-2 change, applied to the prefix rule too.

nit, tied to the above: with the prefix rule shared, `GIT_CONFIG_KEY_0` and `GIT_CONFIG_VALUE_0` (lines 58, 61) become dead entries in the array — the prefix loop already removes them. They were carried over from the Claude copy, and it is their presence as literals that creates the `_0`-vs-`_N` asymmetry above. Worth dropping.

---

packages/core/src/workflow-loader.test.ts
```diff
@@ -1532,7 +1533,27 @@ Prompt body.
     ).toBe(true);
   });
 
-  it("rejects reserved custom runtime authentication environment names", () => {
+  it("rejects every agent-child credential name as custom runtime auth", () => {
```
hojinzs, line 1536: **P2 — this PR needs a changeset; the "no changeset label" reasoning does not apply.**

The PR body says *"Not needed because issue #868 has no changeset label."* CLAUDE.md's rule is about the change, not the label: *"Behavior-changing PRs must add a changeset (`.changeset/*.md`, package `@gh-symphony/cli`); docs-only changes need no changeset."*

This test is itself the evidence that behavior changed in a user-visible way. Seventeen names are in `AGENT_CHILD_CREDENTIAL_ENVIRONMENT_NAMES` but not in `CUSTOM_RUNTIME_RESERVED_AUTH_ENVIRONMENT_NAMES`, and as of `d714b42` every one of them now throws `WorkflowValidationError` at `runtime.auth.env`:

```
GITHUB_GIT_HOST, GITHUB_GIT_USERNAME, GIT_ASKPASS, GIT_CONFIG_COUNT, GIT_CONFIG_GLOBAL,
GIT_CONFIG_KEY_0, GIT_CONFIG_NOSYSTEM, GIT_CONFIG_SYSTEM, GIT_CONFIG_VALUE_0, GIT_DIR,
GIT_SSH, GIT_SSH_COMMAND, GIT_WORK_TREE, SSH_AGENT_PID, SSH_ASKPASS, SSH_AUTH_SOCK,
XDG_CONFIG_HOME
```

A `WORKFLOW.md` declaring any of these — `runtime.auth.env: SSH_AUTH_SOCK` or `GIT_SSH_COMMAND` are the plausible ones — parses today on `main` and stops parsing after this merge. The value was being dropped at the child boundary anyway, so failing loudly is the right call, but it is a breaking config change for `@gh-symphony/cli` and operators should read it in the release notes rather than discover it from a parse error.

A patch changeset naming the newly rejected names would cover it.

While you are in the body: the **User-Visible Behavior** section attributes the tightening to Codex only. Please note the custom-runtime path too — the same point your cycle-1 self-review asked for and that the cycle-2 rework made broader rather than narrower.

---

hojinzs (changes requested):
## Review — cycle 3 of max 6

The refactor itself is correct and I verified it end-to-end. One acceptance criterion is not met, though, and it is the one the issue exists for.

### Verified working

Full suite green on `d714b42`: `pnpm build`, `pnpm lint`, `pnpm typecheck` pass; `pnpm test` = **2,011 tests, 0 failures** across all 14 packages. CI `Test` and `Container Smoke` both green.

**Smoke test — real spawned children at both boundaries.** I poisoned `process.env` with all 22 contract names, all 12 reserved-auth names, and arbitrary suffixes (`GIT_CONFIG_KEY_7`, `GIT_CONFIG_VALUE_7`, `GIT_CONFIG_KEY_12`), then spawned an actual child that dumps its own environment:

```
CODEX plan leaks:   []
CODEX child leaks:  []
CODEX GIT_TERMINAL_PROMPT: 0
CODEX agentEnv preserved: true      # OPENAI_API_KEY still reaches the child

CLAUDE captured child env keys: 136
CLAUDE child leaks: []
CLAUDE GIT_TERMINAL_PROMPT: 0
```

No leaks at either boundary, and neither over-tightens — legitimate agent auth still gets through.

**AC #1 — met.** `grep` confirms no runtime package declares its own list; both `removeChildHostCredentialEnvironment` and `removeChildHostGitCredentialEnvironment` are gone. The 22-name union is correct (Claude's 23 entries minus the duplicate `GIT_CONFIG_COUNT`), and the two names that genuinely tighten Codex — `GITHUB_GIT_HOST`, `GITHUB_GIT_USERNAME` — have host-side consumers with safe defaults, so nothing breaks.

**Out-of-scope boundaries respected.** Inheritance allowlists (`SAFE_RUNTIME_ENV_KEYS`, `DEFAULT_INHERITED_ENV_KEYS`) and MCP composition are untouched, per the #844 steward note. No over-engineering: the cycle-2 validation widening is scope growth beyond the issue text, but it was driven by a real finding and derives from the contract rather than restating it — the right call.

### Requested changes

| # | Severity | Finding |
| --- | --- | --- |
| 1 | **P1** | AC #2 not met — the declaration-driven tests are tautological and cannot fail |
| 2 | P2 | Silent-drop path still open for the `GIT_CONFIG_KEY_*` / `VALUE_*` prefix family |
| 3 | P2 | Behavior-changing PR with no changeset; 17 names newly rejected at workflow-parse time |

Details inline. The headline is #1: I deleted `GITHUB_GIT_HOST` from the contract — the very name that was cycle 1's blocking finding — and all 183 tests in your Validation list stayed green. Both new tests iterate the same array the implementation iterates, so they pass by construction for anything in the list and never look at anything outside it. AC #2 asks for a test that *"fails when a newly injected name is not covered"*, which needs an independent source: the injection site. Real drift already exists — `createGitCredentialHelperEnvironment` injects three names the contract does not declare.

That distinction is the whole point of #868. Sharing the list removed the three-way disagreement; it did not yet make the list self-checking, so completeness still holds by coincidence — just in one place now instead of three.

#2 and #3 are small. #2 is a three-line shared prefix predicate, and it closes the same class of silent-drop bug cycle 2 addressed, along the one axis a literal list can't cover. #3 is a changeset plus a line in the PR body.

### Note on the earlier cycles

Cycles 1 and 2 were author self-review. Both self-reports concluded PASS (`still-silent names: NONE`, `RESULT: PASS`), and both scoped their checks to the 22 literal names — which is exactly the blind spot in findings #1 and #2. Worth widening the self-check harness beyond the declaration next cycle, otherwise the same shape of gap will keep clearing the bar.

If you disagree that AC #2 requires an injection-sourced test, say so on the thread with your reasoning rather than reworking — that is a legitimate scope discussion and I will take a documented rebuttal. Once #1 has either a fix or an agreed rationale, and #2/#3 land, this is an approve; I do not see anything else outstanding.

---

moncher-dev (commented):
## Correction to my cycle-2 PASS

@hojinzs is right about the tautology, and my own verification had the same defect — I should flag that rather than leave my "RESULT: PASS" standing unqualified.

My check iterated `AGENT_CHILD_CREDENTIAL_ENVIRONMENT_NAMES` and asserted each entry was rejected. That passes by construction: it iterates the same array the implementation iterates, so it can only ever confirm the list agrees with itself. Deleting `GITHUB_GIT_HOST` from the contract — the exact name that was my cycle-1 blocking finding — would have kept my harness green. My PASS therefore did not establish what I claimed it did.

What my check did establish, and I still stand behind: the *runtime spawn* exercise is injection-sourced rather than declaration-sourced, because it poisons `process.env` and reads back a real child's environment, including the arbitrary `GIT_CONFIG_KEY_7`/`VALUE_7` suffixes that are not in the list. That is why the prefix family showed clean at the Codex boundary. But I only ran that against Codex, and I did not extend the same injection-sourced approach to the custom-runtime path, which is where finding #2 lives.

No action needed from the worker beyond the cycle already in flight — this is my record correction, not a new request. For my next re-test I will derive the expected set from the injection sites (`createGitCredentialHelperEnvironment` and the spawn boundaries) rather than from the contract array, so the harness can actually fail.

packages/runtime-codex/src/runtime.test.ts
moncher-dev: Resolved in `f420e3f`. The new `createGitCredentialHelperEnvironment` test derives names from the injection function and independently checks the literal credential contract, reserved-auth declaration, and dynamic Git-config pattern. The existing Codex and Claude boundary tests remain in place for wiring coverage, and the architecture wording now distinguishes these roles.

packages/core/src/runtime/custom-child-env.ts
moncher-dev: Resolved in `f420e3f`. Core now exports `isAgentChildCredentialEnvironmentName`, which recognizes both declared literals and every `GIT_CONFIG_KEY_*` / `GIT_CONFIG_VALUE_*` suffix. Custom-runtime validation and final child stripping both call it, with workflow-loader regression cases for suffixes 1, 7, and 12.

packages/core/src/workflow-loader.test.ts
moncher-dev: Resolved in `f420e3f`. Added `.changeset/quiet-runtimes-strip-credentials.md` as a patch changeset for `@gh-symphony/cli`, describing the fail-loud custom-runtime authentication behavior. The refreshed PR body also calls out the custom-runtime validation change.

moncher-dev (commented):
## Re-test at `f420e3f2` — all three findings verified fixed, ready for @hojinzs

This time I built the harness so it could actually fail, per your note on cycles 1–2 and my correction above.

**Suite:** `pnpm install --frozen-lockfile && pnpm build && pnpm test` — green across all 14 packages.

### Finding #1 (P1, tautological tests) — fixed, and I proved it by mutation

The new `classifies every injected credential name for child stripping` test is genuinely injection-sourced: it derives its name list by *calling* `createGitCredentialHelperEnvironment` and asserts the contract covers what that function actually injects. So I ran your experiment rather than trusting mine:

```
# deleted "GITHUB_GIT_HOST" from AGENT_CHILD_CREDENTIAL_ENVIRONMENT_NAMES, rebuilt core
× createGitCredentialHelperEnvironment > classifies every injected credential name for child stripping
  Tests  1 failed | 42 passed (43)

# restored
  Tests  43 passed (43)
```

The test now fails on exactly the drift that previously passed silently. That is AC #2 met — the list is self-checking against a real injection site, not against itself.

### Finding #2 (P2, prefix family) — fixed

`isAgentChildCredentialEnvironmentName` is now a shared predicate covering both the literal names and the `GIT_CONFIG_KEY_*` / `GIT_CONFIG_VALUE_*` prefixes, and both the validation path and the strip path call it. Probed with suffixes invented outside the contract:

```
prefix-family probes: 5 (KEY_0, VALUE_0, KEY_7, VALUE_12, KEY_999) | still silent: NONE
legit custom auth key preserved: true
GIT_CONFIGURED_THING not over-rejected: true
```

Arbitrary suffixes now reject at parse time rather than being silently dropped, and the predicate does not over-match a merely `GIT_CONFIG`-ish name.

### Finding #3 (P2, changeset) — present

`.changeset/quiet-runtimes-strip-credentials.md`, `@gh-symphony/cli` patch, describing the newly rejected names.

### Still holding

Re-ran the real spawned-child exercise through `launchCodexAppServer` at this head: `SPAWNED CHILD leaks: []`, `GIT_TERMINAL_PROMPT: 0`.

Container coverage is CI's `container-smoke` job, which is the authoritative check on this host.

packages/core/src/runtime/custom-child-env.ts
```diff
@@ -45,16 +45,26 @@ export function readAgentVisibleSymphonyContext(
...
+  "AGENT_CREDENTIAL_CACHE_PATH",
+  "GITHUB_GIT_HOST",
+  "GITHUB_GIT_USERNAME",
   "GIT_ASKPASS",
   "GIT_CONFIG_COUNT",
   "GIT_CONFIG_GLOBAL",
+  "GIT_CONFIG_KEY_0",
   "GIT_CONFIG_NOSYSTEM",
   "GIT_CONFIG_SYSTEM",
+  "GIT_CONFIG_VALUE_0",
```
hojinzs, line 61: **nit (non-blocking) — `GIT_CONFIG_KEY_0` / `GIT_CONFIG_VALUE_0` are now dead entries.**

Carried over from my cycle-3 comment; not addressed and not rebutted, so recording it rather than holding the PR on it.

Now that `isAgentChildCredentialEnvironmentName` owns the prefix rule and both validation and stripping route through it, these two literals are unreachable — `startsWith("GIT_CONFIG_KEY_")` already matches `GIT_CONFIG_KEY_0`. I confirmed the asymmetry they originally caused is gone: `_0`, `_1`, `_7` and `_12` now all reject at parse time and all throw from `buildCustomRuntimeChildEnvironment`, so removing them changes nothing behaviourally.

The one thing they still buy is a literal-list sample of the dynamic family for the declaration-driven boundary tests. That is a real if small benefit, so keeping them is a defensible call — just worth a comment saying so, otherwise the next reader hits the same "is this redundant?" question. Fine to leave for a follow-up either way.

---

packages/runtime-codex/src/runtime.test.ts
```diff
@@ -517,6 +537,36 @@ describe("parseAgentCommand", () => {
...
+          "https://broker.example/api/workspaces/workspace-123/runtime-credentials",
+        githubTokenBrokerSecret: "runtime-secret",
+        githubTokenCachePath: "/workspace-runtime/.github-token.json",
+        tokenBrokerTimeoutMs: 1_000,
+      })
+    ).filter(
+      (name) =>
+        name !== "GIT_TERMINAL_PROMPT" &&
+        name !== "GITHUB_TOKEN_BROKER_TIMEOUT_MS"
+    );
```
hojinzs, line 554: **nit (non-blocking) — the two exclusions are the one hand-maintained part of an otherwise self-checking test; give them a reason in the source.**

This filter is the escape hatch in the drift detector: anything listed here is exempt from the coverage assertion forever, and neither name carries a justification. I verified both are legitimate — `GIT_TERMINAL_PROMPT` is set rather than stripped (`custom-child-env.ts:194`), and `GITHUB_TOKEN_BROKER_TIMEOUT_MS` is `String(tokenBrokerTimeoutMs)`, a plain number with no secret content, so it is the only injected name that intentionally reaches the child. But that took reading three files to establish, and the next person adding a name here will have no signal about what the bar is.

A one-line comment — "not credentials: the first is asserted rather than stripped, the second is a non-secret timeout" — keeps the exemption list from quietly becoming the place drift hides.

---

hojinzs (approved):
## Approve — cycle 4 of max 6

All three cycle-3 findings are fixed. I verified each by mutation and repro rather than by re-reading the code, since the recurring failure mode on this PR has been checks that pass by construction.

### Finding #1 (P1) — AC #2 now genuinely met

The new injection-sourced test at `runtime.test.ts:540` fails in **both** drift directions. I ran the exact mutation that was green in cycle 3:

```
$ sed -i '/"GITHUB_GIT_HOST",/d' packages/core/src/runtime/custom-child-env.ts
FAIL  runtime.test.ts > classifies every injected credential name for child stripping
Tests  1 failed | 187 passed (188)
```

Cycle 3 reported `183 passed` on this same mutation. The detector now bites.

And the other half of the AC — "fails when a **newly injected** name is not covered" — added `NEW_HELPER_SECRET` to `createGitCredentialHelperEnvironment`:

```
AssertionError: expected [ 'GITHUB_GIT_HOST', …(8) ] to deeply equal [ …(9) ]
Tests  1 failed | 42 passed (43)
```

That is the property #868 was filed to establish: completeness no longer holds by coincidence, because the injection site and the contract are now checked against each other.

### Finding #2 (P2) — silent-drop path closed on the prefix axis

Re-ran my cycle-3 repro across the literals, the dynamic family, and a control:

```
GIT_CONFIG_KEY_0     parseRejects=true  builderThrows=true  reaches=false  SILENT=false
GIT_CONFIG_KEY_1     parseRejects=true  builderThrows=true  reaches=false  SILENT=false
GIT_CONFIG_VALUE_1   parseRejects=true  builderThrows=true  reaches=false  SILENT=false
GIT_CONFIG_KEY_7     parseRejects=true  builderThrows=true  reaches=false  SILENT=false
GIT_CONFIG_VALUE_12  parseRejects=true  builderThrows=true  reaches=false  SILENT=false
GITHUB_GIT_HOST      parseRejects=true  builderThrows=true  reaches=false  SILENT=false
SSH_AUTH_SOCK        parseRejects=true  builderThrows=true  reaches=false  SILENT=false
XDG_CONFIG_HOME      parseRejects=true  builderThrows=true  reaches=false  SILENT=false
MY_CUSTOM_TOKEN      parseRejects=false builderThrows=false reaches=true   SILENT=false
```

No silent names on any axis, and the control confirms it did not over-tighten — a legitimate custom auth key still reaches the child.

### Finding #3 (P2) — changeset landed

`.changeset/quiet-runtimes-strip-credentials.md`, patch on `@gh-symphony/cli`, describing the fail-loud rejection. PR body's behavior section now names the custom-runtime path.

### Smoke test — real spawned children, injection-sourced

Poisoned `process.env` with the actual `createGitCredentialHelperEnvironment` output plus every declared name and arbitrary suffixes (`_7`, `_12`), then spawned a real child that dumps its own environment:

```
CODEX plan leaks:            []
CODEX child leaks:           []
CODEX GIT_TERMINAL_PROMPT: 0 | OPENAI_API_KEY preserved: true
CLAUDE child env keys:       137
CLAUDE child leaks:          []
CLAUDE GIT_TERMINAL_PROMPT: 0
```

Clean at both boundaries, and neither over-tightens.

### Suite

`pnpm build`, `pnpm lint`, `pnpm typecheck` pass. `pnpm test` = **2,022 tests, 0 failures** across all packages. CI `Test` and `Container Smoke` green on `f420e3f`.

### Requirements and scope

**AC #1 met** — `grep` confirms only `packages/core` declares the list; both `removeChildHostCredentialEnvironment` and `removeChildHostGitCredentialEnvironment` are gone. **AC #2 met**, per the mutations above.

**No over-engineering.** Eight files, none outside the issue's scope. Inheritance allowlists (`SAFE_RUNTIME_ENV_KEYS`, `DEFAULT_INHERITED_ENV_KEYS`), tracker secret declarations, and MCP composition are untouched, per the #844 steward note. The one arbitrary-decision candidate is the cycle-2 validation widening, which I accepted then and still do: it derives from the contract rather than restating it, and it was driven by a real finding.

I also checked the two things a refactor like this most easily loses. `GIT_TERMINAL_PROMPT = "0"` survived the move into the shared function (`custom-child-env.ts:194`) — both deleted copies set it, and dropping it would have let the child prompt interactively instead of failing fast. And the two names that genuinely tighten Codex, `GITHUB_GIT_HOST` / `GITHUB_GIT_USERNAME`, have only host-side consumers (`git-transport.ts`, `git-credential-helper.ts`) with safe defaults; the child never had the `GIT_CONFIG_KEY_*` helper wiring to use them with, so nothing regresses.

Two nits inline, both non-blocking and fine as follow-ups. Approving.

---

PR #875, opened 2026-09-05
docs: map credential broker footprint before removal

moncher-dev:
## Issues

- Closes #874

## Summary

- Maps the two credential-broker protocols and traces the GitHub token path through host-side GraphQL and branch publication.
- Separates 19 production files into 6 credential data-path consumers, 1 Claude validation-only dependency, and 12 supporting surfaces, with reproducible counts for tests and documentation.
- Recommends three ordered removal issues that preserve the tracker control plane and both worker publication lifecycles.

## Change-point diagram

- Project/daemon credentials → GitHub tracker adapter → host GraphQL and worker/orchestrator Git transport
- Agent credential broker → Codex provider authentication; Claude preflight validation only
- Evidence and boundaries → `docs/reports/2026-09-05-credential-broker-footprint.md`

## Start here

- `docs/reports/2026-09-05-credential-broker-footprint.md:9` — executive conclusion and the two-broker distinction
- `docs/reports/2026-09-05-credential-broker-footprint.md:34` — reproducible dependency/mention split
- `docs/reports/2026-09-05-credential-broker-footprint.md:77` — timeout boundaries across both worker publication lifecycles
- `docs/reports/2026-09-05-credential-broker-footprint.md:107` — ordered three-issue recommendation

## User-Visible Behavior / Operational Impact

- None. This PR adds a read-only investigation report and changes no runtime, CLI, configuration, or package behavior.

## Validation

- `pnpm exec prettier --check docs/reports/2026-09-05-credential-broker-footprint.md` — pass
- `pnpm lint` — pass
- `git diff --check origin/main...HEAD` — pass
- `git diff --name-status origin/main...HEAD` — pass: exactly one added report file
- Tests, typecheck, build, and Docker E2E — not applicable to a documentation-only investigation with no behavior change

## Changeset

- Not needed because no package or user-visible behavior changes.

## Risks & rollback

- Risk is limited to an incorrect investigation conclusion; reviewers should focus on the six-plus-one consumer classification and timeout boundaries across both publication lifecycles.
- Rollback is removal of the single report file.

## Changed files

- `docs/reports/2026-09-05-credential-broker-footprint.md` — evidence-backed footprint, replacement floor, timeout analysis, and removal recommendation

## Post-merge / human validation

- [ ] Confirm the three proposed follow-up issue boundaries before beginning Phase 4 removal.

## Security

- [ ] No real tokens, private keys, `.env` files, or generated installation tokens are committed

PR #877, opened 2026-09-05
fix(tracker-github): reconcile tracker secret declarations

### LINKED ISSUE #869: fix(tracker-github): reconcile tracker secret declarations with what the adapters inject
Split 2 of 5 from #844 (see the triage comment there). Parent epic: #847, Phase 3.

## Problem

Tracker secret names have a declared source and two hardcoded copies that describe a different set.

- Codex derives them from `SYMPHONY_TRACKER_SECRET_ENVIRONMENT_NAMES` (`packages/runtime-codex/src/launcher.ts:64-70`). This is the declared path and is correct.
- The Claude adapter takes the declared names and *then also* hardcodes `GH_TOKEN`, `GH_ENTERPRISE_TOKEN`, `GITHUB_TOKEN`, `GITHUB_GRAPHQL_TOKEN`, `GITHUB_TOKEN_BROKER_SECRET`, `LINEAR_API_KEY` and `LINEAR_AUTHORIZATION` (`packages/runtime-claude/src/adapter.ts:694-700`).
- The GitHub adapter's own `secretEnvironmentNames()` (`packages/tracker-github/src/orchestrator-adapter.ts:44-52`) omits the broker URL and cache path that #827 now injects.

So the declaration and the hardcoded lists disagree, and the declaration is the one that is incomplete.

## Scope

- Add the broker URL and cache path (and anything else #827 injects) to the GitHub tracker adapter's `secretEnvironmentNames()` declaration.
- Audit the Linear adapter's declaration for the same gap.
- Delete the hardcoded tracker-secret list in the Claude adapter so both runtimes read the adapter declaration only.
- Test: the set the runtime strips equals the set the adapter declares, for each tracker adapter.

## Out of scope

- The credential-strip contract itself (split 1).
- Inheritance allowlists.
- MCP composition removal — composition is kept and repaired, gated by source.

## Acceptance criteria

- Tracker secret names come from the adapter declaration only; no hardcoded copies remain in either runtime.
- Each adapter's declaration covers every name it actually injects, asserted by test.

## Sequencing

After split 1 (#868), which establishes the shared strip contract this reads from.

moncher-dev:
## Issues

- Closes #869

## Summary

- Adds regression coverage proving GitHub and Linear tracker declarations include every credential key their adapters inject.
- Verifies both Codex and Claude strip synthetic tracker credentials supplied exclusively through adapter declarations.
- Strengthens adapter coverage so it fails if credential resolution unexpectedly returns nothing.

## Change-point diagram

- tracker-github / tracker-linear resolver outputs → pinned non-empty credential maps → declared secret-name superset assertions
- tracker adapter declarations → `SYMPHONY_TRACKER_SECRET_ENVIRONMENT_NAMES` → Codex and Claude child-environment stripping tests

## Start here

- `packages/tracker-github/src/tracker-github.test.ts:1262` — GitHub injection/declaration contract and resolver-branch fixtures
- `packages/tracker-linear/src/tracker-linear.test.ts:1490` — Linear injection/declaration contract
- `packages/runtime-codex/src/runtime.test.ts` — Codex declaration-driven stripping
- `packages/runtime-claude/src/adapter.test.ts` — Claude declaration-driven stripping

- None. The source corrections are already present on the base branch; this PR adds durable regression coverage.

## Validation

- `pnpm exec prettier --check packages/tracker-github/src/tracker-github.test.ts packages/tracker-linear/src/tracker-linear.test.ts` — pass
- `pnpm exec vitest run packages/tracker-github/src/tracker-github.test.ts packages/tracker-linear/src/tracker-linear.test.ts` — pass (163 tests)
- `pnpm lint` — pass
- `pnpm test` — pass
- `pnpm typecheck` — pass
- `pnpm build` — pass
- Docker E2E — not applicable: the diff is deterministic unit-test coverage only and changes no integration behavior

## Changeset

- Not needed because the PR changes tests only and has no shipped behavior.

## Risks & rollback

- Low risk: four test files only, with no production code changes.
- Reviewer focus: credential fixtures must enumerate every resolver return branch and remain non-empty.
- Rollback by reverting commits `6102416` and `a7cbe93`.

## Changed files

- `packages/tracker-github/src/tracker-github.test.ts` — asserts declared GitHub secrets cover pinned direct and broker injection results
- `packages/tracker-linear/src/tracker-linear.test.ts` — asserts declared Linear secrets cover pinned injection results
- `packages/runtime-codex/src/runtime.test.ts` — asserts Codex strips adapter-declared tracker credentials only
- `packages/runtime-claude/src/adapter.test.ts` — asserts Claude strips adapter-declared tracker credentials only

- [ ] None

## Security

### DIFF
--- a/packages/runtime-claude/src/adapter.test.ts
+++ b/packages/runtime-claude/src/adapter.test.ts
@@ -75,6 +75,39 @@ describe("ClaudePrintRuntimeAdapter", () => {
     );
   });
 
+  it("strips tracker credentials from the adapter declaration", async () => {
+    const calls: Array<NodeJS.ProcessEnv | undefined> = [];
+    const { child, stdout, stderr } = createStubChild();
+    const adapter = new ClaudePrintRuntimeAdapter(
+      {
+        workingDirectory: "/workspace",
+        env: {
+          TRACKER_ADAPTER_SECRET: "secret",
+          UNDECLARED_TRACKER_VALUE: "visible",
+          SYMPHONY_TRACKER_SECRET_ENVIRONMENT_NAMES: JSON.stringify([
+            "TRACKER_ADAPTER_SECRET",
+          ]),
+        },
+      },
+      {
+        spawnImpl: (_command, _args, options) => {
+          calls.push(options.env);
+          queueMicrotask(() => {
+            stdout.end();
+            stderr.end();
+            child.emit("close", 0, null);
+          });
+          return child;
+        },
+      }
+    );
+
+    await adapter.spawnTurn({ messages: [] });
+
+    expect(calls[0]?.TRACKER_ADAPTER_SECRET).toBeUndefined();
+    expect(calls[0]?.UNDECLARED_TRACKER_VALUE).toBe("visible");
+  });
+
   it("strips every declared credential name at the agent-child boundary", async () => {
     const calls: Array<NodeJS.ProcessEnv | undefined> = [];
     const { child, stdout, stderr } = createStubChild();
--- a/packages/tracker-github/src/tracker-github.test.ts
+++ b/packages/tracker-github/src/tracker-github.test.ts
@@ -1282,6 +1282,15 @@ Prompt`,
         daemon: {},
       }
     );
+    // Keep one fixture per resolveWorkerCredentials return branch.
+    expect(directCredentials).toEqual({
+      GITHUB_GRAPHQL_TOKEN: "github-token",
+    });
+    expect(brokerCredentials).toEqual({
+      GITHUB_TOKEN_BROKER_URL: "https://broker.example/token",
+      GITHUB_TOKEN_BROKER_SECRET: "broker-secret",
+      GITHUB_TOKEN_CACHE_PATH: "/runtime/github-token.json",
+    });
     const injectedNames = new Set([
       ...Object.keys(directCredentials ?? {}),
       ...Object.keys(brokerCredentials ?? {}),
--- a/packages/tracker-github/src/tracker-github.test.ts
+++ b/packages/tracker-github/src/tracker-github.test.ts
@@ -1259,6 +1259,39 @@ Prompt`,
     });
   });
 
+  it("declares every GitHub credential environment name it can inject", () => {
+    const adapter = resolveTrackerAdapter({
+      adapter: "github-project",
+      bindingId: "project-123",
+    });
+    const directCredentials = adapter.resolveWorkerCredentials?.(
+      makeProjectConfig(),
+      {
+        project: { GITHUB_GRAPHQL_TOKEN: "github-token" },
+        daemon: {},
+      }
+    );
+    const brokerCredentials = adapter.resolveWorkerCredentials?.(
+      makeProjectConfig(),
+      {
+        project: {
+          GITHUB_TOKEN_BROKER_URL: "https://broker.example/token",
+          GITHUB_TOKEN_BROKER_SECRET: "broker-secret",
+          GITHUB_TOKEN_CACHE_PATH: "/runtime/github-token.json",
+        },
+        daemon: {},
+      }
+    );
+    const injectedNames = new Set([
+      ...Object.keys(directCredentials ?? {}),
+      ...Object.keys(brokerCredentials ?? {}),
+    ]);
+
+    expect(adapter.secretEnvironmentNames()).toEqual(
+      expect.arrayContaining([...injectedNames])
+    );
+  });
+
   it("propagates the configured GitHub GraphQL endpoint into worker env", () => {
     const adapter = resolveTrackerAdapter({
       adapter: "github-project",
--- a/packages/tracker-linear/src/tracker-linear.test.ts
+++ b/packages/tracker-linear/src/tracker-linear.test.ts
@@ -1499,6 +1499,10 @@ Prompt`,
       }
     );
 
+    expect(credentials).toEqual({
+      LINEAR_AUTHORIZATION: "Bearer lin-authorization",
+      LINEAR_API_KEY: "lin_api_key",
+    });
     expect(linearTrackerAdapter.secretEnvironmentNames()).toEqual(
       expect.arrayContaining(Object.keys(credentials ?? {}))
     );
--- a/packages/tracker-linear/src/tracker-linear.test.ts
+++ b/packages/tracker-linear/src/tracker-linear.test.ts
@@ -1487,6 +1487,23 @@ Prompt`,
     expect(env).toEqual({ LINEAR_API_KEY: "lin_project_key" });
   });
 
+  it("declares every Linear credential environment name it can inject", () => {
+    const credentials = linearTrackerAdapter.resolveWorkerCredentials?.(
+      makeProject(),
+      {
+        project: {
+          LINEAR_API_KEY: "lin_api_key",
+          LINEAR_AUTHORIZATION: "Bearer lin-authorization",
+        },
+        daemon: {},
+      }
+    );
+
+    expect(linearTrackerAdapter.secretEnvironmentNames()).toEqual(
+      expect.arrayContaining(Object.keys(credentials ?? {}))
+    );
+  });
+
   it("defaults blank tracker apiUrl to the Linear GraphQL endpoint", () => {
     const env = linearTrackerAdapter.buildWorkerEnvironment(
       makeProject({ apiUrl: "   " }),
--- a/packages/runtime-codex/src/runtime.test.ts
+++ b/packages/runtime-codex/src/runtime.test.ts
@@ -121,6 +121,21 @@ describe("createCodexDynamicToolSpecs", () => {
 });
 
 describe("buildCodexRuntimePlan", () => {
+  it("strips tracker credentials from the adapter declaration", () => {
+    const plan = buildCodexRuntimePlan({
+      projectId: "workspace-123",
+      workingDirectory: "/tmp/workspace-123",
+      trackerSecretEnvironmentNames: ["TRACKER_ADAPTER_SECRET"],
+      extraEnv: {
+        TRACKER_ADAPTER_SECRET: "secret",
+        UNDECLARED_TRACKER_VALUE: "visible",
+      },
+    });
+
+    expect(plan.env.TRACKER_ADAPTER_SECRET).toBeUndefined();
+    expect(plan.env.UNDECLARED_TRACKER_VALUE).toBe("visible");
+  });
+
   it("strips every declared credential name at the agent-child boundary", () => {
     const injectedCredentials = Object.fromEntries(
       AGENT_CHILD_CREDENTIAL_ENVIRONMENT_NAMES.map((name) => [name, "secret"])


moncher-dev:
## Board review — PR #877 (issue #869)

Smoke-tested in a scratch worktree at head `6102416` against `main` = `519b1cfc`.

### Gate results
- `pnpm install --frozen-lockfile` / `pnpm build` / `pnpm test` / `pnpm typecheck` / `pnpm lint` — all pass. Full suite: every package's test files pass, 0 failed.
- CI: `Test` pass, `Container Smoke` pass. `MERGEABLE` / `CLEAN`.

### Source-level acceptance criteria of #869 verified on `main`, not just asserted
The PR is test-only, and its claim is that #827 + #868 already landed the source fix. I checked that on `main` rather than taking the claim:
- `packages/tracker-github/src/orchestrator-adapter.ts` — `secretEnvironmentNames()` now includes `GITHUB_TOKEN_BROKER_URL`, `GITHUB_TOKEN_BROKER_SECRET`, `GITHUB_TOKEN_CACHE_PATH`, i.e. every key `resolveWorkerCredentials()` can return on both the direct-token and broker branches.
- `packages/runtime-claude/src/adapter.ts:702` — the hardcoded `GH_TOKEN` / `GITHUB_*` / `LINEAR_*` tracker list named in the issue is gone; the strip set is now `declaredNames` (from `SYMPHONY_TRACKER_SECRET_ENVIRONMENT_NAMES`) ∪ MCP-sourced names ∪ the shared `CUSTOM_RUNTIME_RESERVED_AUTH_ENVIRONMENT_NAMES` from `@gh-symphony/core` (split 1, #873). That shared constant is the contract, not a per-runtime copy, so "no hardcoded copies remain in either runtime" holds.
- `packages/runtime-codex/src/runtime.ts:662` — same shape.

### The tests are load-bearing (mutation-tested, not just read)
I reintroduced each regression #869 describes and confirmed the new test fails, then reverted:
1. Dropped `GITHUB_TOKEN_CACHE_PATH` from the GitHub declaration → `declares every GitHub credential environment name it can inject` fails with the exact missing name. This is the original #869 defect.
2. Dropped `LINEAR_AUTHORIZATION` from the Linear declaration → Linear declaration test fails.
3. Forced `declaredNames = []` in the Claude adapter (a regression back to a hardcoded list) → `strips tracker credentials from the adapter declaration` fails, `expected 'secret' to be undefined`.
4. Removed `...secretEnvironmentNames` from the Codex strip set → the Codex counterpart fails identically.

All four also confirm the negative half: `UNDECLARED_TRACKER_VALUE` survives, so the strip is declaration-driven and not a blanket wipe.

### Non-blocking observation (not for this PR)
`packages/runtime-codex/src/runtime.ts:665` still strips the literal `"LINEAR_GRAPHQL_URL"` inline, and the Claude runtime does not — a tracker-specific literal living in a runtime, asymmetric between the two. It is a non-secret URL, absent from the Linear declaration, and predates this PR, so it is out of scope here; noting it in case it belongs to a later split of #844.

Local `run-standalone-project-e2e.sh` was not run: Docker is intentionally absent on this host and CI's `container-smoke` is the authoritative container check — it passed.

Recommend merge. @hojinzs added as reviewer.

packages/tracker-github/src/tracker-github.test.ts
```diff
@@ -1259,6 +1259,39 @@ Prompt`,
...
+        },
+        daemon: {},
+      }
+    );
+    const injectedNames = new Set([
+      ...Object.keys(directCredentials ?? {}),
+      ...Object.keys(brokerCredentials ?? {}),
+    ]);
+
+    expect(adapter.secretEnvironmentNames()).toEqual(
```
hojinzs, line 1299: **P2 — this assertion can pass vacuously, which defeats the guard this PR exists to add.**

`expect.arrayContaining([])` is a tautology. `injectedNames` is derived entirely from what `resolveWorkerCredentials` happens to return, and both call sites are `?? {}`-guarded — so if the resolver ever returns nothing, `injectedNames` is empty and this test goes green while asserting nothing.

I mutation-tested it rather than reasoning about it. Adding an early `return undefined;` at the top of `resolveWorkerCredentials` in `packages/tracker-github/src/orchestrator-adapter.ts`:

```
✓ packages/tracker-github/src/tracker-github.test.ts (115 tests | 114 skipped) 4ms
  Tests  1 passed | 114 skipped (115)
```

The declaration-coverage test passed with the credential path completely gutted.

For contrast, the defect #869 actually describes *is* caught — dropping `"GITHUB_TOKEN_CACHE_PATH"` from `secretEnvironmentNames()` fails with exactly the missing name. So the test works in the direction it was written for; it just has no floor under it.

Pinning the inputs fixes it in two lines:

```ts
expect(directCredentials).toEqual({ GITHUB_GRAPHQL_TOKEN: "github-token" });
expect(brokerCredentials).toEqual({
  GITHUB_TOKEN_BROKER_URL: "https://broker.example/token",
  GITHUB_TOKEN_BROKER_SECRET: "broker-secret",
  GITHUB_TOKEN_CACHE_PATH: "/runtime/github-token.json",
});
```

That keeps the `arrayContaining` check (declaration ⊇ injected is the right direction for the AC — the declaration legitimately also carries `GH_TOKEN` / `GH_ENTERPRISE_TOKEN` / `GITHUB_TOKEN`) while making the premise load-bearing.

---

packages/tracker-linear/src/tracker-linear.test.ts
```diff
@@ -1487,6 +1487,23 @@ Prompt`,
...
+      {
+        project: {
+          LINEAR_API_KEY: "lin_api_key",
+          LINEAR_AUTHORIZATION: "Bearer lin-authorization",
+        },
+        daemon: {},
+      }
+    );
+
+    expect(linearTrackerAdapter.secretEnvironmentNames()).toEqual(
```
hojinzs, line 1506: **P2 — same vacuous-pass shape as the GitHub counterpart.**

`Object.keys(credentials ?? {})` collapses to `[]` if `resolveWorkerCredentials` returns `undefined`, and `expect.arrayContaining([])` then passes unconditionally. One line restores the floor:

```ts
expect(credentials).toEqual({
  LINEAR_AUTHORIZATION: "Bearer lin-authorization",
  LINEAR_API_KEY: "lin_api_key",
});
```

Worth noting this fixture only exercises the branch where *both* `LINEAR_AUTHORIZATION` and `LINEAR_API_KEY` are set, which is also the only branch that returns both keys — so it does cover the full declared surface. Pinning it makes that intent explicit rather than incidental.

---

packages/tracker-github/src/tracker-github.test.ts
```diff
@@ -1259,6 +1259,39 @@ Prompt`,
...
+      adapter: "github-project",
+      bindingId: "project-123",
+    });
+    const directCredentials = adapter.resolveWorkerCredentials?.(
+      makeProjectConfig(),
+      {
+        project: { GITHUB_GRAPHQL_TOKEN: "github-token" },
+        daemon: {},
+      }
+    );
```
hojinzs, line 1273: **P3 (non-blocking) — the branch enumeration is invisible coupling.**

This test covers `resolveWorkerCredentials` by hand-picking two input shapes that happen to map 1:1 onto the resolver's two return branches (direct `GITHUB_GRAPHQL_TOKEN`, and the broker trio). That is correct today, but nothing links the fixture list to the branch list: a third branch injecting a new key would slip past this test silently, which is precisely the class of drift #869 was filed about.

A one-line comment naming the invariant would be enough, e.g. `// One fixture per return branch of resolveWorkerCredentials — add one when a branch is added.` Restructuring into a fixture table would also work but isn't worth the churn for two branches.

---

packages/runtime-claude/src/adapter.test.ts
```diff
@@ -75,6 +75,39 @@ describe("ClaudePrintRuntimeAdapter", () => {
     );
   });
 
+  it("strips tracker credentials from the adapter declaration", async () => {
```
hojinzs, line 78: This one is genuinely load-bearing — I checked that the assertion cannot be satisfied by a second path. `stripTrackerSecrets` unions `declaredNames` with `collectMcpSecretEnvironmentNames(...)`, and the latter is passed `secretEnvironmentNames: declaredNames`, so at a glance it looks like it might echo the declared names back and keep this test green even if `...declaredNames` were removed from the strip set. It does not: `collectMcpSecretEnvironmentNames` returns the `${VAR}` *source* names referenced by MCP configs (`packages/core/src/runtime/mcp-compose.ts:53-58`), using `declaredSecrets` only as a filter on the keys. So dropping `...declaredNames` really does break this test. Good.

nit (pre-existing, not introduced here): that same call reads `~/.gh-symphony/mcp.json` through `homedir()` with no `homeDir` override, so this test's environment is not fully hermetic. Harmless for these two assertions — `TRACKER_ADAPTER_SECRET` is stripped either way and `UNDECLARED_TRACKER_VALUE` would have to appear as an MCP source name to be affected — but worth knowing if these tests ever go flaky on a developer machine.

Also confirming the new test does not overlap the existing `strips every declared credential name at the agent-child boundary` below it: that one drives `AGENT_CHILD_CREDENTIAL_ENVIRONMENT_NAMES`, this one drives the `SYMPHONY_TRACKER_SECRET_ENVIRONMENT_NAMES` declaration path. Distinct paths, both worth having.

---

hojinzs (changes requested):
## Review round 1/6 — verified independently at `6102416`

I re-ran the verification from source rather than relying on the board-review comment, and reproduced its main claims. Two P2 test-strength defects block; everything else is non-blocking.

### Requirement coverage vs #869

| --- | --- |
| Each adapter's declaration covers every name it injects, **asserted by test** | ✅ Met — mutation-verified |
| Tracker secret names come from the adapter declaration only; no hardcoded copies remain in either runtime | ⚠️ Met at the runtime-file level; see below |

I confirmed on `main` that the three source fixes #869 asked for did land ahead of this PR, so a test-only PR here is the right shape:
- `packages/tracker-github/src/orchestrator-adapter.ts:44-54` — declaration now carries `GITHUB_TOKEN_BROKER_URL` / `GITHUB_TOKEN_BROKER_SECRET` / `GITHUB_TOKEN_CACHE_PATH`, matching every key `resolveWorkerCredentials` can return on either branch.
- `packages/runtime-claude/src/adapter.ts:702-713` — the per-runtime hardcoded `GH_TOKEN` / `GITHUB_*` / `LINEAR_*` list named in the issue is gone.
- `packages/runtime-codex/src/runtime.ts:611-619, 662-666` — same shape.

**On the second criterion.** The list wasn't deleted so much as relocated: `CUSTOM_RUNTIME_RESERVED_AUTH_ENVIRONMENT_NAMES` (`packages/core/src/runtime/custom-child-env.ts:74-87`) still hardcodes `GH_TOKEN`, `GH_ENTERPRISE_TOKEN`, `GITHUB_TOKEN`, `GITHUB_GRAPHQL_TOKEN`, `GITHUB_TOKEN_BROKER_SECRET`, `LINEAR_API_KEY`, `LINEAR_AUTHORIZATION` — the exact set #869 quoted — and both runtimes union it into the strip set. So "declaration only" is not literally true, and tracker-specific names now sit in `packages/core`, which cuts against CLAUDE.md's *"Keep tracker-specific behavior out of core layers."*

I am **not** asking for a change here: it arrived via #873, it is a strict superset of the declarations (fail-safe, no leak), and widening this PR to chase it would be scope creep. But it does mean the declaration is not the sole source of truth, and no test in this PR can detect that — the new Claude/Codex cases use a synthetic `TRACKER_ADAPTER_SECRET` precisely because a real tracker name would be stripped by the core constant regardless. Worth a follow-up issue under #844 rather than a silent close of #869.

### Smoke test

Built and ran the suite locally on this head (`pnpm install --frozen-lockfile` → `pnpm build` → targeted vitest):

```
✓ packages/tracker-github/src/tracker-github.test.ts  (115 tests) 149ms
✓ packages/runtime-claude/src/adapter.test.ts          (38 tests) 165ms
✓ packages/runtime-codex/src/runtime.test.ts
✓ packages/tracker-linear/src/tracker-linear.test.ts   (48 tests)  53ms
  Test Files  4 passed (4)   Tests  245 passed (245)
```

CI on `6102416`: `Test` ✅, `Container Smoke` ✅. `mergeable_state: clean`. No API or UI surface changes, so container smoke is the right ceiling here — agreed that the Docker E2E script isn't applicable to a test-only diff.

### Mutation testing

The point of this PR is assertion strength, so I mutated the source instead of reading the tests:

1. Dropped `"GITHUB_TOKEN_CACHE_PATH"` from the GitHub declaration → test fails naming exactly that key. **This is the original #869 defect, and it is caught.** ✅
2. Gutted `resolveWorkerCredentials` (early `return undefined;`) → the same test **passes**. ❌ See the P2 below.

### Blocking (P2)

Both are one- to two-line fixes in the tests only; no source change.

- `tracker-github.test.ts:1285-1292` — `expect.arrayContaining([])` is a tautology, and `injectedNames` is `?? {}`-guarded on both sides, so the guard evaporates without signal if the credential path ever returns nothing.
- `tracker-linear.test.ts:1502-1504` — identical shape.

A contract test whose whole job is "the declaration covers everything injected" shouldn't be able to go green with nothing injected. Pin the resolved credentials with `toEqual` and the `arrayContaining` check keeps working as intended.

### Non-blocking

- **P3** — the GitHub test's fixtures map 1:1 onto the resolver's return branches with nothing linking them; a future third branch slips past silently. A one-line comment suffices.
- **nit** — `collectMcpSecretEnvironmentNames` reads `~/.gh-symphony/mcp.json` via `homedir()` with no override, so the Claude tests aren't fully hermetic. Pre-existing.
- Agreed the inline `"LINEAR_GRAPHQL_URL"` literal at `runtime.ts:665` (Codex-only, absent from Claude) is out of scope here.

### Scope discipline

No over-engineering and no unrequested decisions: 98 added lines, four test files, zero source changes, no speculative abstraction. Omitting the changeset is correct per CLAUDE.md — no shipped behavior changes.

Push the two P2 fixes and I'll approve on the next round.

---

packages/tracker-github/src/tracker-github.test.ts
moncher-dev: Resolved in `a7cbe93`: the test now pins both the direct-token and broker credential maps before deriving `injectedNames`, so an undefined or empty resolver result fails explicitly.

packages/tracker-linear/src/tracker-linear.test.ts
moncher-dev: Resolved in `a7cbe93`: the Linear test now pins the expected authorization and API-key credential map before checking declaration coverage, preventing a vacuous empty-array pass.

packages/tracker-github/src/tracker-github.test.ts
moncher-dev: Addressed in `a7cbe93`: added a concise invariant comment requiring one fixture per `resolveWorkerCredentials` return branch.

packages/runtime-claude/src/adapter.test.ts
moncher-dev: Acknowledged. The MCP home-directory isolation concern is pre-existing and out of scope for this test-only rework; the declaration-driven assertion remains load-bearing as verified in the review.

hojinzs (approved):
## Review round 2/6 — Approve, verified at `a7cbe93`

Both round-1 P2s are fixed, and I confirmed that by re-running the exact mutation that defeated them rather than reading the diff or trusting the thread replies.

### Round-1 findings — all closed

| # | Finding | Status |
| --- | --- | --- |
| P2 | `tracker-github.test.ts` — vacuous `arrayContaining([])` pass | ✅ Fixed, mutation-verified |
| P2 | `tracker-linear.test.ts` — same shape | ✅ Fixed, mutation-verified |
| P3 | GitHub fixtures ↔ resolver branches unlinked | ✅ Comment added |
| nit | `collectMcpSecretEnvironmentNames` non-hermetic `homedir()` | ✅ Correctly declined as pre-existing |

**The P2 fix is load-bearing, not cosmetic.** In round 1, adding an early `return undefined;` to `resolveWorkerCredentials` let the declaration-coverage test pass with the credential path gutted. Re-run on this head:

```
× declares every GitHub credential environment name it can inject
  → expected undefined to deeply equal { Object (GITHUB_GRAPHQL_TOKEN) }
× declares every Linear credential environment name it can inject
  → expected undefined to deeply equal { …(2) }
```

Both now fail. The floor is real. And the test still catches the defect it was written for — dropping `"GITHUB_TOKEN_CACHE_PATH"` from `secretEnvironmentNames()` fails naming exactly that key, which is the original #869 regression.

### Scope discipline on the round-2 delta

`6102416..a7cbe93` is 15 added lines across two test files: two `toEqual` pins and one comment. Exactly what was asked, nothing adjacent picked up along the way. No over-engineering, no unrequested decisions, no speculative abstraction.

### Requirement coverage vs #869

| --- | --- |
| Each adapter's declaration covers every name it injects, asserted by test | ✅ Met, mutation-verified in both directions |
| Tracker secret names come from the adapter declaration only; no hardcoded copies remain in either runtime | ✅ Met at the runtime level |

Re-confirmed on this head that no tracker-name literals survive in either runtime source — `grep` for `"GH_TOKEN"` / `"GH_ENTERPRISE_TOKEN"` / `"LINEAR_API_KEY"` / `"LINEAR_AUTHORIZATION"` across `runtime-claude/src` and `runtime-codex/src` (excluding tests) returns nothing. `stripTrackerSecrets` is now `declaredNames` ∪ MCP-sourced names ∪ the shared core constant.

### Smoke test

No API or UI surface is touched, so the local suite plus CI container smoke is the right ceiling here; there is no preview environment to exercise.

- `pnpm build` ✅ · `pnpm lint` ✅ · `pnpm typecheck` ✅
- `pnpm test` ✅ — **2,026 tests, 14 packages, 0 failed**
- Four touched files: 245 passed (245)
- CI on `a7cbe93`: `Test` ✅, `Container Smoke` ✅
- `pnpm format` reports 72 files, but the identical 72 fail on `main` — pre-existing, and none of this PR's four files are among them. Not this PR's.
- Base is behind `main` by one docs-only commit (#875), so no conflict risk.

Omitting the changeset remains correct per CLAUDE.md: tests only, no shipped behavior.

### Carried forward, not blocking

The round-1 observation stands unchanged and I am explicitly *not* gating on it: `CUSTOM_RUNTIME_RESERVED_AUTH_ENVIRONMENT_NAMES` (`packages/core/src/runtime/custom-child-env.ts`) still hardcodes the exact tracker-name set #869 quoted, so "declaration only" is true per-runtime but not end-to-end, and tracker-specific names sit in `packages/core` against CLAUDE.md's *"keep tracker-specific behavior out of core layers."* It arrived via #873, it is a fail-safe superset, and chasing it here would be scope creep. Suggest tracking it as a follow-up under #844 rather than closing #869 silently — your call, I have not filed one.

The `"LINEAR_GRAPHQL_URL"` literal at `runtime-codex/src/runtime.ts:665` (Codex-only, asymmetric with Claude) belongs to that same follow-up.

Approving. Two rounds, both P2s closed with minimal diffs — no reason to hold this further.

---

PR #880, opened 2026-09-05
refactor(core): share agent child environment constructor

### LINKED ISSUE #870: refactor(core): build the agent child's explicit environment through one shared constructor
Split 3 of 5 from #844 (see the triage comment there). Parent epic: #847, Phase 3.

## Problem

Every value the agent child must see has to be added to both runtimes by hand. #825 adds `DOCKER_CONFIG` to both; #812 adds `SYMPHONY_ASSIGNED_BRANCH` to both. Each runtime builds the child's explicit assignments in its own code, so the two constructions drift and every new value costs two edits and can be forgotten in one of them.

This is the recurring cost #844 exists to remove.

## Scope

- Add a shared explicit child-environment constructor in `packages/core` that both runtimes use to build the agent child's assignments (the host-constructed values, applied after the strip from #868).
- Migrate `packages/runtime-codex` and `packages/runtime-claude` onto it and delete their per-runtime construction.
- Test that both runtimes produce the same child assignment set for the same inputs, so a value added once appears in both.

## Out of scope

- Consolidating the five *inheritance* allowlists into one data module — explicitly out of scope on #844.
- A declarative non-secret configuration channel to the agent child — explicitly out of scope on #844; name-based validation is not sound for the motivating value class (paths).
- MCP composition removal — composition is kept and repaired, gated by source.

## Acceptance criteria

- Both runtimes build their explicit child assignments through the shared helper.
- Adding a new host-constructed value requires one edit, and a test proves both runtimes pick it up.

## Sequencing

After split 1 (#868) and split 2 (#869).

moncher-dev:
## Issues

- Closes #870

## Summary

- TL;DR: Centralize built-in agent child environment assignments in core so Codex and Claude receive the same host-constructed values.
- Apply declared-secret exclusions only to source-derived agent-visible context, while pinned child-isolation paths always win.
- Isolate `HOME`, `USERPROFILE`, and runtime config homes consistently, preserve sanitized Symphony context, and add a CLI patch changeset.

## Change-point diagram

- `packages/core/src/runtime/agent-child-env.ts` → filters source context → appends pinned isolation assignments
- `packages/runtime-codex` + `packages/runtime-claude` → collect runtime-specific removal sets → call the shared constructor
- Core and runtime tests → prove the shared set, context exclusion, and pinned-value collision behavior

## Start here

- `packages/core/src/runtime/agent-child-env.ts:5` — shared assignment constructor and precedence boundary
- `packages/runtime-codex/src/runtime.ts:657` — Codex removal set and final pinned `CODEX_HOME`
- `packages/runtime-claude/src/adapter.ts:654` — Claude inherit/non-inherit construction

- Built-in agent children consistently receive isolated `HOME` and `USERPROFILE`, plus shared GitHub CLI and Docker config paths.
- A Symphony context name declared secret remains absent, but a collision with a pinned isolation name cannot disable child isolation.
- Custom runtime construction remains a separate boundary and is intentionally unchanged.

## Validation

- `pnpm exec vitest run packages/core/src/runtime/agent-child-env.test.ts packages/runtime-codex/src/runtime.test.ts packages/runtime-claude/src/adapter.test.ts` — pass (84 tests, including after merging current `main`)
- `pnpm lint` — pass
- `pnpm test` — pass
- `pnpm typecheck` — pass
- `pnpm build` — pass
- Docker E2E / blackbox — not applicable: internal built-in runtime environment construction is covered at the exact spawn-plan boundary and does not change orchestration dispatch or lifecycle behavior

## Changeset

- `.changeset/fuzzy-cats-guard.md` (`patch`)

## Risks & rollback

- Risk is limited to environment precedence at the built-in agent-child boundary; regression tests cover both runtime paths, declared-secret overlap, and pinned-name collisions.
- Roll back commits `0bb0fd6`, `efa724b`, and `70ef0ce` together to restore per-runtime construction.

## Changed files

- `packages/core/src/runtime/agent-child-env.ts` and test — shared constructor, source-only exclusions, and exact assignment coverage
- `packages/core/src/runtime/index.ts` — export the constructor
- `packages/runtime-codex/src/runtime.ts` and test — migrate Codex and preserve removed context names plus pinned `CODEX_HOME`
- `packages/runtime-claude/src/adapter.ts` and test — migrate both Claude environment paths and preserve pinned isolation values
- `docs/architecture.md` — map the shared built-in boundary and distinct custom-runtime boundary
- `.changeset/fuzzy-cats-guard.md` — patch release note

- [ ] None beyond normal release review

## Security


moncher-dev:
Smoke-tested at head `55d4df0f` in a clean worktree. Ready for your review.

**Bar**: `pnpm install --frozen-lockfile` → `pnpm build` → `pnpm test` (all packages, exit 0; orchestrator 362, cli 566) → `pnpm lint` (0) → `pnpm typecheck` (0). CI `Test` and `Container Smoke` both pass on this head.

**Exercised for real, not asserted.** I built both runtimes' child environments from the compiled `dist` under a deliberately hostile host env (`HOME=/Users/operator`, `USERPROFILE=/Users/operator`, a *stale* `SYMPHONY_ISSUE_ID=STALE-999` simulating cross-issue leakage, plus `GIT_ASKPASS`, `GIT_CONFIG_GLOBAL`, `AGENT_CREDENTIAL_BROKER_SECRET`, `GITHUB_TOKEN`), passing the run-scoped values through `extraEnv`/`env` including a clone URL with an embedded `ghs_` token. For Codex I then spawned a real child process with the computed `plan.env` and printed what the child actually saw; for Claude I drove `ClaudePrintRuntimeAdapter.spawnTurn` through the injected `spawnImpl` and captured the env handed to spawn. I ran the identical script against the `main` build (`bca2fcbb`) for a differential.

Result — the child sees `HOME` = `USERPROFILE` = child home, `GH_CONFIG_DIR`, `DOCKER_CONFIG`, `CODEX_HOME` under it, `GIT_TERMINAL_PROMPT=0`, run-scoped `SYMPHONY_ISSUE_ID=CURRENT-870` winning over the stale host value, and every credential variable absent.

**Two real fixes the PR body undersells**, both visible only in the differential:

1. *Codex leaked an installation token to the agent child on `main`.* Because `readAgentVisibleSymphonyContext` read only `process.env` and was spread **early**, the raw `extraEnv` value won and the child received `TARGET_REPOSITORY_CLONE_URL = https://x-access-token:ghs_SECRET@github.com/...`. On this branch the sanitizer reads `extraEnv`/`agentEnv` and is applied **last**, so the child receives `https://github.com/hojinzs/github-symphony.git`. The Claude path was already correct; Codex was not.
2. *Claude leaked the operator's `USERPROFILE` on `main`* (`/Users/operator`) while `HOME` was correctly isolated — a hole in the `inheritProcessEnv` path. Now pinned to the child home on both runtimes.

**Two judgment calls for you, neither blocking:**

- **Precedence moved.** The shared assignments are now applied *after* credential stripping and after `orchestratorRunEnv`/`agentEnv`, so the nine agent-visible names can no longer be overridden by those layers. I checked the overlap: `orchestratorRunEnv` sets only `SYMPHONY_ORCHESTRATOR_URL` / `SYMPHONY_RUN_ID` / `SYMPHONY_ORCHESTRATOR_TOKEN`, none of which are in `AGENT_VISIBLE_SYMPHONY_CONTEXT_ENVIRONMENT_NAMES`, and none of the nine are in `AGENT_CHILD_CREDENTIAL_ENVIRONMENT_NAMES`, so nothing stripped is reintroduced and no run-scoped value is clobbered today. It is a real ordering change to be aware of.
- **Codex's context sources widened.** It now reads from `config.extraEnv` / `config.agentEnv`, where previously only `process.env` was consulted and `resolvePreparedAgentEnvironment` filtered `agentEnv` down to `OPENAI_*`. The widening is confined to the nine non-secret allowlisted names (clone URL sanitized), and it is what makes fix #1 above work — but it is a behavior change beyond a pure refactor, and #870 carries no `changeset:*` label. Worth deciding whether the agent-visible env change deserves a changeset.

packages/runtime-codex/src/runtime.ts
```diff
@@ -667,6 +663,13 @@ export function buildCodexRuntimePlan(
     delete plan.env[name];
   }
   stripCredentialEnvironmentForAgentChild(plan.env);
+  Object.assign(
+    plan.env,
+    buildAgentChildEnvironmentAssignments({
+      childHome,
+      sources: [process.env, config.extraEnv, config.agentEnv],
+    })
+  );
```
hojinzs, line 674: **P1 — the declared-secret strip is no longer authoritative; stripped names are reintroduced here.**

The delete loop just above removes `secretEnvironmentNames` (tracker-declared + MCP-derived source names) from `plan.env`. `buildAgentChildEnvironmentAssignments` then re-reads `process.env` / `config.extraEnv` / `config.agentEnv` and assigns unconditionally, so any of the nine `AGENT_VISIBLE_SYMPHONY_CONTEXT_ENVIRONMENT_NAMES` that was designated a secret comes straight back.

Verified as a differential against `main`, same inputs (`trackerSecretEnvironmentNames: ["TARGET_REPOSITORY_URL"]`, value supplied via `extraEnv`):

```
main    → plan.env.TARGET_REPOSITORY_URL === undefined                (stripped)
55d4df0 → plan.env.TARGET_REPOSITORY_URL === "https://github.com/o/r" (reintroduced)
```

I'm not claiming a live secret escape: the two tracker adapters declare only token names, and the nine agent-visible names are non-secret by construction with the clone URL sanitized. But `secretEnvironmentNames` also carries MCP-derived source names, which are operator-controlled (`~/.gh-symphony/mcp.json`) and repo-controlled under `trustRepoConfig` — so an arbitrary name *can* land in that set, and the strip that `main` enforced silently stops holding. On a PR whose subject is the child security boundary, "applied last" needs to mean "last, minus what was removed".

Cheapest symmetric fix, and it keeps the one-edit property the issue asks for — give the constructor an exclusion input rather than open-coding the subtraction at each call site:

```ts
Object.assign(
  plan.env,
  buildAgentChildEnvironmentAssignments({
    childHome,
    sources: [process.env, config.extraEnv, config.agentEnv],
    excludeNames: removedNames, // the Set already built for the delete loop
  })
);
```

Worth a test that pins it on both runtimes: a name in both the strip set and the agent-visible set must be absent from the child env.

---

packages/runtime-claude/src/adapter.ts
```diff
@@ -654,23 +654,19 @@ function buildClaudeSpawnEnv(options: {
...
       ...options.configEnv,
       ...options.inputEnv,
-      ...agentVisibleContext,
     };
     stripTrackerSecrets(env, options.workingDirectory, options.configEnv);
-    env.HOME = options.childHome;
-    env.GH_CONFIG_DIR = join(options.childHome, "gh");
-    env.DOCKER_CONFIG = join(options.childHome, ".docker");
     stripCredentialEnvironmentForAgentChild(env);
+    Object.assign(env, explicitAssignments);
```
hojinzs: **P1 (same defect as the Codex call site) — `explicitAssignments` is applied after `stripTrackerSecrets`, undoing it for the nine agent-visible names.**

Applies to both paths: here (inherit) and again on line 685 (non-inherit). On `main` the agent-visible context was spread *before* `stripTrackerSecrets`, so the strip governed it. Now it runs after and wins.

Differential, driving `spawnTurn` through an injected `spawnImpl` with `SYMPHONY_TRACKER_SECRET_ENVIRONMENT_NAMES=["TARGET_REPOSITORY_URL"]` and that name present in `configEnv`:

```
main    → TARGET_REPOSITORY_URL === undefined                (stripped)
55d4df0 → TARGET_REPOSITORY_URL === "https://github.com/o/r" (reintroduced)
```

Claude's exposure is a bit wider than Codex's, because `stripTrackerSecrets` reads its name list from `SYMPHONY_TRACKER_SECRET_ENVIRONMENT_NAMES` in the env itself plus `collectMcpSecretEnvironmentNames`, both operator/repo-supplied. Same fix as on the Codex side — subtract the stripped names from the assignments before applying them, ideally via an `excludeNames` option on the shared constructor so the two runtimes stay identical.

Credit where due: the reordering is also what fixes the real `USERPROFILE` hole. I confirmed on `main` that the *non-inherit* path — the one the test calls "returns only the isolated child boundary" — handed the child `USERPROFILE=/Users/operator` while `HOME` was correctly isolated, because `DEFAULT_INHERITED_ENV_KEYS` copies it in and only `HOME` was pinned. That's broader than the inherit-path-only framing in your PR comment, and it's a genuine isolation fix. Keep the new ordering; just make it respect the strip.

---

packages/core/src/runtime/agent-child-env.ts
```diff
@@ -0,0 +1,16 @@
+import { join } from "node:path";
+import { readAgentVisibleSymphonyContext } from "./custom-child-env.js";
+
+/** Builds host-constructed environment assignments shared by agent runtimes. */
+export function buildAgentChildEnvironmentAssignments(options: {
+  childHome: string;
+  sources?: ReadonlyArray<NodeJS.ProcessEnv | undefined>;
+}): NodeJS.ProcessEnv {
+  return {
+    ...readAgentVisibleSymphonyContext(...(options.sources ?? [])),
```
hojinzs: **P3 — `sources` is optional but never omitted; the `?? []` fallback is dead.**

Both call sites pass `sources`. The optional marker plus the `?? []` fallback is the uncovered branch the coverage bot flagged on this new file (branches 50%). Make it required — a shared constructor whose whole point is a single definition site shouldn't have a silently-empty-context mode:

```ts
export function buildAgentChildEnvironmentAssignments(options: {
  childHome: string;
  sources: ReadonlyArray<NodeJS.ProcessEnv | undefined>;
  excludeNames?: Iterable<string>;
}): NodeJS.ProcessEnv {
```

`excludeNames` is the P1 fix from the two runtime call sites; landing it here is what keeps "one edit" true.

---

packages/runtime-codex/src/runtime.test.ts
```diff
@@ -236,6 +236,7 @@ describe("buildCodexRuntimePlan", () => {
     expect(plan.tools).toEqual([]);
     expect(plan.env).toMatchObject({
       HOME: "/tmp/runtime-123/child-home",
+      USERPROFILE: "/tmp/runtime-123/child-home",
```
hojinzs, line 243: **nit** — the acceptance criterion is "a value added once appears in both", but this asserts the four shared names by hand (and with `toMatchObject`, so it's a subset check). Adding a tenth host-constructed value still means editing this assertion and the Claude one.

Structurally you're already safe, since both runtimes now call the constructor — but a direct parity assertion would make the criterion actually tested rather than incidental:

```ts
expect(plan.env).toMatchObject(
  buildAgentChildEnvironmentAssignments({ childHome, sources: [...] })
);
```

Low value; take it or leave it.

---

hojinzs (changes requested):
Reviewed at head `55d4df0f`. Round 1 of at most 6.

The refactor itself is the right shape and stays inside #870's scope — no over-engineering, no drive-by decisions. One thing must change before merge.

## Verdict: request changes

**P1 — the reordering silently defeats the declared-secret strip.** Both runtimes now apply the shared assignments *after* their secret-removal step, and the constructor re-reads the same sources, so any of the nine `AGENT_VISIBLE_SYMPHONY_CONTEXT_ENVIRONMENT_NAMES` that a tracker adapter or an MCP config designates as a secret is deleted and then put back. Inline on both call sites with the `main` differential. No live secret escape today — the nine are non-secret by construction and the clone URL is sanitized — but `secretEnvironmentNames` also carries operator- and (under `trustRepoConfig`) repo-supplied MCP source names, so an arbitrary name can reach that set, and the invariant `main` enforced no longer holds. Fix is a few lines; suggested shape is an `excludeNames` option on the shared constructor so both runtimes stay identical.

Also **P2 (changeset)** and **P3 (a third construction site)** below, plus two inline P3/nits.

## Requirements vs #870

| --- | --- |
| Both runtimes build explicit child assignments through the shared helper | ✅ |
| Per-runtime construction deleted | ✅ |
| Adding a host-constructed value requires one edit | ⚠️ see P3 |
| A test proves both runtimes pick it up | ✅ structurally (both call the helper); assertions are hand-written per runtime — nit inline |

Out-of-scope items on the issue (allowlist consolidation, declarative non-secret channel, MCP composition removal) are all correctly left alone.

## Smoke test

Clean checkout of `55d4df0f`, `pnpm install --frozen-lockfile` → `pnpm build` (success) → core 354 / runtime-codex 81 / runtime-claude 110 tests, all passing. No preview environment for this change, so I exercised the two child-environment builders directly and ran the identical probes against the `main` build (`bca2fcb`) for a differential — hostile host env (`HOME`/`USERPROFILE=/Users/operator`, stale `SYMPHONY_ISSUE_ID=STALE-999`, `GIT_ASKPASS`, `GIT_CONFIG_GLOBAL`, `AGENT_CREDENTIAL_BROKER_SECRET`, `GITHUB_TOKEN`), run-scoped values through `extraEnv`, Claude driven through `spawnTurn` with an injected `spawnImpl`.

| | `main` | `55d4df0f` |
| --- | --- | --- |
| Codex `USERPROFILE` | absent | child home ✅ |
| Claude `USERPROFILE` (non-inherit) | `/Users/operator` | child home ✅ |
| Codex clone URL from `extraEnv` | `https://x-access-token:ghs_SECRET@…` | sanitized ✅ |
| Run-scoped `SYMPHONY_ISSUE_ID` over stale host value | wins | wins ✅ |
| Credential vars in child | absent | absent ✅ |
| Declared-secret name ∩ agent-visible name | stripped | **reintroduced** ❌ |

`HOME` / `GH_CONFIG_DIR` / `DOCKER_CONFIG` / `CODEX_HOME` / `GIT_TERMINAL_PROMPT=0` all correct on both runtimes.

## Answers to the two judgment calls you raised

**Precedence.** Your overlap analysis is right and I re-derived it: `orchestratorRunEnv` sets only `SYMPHONY_ORCHESTRATOR_URL` / `SYMPHONY_RUN_ID` / `SYMPHONY_ORCHESTRATOR_TOKEN`, none of the nine, and none of the nine are in `AGENT_CHILD_CREDENTIAL_ENVIRONMENT_NAMES`. What that analysis misses is the *other* strip — the `secretEnvironmentNames` delete loop in Codex and `stripTrackerSecrets` in Claude — which is where the P1 lives. Keep the new ordering; make it subtract.

**Changeset — yes, add one (P2).** The PR body reasons from the absence of a `changeset:*` label on #870, but CLAUDE.md keys the rule on behavior: "Behavior-changing PRs must add a changeset." This changes what the agent child actually receives on both runtimes (Codex gains `USERPROFILE`; Claude stops handing the operator's host `USERPROFILE` to an otherwise-isolated child). A patch changeset for `@gh-symphony/cli` describing the `USERPROFILE` pinning is the honest record.

One correction to the premise, since it affects how you size that changeset: **claim #1 is real at the function boundary but not reachable from any production caller.** `runtime-codex/src/launcher.ts:44` already builds `extraEnv` through `readAgentVisibleSymphonyContext(env)`, so the clone URL arrives pre-sanitized; and `orchestrator/src/runtime-factory.ts:171` passes `context.env`, whose `TARGET_REPOSITORY_CLONE_URL` is `deriveCloneUrl(repository.url)` — a GitHub API URL with no userinfo. I reproduced your result only by injecting a tokenized URL into `extraEnv` directly, which nothing in the orchestrator does. So it's defense-in-depth hardening, not a leak that was occurring. Claim #2 is the one that was actually live, and it's slightly *bigger* than you described — see the Claude inline.

## P3 — a third construction site still exists

`packages/core/src/runtime/custom-child-env.ts:151-153` still hand-assigns `HOME` / `USERPROFILE` / `GH_CONFIG_DIR`, and omits `DOCKER_CONFIG` — so the custom-runtime child already diverges from the two agent runtimes on a host-constructed value, and "one edit" isn't true while it stands. #870's scope names only the two runtimes, so I'm not asking you to widen the PR. Either migrate it in a follow-up, or, if the `DOCKER_CONFIG` omission is deliberate, say so in the `§17.5` row you touched in `docs/architecture.md` so the next person doesn't read the shared constructor as universal.

The `docs/architecture.md` update is otherwise accurate.

Fix the P1 and add the changeset and I'll approve.

---

packages/runtime-codex/src/runtime.ts
moncher-dev: Fixed in `efa724b`. The Codex strip set is now retained as `removedEnvironmentNames` and passed to the shared constructor via `excludeNames`, so last-applied host assignments cannot restore a declared secret. The runtime test pins `TARGET_REPOSITORY_URL` as both agent-visible and tracker-secret and verifies it remains absent.

packages/runtime-claude/src/adapter.ts
moncher-dev: Fixed in `efa724b` for both Claude environment paths. `stripTrackerSecrets` now returns the complete removed-name set, and both inherit and isolated paths pass that set to the shared constructor. The Claude spawn test covers an agent-visible `TARGET_REPOSITORY_URL` declared secret and verifies it stays absent.

packages/core/src/runtime/agent-child-env.ts
moncher-dev: Fixed in `efa724b`. `sources` is now required and the dead empty-source fallback is removed. The constructor also owns the optional `excludeNames` subtraction and has direct coverage for excluding both an agent-visible context name and a pinned host assignment.

packages/runtime-codex/src/runtime.test.ts
moncher-dev: Addressed through the shared-constructor contract and boundary regressions in `efa724b`. I kept the runtime assertions focused on public child-environment behavior: both runtime tests feed the same overlapping agent-visible/secret name and assert absence, while the core constructor test exactly asserts the complete assignment set. Since both runtimes call that constructor, a newly added assignment is defined and exhaustively tested once without duplicating the full set in both adapters.

packages/core/src/runtime/agent-child-env.ts
```diff
@@ -0,0 +1,23 @@
...
+    HOME: options.childHome,
+    USERPROFILE: options.childHome,
+    GH_CONFIG_DIR: join(options.childHome, "gh"),
+    DOCKER_CONFIG: join(options.childHome, ".docker"),
+  };
+
+  for (const name of options.excludeNames ?? []) {
+    delete assignments[name];
+  }
+
```
hojinzs: **P2 — `excludeNames` is applied to the pinned isolation assignments too, so a declared secret name can now switch off child-home isolation on Claude.**

The `excludeNames` subtraction is the right fix for the agent-visible context (that was the P1). But it runs over the whole `assignments` object, and `assignments` holds two categories that are not alike:

- the nine `AGENT_VISIBLE_SYMPHONY_CONTEXT_ENVIRONMENT_NAMES` — read from operator/repo-supplied sources, and legitimately subtractable, because the strip is what decides whether they may cross;
- `HOME` / `USERPROFILE` / `GH_CONFIG_DIR` / `DOCKER_CONFIG` — host-constructed constants that *are* the boundary. Nothing in the strip set can make it correct to withhold them; withholding them is what removes the isolation.

Because `removedEnvironmentNames` carries MCP-derived source names (`collectMcpSecretEnvironmentNames`, operator-controlled via `~/.gh-symphony/mcp.json` and repo-controlled under `trustRepoConfig`), an arbitrary name can land in that set — the same reachability you accepted for the P1 fix. `mcp-compose.ts:53-58` collects the *value* side of any server `env` entry whose key is a declared secret, so `{"env": {"GITHUB_TOKEN": "${HOME}"}}` puts `HOME` into the set.

Verified as a differential, driving `spawnTurn` through an injected `spawnImpl` with `SYMPHONY_TRACKER_SECRET_ENVIRONMENT_NAMES=["HOME"]`:

```
              inherit=true                          inherit=false
main    → HOME=/runtime/run-123/child-home     HOME=/runtime/run-123/child-home
efa724b → HOME=undefined                       HOME=undefined
```

On `main` the Claude paths assigned `env.HOME = childHome` *after* `stripTrackerSecrets`, so the strip could not reach it. Moving the assignment behind `excludeNames` hands that power to the strip set. The child then launches with no `HOME` at all, and consumers that fall back to the passwd entry rather than failing closed — Node's `os.homedir()`, which is how the Claude CLI resolves `~/.claude` — land on the *host* home instead of the isolated one. That is the inverse of what this file exists to guarantee.

Codex is not a regression here; I confirmed it already dropped `HOME` on `main` (the literal was assigned before the delete loop, `codex HOME=undefined` on both). So this is one fix that closes both.

Subtract from the context read only, and let the pinned values keep the last word:

```ts
export function buildAgentChildEnvironmentAssignments(options: {
  childHome: string;
  sources: ReadonlyArray<NodeJS.ProcessEnv | undefined>;
  excludeNames?: Iterable<string>;
}): NodeJS.ProcessEnv {
  const context = readAgentVisibleSymphonyContext(...options.sources);
  for (const name of options.excludeNames ?? []) {
    delete context[name];
  }

  return {
    ...context,
    HOME: options.childHome,
    USERPROFILE: options.childHome,
    GH_CONFIG_DIR: join(options.childHome, "gh"),
    DOCKER_CONFIG: join(options.childHome, ".docker"),
  };
}
```

The second case in `agent-child-env.test.ts` currently pins the behaviour I'm arguing against — it asserts `USERPROFILE` is removed when named in `excludeNames`. Worth inverting: a declared-secret collision on a pinned name must not be able to suppress the isolation value, while `TARGET_REPOSITORY_URL` in the same call still stays out.

---

packages/runtime-codex/src/runtime.ts
```diff
@@ -643,30 +643,35 @@ export function buildCodexRuntimePlan(
...
-      ...readAgentVisibleSymphonyContext(process.env),
       ...config.extraEnv,
       ...config.agentEnv,
       CODEX_PROJECT_ID: config.projectId,
       GITHUB_PROJECT_ID: config.githubProjectId ?? "",
       ...orchestratorRunEnv,
       ...agentEnv,
-      HOME: childHome,
-      GH_CONFIG_DIR: join(childHome, "gh"),
       CODEX_HOME: join(childHome, ".codex"),
```
hojinzs: **nit** — `CODEX_HOME` is now the odd one out.

It is a host-constructed child path just like the four the shared constructor owns, but it stays in the object literal, i.e. *before* the delete loop, while `HOME` / `GH_CONFIG_DIR` / `DOCKER_CONFIG` moved to *after* it. So the four are reapplied and `CODEX_HOME` is not — the same declared-secret collision that the P2 above describes silently produces a child with `CODEX_HOME` gone but the others pinned.

Not worth widening the shared constructor for (it is genuinely Codex-only, and doesn't belong in a set that Claude also consumes). Reassigning it alongside the `Object.assign` below would make the boundary uniform in one line:

```ts
Object.assign(plan.env, buildAgentChildEnvironmentAssignments({ ... }), {
  CODEX_HOME: join(childHome, ".codex"),
});
```

Take it or leave it — it only matters once the P2 is settled.

---

hojinzs (changes requested):
## Round 2 — one P2 left, everything from round 1 verified fixed

### Round 1 follow-up: all three confirmed closed, not just claimed

I re-ran each finding as a differential rather than reading the diff:

| Round 1 | Status at `efa724b` |
| --- | --- |
| P1 Codex — strip reintroduced by last-applied assignments | **Fixed.** `TARGET_REPOSITORY_URL` declared secret + agent-visible → absent from `plan.env`. |
| P1 Claude — same defect, both paths | **Fixed.** Absent on `inherit=true` and `inherit=false`. |
| P3 — dead `?? []` on optional `sources` | **Fixed.** Required, fallback gone. |
| nit — hand-asserted parity | Declined with reasoning on the thread. Accepted — see below. |

On the declined nit: the reasoning holds, and I checked the criterion directly rather than taking it on structure. A parity assertion projecting the full shared surface out of both runtimes for identical inputs **passes on `efa724b` and fails on `main`** — so "adding a value once appears in both" is now genuinely true. I'd still rather see that assertion in the tree than in my scratch dir, but it is not a blocker.

### Requirements vs. issue #870

- Shared constructor in `core`, both runtimes migrated, per-runtime construction deleted — done.
- One-edit property — holds; verified by the parity differential above.
- Out-of-scope items (inheritance allowlists, declarative non-secret channel, MCP composition) all correctly left alone. No overengineering: the constructor is 20 lines and takes exactly the two inputs the call sites need.

`USERPROFILE` is a scope addition beyond the issue text, but it's a real isolation hole (`DEFAULT_INHERITED_ENV_KEYS` was copying the host value into the Claude non-inherit path while `HOME` was pinned), it's the natural blast radius of this refactor, and the changeset calls it out. Fine to carry here.

### Smoke test

No preview environment on this PR, so black-box locally at the spawn-plan boundary — `buildCodexRuntimePlan` directly, and `buildClaudeSpawnEnv` driven through `spawnTurn` with an injected `spawnImpl`, on `efa724b` vs `main`:

- cross-runtime parity of the shared assignment surface — **passes on head, fails on main** ✅
- declared secret ∩ agent-visible absent — Codex ✅, Claude inherit ✅, Claude non-inherit ✅
- clone-URL userinfo sanitized when supplied through `extraEnv` — ✅ (an incidental improvement: on `main` the raw `extraEnv` spread landed *after* the context read on Codex, so userinfo survived; now it doesn't)
- pinned-assignment suppression — ❌ see the P2

Repo suites: `packages/core/src/runtime`, `packages/runtime-codex`, `packages/runtime-claude` → 227 passed. CI on `efa724b`: Test ✅, Container Smoke ✅.

### Blocking

**P2 — `packages/core/src/runtime/agent-child-env.ts`.** `excludeNames` subtracts from the pinned isolation assignments as well as the agent-visible context, so an operator/repo-reachable name colliding with `HOME` now suppresses child-home isolation on Claude where `main` pinned it. Differential and a three-line fix on the inline comment. Same reachability argument that carried the round-1 P1, so I'm holding it to the same bar.

One `nit` on `CODEX_HOME` asymmetry, take it or leave it.

Fix the P2 and I'll approve — I don't have further concerns beyond it.

---

packages/core/src/runtime/agent-child-env.ts
moncher-dev: Fixed in `0bb0fd6`. The shared constructor now removes excluded names only from the source-derived agent-visible context, then applies `HOME`, `USERPROFILE`, `GH_CONFIG_DIR`, and `DOCKER_CONFIG` as unconditional pinned isolation assignments. The core regression now excludes `TARGET_REPOSITORY_URL` and `USERPROFILE` together and verifies that only the source context is removed while `USERPROFILE` remains pinned. Both runtime collision tests pass.

packages/runtime-codex/src/runtime.ts
moncher-dev: Addressed in `0bb0fd6`. I accepted the consistency nit: `CODEX_HOME` is now assigned alongside the shared pinned values after the removal loop, so a declared-secret collision cannot suppress the Codex-specific isolated home. The Codex boundary test includes both `HOME` and `CODEX_HOME` in the declared-secret set and verifies both remain pinned.

packages/runtime-claude/src/adapter.ts
```diff
@@ -684,12 +687,20 @@ function buildClaudeSpawnEnv(options: {
...
+  );
   stripCredentialEnvironmentForAgentChild(env);
+  Object.assign(
+    env,
+    buildAgentChildEnvironmentAssignments({
+      childHome: options.childHome,
+      sources: [process.env, options.configEnv, options.inputEnv],
+      excludeNames: removedEnvironmentNames,
+    })
+  );
```
hojinzs, line 703: **nit** (non-blocking, do it in the follow-up if you prefer) — `HOME` and `USERPROFILE` in `DEFAULT_INHERITED_ENV_KEYS` (line 636) are now dead entries on this path.

Both are read out of `process.env` at line 682, survive `Object.assign(env, configEnv, inputEnv)` and `stripTrackerSecrets`, and are then unconditionally overwritten by the pinned assignments here. I checked that nothing between the two points reads either name — `stripTrackerSecrets` reads only `SYMPHONY_TRACKER_SECRET_ENVIRONMENT_NAMES` off `env` and `repositoryDir`/`projectDir` off arguments, and `stripCredentialEnvironmentForAgentChild` only deletes. So the inherit list can drop both without changing the child env.

Worth doing only because leaving them reads as though the host value still matters on the isolated path — which is exactly the misreading that produced the `USERPROFILE` hole this PR fixes. Zero behavioral risk either way; I confirmed the child env is byte-identical with them removed.

---

hojinzs (approved):
## Round 3 — approved. The round-2 P2 is genuinely closed, and the `main` differential is clean.

Reviewed at head `3469d2f` (the merge of current `main`). Round 3 of at most 6.

### Round 2 follow-up — verified by differential, not by reading the diff

I re-ran the exact probe that produced the round-2 P2, driving `buildCodexRuntimePlan` directly and `buildClaudeSpawnEnv` through `spawnTurn` with an injected `spawnImpl`, with `HOME`, `USERPROFILE`, `GH_CONFIG_DIR`, `DOCKER_CONFIG`, `CODEX_HOME` **and** `TARGET_REPOSITORY_URL` all declared secret at once — i.e. a collision on every pinned name simultaneously:

| Pinned name | Codex | Claude `inherit=false` | Claude `inherit=true` |
| --- | --- | --- | --- |
| `HOME` | child home ✅ | child home ✅ | child home ✅ |
| `USERPROFILE` | child home ✅ | child home ✅ | child home ✅ |
| `GH_CONFIG_DIR` | `…/gh` ✅ | `…/gh` ✅ | `…/gh` ✅ |
| `DOCKER_CONFIG` | `…/.docker` ✅ | `…/.docker` ✅ | `…/.docker` ✅ |
| `CODEX_HOME` | `…/.codex` ✅ | n/a | n/a |

The pinned surface is **byte-identical between the clean run and the all-names-collide run** on all three paths. Splitting the subtraction so `excludeNames` reaches only the source-derived context is the right cut, and the `CODEX_HOME` nit being taken means the Codex boundary is uniform rather than one-name-special.

Critically, the round-1 P1 did **not** regress while fixing this: `TARGET_REPOSITORY_URL`, agent-visible *and* declared secret, is still absent from all three child environments. The two invariants — "a declared secret cannot cross" and "the strip cannot switch off isolation" — now hold at the same time, which is the whole point.

Round-1 P3 (`custom-child-env.ts` as a third construction site) is discharged the documented way: the `§17.5` row now states the custom constructor is separate and does not share the agent-runtime assignment set. Good enough — the next reader won't mistake the shared constructor for universal.

### `main` differential — 12 changes, all in the safe direction

Same probe against `bcf80f6`. Every single difference is head being stricter; there is no cell where `main` is safer than head:

| | `main` | `3469d2f` |
| --- | --- | --- |
| Codex `USERPROFILE` | absent | child home |
| Claude `USERPROFILE` (both paths) | `/Users/operator` | child home |
| Codex `TARGET_REPOSITORY_CLONE_URL` | `https://x-access-token:ghs_SECRET@…` | sanitized |
| Codex pinned names under secret collision | **all dropped** (`HOME`, `GH_CONFIG_DIR`, `DOCKER_CONFIG`, `CODEX_HOME` = undefined) | all pinned |
| Claude `USERPROFILE` under secret collision | dropped | pinned |

Everything else — credential vars absent, `GIT_TERMINAL_PROMPT=0`, run-scoped `SYMPHONY_ISSUE_ID` winning over a stale host value — is unchanged from `main`.

Two notes on how to read that table, since they affect how the release is described rather than whether it ships:

- The clone-URL sanitization is **hardening, not a leak that was occurring** — I re-confirmed the round-1 reachability finding at this head. `runtime-codex/src/launcher.ts:44` already builds `extraEnv` through `readAgentVisibleSymphonyContext`, and `orchestrator/src/runtime-factory.ts` passes `context.env` whose clone URL is `deriveCloneUrl(repository.url)` with no userinfo. `buildCodexRuntimePlan` has no other production caller. I reproduce the tokenized value only by injecting it into `extraEnv` directly, which nothing does.
- The row that *was* live is the Claude `USERPROFILE` one, and the Codex collision row is a real robustness gain on a path `main` silently got wrong.

The changeset covers the `USERPROFILE` pinning and the secret-exclusion preservation. It doesn't mention the clone-URL hardening — given it isn't reachable, omitting it is defensible, so I'm not asking for a reword.

### Requirements vs. issue #870

| --- | --- |
| Both runtimes build explicit child assignments through the shared helper | ✅ |
| Per-runtime construction deleted | ✅ |
| Adding a host-constructed value requires one edit | ✅ |
| A test proves both runtimes pick it up | ✅ |

No over-engineering: the constructor is 22 lines with exactly the three inputs the call sites need. All three out-of-scope items on the issue (inheritance-allowlist consolidation, a declarative non-secret channel, MCP composition removal) are correctly left alone. The `USERPROFILE` scope addition remains the right call — it's the natural blast radius of the refactor and the changeset records it.

### Validation

`pnpm install --frozen-lockfile` → `pnpm build` → `pnpm lint` (0) → `pnpm typecheck` (0) → `pnpm test` (full monorepo, exit 0; orchestrator 362, cli 567). Targeted suites `packages/core/src/runtime` + `runtime-codex` + `runtime-claude`: 227 passed. CI on `3469d2f`: **Test ✅, Container Smoke ✅**. No preview environment applies to an in-process spawn-plan boundary, so the smoke test is the black-box probe described above rather than a deployed surface.

The regression tests in the tree pin what matters: the core test asserts the exact assignment set with a pinned-name collision, and both runtime tests feed an overlapping agent-visible/secret name and assert absence while the pinned values survive.

One `nit` inline about two now-dead entries in `DEFAULT_INHERITED_ENV_KEYS` — I verified removing them leaves the full child environment identical across all six configurations, so it's cosmetic. Take it or leave it; it is not worth another round.

Nothing blocking remains. Approving.

---

PR #884, opened 2026-09-05
fix(cli): report effective workflow timeouts

### LINKED ISSUE #882: fix(cli): workflow validate reports codex timeouts the runtime does not use
## Problem

`gh-symphony workflow validate` prints timeout values that are **not the ones the runtime uses**, on exactly the knobs an operator checks after a stall incident.

Against this repository's `WORKFLOW.md` the command reports:

```
codex.read_timeout_ms=5000
codex.stall_timeout_ms=300000
codex.turn_timeout_ms=3600000
```

while the file declares:

```yaml
runtime:
  timeouts:
    read_timeout_ms: 30000
    turn_timeout_ms: 3600000
    stall_timeout_ms: 900000
```

The reported `read_timeout_ms` is 6× smaller and `stall_timeout_ms` 3× smaller than what actually runs. `turn_timeout_ms` matches only by coincidence — the declared value equals the default.

## Cause

There are two front-matter locations for the same three knobs, and `validate` reports the one this file does not use.

- `packages/core/src/workflow/parser.ts:295-312` parses a top-level `codex:` block into `workflow.codex`, defaulting each field.
- `packages/core/src/workflow/parser.ts:1138-1152` parses `runtime.timeouts` into `workflow.runtime.timeouts`.
- `packages/core/src/workflow/config.ts:285-289` — `resolveWorkflowRuntimeTimeouts` returns `workflow.runtime?.timeouts ?? workflow.codex`, so `runtime.timeouts` **wins** wherever it is present.
- `packages/cli/src/commands/workflow.ts:881` builds the report summary from `workflow.codex.*` unconditionally, and `:923` prints it.

So the runtime is correct — `packages/orchestrator/src/service.ts:3333` resolves through `resolveWorkflowRuntimeTimeouts` and forwards the result as `SYMPHONY_READ_TIMEOUT_MS` / `SYMPHONY_TURN_TIMEOUT_MS` at `:3406-3407`. Only the report is wrong.

## Why it matters

The command's whole purpose is to answer "what will actually run". These particular values are what an operator reaches for after agent turns time out — on 2026-09-04 this host lost five consecutive worker runs to timeouts, and `read_timeout_ms` was raised to 30000 in response. An operator reading `validate` today would conclude that raise never landed, and could re-apply a fix that is already in place or chase a misconfiguration that does not exist. The operator very nearly did exactly that today before tracing the parser.

The failure is silent: `validate` exits successfully and prints a plausible number.

## Acceptance criteria

- [ ] `workflow validate` reports the timeout values the runtime will actually use — that is, the result of `resolveWorkflowRuntimeTimeouts`, not `workflow.codex` unconditionally.
- [ ] When a file declares `runtime.timeouts`, the reported values equal the declared ones. Verified against this repository's own `WORKFLOW.md`, which currently reproduces the defect.
- [ ] When a file declares only the legacy `codex:` block, the reported values still equal that block's values — no regression for that shape.
- [ ] When a file declares neither, the reported values are the documented defaults.
- [ ] A test pins the precedence, and fails if the report is rewired back to `workflow.codex` unconditionally.
- [ ] The label the report prints does not claim `codex.*` when the value came from `runtime.timeouts`, so the source is unambiguous to a reader.

## Out of scope

Removing either front-matter location, changing the precedence in `resolveWorkflowRuntimeTimeouts`, and any change to the timeout values themselves. This issue only makes the report tell the truth.

## Evidence

Traced 2026-09-05 by the operator while verifying the daemon after `policy(land)` (#881, `bcf80f6c`) merged. `workflow validate` was run from `~/moncher-stack/github-symphony` against the symlinked `WORKFLOW.md`.

moncher-dev:
## Issues

- Closes #882

## Summary

- TL;DR: make `workflow validate` report the effective timeout values used by the runtime.
- Resolve precedence through the shared core helper, expose a stable three-field JSON object, and identify the effective configuration source.
- Cover explicit runtime settings, legacy `codex` fallback, parser-materialized runtime defaults, and documented defaults.

## Change-point diagram

- `WORKFLOW.md` → core parser → `resolveWorkflowRuntimeTimeouts` → narrowed CLI validation summary → text/JSON report

## Start here

- `packages/cli/src/commands/workflow.ts:847` — resolves and narrows effective runtime timeouts for the report.
- `packages/cli/src/commands/workflow.test.ts:125` — pins precedence, all four front-matter shapes, sources, and the JSON key set.

- Text output reports the values the runtime uses under `runtime.timeouts.*` and identifies `runtime.timeouts` versus `codex/defaults`.
- JSON output moves effective values from the misleading `summary.codex.*TimeoutMs` fields to `summary.runtimeTimeouts.*`; no compatibility aliases preserve the incorrect fields.
- Runtime execution and timeout precedence are unchanged. The separate parser/resolver precedence defect is tracked in #888.

## Validation

- `pnpm exec prettier --check packages/cli/src/commands/workflow.ts packages/cli/src/commands/workflow.test.ts packages/cli/README.md .changeset/tidy-pandas-report.md` — pass
- `pnpm exec vitest run packages/cli/src/commands/workflow.test.ts` — pass (23 tests)
- Repository `WORKFLOW.md` smoke validation — pass (`30000`, `900000`, `3600000`, source `runtime.timeouts`)
- `pnpm lint` — pass
- `pnpm test` — pass
- `pnpm typecheck` — pass
- `pnpm build` — pass
- Docker E2E — not applicable: CLI reporting only; runtime/integration behavior is unchanged

## Changeset

- `.changeset/tidy-pandas-report.md` (patch)

## Risks & rollback

- Low risk: the change reuses the runtime resolver and explicitly narrows its result. JSON consumers must migrate to the documented truthful keys; rollback is commits `2a51e85` and `fc43c5b`.

## Changed files

- `packages/cli/src/commands/workflow.ts` — report narrowed effective timeouts and their source.
- `packages/cli/src/commands/workflow.test.ts` — cover precedence, fallback, defaults, mixed runtime configuration, labels, and JSON shape.
- `README.md` — document effective validation output.
- `packages/cli/README.md` — document timeout semantics and the JSON key migration.
- `docs/configuration.md` — document timeout precedence and worker parity.
- `.changeset/tidy-pandas-report.md` — release-note the CLI behavior fix.

- [ ] None

## Security

### DIFF
--- a/.changeset/tidy-pandas-report.md
+++ b/.changeset/tidy-pandas-report.md
@@ -1,0 +1,5 @@
+---
+"@gh-symphony/cli": patch
+---
+
+Report the effective runtime timeout values and their configuration source from `workflow validate`, including a stable three-field JSON timeout object (#882).
--- a/packages/cli/src/commands/workflow.test.ts
+++ b/packages/cli/src/commands/workflow.test.ts
@@ -138,6 +138,7 @@ codex:
   stall_timeout_ms: 60000
   turn_timeout_ms: 120000`,
       expected: [30000, 900000, 1800000],
+      expectedSource: "runtime.timeouts",
     },
     {
       name: "legacy codex timeout fallback",
@@ -146,15 +147,30 @@ codex:
   stall_timeout_ms: 60000
   turn_timeout_ms: 120000`,
       expected: [7000, 60000, 120000],
+      expectedSource: "codex/defaults",
+    },
+    {
+      name: "runtime defaults over legacy codex timeouts",
+      frontMatter: `runtime:
+  kind: codex-app-server
+  command: codex
+  args: [app-server]
+codex:
+  read_timeout_ms: 7000
+  stall_timeout_ms: 60000
+  turn_timeout_ms: 120000`,
+      expected: [5000, 300000, 3600000],
+      expectedSource: "runtime.timeouts",
     },
     {
       name: "documented timeout defaults",
       frontMatter: "codex:\n  command: codex app-server",
       expected: [5000, 300000, 3600000],
+      expectedSource: "codex/defaults",
     },
   ])(
     "reports $name as effective runtime timeouts",
-    async ({ frontMatter, expected }) => {
+    async ({ frontMatter, expected, expectedSource }) => {
       const root = await mkdtemp(join(tmpdir(), "workflow-validate-timeouts-"));
       const workflowPath = join(root, "WORKFLOW.md");
       const stdout = captureWrites(process.stdout);
@@ -186,6 +202,32 @@ codex:
         `runtime.timeouts.turn_timeout_ms=${expected[2]}`
       );
       expect(stdout.output()).not.toContain("codex.read_timeout_ms=");
+      expect(stdout.output()).toContain(`(source: ${expectedSource})`);
+
+      const jsonStdout = captureWrites(process.stdout);
+      try {
+        await workflowCommand(["validate", "--file", workflowPath], {
+          configDir: root,
+          verbose: false,
+          json: true,
+          noColor: false,
+        });
+      } finally {
+        jsonStdout.restore();
+      }
+
+      const report = JSON.parse(jsonStdout.output()) as {
+        summary: {
+          runtimeTimeouts: Record<string, number>;
+          runtimeTimeoutSource: string;
+        };
+      };
+      expect(report.summary.runtimeTimeouts).toEqual({
+        readTimeoutMs: expected[0],
+        stallTimeoutMs: expected[1],
+        turnTimeoutMs: expected[2],
+      });
+      expect(report.summary.runtimeTimeoutSource).toBe(expectedSource);
     }
   );
 
--- a/packages/cli/src/commands/workflow.test.ts
+++ b/packages/cli/src/commands/workflow.test.ts
@@ -122,6 +122,73 @@ describe("workflow command handler", () => {
     expect(stdout.output()).toContain("active_states=Ready, In progress");
   });
 
+  it.each([
+    {
+      name: "runtime timeout precedence",
+      frontMatter: `runtime:
+  kind: codex-app-server
+  command: codex
+  args: [app-server]
+  timeouts:
+    read_timeout_ms: 30000
+    stall_timeout_ms: 900000
+    turn_timeout_ms: 1800000
+codex:
+  read_timeout_ms: 7000
+  stall_timeout_ms: 60000
+  turn_timeout_ms: 120000`,
+      expected: [30000, 900000, 1800000],
+    },
+    {
+      name: "legacy codex timeout fallback",
+      frontMatter: `codex:
+  read_timeout_ms: 7000
+  stall_timeout_ms: 60000
+  turn_timeout_ms: 120000`,
+      expected: [7000, 60000, 120000],
+    },
+    {
+      name: "documented timeout defaults",
+      frontMatter: "codex:\n  command: codex app-server",
+      expected: [5000, 300000, 3600000],
+    },
+  ])(
+    "reports $name as effective runtime timeouts",
+    async ({ frontMatter, expected }) => {
+      const root = await mkdtemp(join(tmpdir(), "workflow-validate-timeouts-"));
+      const workflowPath = join(root, "WORKFLOW.md");
+      const stdout = captureWrites(process.stdout);
+
+      await writeFile(
+        workflowPath,
+        `---\ntracker:\n  kind: github-project\n${frontMatter}\n---\nPrompt {{ issue.identifier }}\n`,
+        "utf8"
+      );
+
+      try {
+        await workflowCommand(["validate", "--file", workflowPath], {
+          configDir: root,
+          verbose: false,
+          json: false,
+          noColor: false,
+        });
+      } finally {
+        stdout.restore();
+      }
+
+      expect(stdout.output()).toContain(
+        `runtime.timeouts.read_timeout_ms=${expected[0]}`
+      );
+      expect(stdout.output()).toContain(
+        `runtime.timeouts.stall_timeout_ms=${expected[1]}`
+      );
+      expect(stdout.output()).toContain(
+        `runtime.timeouts.turn_timeout_ms=${expected[2]}`
+      );
+      expect(stdout.output()).not.toContain("codex.read_timeout_ms=");
+    }
+  );
+
   it("prints a typed error when a removed flat tracker key is configured", async () => {
     const root = await mkdtemp(join(tmpdir(), "workflow-validate-priority-"));
     const workflowPath = join(root, "WORKFLOW.md");
--- a/packages/cli/src/commands/workflow.ts
+++ b/packages/cli/src/commands/workflow.ts
@@ -112,6 +112,7 @@ type WorkflowValidationReport = {
       stallTimeoutMs: number;
       turnTimeoutMs: number;
     };
+    runtimeTimeoutSource: "runtime.timeouts" | "codex/defaults";
     hooks: {
       afterCreate: string | null;
       beforeRun: string | null;
@@ -846,6 +847,7 @@ function validateWorkflow(
         return "pass" as const;
       })()
     : ("skip" as const);
+  const effectiveTimeouts = resolveWorkflowRuntimeTimeouts(workflow);
 
   return {
     ok: true,
@@ -882,7 +884,14 @@ function validateWorkflow(
         threadSandbox: workflow.codex.threadSandbox,
         turnSandboxPolicy: workflow.codex.turnSandboxPolicy,
       },
-      runtimeTimeouts: resolveWorkflowRuntimeTimeouts(workflow),
+      runtimeTimeouts: {
+        readTimeoutMs: effectiveTimeouts.readTimeoutMs,
+        stallTimeoutMs: effectiveTimeouts.stallTimeoutMs,
+        turnTimeoutMs: effectiveTimeouts.turnTimeoutMs,
+      },
+      runtimeTimeoutSource: workflow.runtime
+        ? "runtime.timeouts"
+        : "codex/defaults",
       hooks: {
         afterCreate: workflow.hooks.afterCreate,
         beforeRun: workflow.hooks.beforeRun,
@@ -921,9 +930,9 @@ Runtime
   codex.approval_policy=${report.summary.codex.approvalPolicy ?? "unset"}
   codex.thread_sandbox=${report.summary.codex.threadSandbox ?? "unset"}
   codex.turn_sandbox_policy=${report.summary.codex.turnSandboxPolicy ?? "unset"}
-  runtime.timeouts.read_timeout_ms=${report.summary.runtimeTimeouts.readTimeoutMs}
-  runtime.timeouts.stall_timeout_ms=${report.summary.runtimeTimeouts.stallTimeoutMs}
-  runtime.timeouts.turn_timeout_ms=${report.summary.runtimeTimeouts.turnTimeoutMs}
+  runtime.timeouts.read_timeout_ms=${report.summary.runtimeTimeouts.readTimeoutMs} (source: ${report.summary.runtimeTimeoutSource})
+  runtime.timeouts.stall_timeout_ms=${report.summary.runtimeTimeouts.stallTimeoutMs} (source: ${report.summary.runtimeTimeoutSource})
+  runtime.timeouts.turn_timeout_ms=${report.summary.runtimeTimeouts.turnTimeoutMs} (source: ${report.summary.runtimeTimeoutSource})
 
 Hooks
   after_create=${report.summary.hooks.afterCreate ?? "unset"}
--- a/packages/cli/src/commands/workflow.ts
+++ b/packages/cli/src/commands/workflow.ts
@@ -7,6 +7,7 @@ import {
   WorkflowValidationError,
   renderPrompt,
   resolveWorkflowExecutionPhase,
+  resolveWorkflowRuntimeTimeouts,
   type TrackedIssue,
 } from "@gh-symphony/core";
 import {
@@ -105,6 +106,8 @@ type WorkflowValidationReport = {
       approvalPolicy: string | null;
       threadSandbox: string | null;
       turnSandboxPolicy: string | null;
+    };
+    runtimeTimeouts: {
       readTimeoutMs: number;
       stallTimeoutMs: number;
       turnTimeoutMs: number;
@@ -878,10 +881,8 @@ function validateWorkflow(
         approvalPolicy: workflow.codex.approvalPolicy,
         threadSandbox: workflow.codex.threadSandbox,
         turnSandboxPolicy: workflow.codex.turnSandboxPolicy,
-        readTimeoutMs: workflow.codex.readTimeoutMs,
-        stallTimeoutMs: workflow.codex.stallTimeoutMs,
-        turnTimeoutMs: workflow.codex.turnTimeoutMs,
       },
+      runtimeTimeouts: resolveWorkflowRuntimeTimeouts(workflow),
       hooks: {
         afterCreate: workflow.hooks.afterCreate,
         beforeRun: workflow.hooks.beforeRun,
@@ -920,9 +921,9 @@ Runtime
   codex.approval_policy=${report.summary.codex.approvalPolicy ?? "unset"}
   codex.thread_sandbox=${report.summary.codex.threadSandbox ?? "unset"}
   codex.turn_sandbox_policy=${report.summary.codex.turnSandboxPolicy ?? "unset"}
-  codex.read_timeout_ms=${report.summary.codex.readTimeoutMs}
-  codex.stall_timeout_ms=${report.summary.codex.stallTimeoutMs}
-  codex.turn_timeout_ms=${report.summary.codex.turnTimeoutMs}
+  runtime.timeouts.read_timeout_ms=${report.summary.runtimeTimeouts.readTimeoutMs}
+  runtime.timeouts.stall_timeout_ms=${report.summary.runtimeTimeouts.stallTimeoutMs}
+  runtime.timeouts.turn_timeout_ms=${report.summary.runtimeTimeouts.turnTimeoutMs}
 
 Hooks
   after_create=${report.summary.hooks.afterCreate ?? "unset"}
--- a/README.md
+++ b/README.md
@@ -801,7 +801,7 @@ gh-symphony workflow init --non-interactive --project PVT_xxx --output WORKFLOW.
 gh-symphony workflow init --non-interactive --project PVT_xxx --dry-run
 ```
 
-`gh-symphony workflow validate` parses the target file, strictly renders the prompt body and continuation guidance with canonical sample variables, and prints a compact runtime/lifecycle summary.
+`gh-symphony workflow validate` parses the target file, strictly renders the prompt body and continuation guidance with canonical sample variables, and prints a compact runtime/lifecycle summary. Its `runtime.timeouts.*` values are the effective runtime settings: `runtime.timeouts` takes precedence over the legacy `codex.*_timeout_ms` fields, with documented defaults used when neither is configured.
 
 `gh-symphony workflow preview --issue owner/repo#123` is the fastest validation step after `workflow init`: it resolves the active managed project (or `--project-id`) and renders the exact worker prompt from the live GitHub Project issue. Linear workflows can preview a single issue with `gh-symphony workflow preview ENG-123`, which routes through the configured Linear tracker adapter and `LINEAR_API_KEY`. Keep `--sample <path-to-json>` for fixture-based debugging, and use `--attempt <n>` to inspect retry prompts before changing policy files.
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -83,6 +83,11 @@ negative, that threshold is disabled, but the orchestrator still applies a hard
 30-minute elapsed-run fallback to prevent an indefinitely stuck worker. This is
 an intentional implementation-defined safety limit.
 
+`runtime.timeouts` is the preferred timeout configuration and takes precedence
+over the legacy `codex.*_timeout_ms` fields. `gh-symphony workflow validate`
+prints the effective values under `runtime.timeouts.*`, matching the values the
+orchestrator injects into a worker.
+
 Hooks are opt-in repository-local extensions, not shell snippets. Each hook
 value must be a path to an executable script; shell syntax and inline commands
 are rejected. Set `SYMPHONY_ALLOW_WORKFLOW_HOOKS=1` (or `true`) in the host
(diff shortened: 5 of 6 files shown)


moncher-dev (commented):
@hojinzs ready for your review — verified at head `fc43c5bd`. This one I exercised for real rather than reading; all three checks pass.

## Build

`pnpm install --frozen-lockfile && pnpm build && pnpm test` in a clean worktree of this head. One failure, `core > executeWorkspaceHook > drains large stdout without waiting for the timeout` (`expected 'timeout' to be 'success'`) — a load-sensitive timing test on a host at load 21, in a package this PR does not touch (the diff is `packages/cli` only). Re-run in isolation: passes in 543ms. CI is green on both jobs. Flake, not this change.

`packages/cli/src/commands/workflow.test.ts`: 22 passed.

## 1. The reported bug is fixed

Ran the built CLI from both branches against the real `~/moncher-stack/github-symphony/WORKFLOW.md`, which declares `runtime.timeouts.read_timeout_ms: 30000` / `stall_timeout_ms: 900000`:

```
main    codex.read_timeout_ms=5000              <- wrong
        codex.stall_timeout_ms=300000           <- wrong
PR884   runtime.timeouts.read_timeout_ms=30000  <- matches WORKFLOW.md
        runtime.timeouts.stall_timeout_ms=900000
```

The runtime was always using 30000/900000; only the report was wrong. It now agrees.

## 2. Legacy shape still reports its own values

Built a fixture with a top-level `codex:` block only (7777 / 88888 / 999999) and no `runtime:` block:

```
main    codex.read_timeout_ms=7777
PR884   runtime.timeouts.read_timeout_ms=7777
```

Preserved.

## 3. The label no longer lies about its source

`codex.read_timeout_ms=` is gone for values that came from `runtime.timeouts`. The label now names the effective slot.

---

## Two things for you to decide

**a) No changeset.** CLAUDE.md: "Behavior-changing PRs must add a changeset (`.changeset/*.md`, package `@gh-symphony/cli`)." This changes released `workflow validate` output and adds none — #886 added one for a comparable change. As it stands this fix either rides an unrelated release unattributed, or fails to trigger one. The Codex reviewer flagged the same. Your call given #883 is open.

**b) `--json` schema change.** `summary.codex.{readTimeoutMs,stallTimeoutMs,turnTimeoutMs}` are removed and replaced by `summary.runtimeTimeouts.*`. Confirmed against `--json` output. Codex asked for compatibility aliases; I'd argue against them — the old keys held a value the runtime did not use, so a consumer reading a now-missing key fails loudly instead of silently trusting a wrong number. But it is a breaking change to a released command's output and worth being deliberate about.

## Separate latent bug found while testing (not this PR's)

`resolveWorkflowRuntimeTimeouts` returns `runtime?.timeouts ?? codex`, but the parser materializes `runtime.timeouts` with defaults whenever a `runtime:` block is present. So the `?? codex` fallback is dead for any workflow that has a `runtime:` block without a `timeouts:` sub-block:

```
front-matter: runtime: {kind: codex-app-server, ...}  +  codex: {read_timeout_ms: 7777, ...}
parsed runtime.timeouts -> {readTimeoutMs: 5000, stallTimeoutMs: 300000}   # defaults
resolved                -> 5000 / 300000                                    # 7777 silently discarded
```

The orchestrator forwards this resolver, so the *runtime* also ignores the declared 7777 — the report is right and the behavior is wrong. Only reachable mid-migration (both blocks present), but it fails silently. Filing it separately.

packages/cli/src/commands/workflow.ts
```diff
@@ -878,10 +881,8 @@ function validateWorkflow(
         approvalPolicy: workflow.codex.approvalPolicy,
         threadSandbox: workflow.codex.threadSandbox,
         turnSandboxPolicy: workflow.codex.turnSandboxPolicy,
-        readTimeoutMs: workflow.codex.readTimeoutMs,
-        stallTimeoutMs: workflow.codex.stallTimeoutMs,
-        turnTimeoutMs: workflow.codex.turnTimeoutMs,
       },
+      runtimeTimeouts: resolveWorkflowRuntimeTimeouts(workflow),
```
hojinzs: **P1 — this assignment leaks the whole `codex` block into `--json`, and the emitted schema now depends on front-matter shape.**

`resolveWorkflowRuntimeTimeouts` is `workflow.runtime?.timeouts ?? workflow.codex` (`packages/core/src/workflow/config.ts:285-289`). On the fallback branch it returns the *entire* `workflow.codex` object, not a three-field timeouts object. `WorkflowValidationReport` declares only three fields here, but TypeScript's structural typing accepts the wider object and `runValidate` serializes `report` verbatim — so the extra keys survive to stdout.

Verified against the built CLI at this head:

```
# runtime.timeouts declared  -> summary.runtimeTimeouts
{ "turnTimeoutMs": 3600000, "readTimeoutMs": 30000, "stallTimeoutMs": 900000 }

# legacy codex: only         -> summary.runtimeTimeouts
{ "command": "codex app-server", "approvalPolicy": null, "threadSandbox": null,
  "turnSandboxPolicy": null, "turnTimeoutMs": 999999, "readTimeoutMs": 7777,
  "stallTimeoutMs": 88888 }
```

Three keys or seven, depending on whether the file has a `runtime:` block. `command` / `approvalPolicy` / `threadSandbox` / `turnSandboxPolicy` are duplicated out of `summary.codex` into a field named `runtimeTimeouts`, which is both wrong-by-name and an unstable contract for the `--json` consumers the schema discussion below is about. The text report is unaffected — it reads the three fields by name — which is why the existing tests pass.

One-line fix, keeping the shared resolver:

```ts
const effectiveTimeouts = resolveWorkflowRuntimeTimeouts(workflow);
// ...
runtimeTimeouts: {
  readTimeoutMs: effectiveTimeouts.readTimeoutMs,
  stallTimeoutMs: effectiveTimeouts.stallTimeoutMs,
  turnTimeoutMs: effectiveTimeouts.turnTimeoutMs,
},
```

Worth a `--json` assertion in the new `it.each` so the key set is pinned, not just the stdout lines.

---

packages/cli/src/commands/workflow.ts
```diff
@@ -920,9 +921,9 @@ Runtime
   codex.approval_policy=${report.summary.codex.approvalPolicy ?? "unset"}
   codex.thread_sandbox=${report.summary.codex.threadSandbox ?? "unset"}
   codex.turn_sandbox_policy=${report.summary.codex.turnSandboxPolicy ?? "unset"}
-  codex.read_timeout_ms=${report.summary.codex.readTimeoutMs}
-  codex.stall_timeout_ms=${report.summary.codex.stallTimeoutMs}
-  codex.turn_timeout_ms=${report.summary.codex.turnTimeoutMs}
+  runtime.timeouts.read_timeout_ms=${report.summary.runtimeTimeouts.readTimeoutMs}
```
hojinzs: **P3 — the label now misattributes in the opposite direction.**

Acceptance criterion #6 asks that the label not claim `codex.*` for a value that came from `runtime.timeouts`, "so the source is unambiguous to a reader". This satisfies the letter of it, but inverts the ambiguity rather than removing it. Verified with a legacy-only fixture (no `runtime:` block, `codex.read_timeout_ms: 7777`):

```
runtime.timeouts.read_timeout_ms=7777
```

The value is declared under `codex:` and reported under `runtime.timeouts.*`. An operator on a half-migrated file who greps `WORKFLOW.md` for `runtime.timeouts` finds nothing and is back in the same "where did this number come from" loop that #882 describes — just from the other side.

`runtime.timeouts.*` is the right *effective slot* name, so I'd keep it and add the source rather than rename:

```
runtime.timeouts.read_timeout_ms=7777 (source: codex)
```

Cheap, and it makes the migration state visible at exactly the moment the operator is looking. Non-blocking — say so and I'll drop it if you'd rather keep the report a flat key=value grid.

---

packages/cli/src/commands/workflow.test.ts
```diff
@@ -122,6 +122,73 @@ describe("workflow command handler", () => {
...
+      frontMatter: `codex:
+  read_timeout_ms: 7000
+  stall_timeout_ms: 60000
+  turn_timeout_ms: 120000`,
+      expected: [7000, 60000, 120000],
+    },
+    {
+      name: "documented timeout defaults",
+      frontMatter: "codex:\n  command: codex app-server",
+      expected: [5000, 300000, 3600000],
```
hojinzs, line 168: **P3 — the fourth shape is the one most likely to surprise, and it is untested.**

The three cases here cover `runtime.timeouts` wins / legacy-only / neither. The missing shape is `runtime:` present **without** a `timeouts:` sub-block, alongside a `codex:` block — because `parser.ts:1143-1151` materializes `runtime.timeouts` with defaults whenever any `runtime:` block exists, so `?? workflow.codex` never fires for it. Verified against the built CLI:

```yaml
runtime: { kind: codex-app-server, command: codex, args: [app-server] }
codex:   { read_timeout_ms: 7777, stall_timeout_ms: 88888, turn_timeout_ms: 999999 }
```
```
runtime.timeouts.read_timeout_ms=5000
runtime.timeouts.stall_timeout_ms=300000
```

To be clear, **this PR is right here** — the orchestrator resolves through the same helper (`service.ts:3333`, `:5741`), so 5000/300000 is genuinely what runs, and the report is telling the truth. Changing that precedence is explicitly out of scope for #882, and filing the underlying core defect separately is the right call.

But that makes it worth a fourth `it.each` entry pinning `expected: [5000, 300000, 3600000]` for this shape. It documents that the mixed shape resolves to defaults rather than to `codex`, and when the core resolver is eventually fixed, this test flips and forces a deliberate decision instead of silently following.

---

hojinzs (changes requested):
Reviewed at head `fc43c5bd`. Built the CLI and exercised `workflow validate` against four front-matter shapes rather than reading the diff alone. **All six acceptance criteria of #882 are met, and the reported defect is genuinely fixed** — but the new `runtimeTimeouts` field leaks the whole `codex` block into `--json`, so I'm holding on that one line.

## Smoke test — `packages/cli/dist/index.js workflow validate`

| shape | reported | verdict |
|---|---|---|
| this repo's `WORKFLOW.md` (`runtime.timeouts` declared 30000/900000/3600000) | `30000` / `900000` / `3600000` | matches the declared values — **AC 1, 2 met**, the #882 defect is gone |
| legacy `codex:` only (7777/88888/999999) | `7777` / `88888` / `999999` | no regression — **AC 3 met** |
| neither block | `5000` / `300000` / `3600000` | documented defaults — **AC 4 met** |
| `runtime:` without `timeouts:` + `codex:` values | `5000` / `300000` / `3600000` | truthful; see the test-file comment |

AC 5 (a test pins precedence) and AC 6 (the label no longer says `codex.*`) are both satisfied — the `it.each` block would fail if the report were rewired back to `workflow.codex`, and `not.toContain("codex.read_timeout_ms=")` pins the label.

Scope is clean: five files, no speculative refactor, the shared resolver reused rather than re-implemented, and the tempting-but-out-of-scope precedence change correctly left alone. `packages/cli/src/commands/workflow.test.ts` 22/22 pass; `pnpm lint`, `pnpm typecheck`, `pnpm build` pass; CI green on both jobs. (`pnpm format` flags 72 files here, all untouched by this PR — a trailing-comma drift in my container, not yours. This PR's five files pass `prettier --check`.)

## Findings

- **P1 — `runtimeTimeouts: resolveWorkflowRuntimeTimeouts(workflow)`** (`workflow.ts:885`). On the fallback branch the resolver returns the whole `workflow.codex` object; the declared three-field type doesn't survive to runtime, and `--json` serializes it verbatim. `summary.runtimeTimeouts` has 3 keys with a `runtime:` block and 7 without — including `command` and `approvalPolicy` duplicated under a field named "timeouts". Inline comment has the evidence and a one-line fix. This is the only thing blocking.
- **P3** — the label misattributes in the reverse direction for legacy-only files (`runtime.timeouts.read_timeout_ms=7777` for a value declared under `codex:`). Suggested a `(source: codex)` suffix. Non-blocking.
- **P3** — the mixed shape (row 4 above) has no test. Suggested a fourth `it.each` entry.

## On the two open questions

**a) Changeset — I'd add one.** Your reading is defensible: the land gate is "changeset present *if labeled*" (`.codex/skills/land/SKILL.md:131`) and #882 carries no label, so nothing will block you. But that's the enforcement threshold, not the authoring rule — CLAUDE.md says "Behavior-changing PRs must add a changeset" and CONTRIBUTING.md says "when the change affects released package behavior." This changes released `workflow validate` output in two ways. With `fuzzy-cats-guard.md` pending, this fix ships in that release with no changelog line, and the operator who hit #882 has no way to tell which version fixed it. A patch changeset costs one file.

**b) `--json` compatibility aliases — I agree with you, no aliases.** The old keys held a value the runtime did not use; a consumer reading a now-missing key fails loudly instead of silently trusting a wrong number, which is the whole point of #882. Codex's concern is real but the alias is the worse resolution. What's missing is that the break is undocumented: all three doc edits describe the text output only, and none of README.md, `packages/cli/README.md`, or `docs/configuration.md` mentions that `summary.codex.{read,stall,turn}TimeoutMs` moved to `summary.runtimeTimeouts.*`. Put it in the changeset from (a) and a line in `packages/cli/README.md`. Fixing the P1 first makes that documented schema stable rather than shape-dependent.

## The latent core bug you found

Confirmed independently — `parser.ts:1143-1151` materializes `runtime.timeouts` with defaults for any `runtime:` block, so `?? workflow.codex` in `resolveWorkflowRuntimeTimeouts` is unreachable whenever a `runtime:` block exists, and a declared `codex.read_timeout_ms` is silently discarded by the *runtime*, not just the report. Right call keeping it out of this PR — #882 puts the precedence explicitly out of scope. Please link the follow-up issue here when you file it.

Fix the P1 and take or wave off the two P3s and I'll approve on the next pass.

---

packages/cli/src/commands/workflow.ts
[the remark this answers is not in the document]
moncher-dev: Resolved by documenting the intentional JSON migration in `packages/cli/README.md` and the PR body. We are not retaining compatibility aliases: the removed `summary.codex.*TimeoutMs` fields could expose values the runtime did not use, so preserving them would keep the silent-misconfiguration failure that #882 fixes. Effective values now live only under `summary.runtimeTimeouts.*`.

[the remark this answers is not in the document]
moncher-dev: Resolved in `2a51e85`: added `.changeset/tidy-pandas-report.md` as a patch changeset for `@gh-symphony/cli`, describing the effective timeout report and stable JSON object.

moncher-dev: Resolved in `2a51e85`: `resolveWorkflowRuntimeTimeouts` is evaluated once, then only `readTimeoutMs`, `stallTimeoutMs`, and `turnTimeoutMs` are copied into `summary.runtimeTimeouts`. The parameterized test now parses `--json` for all four shapes and asserts exact object equality, so unrelated legacy `codex` keys cannot leak.

packages/cli/src/commands/workflow.ts
moncher-dev: Accepted and resolved in `2a51e85`: text output retains the effective `runtime.timeouts.*` slot names and appends an explicit source. Legacy/default fallback reports `(source: codex/defaults)`; runtime configurations report `(source: runtime.timeouts)`. The JSON report exposes the same provenance as `runtimeTimeoutSource`.

packages/cli/src/commands/workflow.test.ts
moncher-dev: Accepted and resolved in `2a51e85`: added the mixed `runtime`-without-`timeouts` plus legacy `codex` case. It pins the current runtime behavior at `5000 / 300000 / 3600000`, asserts `source: runtime.timeouts`, and verifies the exact three-key JSON object. The underlying out-of-scope resolver behavior remains tracked in #888.

packages/cli/src/commands/workflow.ts
```diff
@@ -920,9 +930,9 @@ Runtime
   codex.approval_policy=${report.summary.codex.approvalPolicy ?? "unset"}
   codex.thread_sandbox=${report.summary.codex.threadSandbox ?? "unset"}
   codex.turn_sandbox_policy=${report.summary.codex.turnSandboxPolicy ?? "unset"}
-  codex.read_timeout_ms=${report.summary.codex.readTimeoutMs}
-  codex.stall_timeout_ms=${report.summary.codex.stallTimeoutMs}
-  codex.turn_timeout_ms=${report.summary.codex.turnTimeoutMs}
+  runtime.timeouts.read_timeout_ms=${report.summary.runtimeTimeouts.readTimeoutMs} (source: ${report.summary.runtimeTimeoutSource})
+  runtime.timeouts.stall_timeout_ms=${report.summary.runtimeTimeouts.stallTimeoutMs} (source: ${report.summary.runtimeTimeoutSource})
+  runtime.timeouts.turn_timeout_ms=${report.summary.runtimeTimeouts.turnTimeoutMs} (source: ${report.summary.runtimeTimeoutSource})
```
hojinzs, line 935: **P3 — `codex/defaults` conflates two different answers, and per-report granularity can't express a partially-declared block.**

The source label is one value for all three lines, so it can only say "not `runtime.timeouts`". Verified against the built CLI at this head with a `codex:` block declaring *only* `read_timeout_ms`:

```yaml
codex:
  read_timeout_ms: 7777
```
```
runtime.timeouts.read_timeout_ms=7777 (source: codex/defaults)
runtime.timeouts.stall_timeout_ms=300000 (source: codex/defaults)
runtime.timeouts.turn_timeout_ms=3600000 (source: codex/defaults)
```

7777 came from `codex:`; 300000 and 3600000 are defaults. Three lines, one label, two genuinely different provenances. The operator who greps `WORKFLOW.md` for `stall_timeout_ms` after reading "codex" finds nothing — the smaller version of the loop #882 describes.

This is a real improvement over the previous head and I'm not blocking on it, but the criterion asked for the source to be *unambiguous*, and per-field provenance is what actually delivers that: resolve each field and label it `codex` or `default` individually. Worth a follow-up rather than another round here.

nit, same lines: `(source: …)` repeated identically three times is noise while the label stays per-report — a single `runtime.timeouts source=…` line above the three values reads better.

---

packages/cli/src/commands/workflow.ts
```diff
@@ -878,10 +883,15 @@ function validateWorkflow(
...
-        turnTimeoutMs: workflow.codex.turnTimeoutMs,
       },
+      runtimeTimeouts: {
+        readTimeoutMs: effectiveTimeouts.readTimeoutMs,
+        stallTimeoutMs: effectiveTimeouts.stallTimeoutMs,
+        turnTimeoutMs: effectiveTimeouts.turnTimeoutMs,
+      },
+      runtimeTimeoutSource: workflow.runtime
+        ? "runtime.timeouts"
+        : "codex/defaults",
```
hojinzs, line 894: **P3 — the source is re-derived here instead of coming from the resolver, so it can silently drift when #888 lands.**

`resolveWorkflowRuntimeTimeouts` branches on `workflow.runtime?.timeouts`; this branches on `workflow.runtime`. They agree today, and not by luck — `WorkflowRuntimeConfig.timeouts` is a required field (`config.ts:131`) and the parser always materializes it (`parser.ts:1143-1151`), so `runtime` truthy ⟺ `runtime.timeouts` present. (Which also means the `?? workflow.codex` arm of the resolver is reachable only when `runtime` is absent entirely — the "runtime block, no timeouts sub-block" shape never falls back, exactly as your fourth test case pins.)

The drift risk is that this is a *duplicated* invariant rather than a derived one. #888 is about changing precisely that precedence; when it changes the resolver, nothing here fails to compile and no test in this file necessarily fails — the label just starts lying again, in the same class of way #882 was filed for.

Cheap insurance, and it makes the follow-up in the sibling comment fall out for free: have the resolver return the provenance alongside the values (or add a thin `resolveWorkflowRuntimeTimeoutSource` next to it in core), so the label and the number are read from one place.

Non-blocking — the current output is correct and verified.

---

.changeset/tidy-pandas-report.md
```diff
@@ -0,0 +1,5 @@
+---
+"@gh-symphony/cli": patch
```
hojinzs, line 2: **P3 — `patch` understates a removal of released `--json` keys.**

I agree with your call on the codex bot's alias request: `summary.codex.*TimeoutMs` could report values the runtime never used, so aliasing them forward would preserve exactly the failure #882 exists to kill. Removing them is right.

The severity is the separate question that thread didn't cover. `summary.codex.{readTimeoutMs,stallTimeoutMs,turnTimeoutMs}` were in released `workflow validate --json` output and are now absent — I confirmed against the built CLI at this head that `summary.codex` carries only `approvalPolicy`, `threadSandbox`, `turnSandboxPolicy`. Automation reading those keys gets `undefined`, not a wrong number, which is the better failure but still a break.

A `minor` bump would put the removal in the changelog where a consumer pinning `~` actually sees it before it bites. The one-line body is otherwise good and does mention the new object.

Your call — if CLI `--json` output isn't covered by the package's compat surface, say so and this drops.

---

hojinzs (approved):
**Approve.** Round 2. Every P1/P2 from round 1 is fixed, and I verified each one against the built CLI at `2a51e85` rather than taking the resolution comments at their word.

## Acceptance criteria (#882)

| # | Criterion | Verified |
|---|---|---|
| 1 | Reports `resolveWorkflowRuntimeTimeouts`, not `workflow.codex` | ✅ |
| 2 | `runtime.timeouts` declared → reported = declared | ✅ this repo's `WORKFLOW.md` → `30000 / 900000 / 3600000` |
| 3 | Legacy `codex:` only → no regression | ✅ `7777 / 88888 / 999999` |
| 4 | Neither → documented defaults | ✅ `5000 / 300000 / 3600000` |
| 5 | Test pins precedence, fails on a rewire to `workflow.codex` | ✅ 4-shape `it.each`, values + source + exact JSON key set |
| 6 | Label doesn't misattribute the source | ⚠️ letter met, residual noted below |

The defect in the issue is gone: the exact command and file that reproduced it now print the declared values.

## Smoke test

Built at this head and ran `workflow validate` across all four front-matter shapes, text and `--json`:

```
legacy codex only → 7777/88888/999999   (source: codex/defaults)
runtime.timeouts  → 30000/900000/3600000 (source: runtime.timeouts)
runtime, no timeouts, + codex → 5000/300000/3600000 (source: runtime.timeouts)
neither           → 5000/300000/3600000 (source: codex/defaults)
```

The round-1 **P1 leak is confirmed fixed**: `summary.runtimeTimeouts` is exactly `{readTimeoutMs, stallTimeoutMs, turnTimeoutMs}` in all four shapes — three keys, never seven — and `summary.codex` no longer carries timeout keys at all. Grepped the tree for stale references to the removed keys; docs and tests are consistent.

`pnpm lint`, `pnpm typecheck`, `pnpm build`, and `packages/cli/src/commands/workflow.test.ts` (23 tests) all pass locally. CI green on `2a51e85` (Test, Container Smoke).

## Scope

Tight and appropriate. One resolver call, a three-field narrowing, one provenance field, docs, changeset — no over-engineering. The precedence itself is untouched and correctly deferred to #888, and the fourth test case pins today's behavior so that fix has to be a deliberate decision rather than a silent follow.

## Non-blocking

Three P3s inline: per-field provenance (`codex/defaults` can't describe a partially-declared block), the source label being re-derived rather than returned by the resolver, and `patch` vs `minor` for the `--json` key removal. None is worth another round — take them as follow-ups or reply and drop them.

---

PR #885, opened 2026-09-05
refactor(core): separate tracker declaration provenance

moncher-dev:
## Issues

- Closes #879

## Summary

- Keeps the seven legacy tracker credential names as an unconditional defense-in-depth backstop for Codex, Claude, and custom children, including empty and cross-tracker declaration states.
- Exposes adapter declaration provenance separately so real tracker-name tests can detect declaration drift without the backstop masking the result.
- Preserves serialized declaration forwarding through all runtime factories and documents the intentional core-layer divergence.

## Change-point diagram

- Tracker adapter declaration → serialized worker environment → Codex/Claude/custom runtime filtering
- Core reserved-auth constant → unconditional Codex + Claude + custom-child backstop
- Declaration provenance → `isDeclaredTrackerSecretEnvironmentName` → independently testable adapter coverage

## Start here

- `packages/core/src/runtime/custom-child-env.ts:79` — unconditional fallback and separate declaration predicate
- `packages/runtime-codex/src/runtime.test.ts:152` — all-seven-name agent-boundary probes for empty and Linear-only declarations
- `packages/runtime-claude/src/adapter.test.ts:144` — matching spawn-level Claude probes
- `docs/architecture.md:73` — deliberate divergence and defense-in-depth rationale

## User-Visible Behavior / Operational Impact

- Codex, Claude, and custom children never inherit any of the seven legacy tracker credentials, even when the active adapter declares none or declares only another tracker's names.
- Custom runtime authentication still rejects reserved tracker credential names.
- Adapter declarations remain separately observable and continue to drive coverage for new tracker integrations.

## Validation

- `pnpm exec vitest run packages/core/src/runtime/custom-child-env.test.ts packages/runtime-codex/src/runtime.test.ts packages/runtime-claude/src/adapter.test.ts` — pass (93 tests)
- `pnpm exec vitest run --config test/e2e/claude/vitest.config.ts -t "credential semantics"` — pass (2 tests; 7 skipped by filter)
- `pnpm lint` — pass
- `pnpm test` — pass (all 14 workspace projects); an unrelated workspace-hook timing test failed once, then passed in isolation and on the full rerun
- `pnpm typecheck` — pass
- `pnpm build` — pass
- `pnpm e2e:claude` — documented environment exception #887; two attempts could not start because the shared-host Docker daemon is unavailable
- `git diff --check` — pass

## Changeset

- Not needed because issue #879 has no changeset label.

## Risks & rollback

[...]

PR #886, opened 2026-09-05
docs(configuration): define agent-child environment boundary

### LINKED ISSUE #871: docs(configuration): document the agent-child environment boundary and validate the hook allowlist
Split 4 of 5 from #844 (see the triage comment there). Parent epic: #847, Phase 3.

## Problem

`docs/configuration.md:416-425` gives a precedence table for hooks and worker processes and says nothing about what an **agent child** sees, while the custom-runtime boundary is documented separately at `:490-527`. An operator therefore reasonably expects a project `.env` entry to reach the agent — which is exactly how the `DOCKER_CONFIG` mistake on #825 became plausible.

`SYMPHONY_WORKFLOW_HOOK_ENV_ALLOWLIST` is also neither validated nor documented, and #844 names it as the precedent that must be fixed before any future declarative channel is considered.

## Scope

- Document the **agent-child boundary** in `docs/configuration.md`, next to the existing precedence tables: what the child inherits, what is constructed for it, and what is always stripped.
- State plainly that a project `.env` entry does **not** reach the agent child, and why.
- Validate `SYMPHONY_WORKFLOW_HOOK_ENV_ALLOWLIST` (reject unparseable or unknown entries rather than silently ignoring them) and document it.

## Out of scope

- Any behaviour change to what the child actually receives (splits 1-3).
- Adding a declarative non-secret configuration channel — explicitly out of scope on #844.

## Acceptance criteria

- `docs/configuration.md` documents the agent-child boundary and the hook allowlist variable.
- `SYMPHONY_WORKFLOW_HOOK_ENV_ALLOWLIST` is validated with test coverage for the rejection path.

## Sequencing

After splits 1-3, so the documentation describes the end state rather than the transitional one. Needs a changeset only if the allowlist validation changes observable behaviour.

moncher-dev:
## Issues

- Closes #871

## Summary

- Document the least-privilege agent-child environment boundary and the project-`.env`-only hook allowlist.
- Validate malformed and undefined allowlist names only when a configured, trusted hook will execute.
- Preserve skipped hooks for absent/untrusted commands, while normalizing empty separators and duplicate names.

## Change-point diagram

- Managed project `.env` → execution guard → allowlist validation → approved hook environment
- Worker environment → runtime-specific least-privilege constructor → coding-agent child

## Start here

- `packages/orchestrator/src/service.ts:5053` — execution guard around allowlist validation
- `packages/orchestrator/src/service.test.ts:16107` — skipped-path regressions
- `docs/configuration.md:89` — operator contract
- `docs/configuration.md:397` — agent-child boundary

- Malformed or undefined names fail only when their configured hook is trusted and about to execute.
- Invalid allowlists remain inert when no hook exists or hook execution is trust-disabled.
- Empty separators are ignored and duplicate names are folded together.
- Coding-agent child inputs are unchanged.

## Validation

- `pnpm exec vitest run packages/orchestrator/src/service.test.ts` — pass (221 tests)
- `pnpm lint` — pass
- `pnpm test` — pass
- `pnpm typecheck` — pass
- `pnpm build` — pass
- `./e2e/run-standalone-project-e2e.sh` — unavailable twice at Docker preflight (exit 69; daemon socket unreachable); tracked in #889

## Changeset

- `.changeset/quiet-hooks-validate.md` (`patch`)

## Risks & rollback

- Risk is limited to allowlist parsing immediately before approved hook execution. Revert this PR to restore permissive behavior.

## Changed files

- `packages/orchestrator/src/service.ts` — validate only executing hooks and canonicalize entries
- `packages/orchestrator/src/service.test.ts` — cover parser rejection/normalization and both skip paths
- `docs/configuration.md` — document hook validation timing and the agent-child boundary
- `docs/architecture.md` — map validation to the orchestrator boundary
- `AGENT_TEST.md` — record authoritative coverage
- `.changeset/quiet-hooks-validate.md` — patch release note

## Security


moncher-dev (commented):
Tested at head `6f621603`. `pnpm install --frozen-lockfile && pnpm build && pnpm test` all pass (14 packages, 0 failures), and CI is green. **Requesting changes** on content, not on the build.

This PR's whole value is that the documentation describes what actually runs. One of the two new paragraphs does not.

## The hook allowlist paragraph is wrong about the daemon environment

The new text says:

> To expose additional variables from the **effective project/daemon environment**, set `SYMPHONY_WORKFLOW_HOOK_ENV_ALLOWLIST` to a comma-separated list of names

The daemon environment is not a usable source for either the allowlist variable or its targets. `hookEnv` is built by `buildProjectExecutionEnv` (`packages/orchestrator/src/service.ts:5200`) as:

```
{ ...projectEnv, ...inheritedEnv, ...explicitEnv }
```

- `projectEnv` is `readProjectEnv` — the project `.env` file, and nothing else (`service.ts:5137`).
- `inheritedEnv` is `process.env` filtered by `shouldInheritProcessEnvKey`, i.e. `INHERITED_ENV_ALLOWLIST` = `CI, HOME, LANG, LOGNAME, PATH, PWD, SHELL, TERM, TMPDIR, USER` plus `LC_*` (`service.ts:118`).
- `explicitEnv` is Symphony's injected hook context.

`SYMPHONY_WORKFLOW_HOOK_ENV_ALLOWLIST` is not in that allowlist, so setting it in the daemon environment never reaches `hookEnv` and is silently ignored. A daemon-only *target* is worse than ignored — this PR turns it into a hard failure. Using the parser this PR exports:

```
parseWorkflowHookEnvAllowlist("STAGING_API_HOST", {HOME, PATH, SYMPHONY_WORKFLOW_HOOK_ENV_ALLOWLIST:"STAGING_API_HOST"})
  -> THROWS: names an environment variable that is not defined: STAGING_API_HOST

parseWorkflowHookEnvAllowlist("STAGING_API_HOST", {HOME, STAGING_API_HOST:"x", ...})
  -> ["STAGING_API_HOST"]        # project .env only
```

So an operator who follows this paragraph and puts the variable in the daemon environment does not get a no-op — they get a failing hook. The Codex reviewer flagged this and is right. Either source allowlisted names from `process.env`, or say plainly that this is a project-`.env`-only mechanism.

Worth calling out because this repo's own daemon runs with `SYMPHONY_ALLOW_WORKFLOW_HOOKS=1` and an `after_create` hook.

Secondary, and documented so I am not blocking on it: a trailing comma now throws (`"A_VAR,"` -> `invalid environment variable name: ""`) where it was previously filtered out. The new prose does say malformed entries fail, so this is intentional — just an easy accident to make.

## The child-boundary table

I checked the table against the code as it stands after #868/#869/#870/#877 and it reads accurately today, including the `#agent-visible-symphony-context` anchor, which resolves. One dependency worth noting: the "Tracker-declared secrets ... always stripped" row is only unconditionally true while `CUSTOM_RUNTIME_RESERVED_AUTH_ENVIRONMENT_NAMES` exists in core. PR #885 removes it, which would make seven of those names conditional on the active tracker's declaration. I have asked for changes on #885 for that reason; flagging it here so the two do not land in an order that leaves this table describing something that no longer runs.

## Scope note

The diff for `docs/configuration.md` is 29 additions and 0 deletions. The issue asked for four disagreeing sections to be *replaced* by one child-boundary section; this appends a fifth statement and leaves the existing four in place. If the intent was consolidation, the old text still needs to be reconciled or removed — otherwise the next reader still has five places to compare.

Moving #871 back to `Ready`.

docs/configuration.md
```diff
@@ -89,6 +89,17 @@ are rejected. Set `SYMPHONY_ALLOW_WORKFLOW_HOOKS=1` (or `true`) in the host
 environment to permit hook execution. The gate and path-only rule intentionally
 diverge from the upstream shell-command hook model.
 
+Hook children receive portable process variables (`HOME`, `LANG`, `LOGNAME`,
+`PATH`, `PWD`, `SHELL`, `TERM`, `TMPDIR`, `USER`, and `LC_*`) plus Symphony's
+injected `SYMPHONY_*` hook context. To expose additional variables from the
+effective project/daemon environment, set
+`SYMPHONY_WORKFLOW_HOOK_ENV_ALLOWLIST` to a comma-separated list of names, for
```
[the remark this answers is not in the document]
moncher-dev: Resolved in `2ade8c0`. The allowlist documentation now says to set `SYMPHONY_WORKFLOW_HOOK_ENV_ALLOWLIST` in the managed project `.env` and define additional target names there as well. It explicitly states that daemon-only allowlist settings and target variables are not hook-forwarding sources because only the documented portable process set is inherited from the daemon.

moncher-dev (commented):
@hojinzs re-reviewed at head `2ade8c02`. **The substantive finding is fixed and verified.** Ready for your review.

## Build

`pnpm install --frozen-lockfile && pnpm build && pnpm test` on a clean worktree of this head: **all 14 packages pass, 0 failures.** CI green on both jobs.

## The hook paragraph now matches the code

Previous text claimed the allowlist exposed variables from the "effective project/daemon environment". New text:

> set `SYMPHONY_WORKFLOW_HOOK_ENV_ALLOWLIST` in the managed project `.env` ... and define those names in the same file. A daemon-only allowlist setting or target variable is not a source for hook forwarding because hooks inherit only the portable process variables listed above from the daemon.

That is exactly what `buildProjectExecutionEnv` does. Re-verified against the parser this PR exports, built from `2ade8c02`:

```
allowlist + target in project .env   -> ["STAGING_API_HOST"]
target daemon-only (filtered out)    -> THROWS: names an environment variable that is not defined: STAGING_API_HOST
empty / malformed entry              -> THROWS: contains an invalid environment variable name: ""
```

The doc now warns about the throwing case instead of recommending it. The wording change from "unknown entry" to "names a variable absent from the effective hook environment" is also more accurate, since that is the condition actually tested.

Rejection-path coverage is present in `packages/orchestrator/src/service.test.ts` for both branches (malformed `not-valid`, and absent `MISSPELLED_NAME`), which is the issue's second acceptance criterion.

## Both acceptance criteria on #871 are met

- `docs/configuration.md` documents the agent-child boundary and the hook allowlist variable ✅
- `SYMPHONY_WORKFLOW_HOOK_ENV_ALLOWLIST` is validated with test coverage for the rejection path ✅

I re-checked the child-boundary table against the code after #868/#869/#870/#877 and it still reads accurately, anchor included.

## One scope question for you, and a correction to my own last review

I previously flagged that the `docs/configuration.md` diff is additions-only (now 31/0) and suggested it should have *replaced* the four overlapping sections. Re-reading #871, that is not what the issue asks. Its scope line is "Document the agent-child boundary in `docs/configuration.md`, **next to the existing precedence tables**" — explicitly additive — and its out-of-scope list is about behaviour, not consolidation. So I do not think this PR should be bounced for it, and I am not treating it as a blocker.

If you do want the four sections consolidated, it is worth its own issue: deleting them is a judgement call about which wording survives, and it is easy to lose accurate detail (the broker table's per-variable audience column, for instance, has no equivalent in the new table).

One small pre-existing inaccuracy I noticed while comparing, **not introduced here**: the `GITHUB_TOKEN_BROKER_SECRET` row in *Credential Brokers And Git Access* says it "is declared as a tracker secret and removed from agent children by default". On `main` today it is removed by `CUSTOM_RUNTIME_RESERVED_AUTH_ENVIRONMENT_NAMES` unconditionally, not by the declaration. That row describes the world #885 is trying to create — and #885 is currently failing on exactly that point, so the row is ahead of the code either way. Worth a follow-up once #885 settles.

## Ordering note

The child-boundary table's "Tracker-declared secrets ... always stripped" row is unconditionally true only while the core constant exists. #885 removes it and is still red on the compatibility-mode e2e (re-reviewed at `d5bd4298` just now). This PR is safe to land first; #885 is the one that would need this table revisited.

packages/orchestrator/src/service.ts
```diff
@@ -5064,8 +5064,9 @@ export class OrchestratorService {
           repositoryPath: repositoryDirectory,
           env: hookEnv,
           trusted: isWorkflowHookExecutionAllowed(hookEnv),
-          envAllowlist: parseCommaSeparatedEnvList(
-            hookEnv[WORKFLOW_HOOK_ENV_ALLOWLIST_ENV]
+          envAllowlist: parseWorkflowHookEnvAllowlist(
+            hookEnv[WORKFLOW_HOOK_ENV_ALLOWLIST_ENV],
+            hookEnv
```
hojinzs: **<sub><sub>[P1 Badge](https://img.shields.io/badge/P1-red?style=flat)</sub></sub>  Validation runs before the skip checks, so a typo breaks projects that never run hooks**

`parseWorkflowHookEnvAllowlist(...)` is an **argument** to `executeWorkspaceHook`, so it is evaluated before `executeWorkspaceHook` gets a chance to return `skipped`. It has two skip paths (`packages/core/src/workspace/hooks.ts:243-263`):

1. no hook command configured for this kind → `skipped`
2. `trusted === false` → `skipped` — and this is the **default**, since `SYMPHONY_ALLOW_WORKFLOW_HOOKS` must be explicitly set

In both cases the throw now wins. It lands in `runHook`'s `catch` (`service.ts:5071`) and becomes `outcome: "failure"`, which `isSuccessfulHookResult` (`service.ts:146`) rejects. Downstream that is fatal, not cosmetic:

- `after_create` (`service.ts:3261`) → `rm -rf` of the freshly created workspace, then `throw`
- `before_run` (`service.ts:3330`) → `throw`, dispatch aborted

So an operator who puts `SYMPHONY_WORKFLOW_HOOK_ENV_ALLOWLIST=MISSPELED_NAME` in a project `.env` **and has no hooks at all** now loses dispatch for that project. Before this PR the entry was ignored and everything worked. The PR body says "fail hook execution", but no hook execution is involved in either of these paths.

Verified locally against the real modules — both assertions pass on `2ade8c0`:

```ts
// mirrors the argument-evaluation order of the runHook() call site
const env = { PATH: "/usr/bin", SYMPHONY_WORKFLOW_HOOK_ENV_ALLOWLIST: "MISSPELED_NAME" };

// 1. no hooks configured at all
await expect(runHookCallSite(NO_HOOKS, env, true)).rejects.toThrow(/not defined/);      // ✗ used to be `skipped`

// 2. hook defined but not trust-approved (the default)
await expect(
  runHookCallSite({ ...NO_HOOKS, beforeRun: "./scripts/hook.sh" }, env, false)
).rejects.toThrow(/not defined/);                                                        // ✗ used to be `skipped`
```

Validate only when the hook will actually run. Either move the parse inside `executeWorkspaceHook`, after the `hookCommand` and `trusted` guards, or gate the call site:

```ts
const hookCommand = resolveHookCommand(workflowResolution.workflow.hooks, kind);
const trusted = isWorkflowHookExecutionAllowed(hookEnv);
result = await executeWorkspaceHook({
  ...
  envAllowlist:
    hookCommand && trusted
      ? parseWorkflowHookEnvAllowlist(hookEnv[WORKFLOW_HOOK_ENV_ALLOWLIST_ENV], hookEnv)
      : [],
});
```

Please also add coverage for "bad allowlist + no hook configured stays `skipped`" — the current 217 tests all exercise the parser directly, so neither skip path is covered.

---

packages/orchestrator/src/service.ts
```diff
@@ -6338,14 +6339,33 @@ function parseOwnerProcessId(ownerToken: string): number | null {
...
-    .filter((entry) => /^[A-Z_][A-Z0-9_]*$/.test(entry));
+  const entries = value.split(",").map((entry) => entry.trim());
+  const invalidEntry = entries.find(
+    (entry) => !/^[A-Z_][A-Z0-9_]*$/.test(entry)
+  );
+  if (invalidEntry !== undefined) {
+    throw new Error(
+      `${WORKFLOW_HOOK_ENV_ALLOWLIST_ENV} contains an invalid environment variable name: ${JSON.stringify(invalidEntry)}`
+    );
+  }
```
hojinzs: **<sub><sub>[P2 Badge](https://img.shields.io/badge/P2-orange?style=flat)</sub></sub>  A trailing comma is fatal, and the message names an empty string**

`split(",")` keeps empty segments, so a trailing or doubled separator reaches the regex check:

```ts
parseWorkflowHookEnvAllowlist("A_VAR,", { A_VAR: "x" });
// throws: SYMPHONY_WORKFLOW_HOOK_ENV_ALLOWLIST contains an invalid environment variable name: ""
```

(verified locally on `2ade8c0`)

A trailing separator in a comma-separated list is neither "unparseable" nor "unknown" — the two cases #871 asks you to reject — and `: ""` gives the operator nothing to search for. Combined with the P1 above this turns a stray keystroke into a dispatch failure. Dropping empty segments first keeps the intended strictness while staying forgiving about formatting:

```ts
const entries = value
  .split(",")
  .map((entry) => entry.trim())
  .filter((entry) => entry.length > 0);
```

If failing on an empty entry is deliberate, say so on the thread and give the error its own message rather than interpolating `""` — but note `docs/configuration.md:98` currently documents it as an error, so the docs would need to follow whichever way you go.

---

packages/orchestrator/src/service.test.ts
```diff
@@ -37,11 +37,32 @@ import {
...
+    );
+
+    expect(() =>
+      parseWorkflowHookEnvAllowlist("VALID_NAME,MISSPELLED_NAME", {
+        VALID_NAME: "present",
+      })
+    ).toThrow(
+      "SYMPHONY_WORKFLOW_HOOK_ENV_ALLOWLIST names an environment variable that is not defined: MISSPELLED_NAME"
+    );
+  });
```
hojinzs, line 67: **<sub><sub>[P3 Badge](https://img.shields.io/badge/P3-blue?style=flat)</sub></sub>  The new `describe` block is wedged between two import statements**

This block sits between the `./service.js` import and `import * as trackerAdapters from "./tracker-adapters.js"` on line 65. Hoisting makes it work, but it splits the import list in half and reads as a mistake to the next person in this file. Every other suite here starts after the imports — move it down below line 65.

Also: two unrelated rejection paths are asserted in one `it`, so a regression in the first hides the second. Splitting into `it("rejects a malformed name")` / `it("rejects an undefined name")` matches the surrounding style and gives you the failing case by name.

---

packages/orchestrator/src/service.ts
```diff
@@ -6338,14 +6339,33 @@ function parseOwnerProcessId(ownerToken: string): number | null {
...
+  const unknownEntry = entries.find(
+    (entry) => !Object.hasOwn(environment, entry)
+  );
+  if (unknownEntry !== undefined) {
+    throw new Error(
+      `${WORKFLOW_HOOK_ENV_ALLOWLIST_ENV} names an environment variable that is not defined: ${unknownEntry}`
+    );
+  }
+
+  return entries;
```
hojinzs, line 6383: **<sub><sub>[nit Badge](https://img.shields.io/badge/nit-lightgrey?style=flat)</sub></sub>  Duplicates pass through unchanged**

`parseWorkflowHookEnvAllowlist("A_VAR,A_VAR", { A_VAR: "x" })` returns `["A_VAR", "A_VAR"]` (verified). Harmless downstream — `buildHookExecutionEnv` folds the list into a `Set` — but this is now an exported, validating parser, so `[...new Set(entries)]` would make the return value canonical. Take it or leave it.

---

hojinzs (changes requested):
## Review 1/6 — one P1 blocker, otherwise a well-scoped PR

### Requirements vs. #871

| --- | --- |
| `docs/configuration.md` documents the agent-child boundary | ✅ `docs/configuration.md:397-414`, table + the explicit "a project `.env` entry does **not** reach the agent child" statement |
| `docs/configuration.md` documents the hook allowlist variable | ✅ `docs/configuration.md:92-104` |
| `SYMPHONY_WORKFLOW_HOOK_ENV_ALLOWLIST` validated | ⚠️ implemented, but fires on paths where no hook runs — see P1 |
| Rejection path has test coverage | ⚠️ parser-level only; neither skip path is covered |
| No behaviour change to what the child receives (out of scope) | ✅ documentation only, confirmed against `agent-child-env.ts` / `custom-child-env.ts` |

I spot-checked the doc claims rather than taking them on trust: the portable hook set matches `DEFAULT_HOOK_ENV_KEYS` (`packages/core/src/workspace/hooks.ts:272-282`); the `#agent-visible-symphony-context` anchor resolves to the heading at `docs/configuration.md:634`; and "Codex and Claude also receive a private `DOCKER_CONFIG`" is right — `agent-child-env.ts:20` sets it, `custom-child-env.ts:152-153` deliberately does not. **No over-engineering or unrequested decisions found** — the diff is the issue's scope and nothing more, and the changeset is correctly scoped to `patch`.

The earlier Codex P2 thread is genuinely resolved, not just marked resolved: `2ade8c0` rewrote the paragraph to describe the mechanism as project-`.env`-only, which is what the reviewer asked for.

### Smoke test

No preview environment on this PR, so I ran it locally against the real modules at `2ade8c0`: built the workspace, exercised the `runHook` → `executeWorkspaceHook` composition directly, and re-ran the touched suite.

- `npx vitest run packages/orchestrator/src/service.test.ts` — **217 passed**, matching the PR body
- Targeted reproduction of the composed call path — **surfaced the P1 below**

### Findings

| | Finding |
| --- | --- |
| **P1** | Allowlist validation is evaluated as a call argument, so it runs before `executeWorkspaceHook`'s two `skipped` returns. A typo'd entry now fails dispatch for a project with **no hooks configured**, and for the default-off `SYMPHONY_ALLOW_WORKFLOW_HOOKS` case. Via `after_create` that also `rm -rf`s the freshly created workspace. |
| **P2** | A trailing comma throws `invalid environment variable name: ""` — neither of the two conditions #871 asks you to reject, and undiagnosable from the message. |
| **P3** | The new `describe` block is inserted between two import statements in `service.test.ts`, and packs two rejection paths into one `it`. |
| **nit** | Duplicate entries are returned undeduplicated. |

### What I need for approval

Just the P1 — a guard so the parser runs only when the hook will actually execute, plus a test pinning "bad allowlist + no hook configured stays `skipped`". The P2 is a small follow-on in the same function. P3 and the nit are yours to take or decline; a reasoned "no" on the thread is fine and I won't re-raise them.

Everything else here is ready. The documentation is accurate and the boundary it describes is the right one to write down.

---

packages/orchestrator/src/service.ts
moncher-dev: Resolved in `e7aaa5b`. `runHook` now resolves the command and trust gate before parsing, and passes an empty allowlist for either skip path. The new parameterized integration test verifies both invalid-allowlist + no-command and invalid-allowlist + untrusted-command still dispatch the worker.

packages/orchestrator/src/service.ts
moncher-dev: Resolved in `e7aaa5b`. Empty separators are now ignored, including trailing and doubled commas, and the operator documentation follows that behavior. Actual malformed non-empty names and undefined names remain rejected.

packages/orchestrator/src/service.test.ts
moncher-dev: Resolved in `e7aaa5b`. The tracker adapter import is now contiguous with the other imports, and malformed-name and undefined-name rejection each have an independently named test.

packages/orchestrator/src/service.ts
moncher-dev: Accepted and resolved in `e7aaa5b`. The parser now returns first-seen unique names via `Set`, with coverage combined with the empty-separator normalization case.

hojinzs (approved):
## Review 2/6 — Approved

All four findings from review 1/6 are fixed at `e7aaa5b`, and I verified each against the real modules rather than taking the thread replies on trust.

### Finding disposition

| | Finding | Status |
| --- | --- | --- |
| **P1** | Allowlist parsed as a call argument, firing before both `skipped` returns | ✅ Fixed — `service.ts:5056-5079` hoists `hookCommand`/`trusted` and gates the parse on `hookCommand && trusted` |
| **P2** | Trailing/doubled comma threw `invalid environment variable name: ""` | ✅ Fixed — `.filter((entry) => entry.length > 0)` precedes the regex check |
| **P3** | `describe` wedged between imports; two rejection paths in one `it` | ✅ Fixed — import list is contiguous again, and `rejects a malformed name` / `rejects an undefined name` are separate |
| **nit** | Duplicates returned undeduplicated | ✅ Accepted — `[...new Set(...)]` |

The P1 guard is exactly equivalent to `executeWorkspaceHook`'s two skip returns (`packages/core/src/workspace/hooks.ts:247-266`): same `resolveHookCommand(hooks, kind)` on the same object, and the same `trusted` value that is then passed through. No third skip path exists, so there is no case where validation fires on a path that skips.

### Smoke test

No preview environment, so I ran this locally at `e7aaa5b` — clean `pnpm install --frozen-lockfile && pnpm build` (exit 0), then a direct reproduction of my original P1 repro against the real `executeWorkspaceHook` composed with the new call-site expression:

```
no hooks configured      + bad allowlist -> skipped   ✅ (was: throws)
hook configured/untrusted + bad allowlist -> skipped   ✅ (was: throws)
hook configured + trusted + bad allowlist -> throws    ✅ validation still has teeth
"A_VAR,,A_VAR, "                          -> ["A_VAR"] ✅ P2 + nit
```

The third case matters: gating the parse could have quietly disabled the feature this issue exists to add, and it does not.

- `npx vitest run packages/orchestrator/src/service.test.ts` — **221 passed** (was 217; +4 from this commit), matching the PR body
- `pnpm lint` — pass · `pnpm typecheck` — pass · `pnpm build` — pass
- CI green on both `Test` and `Container Smoke`
- The positive path stays covered end-to-end by `applies allowlisted project .env to approved script hooks` (`service.test.ts:16070`), which runs a real hook script reading the forwarded variables — so the wiring is pinned in both directions now
- `delete process.env.SYMPHONY_ALLOW_WORKFLOW_HOOKS` in the new test is restored by the suite's `afterEach` (`service.test.ts:579-590`), so no cross-test leakage

### Requirements vs. #871

| --- | --- |
| `docs/configuration.md` documents the agent-child boundary | ✅ `:397-414` |
| `docs/configuration.md` documents the hook allowlist variable | ✅ `:92-106`, now including the validation-timing sentence |
| `SYMPHONY_WORKFLOW_HOOK_ENV_ALLOWLIST` validated | ✅ |
| Rejection path has test coverage | ✅ parser-level plus both skip paths as an integration test |
| No behaviour change to what the child receives (out of scope) | ✅ documentation only |

The doc followed the code on the P2 change rather than being left stale — `empty separators are ignored and duplicate names are folded together` replaces the old `an empty or malformed entry` wording. That is the detail that most often gets missed in a fix like this.

**No over-engineering or unrequested decisions.** The diff is the issue's scope and nothing beyond it; the changeset is correctly `patch`.

### One non-blocking note — no response needed

You took the call-site gate, which is the option I offered, so this is an observation and not a request. `resolveHookCommand` is now evaluated twice for every hook — once in `runHook`, once inside `executeWorkspaceHook` — which means `service.ts` has to stay in step with a skip condition owned by `core`. It is correct today. If a third skip condition is ever added to `executeWorkspaceHook`, moving the parse inside it (after its guards) is the version that cannot drift. Not worth a change now.

The `#885` ordering note from the earlier review still stands: this PR is safe to land first, and `#885` is the one that would need the child-boundary table revisited.

---

PR #911, opened 2026-09-05
refactor(cli)!: remove instance registry and instances command

moncher-dev:
## Issues

- Closes #906

## Summary

- Removes the host-global orchestrator instance registry and the visible `gh-symphony instances` command.
- Preserves a hidden removed-command migration stub that directs users to `project list` and `project status`.
- Keeps standalone lifecycle addressing on project folders, daemon PID records, and project locks while documenting the removed cross-runtime-root duplicate-start guard.

## Change-point diagram

- CLI command router/help → `instances` absent from help, hidden migration stub returns guidance
- `project start` → project lock + daemon PID record only
- `project list` → cached project metadata without registry-backed `instance`
- release notes → breaking behavior and upgrade cleanup guidance

## Start here

- `packages/cli/src/index.ts` — hidden migration stub and command registration
- `packages/cli/src/commands/start.ts` — registry-free startup and shutdown
- `packages/cli/src/commands/project.ts` — registry-free project listing
- `.changeset/clean-ravens-rest.md` — complete breaking-change guidance

## User-Visible Behavior / Operational Impact

- `gh-symphony instances` is removed from help and command listings; direct invocation exits 2 with migration guidance.
- `project start --allow-duplicate` and the `project list` JSON `instance` field are removed.
- Cross-runtime-root duplicate-start detection is no longer available; #921 tracks a registry-independent replacement.
- `project start`, `project status`, and `project stop` continue to use project locks and daemon PID records.
- Stale `~/.gh-symphony/instances/` data is unused and may be deleted manually.

## Validation

- `pnpm lint` — pass
- `pnpm test` — pass across all 14 tested workspace packages; CLI 568/568
- `pnpm typecheck` — pass across all 14 workspace packages
- `pnpm build` — pass across all 14 workspace packages
- `pnpm exec vitest run packages/cli/src/index.test.ts` — pass (34 tests)
- `pnpm exec prettier --check packages/cli/src/index.ts packages/cli/src/index.test.ts .changeset/clean-ravens-rest.md` — pass
- `./e2e/run-standalone-project-e2e.sh` — unavailable after two Cycle 1 attempts because Docker was unreachable; documented in #918
- Direct built-CLI host smoke — Cycle 1 pass for `project start --daemon`, `project status`, and `project stop`; PID/lock creation and cleanup verified; no `instances/` directory created

## Changeset

- `.changeset/clean-ravens-rest.md` (`major`)

## Risks & rollback

- Scripts using the removed command, option, or project-list field require migration.
[...]

Issue #847, opened 2026-09-04
[Epic] Reduce Symphony to its spec footprint: remove repo mode, unify the child boundary, move population to a hook


moncher-dev:
Execution epic for the design published as "심포니 스펙 축소" (2026-09-04). The design's own conclusions are the input; this epic is the delivery sequence.

## Standing constraints

- **Dogfood.** All code changes are made by dispatched workers through this board. The operator and the board steward do not edit source directly. The only exceptions are operator-side host actions, which are listed explicitly and marked as such.
- **This orchestrator currently runs in repo-embedded mode**, which the plan removes. The running daemon uses the installed CLI, not merged source, so merging the removal does not break it. The host must nonetheless be migrated to standalone project mode before the daemon is upgraded past the removal.
- **One PR lands per steward cycle**, furthest-behind branch first. Landing siblings together has reset three approvals already.

## Phase 0 — host migration to project mode (operator action)

Prerequisite: #848 adds `repository.slug` to WORKFLOW.md, which standalone mode requires and repo mode ignores.

Then, at a moment when no run is active:

1. Create `~/symphony-projects/github-symphony/`.
2. Symlink `WORKFLOW.md` there to the repository's copy, so the policy stays versioned in git and does not fork.
3. Create `.env` (0600) carrying `GITHUB_GRAPHQL_TOKEN`. Do not carry `DOCKER_CONFIG`; it is inert (#825).
4. `gh-symphony repo stop`, then `gh-symphony project start --daemon --assigned-only --project-dir ~/symphony-projects/github-symphony`.
5. Verify one dispatch end to end before proceeding.

Hook scripts need no move: hooks execute with the issue workspace as their working directory, so `hooks/after_create.sh` continues to resolve inside the checkout. Workspaces relocate under the project folder, which is expected and costs one repopulation.

Standalone mode already defaults to `worktree-cache`, so this also retires the hand-applied populate override recorded on #808.

## Phase 1 — land what is already decided

#826 (publish action; PR #845 still missing the SIGTERM path), #825, #834, #812, and the #839 decision. #811 and its PR #828 are probably deletions rather than merges once repo mode goes, since the `clone` strategy exists to serve repo mode; confirm before spending another review round.

## Phase 2 — remove repo-embedded mode

Three children, sized to one worker run each, landed in order. This is the pivot: every later phase then touches one code path instead of two.

## Phase 3 — boundaries

[...]

PR #912, opened 2026-09-05
refactor(tracker-github): require direct host GraphQL auth

### LINKED ISSUE #902: refactor(tracker-github): resolve host GraphQL auth from a declared direct credential instead of the broker
Phase 4 of epic #847. 브로커 제거 3분할의 **첫 번째**입니다. 근거는 스파이크 보고서 `docs/reports/2026-09-05-credential-broker-footprint.md`이고, 3분할 결정은 #847의 2026-09-05 코멘트에 있습니다.

## 왜 셋으로 쪼갰는가

이 저장소에는 브로커가 **둘** 있습니다. GitHub 토큰 브로커는 호스트 쪽 트래커 GraphQL과 HTTPS Git 발행을 먹이고, 에이전트 크리덴셜 브로커는 런타임에 제공자 키를 먹입니다. 이름과 타입 어휘를 공유할 뿐 프로토콜도 소비자도 다릅니다. 한 이슈로 지우면 트래커 디스패치·에이전트 툴 실행·런타임 시작·복구 발행이 한 폭발 반경에 들어갑니다.

**이것이 먼저인 이유:** Git 발행을 건드리기 전에 보드와 상태 API가 살아 있어야 합니다.

## 무엇을 하는가

호스트의 GitHub GraphQL 인증을 브로커에서 **선언된 직접 호스트 크리덴셜**로 옮깁니다. 트래커 GraphQL에 한해 브로커 선택·페치·캐시·MCP 엔트리 배선·preflight/doctor 메시지와 관련 테스트를 제거합니다.

## 전제 — 이미 확인된 사실

- 워커의 브랜치 푸시는 브로커에 종속되지 않습니다. `GITHUB_GRAPHQL_TOKEN` 직접 토큰 **또는** 브로커 쌍 중 하나면 됩니다 (`packages/worker/src/tracker-credential-preflight.ts:41-57`, `packages/worker/src/git-transport.ts:228-244`).
- 직접 토큰 처리 경로가 이미 존재합니다 (`packages/tracker-github/src/orchestrator-adapter.ts:56-61,109-124`, `packages/tool-github-graphql/src/tool.ts:284-287`).
- 이 호스트의 프로젝트 `.env`에 `GITHUB_GRAPHQL_TOKEN`이 이미 있습니다.

## 상한선 — 스펙 §1107–1110

트래커 크리덴셜이 코딩 에이전트 자식에 상속되지 않을 것, 그리고 환경변수 기반 어댑터가 인증 이름을 선언할 것. 이 둘만 요구되며 브로커는 요구되지 않습니다. 어댑터의 `secretEnvironmentNames()` 선언은 **유지**합니다.

## 완료 조건

- [ ] 호스트 GitHub 트래커 작업과 호스트 소유 `github_graphql` 툴이 직접 호스트 토큰으로 동작한다. 오케스트레이터 폴링과 에이전트 툴 실행 **양쪽**을 증명한다.
- [ ] `GH_TOKEN`, `GH_ENTERPRISE_TOKEN`, `GITHUB_TOKEN`, `GITHUB_GRAPHQL_TOKEN`이 계속 인증 이름으로 선언된다.
- [ ] 자식 경계는 변하지 않는다. #872가 남긴 블랙박스 검증이 계속 통과한다.
- [ ] Git 발행 경로(`buildHostGitEnvironment`)는 **이 이슈에서 건드리지 않는다**. 브로커 쌍을 받는 분기가 그대로 남아 있어도 된다.
- [ ] changeset 추가.

## 범위 밖

Git 발행 이전(P4‑2b), 에이전트 제공자 브로커 제거(P4‑2c), `GITHUB_TOKEN_BROKER_TIMEOUT_MS` 제거. 마지막은 어떤 네트워크 리졸버도 쓰지 않게 된 뒤에만 지웁니다.

## 주의

브로커와 호스트 토큰을 **함께** 지우면 워커가 시작조차 못 하고 보드가 멈춥니다. 이 이슈는 호스트 토큰 경로를 남기는 것이 전제입니다.

moncher-dev:
## Issues

- Closes #902

## Summary

- Require a declared direct host `GITHUB_GRAPHQL_TOKEN` for GitHub tracker polling and the host-owned `github_graphql` tool.
- Remove tracker-tool broker resolution, caching, and MCP wiring while retaining Git publication code for the later #903 slice.
- Preserve all direct and broker secret-name declarations so agent-child and indirect MCP boundaries remain intact.

## Change-point diagram

- Project/daemon environment → GitHub adapter selects direct token → orchestrator polling and host-owned agent tool
- Adapter secret declarations → Codex/Claude child filtering → tracker and broker credentials remain host-only
- Host Git transport → unchanged direct-token-first helper; retained broker branch is documented as currently unreachable

## Start here

- `packages/tracker-github/src/orchestrator-adapter.ts:44` — secret declarations and direct credential resolution
- `packages/tracker-github/src/orchestrator-adapter.ts:95` — host-owned GraphQL tool execution
- `packages/worker/src/tracker-credential-preflight.ts:41` — direct-token startup requirement
- `packages/runtime-codex/src/git-credential-helper.ts:67` — isolated, behavior-preserving Git publication resolver

- GitHub projects must declare `GITHUB_GRAPHQL_TOKEN`; broker-only tracker authentication no longer starts polling or workers.
- Because the required direct token takes precedence in the unchanged Git helper, it must also permit repository pushes.
- The four direct authentication aliases and three retained broker names remain declared for child-boundary filtering.

## Validation

- `pnpm lint` — pass
- `pnpm test` — pass
- `pnpm typecheck` — pass
- `pnpm build` — pass
- Focused tracker/runtime suite — pass (236 tests)
- Docker TC-19/TC-20 startup — documented environment exception #916; Docker daemon unavailable on both attempts before scenario execution

## Changeset

- `.changeset/direct-github-host-auth.md` (`minor`)

## Risks & rollback

- Operational risk: a direct token without repository push scope can poll successfully but fail publication; doctor and documentation call out the required scope.
- Boundary risk: indirect MCP secret names remain protected by retaining broker names in the adapter declaration.
- Rollback: revert this PR to restore broker-backed tracker-tool resolution; no data migration is involved.

## Changed files

- `packages/tracker-github/**` — direct host auth selection, tool execution, and declaration tests
- `packages/tool-github-graphql/**` — remove tracker-tool broker resolution and MCP inputs
- `packages/worker/src/tracker-credential-preflight*` — require direct host token before launch
- `packages/runtime-codex/**`, `packages/runtime-claude/**` — remove stale tracker-tool broker wiring and isolate retained Git helper policy
- `README.md`, `packages/cli/README.md`, `docs/**`, `AGENT_TEST.md` — update operations, architecture, and black-box coverage
- `.changeset/direct-github-host-auth.md` — minor CLI release note

- [ ] Confirm production project `.env` files declare `GITHUB_GRAPHQL_TOKEN` with repository push scope.
- [ ] Re-run TC-19/TC-20 after #916 restores Docker availability.
- [ ] Complete Git publication broker removal in #903; agent provider broker removal remains a separate slice.

## Security


moncher-dev:
Reviewed at `ec1a0ff7` and exercised live, not just read.

**Orchestrator polling on a direct token** — real call against the real board through the built `tracker-github` adapter, `GITHUB_GRAPHQL_TOKEN` only, no broker anywhere:
```
POLL_DIRECT_TOKEN: issues= 347  sample= #792:Done, #703:Done, #696:Done
```

**Host-owned agent tool on a direct token** — `githubProjectTrackerAdapter.executeAgentTool("github_graphql", …)` against api.github.com:
```
AGENT_TOOL_DIRECT_TOKEN: {"data":{"viewer":{"login":"hojinzs"}, …rateLimit cost 1, remaining 4894}}
```

**Broker-only now fails loudly instead of silently working** — same tool with `GITHUB_TOKEN_BROKER_URL`/`_SECRET` and no token:
```
AGENT_TOOL_BROKER_ONLY: GITHUB_GRAPHQL_TOKEN is required.
```

**Auth names still declared**: `secretEnvironmentNames()` still returns `GH_TOKEN`, `GH_ENTERPRISE_TOKEN`, `GITHUB_TOKEN`, `GITHUB_GRAPHQL_TOKEN`, and the test was *tightened* from `expect.arrayContaining([...injectedNames])` to an exact `toEqual([...])` — the #879 failure mode (a list assertion that cannot notice an extra or missing name) is closed here rather than widened.

**Git publication untouched**: the only `packages/worker` files in the diff are `tracker-credential-preflight.ts` and its test. `git-transport.ts`/`buildHostGitEnvironment` are byte-identical to main, and the broker logic the Git path needs was moved intact into `runtime-codex/git-credential-helper.ts` with its 8 existing broker tests (timeout attribution, wrapped cause, non-timeout passthrough) still passing. #903 keeps its scope.

**The tests can fail** — three mutations of the source this PR claims to protect, each went red:
| mutation | red test |
|---|---|
| `resolveGitHubGraphQLToken` returns a fallback instead of throwing | `rejects missing direct host authentication` |
| preflight returns `ok:true` unconditionally | `rejects a GitHub worker without a credential before launch` + 5 more |
| drop `GH_TOKEN`/`GH_ENTERPRISE_TOKEN` from `secretEnvironmentNames` | `declares the direct GitHub authentication names` |

`pnpm build` clean; 355 tests pass across tool-github-graphql, tracker-github, worker preflight, runtime-codex, runtime-claude.

One judgement call for @hojinzs, not blocking: the changeset is `patch`, but this removes a previously supported tracker auth mode (broker-only hosts stop polling on upgrade). #911 uses `major` for removing a command; a `minor` here would read more honestly.

packages/runtime-codex/src/git-credential-helper.ts
```diff
@@ -72,6 +71,99 @@ export async function resolveGitCredential(
...
 }
 
+async function resolveGitCredentialToken(
+  config: GitCredentialHelperConfig,
+  fetchImpl: typeof fetch
+): Promise<string> {
+  if (config.token) {
+    return config.token;
+  }
+  if (!config.tokenBrokerUrl || !config.tokenBrokerSecret) {
```
hojinzs, line 87: **P2 — the "preserved" Git publication broker path is now unreachable, and the docs say otherwise**

This resolver is byte-equivalent to what `resolveGitHubGraphQLToken` did on `main`, so there is no regression *in this function*. But the preflight change one package over makes its broker branch dead in every deployment that can start:

1. `tracker-credential-preflight.ts:42` now hard-requires `GITHUB_GRAPHQL_TOKEN` for every GitHub worker.
2. So `config.token` is always populated by the time Git publication runs.
3. Line 78 returns it before the broker is ever consulted.

Verified at `ec1a0ff` with a stubbed broker + `buildHostGitEnvironment`:

```
SMOKE_4 (broker only)  -> broker contacted, password=brokered-git-token
SMOKE_5 (both present) -> password=direct-token, broker contacted: false
```

Before this PR a broker-only project passed preflight and pushed with the brokered installation token. After it, that project must add `GITHUB_GRAPHQL_TOKEN`, and that token silently takes over publication. If it is narrower than the brokered token (no repo write), the push fails *after* the agent has done its work.

I do **not** think the fix is to reorder this function — #902 explicitly scopes Git publication out, and #903 removes it anyway. Reordering here would be exactly the kind of unrequested change the issue guards against. The problem is that the PR's own new wording claims a compatibility path that no longer exists:

- `docs/configuration.md:478-481` — "Broker endpoint retained for the Git publication compatibility path"
- `packages/cli/README.md:210-211` — "The separate Git publication transport still supports its broker compatibility path"

**Ask:** correct those two sentences to say the broker branch is retained in code but no longer reachable, because every GitHub worker must now declare `GITHUB_GRAPHQL_TOKEN` and that token takes precedence for Git publication — so it must carry push scope. (`doctor` already checks for `repo` scope on it, which limits the blast radius, so this reads as a docs-accuracy fix rather than a code change.)

This also supersedes the Codex P1 on this hunk: its claim that the branch *ordering* changed is wrong, but its conclusion about the effective outcome is right.

---

packages/tracker-github/src/orchestrator-adapter.ts
```diff
@@ -47,9 +47,6 @@ export const githubProjectTrackerAdapter: OrchestratorTrackerAdapter = {
       "GH_ENTERPRISE_TOKEN",
       "GITHUB_TOKEN",
       "GITHUB_GRAPHQL_TOKEN",
-      "GITHUB_TOKEN_BROKER_URL",
-      "GITHUB_TOKEN_BROKER_SECRET",
```
hojinzs, line 51: **P2 — dropping the broker names here narrows the child boundary by one indirect path**

AC #2 of #902 requires the four direct names to *keep* being declared; it never asked for the three broker names to be removed. Removing them is not free, because `secretEnvironmentNames()` feeds two consumers, not one:

1. **Direct strip** — `runtime-codex/src/runtime.ts:651-658` and `runtime-claude/src/adapter.ts:713-724`. Still safe: both also spread `CUSTOM_RUNTIME_RESERVED_AUTH_ENVIRONMENT_NAMES`, which lists all three broker names (`core/src/runtime/custom-child-env.ts:79-91`). The boundary holds here.

2. **Indirect strip** — `collectMcpSecretEnvironmentNames()` (`core/src/runtime/mcp-compose.ts:44-58`) only follows a `$VAR` indirection when the *key* is a declared secret. This one regresses. With a trusted repo `.mcp.json` containing `{"env": {"GITHUB_TOKEN_BROKER_SECRET": "$MY_BROKER_SECRET"}}`, run at `ec1a0ff`:

```
MAIN declaration collects: ["MY_BROKER_SECRET"]
PR   declaration collects: []
```

So `MY_BROKER_SECRET` stays in the Claude child env (`buildClaudeSpawnEnv` inherits full `process.env` when `inheritProcessEnv` is set), where it previously did not. Narrow — it needs `trustRepoConfig` and an indirected `.mcp.json` — but it is a live gap for as long as broker credentials legitimately exist on the host, which this PR says they do until #903.

`docs/architecture.md:74` also still states that adapter declarations are authoritative and the backstop is only defense-in-depth. After this change, broker-secret stripping rests on the backstop alone.

**Ask:** keep `GITHUB_TOKEN_BROKER_URL` / `GITHUB_TOKEN_BROKER_SECRET` / `GITHUB_TOKEN_CACHE_PATH` in this array until #903 removes the credentials themselves. That satisfies AC #2 as written (the four direct names still declared, the exact-match test still valid), leaves the child boundary provably unchanged per AC #3, and costs nothing — the declaration is a strip list, not a resolution list, so it does not resurrect any broker code path.

If you'd rather remove them now, that's a defensible call, but it should be stated as a deliberate divergence in the PR description and `docs/architecture.md` rather than landing silently.

---

packages/runtime-codex/src/runtime.ts
```diff
@@ -189,12 +189,7 @@ type CodexProtocolMessage = Record<string, unknown>;
 export function createGitHubGraphQLToolDefinition(
   config: Pick<
     CodexRuntimeConfig,
-    | "githubToken"
-    | "githubTokenBrokerUrl"
-    | "githubTokenBrokerSecret"
-    | "githubTokenCachePath"
-    | "githubProjectId"
-    | "githubGraphqlApiUrl"
+    "githubToken" | "githubProjectId" | "githubGraphqlApiUrl"
```
hojinzs, line 192: **P3 — leftover broker branch at the only caller of this function**

Narrowing the `Pick` here is right, but `buildCodexRuntimePlan` (line 589-595, outside the diff) still does:

```ts
const usesGitHubTokenBroker = Boolean(
  config.githubTokenBrokerUrl && config.githubTokenBrokerSecret
);
const githubTool = createGitHubGraphQLToolDefinition({
  ...config,
  githubToken: usesGitHubTokenBroker ? undefined : config.githubToken,
});
```

That blanks the direct token whenever a broker is configured — which, since `launcher.ts:41-43` populates those fields from the worker env, is every project that keeps broker credentials for Git publication.

I chased this as a possible P1 and it is **not** one: `plan.tools` is `[]` (line 648) and only `dynamicTools` survives, which carries name/description/schema and drops `env` entirely. The tool executes host-side through `githubProjectTrackerAdapter.executeAgentTool` off `context.environment`. I confirmed the token blanking has no runtime effect.

So it is dead code — but it now reads as "prefer the broker for `github_graphql`", which is the exact opposite of what this PR establishes. Worth deleting in the same sweep, along with `usesGitHubTokenBroker`.

---

packages/tracker-github/src/orchestrator-adapter.ts
```diff
@@ -59,22 +56,8 @@ export const githubProjectTrackerAdapter: OrchestratorTrackerAdapter = {
...
-        };
-        const cachePath = environment.GITHUB_TOKEN_CACHE_PATH?.trim();
-        if (cachePath) {
-          credentials.GITHUB_TOKEN_CACHE_PATH = cachePath;
-        }
-        return credentials;
-      }
     }
-    return {};
+    return {} as Record<string, string>;
```
hojinzs: **nit** — the assertion is unnecessary. `resolveWorkerCredentials` is declared as returning `Record<string, string>` (`core/src/contracts/tracker-adapter.ts:288-294`), so a bare `{}` widens correctly. `main` had `return {};`; the `as` came in with this diff and only suppresses type feedback if a future branch returns the wrong shape.

```suggestion
    return {};
```

---

packages/runtime-codex/src/git-credential-helper.ts
```diff
@@ -1,20 +1,21 @@
-import {
-  resolveGitHubGraphQLToken,
-  type GitHubGraphQLToolConfig,
-} from "@gh-symphony/tool-github-graphql";
 import { writeSync } from "node:fs";
+import { chmod, mkdir, readFile, writeFile } from "node:fs/promises";
+import { dirname } from "node:path";
+import { validateGitHubTokenBrokerUrl } from "@gh-symphony/tool-github-graphql";
```
hojinzs: **P3 — the isolation is incomplete: Git transport still reaches into the GraphQL tool package**

The point of moving the resolver here was to sever Git publication from `tool-github-graphql`. It is still coupled: this file imports `validateGitHubTokenBrokerUrl` from that package, so the broker URL policy — a Git-transport concern now — lives in the GraphQL tool and must stay exported purely for this one consumer.

Per `docs/architecture.md`, keeping tracker-tool concerns out of the Git path is the boundary this PR is asserting, so moving `validateGitHubTokenBrokerUrl` (and its `validateSecureUrl` helper) either alongside this resolver or into `core` would make the isolation real rather than nominal. It would also let #903 delete the broker URL policy with the resolver in one move instead of leaving a dangling export.

Not blocking — the current shape is correct, just not fully isolated as the PR summary states ("isolate the unchanged Git publication broker resolver in the Git helper").

Its own behavior does check out: the URL policy still rejects non-https and private-IP brokers through this path, which I confirmed at `ec1a0ff` (`http://127.0.0.1:…` → `GitHub token broker URL must use https.`).

---

hojinzs (changes requested):
Reviewed at `ec1a0ff` — read the diff, then exercised the changed boundary locally rather than trusting the summary. Round 1 of 6.

The shape of this change is right: 188 added / 634 deleted, almost all deletion, no invention beyond what #902 asked for. The resolver move into the Git helper is the one thing beyond the literal ask and it was forced — the tool no longer holds the broker, and the issue permits the Git branch to stay. I found no over-engineering.

## Requirement coverage against #902

| AC | Verdict |
|---|---|
| Direct token drives both polling **and** host-owned `github_graphql` | ✅ Author demonstrated live polling (347 issues); I independently confirmed `executeAgentTool` sends `Bearer <direct-token>`, and that broker-only config now throws `GITHUB_GRAPHQL_TOKEN is required.` with **zero** network calls — no silent fallback. |
| Four auth names still declared | ✅ And the test was tightened from `arrayContaining` to exact `toEqual`, closing the #879 failure mode rather than widening it. Good. |
| Child boundary unchanged; #872 black-box still passes | ⚠️ The direct strip holds via the `CUSTOM_RUNTIME_RESERVED_AUTH_ENVIRONMENT_NAMES` backstop, but one indirect path regressed — see the `secretEnvironmentNames()` comment. Docker E2E not re-runnable here either (same #916 outage). |
| `buildHostGitEnvironment` untouched | ⚠️ True of the code, byte-for-byte. Not true of the effective behavior — see the `git-credential-helper.ts` comment. |
| Changeset added | ✅ (severity questioned below) |

## Local verification

`pnpm lint`, `pnpm typecheck`, `pnpm build` clean. Full `pnpm test` green (`core` 356, `worker` 241, `tracker-github` 134, `runtime-claude` 112, `runtime-codex` 83, `cli` 571, and the rest). `pnpm format` fails on 72 files, but none are touched by this PR — pre-existing on `main`, not yours.

Six-case smoke run against stubbed GraphQL and broker endpoints:

```
1. direct token          -> Bearer direct-token, {"data":{"viewer":{"login":"smoke-user"}}}
2. broker only           -> throws "GITHUB_GRAPHQL_TOKEN is required.", 0 fetch calls
3. preflight broker-only -> {"ok":false, "…GITHUB_GRAPHQL_TOKEN is required…"}
4. git, broker only      -> broker contacted, password=brokered-git-token
5. git, both present     -> password=direct-token, broker contacted: false
6. project vs daemon     -> {"GITHUB_GRAPHQL_TOKEN":"project-t"}
```

Case 5 is the one that matters and is why I'm not approving yet.

## Findings

- **P2** — `git-credential-helper.ts`: the Git publication broker branch is now unreachable in any deployment that can start, because preflight forces a direct token that then wins at line 78. The code is unchanged; the *docs this PR wrote* claim a compatibility path that no longer functions. Ask is a two-sentence docs correction, not a code change — reordering the resolver would be scope creep into #903.
- **P2** — `orchestrator-adapter.ts`: removing the three broker names from `secretEnvironmentNames()` was not required by AC #2 and costs the indirect `.mcp.json` `$VAR` strip. Demonstrated: `main` collects `["MY_BROKER_SECRET"]`, this PR collects `[]`. Ask is to keep the three names until #903 removes the credentials.
- **P3** — `runtime.ts`: `usesGitHubTokenBroker` at line 589 still blanks the direct token for a tool that no longer has a broker. I chased it as a possible P1 and it is inert (`plan.tools` is `[]`), but it now reads backwards. Delete with the rest.
- **P3** — `git-credential-helper.ts:4`: still imports the broker URL policy from `tool-github-graphql`, so the isolation is nominal rather than real.
- **nit** — `orchestrator-adapter.ts:60`: `{} as Record<string, string>` — the cast is unnecessary.
- **nit** (not inline, the lines aren't in the diff) — `runtime-claude/src/mcp-compose.ts:15-17,26`: `ClaudeMcpTokenEnvironment` still declares `GITHUB_TOKEN_BROKER_URL` / `_SECRET` / `GITHUB_TOKEN_CACHE_PATH` and `SYMPHONY_TRACKER_SECRET_ENVIRONMENT_NAMES`, none of which are read in that file any more after the `stripMcpServerSecretEnvironmentValues` removal. Dropping `stripMcpServerSecretEnvironmentValues` itself is correct — `composeClaudeMcpConfig` reduces `mcpServers` to just `symphony` before writing, so it was already unreachable.

## On the changeset

Agreeing with your own note to @hojinzs: `patch` understates this. The CLI is at `2.0.1`, so semver reads literally — and a broker-only host stops polling *and* fails worker startup on upgrade. That is a removed configuration mode, which is `minor` at the very least. Your call, but I'd rather it be honest in the changelog than discovered at rollout.

## Not blocking, for the record

The Codex P1 on `git-credential-helper.ts:78` is half right. Its stated mechanism — that this PR changed the credential preference order — is wrong; the ordering is identical to `main`. Its conclusion about the effective outcome is right, for the different reason I set out inline. Worth a reply on that thread either way so it doesn't read as unaddressed.

---

Requesting changes on the two P2s only; both are small and specific, and I'd expect this to approve on the next round. The P3s and nits are yours to take or decline with a reason.

---

packages/runtime-codex/src/git-credential-helper.ts
```diff
@@ -72,6 +71,99 @@ export async function resolveGitCredential(
   });
 }
 
+async function resolveGitCredentialToken(
+  config: GitCredentialHelperConfig,
+  fetchImpl: typeof fetch
+): Promise<string> {
+  if (config.token) {
+    return config.token;
+  }
```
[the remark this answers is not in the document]
moncher-dev: Resolved in `1d3c2c6` without reordering the Git helper. The direct-token-first order is unchanged from `main`; the new tracker preflight makes the retained broker branch unreachable. I documented that effective behavior and the required push scope, while leaving Git publication mechanics for #903.

packages/runtime-codex/src/git-credential-helper.ts
moncher-dev: Fixed in `1d3c2c6`. Both operator-facing documents now state that the broker branch remains in code but is currently unreachable because the required direct token takes precedence, and that the token must permit repository pushes. The resolver order remains unchanged.

packages/tracker-github/src/orchestrator-adapter.ts
moncher-dev: Fixed in `1d3c2c6`. `secretEnvironmentNames()` again declares the three broker names in addition to the four direct names. This preserves indirect `$VAR` source-name collection without restoring broker selection or resolution.

packages/runtime-codex/src/runtime.ts
moncher-dev: Fixed in `1d3c2c6`. `buildCodexRuntimePlan` now passes `config.githubToken` directly and no longer computes `usesGitHubTokenBroker`; the host-side dynamic tool remains the only effective path.

packages/tracker-github/src/orchestrator-adapter.ts
moncher-dev: Fixed in `1d3c2c6`; `resolveWorkerCredentials` now returns a bare `{}` and relies on the declared return type.

packages/runtime-codex/src/git-credential-helper.ts
moncher-dev: Fixed in `1d3c2c6`. Broker URL validation and its private-network policy now live in `git-credential-helper.ts`, and `runtime.ts` imports the policy locally. Focused tests cover public HTTPS acceptance and insecure/private rejection.

packages/tracker-github/src/orchestrator-adapter.ts
moncher-dev: Follow-up in `8ed8e27`: the bare `{}` exposed a TypeScript contextual-inference error during the full build, so the method now has an explicit `Record<string, string>` return annotation. This keeps the return expression assertion-free while making the interface contract explicit.

README.md
```diff
@@ -964,10 +964,11 @@ through host dynamic tools; Claude's generated MCP configuration contains only
...
 transport is described in
 [ADR 2026-08-28](docs/adr/2026-08-28_agent-tool-isolation.md).
-At dispatch, tracker adapters resolve host-only credentials from the tenant's
-project `.env` before the daemon environment. Workers use that daemon-resolved
-tracker identity for host tools and Git transport without restoring blanket
-environment inheritance.
+At dispatch, tracker adapters resolve host-only direct credentials from the
+tenant's project `.env` before the daemon environment. Workers use that
+daemon-resolved tracker identity for polling and host tools without restoring
+blanket environment inheritance. Git publication retains its separate broker
```
hojinzs, line 970: **P3 — the previous docs correction landed in two of three places; this one still says the broker path works**

`1d3c2c6` fixed `docs/configuration.md:478-479` and `packages/cli/README.md:209-211`, but this paragraph is a *new* line added by this PR and it still carries the exact wording the earlier P2 asked to retire: "Git publication retains its separate broker compatibility path."

That reads to an operator as "broker-only Git publication still works." It does not. I re-confirmed the mechanism at `8ed8e27` with a stubbed broker:

```
both present -> password=direct-token,       broker contacted: false
broker only  -> password=brokered-git-token, broker contacted: true   (but preflight now rejects this project)
```

`tracker-credential-preflight.ts:42` requires `GITHUB_GRAPHQL_TOKEN` for every GitHub worker, so `config.token` is always set and `git-credential-helper.ts:84` returns before the broker is consulted. The broker branch is retained in code but unreachable in any deployment that can start.

**Ask:** mirror the wording already used in the other two documents — the branch is retained for later removal, the required direct token takes precedence, and it must therefore permit repository pushes. One-line change; not worth another CI cycle on its own, but please land it before merge.

---

packages/runtime-codex/src/git-credential-helper.ts
```diff
@@ -182,3 +280,104 @@ function isTimeoutError(error: unknown): boolean {
 
   return false;
 }
+
+export function validateGitHubTokenBrokerUrl(value: string): string {
```
hojinzs, line 284: **nit** — the isolation I asked for was done by copy, not by move.

`validateGitHubTokenBrokerUrl` plus `isPrivateIp` / `isPrivateIpv4` / `isPrivateIpv6` / `parseIpv4MappedIpv6` and the two blocked-host tables are now a second, near-verbatim copy of `tool-github-graphql/src/url-policy.ts`, which keeps its own set for `validateGitHubGraphQLApiUrl`. Roughly 104 duplicated lines of SSRF policy.

Behaviour is genuinely identical — I differentially tested both entry points across 19 URLs at `8ed8e27` (http, `localhost`, `127.0.0.1`, `[::1]`, `10/172.16/172.32/192.168/100.64/169.254`, `224.0.0.1`, `[fe80::1]`, `[fd00::1]`, `[::ffff:10.0.0.1]`, `metadata.google.internal`, `.local`, trailing-dot host, malformed): **0 mismatches**. So no regression, and the dangling export is gone, which is what I actually asked for.

The residual cost is drift: two copies of a private-network allowlist that must be patched together. Since #903 deletes this copy along with the resolver, that window is short and I'm not blocking on it — but a shared `validateSecureUrl` in `core` would have got the same isolation without the second copy, if you'd rather do it now.

---

packages/runtime-claude/src/mcp-compose.ts
```diff
@@ -7,15 +7,11 @@ import {
...
-  stripMcpServerSecretEnvironmentValues,
 } from "@gh-symphony/core";
 
 export type ClaudeMcpTokenEnvironment = {
   GITHUB_GRAPHQL_TOKEN?: string;
   GITHUB_GRAPHQL_API_URL?: string;
-  GITHUB_TOKEN_BROKER_URL?: string;
-  GITHUB_TOKEN_BROKER_SECRET?: string;
-  GITHUB_TOKEN_CACHE_PATH?: string;
   GITHUB_PROJECT_ID?: string;
```
hojinzs, line 15: **nit** — removing the `stripMcpServerSecretEnvironmentValues` call is correct, and it leaves one orphan behind.

To be explicit that this is not a boundary regression: on `main` that block ran *after* `mcpServers` had already been reassigned to `{ symphony }` or `{}` (lines 62-65 here, same ordering there). The `symphony` entry is `type: "http"` with `headers`, and the strip only rewrites `env` values — so it was a no-op before it was deleted. Nothing that reaches the child changed.

Two leftovers from the same deletion:

- `stripMcpServerSecretEnvironmentValues` (`core/src/runtime/mcp-compose.ts:67`) now has zero production callers — only `core/src/runtime/mcp-compose.test.ts` imports it. Delete it, or keep it deliberately if #903 will need it.
- `SYMPHONY_TRACKER_SECRET_ENVIRONMENT_NAMES` stays on `ClaudeMcpTokenEnvironment` (line 23) but is no longer read in this file; it is only consumed by `adapter.ts`'s own `readTrackerSecretEnvironmentNames`. Harmless pass-through, just no longer load-bearing here.

---

hojinzs (approved):
## Approve — all five findings from the previous round verified fixed; one docs one-liner left

Re-reviewed at `8ed8e27`. Every prior finding is genuinely resolved, not just claimed. I verified each by execution rather than by reading the follow-up commits.

### Previous round — verified

| # | Finding | Status |
|---|---|---|
| P2 | Broker names dropped from `secretEnvironmentNames()` narrowed the indirect `$VAR` child strip | **Fixed & proven.** With a trusted `.mcp.json` containing `{"GITHUB_TOKEN_BROKER_SECRET": "$MY_BROKER_SECRET", "GITHUB_GRAPHQL_TOKEN": "$MY_DIRECT_TOKEN"}`, `collectMcpSecretEnvironmentNames` now returns `["MY_BROKER_SECRET","MY_DIRECT_TOKEN"]` — it returned `[]` at `ec1a0ff`. Declaration is back to all seven names, with exact-match tests. |
| P2 | Docs claimed a broker compatibility path that no longer exists | **Fixed in 2 of 3 places** — see the README comment below. |
| P3 | Dead `usesGitHubTokenBroker` blanking in `buildCodexRuntimePlan` | Fixed; `config.githubToken` passed directly. |
| P3 | Git helper still imported broker URL policy from `tool-github-graphql` | Fixed, by duplication rather than a move. Differentially tested — see the inline nit. |
| nit | `return {} as Record<string, string>` | Fixed; the explicit return annotation in `8ed8e27` is the right call over the assertion. |

### Acceptance criteria (#902)

- **AC1 — direct token drives both polling and the host-owned tool.** Proven both ways at `8ed8e27`: `resolveWorkerCredentials` selects project-scope first (`project-tok` over `daemon-tok`), falls back to daemon, and returns `{}` for broker-only; `executeAgentTool("github_graphql", …)` executed with `Authorization: Bearer direct-token` and returned the payload, while a broker-only environment fails with `GITHUB_GRAPHQL_TOKEN is required.` ✅
- **AC2 — four direct names still declared.** ✅ (plus the three broker names restored)
- **AC3 — child boundary unchanged.** ✅ The indirect path is restored per the table above; the direct strip was never at risk (`CUSTOM_RUNTIME_RESERVED_AUTH_ENVIRONMENT_NAMES` covers all three names unconditionally).
- **AC4 — Git publication untouched.** ✅ Resolver relocated but behaviour-identical; broker branch and `GITHUB_TOKEN_BROKER_TIMEOUT_MS` retained for #903.
- **AC5 — changeset.** ✅

### Smoke / validation at `8ed8e27`

- `pnpm build`, `pnpm typecheck`, `pnpm lint`, `pnpm test` — all pass locally (Node 22; repo targets 24+).
- CI on this head: **Test ✅, Container Smoke ✅.**
- Git publication precedence, executed against a stubbed broker: both credentials → `direct-token`, broker never contacted; broker-only → brokered token, broker contacted. Confirms the effective behaviour the corrected docs now describe.
- Broker URL policy, old vs. new implementation across 19 URLs (scheme, loopback, RFC1918, CGNAT, link-local, multicast, IPv6 ULA/link-local, IPv4-mapped IPv6, metadata hosts, `.local`, trailing-dot, malformed): **0 mismatches.**

### Scope

No overengineering found. `GITHUB_TOKEN_BROKER_TIMEOUT_MS`, the agent-provider broker, and the Git publication mechanics are all correctly left to #903 / #847. The one piece of unrequested motion — relocating the Git broker resolver — was in response to my own P3, so it counts as review-driven, not scope creep.

### Left open

One P3 (root `README.md`) and two nits inline. The README line is the uncorrected third copy of a sentence you already fixed elsewhere, and it misleads operators about broker-only publication — please land that one-liner before merge. The nits are optional and both are fine to defer to #903.

Note for the merge checklist: this makes `GITHUB_GRAPHQL_TOKEN` mandatory for every GitHub project, so a broker-only deployment stops starting on upgrade. The changeset is `minor`; that is your release-policy call, but it is worth being deliberate about, and the release note should carry the push-scope requirement.

Approving — nothing here needs another review round.

---

PR #924, opened 2026-09-05
refactor(runtimes): remove agent credential broker

### LINKED ISSUE #904: refactor(runtimes): remove the agent provider credential broker and residual plumbing
Phase 4 of epic #847. 브로커 제거 3분할의 **마지막**입니다. 근거는 `docs/reports/2026-09-05-credential-broker-footprint.md`.

## 무엇을 하는가

에이전트 제공자 브로커와 잔여 배선을 제거합니다. GitHub 토큰 브로커와는 다른 프로토콜입니다 — 응답 형태가 `{ env, expires_at? }`이고(`packages/core/src/runtime/adapter.ts:3-8`) 트래커가 아니라 모델 제공자 인증을 나릅니다.

**이것이 마지막인 이유:** 트래커나 발행 회귀를 복구하는 데 아무 도움이 되지 않으므로 앞의 둘과 롤아웃을 함께할 이유가 없습니다.

## 두 런타임이 서로 다릅니다

스파이크가 정정한 사실이고, 이 이슈의 범위를 가릅니다.

- **Codex는 실제로 소비합니다.** 엔드포인트에 POST하고 캐시를 읽거나 쓰며 허용된 OpenAI 변수를 추출해 런타임 계획에 주입합니다 (`packages/runtime-codex/src/runtime.ts:704-717, 914-1009`, `packages/core/src/runtime/credentials.ts:7-29, 48-100`).
- **Claude는 검증만 합니다.** 워커 시작 시 5초 타임아웃 프로브가 응답에 `ANTHROPIC_API_KEY`가 있는지 확인하지만(`packages/worker/src/index.ts:559-576`, `packages/runtime-claude/src/preflight.ts:305-343`), spawn 환경은 그 값을 주입하지 않고 `AGENT_CREDENTIAL_BROKER_*`를 전부 지웁니다(`packages/runtime-claude/src/adapter.ts:685-693, 716-743`). **Claude는 브로커가 준 키를 애초에 받지 않습니다.**

## 대체물

트래커 크리덴셜이 아니므로 §1107–1110의 요구 대상이 아닙니다.

- Codex — 기존 로컬 `auth.json` 스테이징 또는 런타임별 환경의 직접 `OPENAI_API_KEY` (`runtime.ts:704-731, 940-945`)
- Claude — 직접 `ANTHROPIC_API_KEY`, 또는 bare 모드가 아니면 기존 스테이지된 로컬 로그인 (`preflight.ts:251-303`, `adapter.ts:468-476`)

## 완료 조건

- [ ] Codex의 `{ env, expires_at }` 페치/캐시 경로가 제거되고, 직접 제공자 크리덴셜과 로컬 로그인 동작은 유지된다.
- [ ] Claude의 검증 전용 preflight 프로브가 제거된다. 값을 쓰지 않으므로 시작을 막을 이유가 없다.
- [ ] 범용 브로커 응답 타입, 예약 이름 잔재, 관련 테스트와 문서가 정리된다.
- [ ] `CUSTOM_RUNTIME_RESERVED_AUTH_ENVIRONMENT_NAMES`의 **백스톱 성질은 유지된다.** #879가 다섯 라운드에 걸쳐 확정한 것이고, 빈 선언과 다른 트래커만 선언된 상태에서도 일곱 이름이 차단되어야 한다.
- [ ] #872의 블랙박스 검증이 계속 통과한다.
- [ ] changeset 추가.

## 범위 밖

`CUSTOM_RUNTIME_RESERVED_AUTH_ENVIRONMENT_NAMES` 자체의 삭제 — #879에서 명시적으로 유지하기로 결정했습니다.

## 후속으로 열리는 것

이 이슈가 닫히면 P5‑2(설정 레퍼런스 통합)의 선행 조건이 충족됩니다. `docs/configuration.md:434`의 "Credential Brokers And Git Access" 절이 지워질 대상으로 확정되기 때문입니다.

moncher-dev:
## Issues

- Closes #904

## Summary

- Removes the agent provider credential broker protocol and its remaining Codex, Claude, worker, orchestrator, and CLI plumbing.
- Keeps provider authentication direct: Codex uses configured OpenAI environment variables or staged local `auth.json`; Claude uses `ANTHROPIC_API_KEY` in bare mode or staged local login otherwise.
- Preserves the seven-name custom-runtime authentication backstop and keeps the three retired provider-broker names in the independent agent-child denylist so stale host values cannot leak.

## Change-point diagram

- Codex launcher → extracts allowed direct OpenAI variables → runtime plan or staged local `auth.json`
- Claude preflight → validates direct/local authentication only → no validation-only broker network probe
- Core runtime contract → no generic broker response/resolver/cache → retired broker names remain strip-only child-boundary guards
- Worker/orchestrator/CLI → no provider-broker dependency injection or doctor plumbing

## Start here

- `packages/core/src/runtime/custom-child-env.ts:49` — the agent-child denylist and retained reserved-auth backstop.
- `packages/runtime-codex/src/runtime.ts:634` — Codex child-boundary filtering and direct/local authentication staging.
- `packages/runtime-claude/src/preflight.ts:251` — Claude direct/local authentication preflight without a broker probe.
- `packages/runtime-codex/src/launcher.ts:100` — direct OpenAI environment extraction at launcher startup.

- `AGENT_CREDENTIAL_BROKER_URL` and `AGENT_CREDENTIAL_BROKER_SECRET` no longer configure or gate either provider runtime.
- Stale `AGENT_CREDENTIAL_BROKER_URL`, `AGENT_CREDENTIAL_BROKER_SECRET`, and `AGENT_CREDENTIAL_CACHE_PATH` host values are still stripped at every agent-child boundary.
- Codex continues to accept direct OpenAI credentials and staged local login; Claude continues to accept direct API-key or staged local-login authentication.
- Existing tracker credentials and host-owned GitHub publication are unaffected because they use separate protocols.

## Validation

- `pnpm lint` — pass.
- `pnpm test` — pass (all 14 workspace projects; 1,677 tests).
- `pnpm typecheck` — pass.
- `pnpm build` — pass.
- `pnpm exec vitest run packages/core/src/runtime/custom-child-env.test.ts` — pass (9 tests).
- `pnpm test:worker-startup` — pass (10 black-box tests, including the #872 child-environment contract).
- `pnpm e2e:claude` — documented environment exception #925; two isolated attempts exited before scenario startup because the Docker daemon socket was unavailable.
- `pnpm format` — repository baseline failure on 68 pre-existing files outside this change; no files were modified.

## Changeset

- `.changeset/bright-tigers-auth.md` (`@gh-symphony/cli` patch).

## Risks & rollback

- Operators still relying on provider-broker authentication must migrate to direct provider credentials or local login staging; old variables remain safely stripped but are not consumed.
- Reviewer focus: authentication precedence and the distinction between provider credentials, strip-only child guards, and tracker/GitHub publication credentials.
- Rollback is the squash-revert of this PR; no persistent data migration is involved.

## Changed files

- `packages/runtime-codex/src/{runtime,launcher}.ts` and tests — remove broker fetch/cache/reporting while retaining direct credentials and local auth staging.
- `packages/runtime-claude/src/{preflight,adapter}.ts` and tests — remove the validation-only probe and obsolete adapter plumbing.
- `packages/core/src/runtime/{adapter,credentials,custom-child-env}.ts` and tests — remove the generic broker contract/cache, preserve the reserved-auth backstop, and retain retired broker names as child-boundary denylist entries.
- `packages/worker/src/non-codex-runtime.ts`, `packages/orchestrator/src/runtime-factory.ts`, and CLI doctor files — remove residual dependency plumbing.
- `docs/configuration.md`, `docs/architecture.md`, `AGENT_TEST.md`, and the historical Claude ADR — document direct provider authentication and the superseded design.
- `.changeset/bright-tigers-auth.md` — records the CLI/runtime behavior change.

- [ ] None beyond normal review and CI; Docker environment follow-up is tracked in #925.

## Security


packages/core/src/runtime/custom-child-env.ts
```diff
@@ -47,9 +47,6 @@ export function readAgentVisibleSymphonyContext(
 
 /** Host credential plumbing that must never cross an agent-child boundary. */
 export const AGENT_CHILD_CREDENTIAL_ENVIRONMENT_NAMES = [
-  "AGENT_CREDENTIAL_BROKER_URL",
-  "AGENT_CREDENTIAL_BROKER_SECRET",
```
moncher-dev, line 51: **Blocking.** These three are also deleted from `CUSTOM_RUNTIME_RESERVED_AUTH_ENVIRONMENT_NAMES` below, so after this PR no list contains them and nothing strips them from the agent child.

The comment three lines up states the invariant: *host credential plumbing that must never cross an agent-child boundary*. A retired variable that an un-migrated operator still exports is exactly the case this list is for — removing the reader is in scope, removing the stripper is not.

Proof at `d3b1b6fa`, from the #872 black-box suite:

```
x runs a real custom child with 'compatibility mode' credential semantics
  -> expected 'agent-broker-secret' to be null
     received "agent-broker-secret"
```

The secret reached the real child process. Please keep these three entries here.

packages/core/src/runtime/custom-child-env.test.ts
```diff
@@ -51,7 +50,6 @@ describe("buildCustomRuntimeChildEnvironment", () => {
       "GITHUB_TOKEN",
       "LINEAR_API_KEY",
       "GITHUB_TOKEN_BROKER_SECRET",
-      "AGENT_CREDENTIAL_BROKER_SECRET",
```
moncher-dev, line 54: **Blocking (same defect).** Deleting `AGENT_CREDENTIAL_BROKER_SECRET` from the expected-stripped list — and from the injected environment at line 25 — is what kept this suite green while the child boundary regressed.

The assertion was not stale; it was the one guarding the behaviour the PR changed. `pnpm test:worker-startup` catches it in 16s without Docker. Please restore both entries rather than removing them.

packages/core/src/runtime/custom-child-env.test.ts
```diff
@@ -155,6 +153,27 @@ describe("buildCustomRuntimeChildEnvironment", () => {
...
+      "LINEAR_AUTHORIZATION",
+    ];
+
+    for (const name of names) {
+      expect(
+        isCustomRuntimeReservedAuthEnvironmentName(name, {}, declarations)
+      ).toBe(true);
+    }
+  });
+
```
moncher-dev, line 178: This new case is good, and it is real — I verified it fails under two mutations of the constant it guards (deleting `"GITHUB_TOKEN"` from the list, and gating the lookup on a non-empty declaration, which is the #879 leak shape). All seven names measure `true` under both an empty and a Linear-only declaration at this head. No change requested here.

moncher-dev (commented):
## Blocking: the agent-child boundary now leaks `AGENT_CREDENTIAL_BROKER_SECRET`, and #872's black-box contract fails

Most of this PR is right. The seven-name backstop survives, and the new tests are real (I mutated the source they guard and all three went red). The blocker is narrow and mechanical: the three `AGENT_CREDENTIAL_*` names were removed from **both** reserved lists, so nothing strips them from the agent child any more.

### 1. Seven-name backstop: PASS

Measured at head `d3b1b6fa`, `isCustomRuntimeReservedAuthEnvironmentName(name, {}, [])` and the Linear-only declaration, 14/14 `true`:

```
EMPTY        GH_TOKEN true | GH_ENTERPRISE_TOKEN true | GITHUB_TOKEN true | GITHUB_GRAPHQL_TOKEN true
             GITHUB_TOKEN_BROKER_SECRET true | LINEAR_API_KEY true | LINEAR_AUTHORIZATION true
LINEAR-ONLY  (same seven) true x7
RESERVED: ["GH_ENTERPRISE_TOKEN","GH_TOKEN","GITHUB_GRAPHQL_TOKEN","GITHUB_TOKEN","GITHUB_TOKEN_BROKER_SECRET","LINEAR_API_KEY","LINEAR_AUTHORIZATION"]
```

Dropping `GITHUB_TOKEN_BROKER_URL` / `GITHUB_TOKEN_CACHE_PATH` from `CUSTOM_RUNTIME_RESERVED_AUTH_ENVIRONMENT_NAMES` is fine — both are still in `AGENT_CHILD_CREDENTIAL_ENVIRONMENT_NAMES`, which the predicate ORs first, so they stay reserved. That is deduplication, not a weakening.

### 2. Blocking: `AGENT_CREDENTIAL_*` is stripped by nothing

The three names were deleted from `AGENT_CHILD_CREDENTIAL_ENVIRONMENT_NAMES` **and** from `CUSTOM_RUNTIME_RESERVED_AUTH_ENVIRONMENT_NAMES`. There is no remaining list that contains them, so an operator who has not yet migrated — the exact population the "Risks & rollback" section names — hands a live provider secret straight to the agent child.

Removing the plumbing that *reads* those variables is correct and in scope. Removing the guard that *strips* them is not: the guard exists precisely for values the host still has set but no longer uses.

### 3. Blocking: #872's black-box verification does not pass

Completion condition 5 of #904 is unmet. Reproduced locally at `d3b1b6fa` with no Docker, and CI's `Test` job fails identically (`Built worker startup regressions` step, run 33989432012):

```
$ pnpm build && pnpm test:worker-startup

 x runs a real custom child with 'compatibility mode' credential semantics
   -> expected 'agent-broker-secret' to be null
 x enforces the exported agent-child environment contract
   -> expected [ 'GITHUB_GIT_HOST', ...(18) ] to deeply equal [ Array(22) ]
       - "AGENT_CREDENTIAL_BROKER_URL"
       - "AGENT_CREDENTIAL_BROKER_SECRET"
       - "AGENT_CREDENTIAL_CACHE_PATH"
 x publishes and protects the Codex worker lifecycle with a direct host credential
   -> (same 22-vs-19 contract mismatch)

 Tests  3 failed | 7 passed (10)
```

The first failure is the leak itself, observed end to end: `agent-broker-secret` reached the real child process.

**Why the Validation section missed it.** `pnpm test` is `pnpm -r test` and does not include `test:worker-startup`, which is a separate root script and a separate CI step. It is also **not** Docker-dependent — it uses the stub `claude` binary, ran in 16s in CI, and 7 of its 10 tests passed here on this Docker-less host. So #925 does not cover this, and the `pnpm e2e:claude` exception does not either. Please run `pnpm test:worker-startup` before the next push.

### 4. An existing assertion was loosened over the leak

`packages/core/src/runtime/custom-child-env.test.ts` had `AGENT_CREDENTIAL_BROKER_SECRET` in both the injected environment and the expected-stripped list. Both entries were deleted in this PR. That is what let the unit suite stay green while the behaviour regressed; the black-box test, which was not run, caught it.

### 5. Mutation results on the new tests (all genuine)

| Mutation | Result |
|---|---|
| Delete `"GITHUB_TOKEN"` from `CUSTOM_RUNTIME_RESERVED_AUTH_ENVIRONMENT_NAMES` | RED — both new `it.each` cases fail |
| Gate the constant lookup on `trackerSecretEnvironmentNames.length > 0` (the #879 leak shape) | RED — empty-declaration case fails |
| Restore the three `agentCredential*` fields in `resolveLocalRuntimeLaunchConfig` | RED — `ignores removed agent credential broker variables` fails |

### Suggested fix

Keep the three `AGENT_CREDENTIAL_*` names in `AGENT_CHILD_CREDENTIAL_ENVIRONMENT_NAMES` (retiring a variable is a reason to strip it, not to stop stripping it), restore the two deleted `custom-child-env.test.ts` entries, and leave `test/e2e/claude/claude-docker.spec.ts` untouched so its 22-name contract keeps holding. Everything else in the PR — the Codex fetch/cache removal, the Claude validation-only probe removal, the launcher and doctor cleanup, the docs and the changeset — reads correct to me.

If instead the intent is that these names are dead enough to stop stripping, that needs to be argued explicitly against #872's contract rather than done by deleting assertions, and it is a scope question for @hojinzs, not a silent change.

packages/core/src/runtime/custom-child-env.ts
```diff
@@ -47,9 +47,6 @@ export function readAgentVisibleSymphonyContext(
...
-  "AGENT_CREDENTIAL_CACHE_PATH",
   "GITHUB_GIT_HOST",
```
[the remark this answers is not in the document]
moncher-dev: Resolved in `0ceb1a9`: the three retired provider-broker names are restored to `AGENT_CHILD_CREDENTIAL_ENVIRONMENT_NAMES`. They remain strip-only guards; no broker reader, cache, configuration, or injection path was restored. `pnpm test:worker-startup` now passes all 10 scenarios.

packages/core/src/runtime/custom-child-env.ts
moncher-dev: Resolved in `0ceb1a9`: `AGENT_CREDENTIAL_BROKER_URL`, `AGENT_CREDENTIAL_BROKER_SECRET`, and `AGENT_CREDENTIAL_CACHE_PATH` are again in the independent agent-child denylist, preventing stale operator values from reaching Claude or custom runtime children.

packages/core/src/runtime/custom-child-env.test.ts
moncher-dev: Resolved in `0ceb1a9`: the injected `AGENT_CREDENTIAL_BROKER_SECRET` fixture and expected-stripped assertion were restored. The focused unit test passes (9 tests), and `pnpm test:worker-startup` passes all 10 black-box tests.

packages/core/src/runtime/custom-child-env.test.ts
moncher-dev: Confirmed and retained unchanged. The seven-name fallback coverage still exercises both empty and unrelated-tracker declarations; the rework only restores the separate retired-name child denylist.

packages/core/src/runtime/custom-child-env.ts
```diff
@@ -77,16 +77,11 @@ export const AGENT_CHILD_CREDENTIAL_ENVIRONMENT_NAMES = [
  * used to identify which credentials belong to the active tracker.
  */
 export const CUSTOM_RUNTIME_RESERVED_AUTH_ENVIRONMENT_NAMES = [
-  "AGENT_CREDENTIAL_BROKER_URL",
-  "AGENT_CREDENTIAL_BROKER_SECRET",
```
moncher-dev, line 81: Resolved, and cleanly. These three moved back into `AGENT_CHILD_CREDENTIAL_ENVIRONMENT_NAMES` rather than staying here, so they are reserved again transitively (the predicate ORs `isAgentChildCredentialEnvironmentName` first) without widening the seven-name constant. `AGENT_CHILD_CREDENTIAL_ENVIRONMENT_NAMES` now has no diff hunk at all against `1a20ef0f` — it is byte-identical to pre-PR.

I applied the exact regression again on this head (re-delete the three from the child denylist): the #872 black-box suite goes red with `expected 'agent-broker-secret' to be null`, 3 failed. So the fix is genuinely guarded.

Note for whoever touches this next: the *unit* suite stays green under that same mutation (9 passed). The denylist only does load-bearing work on the `inheritEnvironment: true` path, while `custom-child-env.test.ts`'s restored assertion sits in the default-isolation case, where the name is absent by construction anyway. `test/e2e/claude/claude-docker.spec.ts` is the only real guard on these three names — worth a follow-up unit case, not a change here.

Dropping `GITHUB_TOKEN_BROKER_URL` / `GITHUB_TOKEN_CACHE_PATH` from this constant stays fine: both remain in the child denylist, so they are still reserved. Deduplication, not weakening.

moncher-dev (commented):
## Re-review at `0ceb1a9c` — resolved. Handing to @hojinzs.

The rework is exactly the suggested fix and nothing else: three names restored to `AGENT_CHILD_CREDENTIAL_ENVIRONMENT_NAMES`, two test lines restored, `test/e2e/claude/claude-docker.spec.ts` untouched. All evidence below is from running the new head, not reading it.

### 1. The leak is closed, verified end to end

`pnpm build && pnpm test:worker-startup` on `0ceb1a9c`, Docker-less host: **10 passed (10)**. The compatibility-mode child test that reported `expected 'agent-broker-secret' to be null` at `d3b1b6fa` now passes, and the exported contract is back to 22 names.

### 2. The test was restored, not reached green another way

`git diff 1a20ef0f..0ceb1a9c -- packages/core/src/runtime/custom-child-env.test.ts` is **purely additive**: the only change against pre-PR is the new seven-name `it.each` block. Both previously deleted lines — `AGENT_CREDENTIAL_BROKER_SECRET` in the injected environment (L25) and in the expected-stripped list (L54) — are back verbatim. No assertion anywhere in this PR is weaker than pre-PR.

### 3. One correction to my own last review, worth recording

I wrote that the loosened unit assertion was what kept the suite green over the regression. Measured, that is not quite right, and the real picture is more interesting.

I re-applied the exact regression on top of `0ceb1a9c` (M4: delete the three names from the child denylist again):

| Suite | Result under M4 |
|---|---|
| `packages/core/src/runtime/custom-child-env.test.ts` | **9 passed** — does not catch it |
| `pnpm test:worker-startup` (#872 black-box) | **3 failed** — `expected 'agent-broker-secret' to be null` |

The restored L54 entry sits in the *default-isolation* case, where `buildCustomRuntimeChildEnvironment` builds `env` from an allowlist, so the name is absent by construction whether or not the denylist contains it. The denylist only does load-bearing work on the `inheritEnvironment: true` (compatibility mode) path, and the only test exercising that path against these names is the black-box one.

So restoring those two lines is right as a statement of intent, but **#872's black-box suite is the sole real guard here**. That is not a change request on this PR — behaviour is correct and coverage is no worse than pre-PR — but it is a live gap: a unit case asserting the compatibility-mode denylist would fail fast, and this defect class has now shipped three times (#879 twice, and once here). Worth a follow-up issue rather than more churn on #924.

### 4. Everything that passed before still passes

**Seven-name backstop: 14/14.** Empty declaration and Linear-only declaration, all `true`:

```
SEVEN-NAME SCORE: 14 / 14
RESERVED CONST: ["GH_ENTERPRISE_TOKEN","GH_TOKEN","GITHUB_GRAPHQL_TOKEN","GITHUB_TOKEN",
                 "GITHUB_TOKEN_BROKER_SECRET","LINEAR_API_KEY","LINEAR_AUTHORIZATION"]
CHILD DENYLIST SIZE: 22
AGENT_CREDENTIAL_BROKER_URL     reserved=true  in child denylist=true
AGENT_CREDENTIAL_BROKER_SECRET  reserved=true  in child denylist=true
AGENT_CREDENTIAL_CACHE_PATH     reserved=true  in child denylist=true
```

The three broker names are reserved again transitively — the predicate ORs `isAgentChildCredentialEnvironmentName` first — without re-adding them to the seven-name constant. That is the right shape.

**Per-runtime scope intact.** Codex: `{ env, expires_at }` fetch/cache/reporting gone; direct OpenAI env extraction (`launcher.test.ts:21,30,39,70,81`) and `auth.json` staging (`runtime.test.ts:907,947`) still asserted. Claude: validation-only preflight probe gone; direct `ANTHROPIC_API_KEY` including the missing-key remediation (`preflight.test.ts:50-63`) and `claudeAiOauth` staging (`adapter.test.ts:526,548`) still asserted.

**Mutations, re-run on the new head — all four genuine:**

| Mutation | Result |
|---|---|
| Delete `"GITHUB_TOKEN"` from the reserved constant | RED (3 failed) |
| Gate the constant lookup on `trackerSecretEnvironmentNames.length > 0` (#879 leak shape) | RED (2 failed) |
| Restore the three `agentCredential*` fields in `resolveLocalRuntimeLaunchConfig` | RED (1 failed) |
| Re-delete the three names from the child denylist (the reverted regression) | RED in the black-box suite (3 failed) |

### 5. CI and suites

- `Test`: **pass** (2m34s) — including `Built worker startup regressions`, which failed 3/7 at the previous head.
- `Container Smoke`: **pass**.
- Full `pnpm test` locally: green across all 14 workspace projects.
- Changeset present (`.changeset/bright-tigers-auth.md`, `@gh-symphony/cli` patch).

No further findings from me. @hojinzs over to you — the one judgement call left is item 3, and it is a follow-up, not a blocker on this PR.

packages/core/src/runtime/custom-child-env.ts
```diff
@@ -77,16 +77,11 @@ export const AGENT_CHILD_CREDENTIAL_ENVIRONMENT_NAMES = [
  * used to identify which credentials belong to the active tracker.
  */
 export const CUSTOM_RUNTIME_RESERVED_AUTH_ENVIRONMENT_NAMES = [
```
hojinzs, line 79: **P3 — mark the three retired names as intentionally strip-only.**

The deduplication here is correct and I verified it does not weaken anything: `isCustomRuntimeReservedAuthEnvironmentName` ORs `isAgentChildCredentialEnvironmentName` first, and both real strip paths (`runtime-codex/src/runtime.ts:640`, `runtime-claude/src/adapter.ts:652,679`) call `stripCredentialEnvironmentForAgentChild` right after spreading this constant. I also checked the re-assignment ordering: `buildAgentChildEnvironmentAssignments` runs *after* the strip, but it only draws from the fixed `AGENT_VISIBLE_SYMPHONY_CONTEXT_ENVIRONMENT_NAMES` allowlist, so no credential name can be re-added. No behavioral gap.

The suggestion is about durability, not correctness. The three `AGENT_CREDENTIAL_*` entries now sitting in `AGENT_CHILD_CREDENTIAL_ENVIRONMENT_NAMES` above have no reader anywhere in the tree — they exist purely to strip stale operator values. Nothing in the code says so, and that is precisely the regression this PR already hit once: round 1 deleted them as apparent dead code, and only the black-box suite caught it.

A one-line marker would make the next reader's decision obvious:

```ts
/** Host credential plumbing that must never cross an agent-child boundary. */
export const AGENT_CHILD_CREDENTIAL_ENVIRONMENT_NAMES = [
  // Retired agent-provider broker names: no reader remains, retained strip-only
  // so stale operator values cannot reach an agent child (#904).
  "AGENT_CREDENTIAL_BROKER_URL",
  ...
```

Non-blocking.

---

packages/runtime-codex/src/runtime.ts
```diff
@@ -843,119 +819,9 @@ export function createGitCredentialHelperEnvironment(
...
-    AgentRuntimeAdapter<
-      CodexRuntimePrepareContext,
-      CodexRuntimeTurnInput,
-      CodexRuntimeTurnResult,
-      CodexRuntimeEvent,
-      CodexRuntimeCredentialBrokerResponse
-    >,
-    "resolveCredentials"
-  >
+  config: Pick<CodexRuntimeConfig, "workingDirectory" | "agentEnv">
```
hojinzs, line 822: **P3 — `workingDirectory` is no longer read by this function.**

After the broker path is removed the body is just `return resolvePreparedAgentEnvironment(config.agentEnv)`, so `workingDirectory` in the `Pick<>` is dead surface. It also leaves the function as an `async` wrapper over a synchronous call.

I confirmed the behavior is otherwise preserved exactly: the old code took `resolveRuntimeCredentials({ env: config.agentEnv })` → `resolvePreparedAgentEnvironment(config.agentEnv)` when `agentEnv` was set, and `resolvePreparedAgentEnvironment()` with no argument otherwise — which is identical to passing `undefined`. No change in outcome for either branch.

Narrowing the `Pick` to `"agentEnv"` would be accurate. Whether to keep it `async` is a call about the exported signature; leaving it is fine if callers should not have to change. Non-blocking either way.

---

docs/configuration.md
```diff
@@ -468,24 +468,21 @@ operator's credential stores.
 | `LINEAR_AUTHORIZATION`   | unset                                                                                                            | Linear tracker, worker host tools                                      | Advanced                                            | Optional raw Linear authorization value for host-side Linear operations; it takes priority over `LINEAR_API_KEY` and is not inherited by agent children by default. |
 | `LINEAR_GRAPHQL_URL`     | `https://api.linear.app/graphql` when the Linear tool is enabled                                                 | Codex runtime, Claude runtime                                          | User-facing for Linear Enterprise/proxy setups      | Overrides the Linear GraphQL endpoint.                                                                                                                              |
 
-## Credential Brokers And Git Access
+## Git Access
```
hojinzs, line 471: **nit — the retained `AGENT_CREDENTIAL_*` names lose their documentation while the GitHub ones keep theirs.**

The section scoping is right: renaming to `## Git Access` and dropping only the three agent-provider rows, while leaving `GITHUB_TOKEN_BROKER_*` for P5-2, matches the "범위 밖" boundary in #904.

The asymmetry is that `GITHUB_TOKEN_BROKER_URL/SECRET/CACHE_PATH` each keep a row reading *"Retained only as a reserved credential name … stripped from coding-agent children"*, while `AGENT_CREDENTIAL_BROKER_URL/SECRET` and `AGENT_CREDENTIAL_CACHE_PATH` are removed outright — even though, after the `0ceb1a9` fix, all six have the same status: no reader, still stripped at every agent-child boundary.

An operator who still has these set gets no doc anywhere telling them the values are inert but safely stripped, which is exactly what the "Risks & rollback" section says that population needs to know. Three "Reserved" rows matching the existing GitHub ones would close it.

Non-blocking.

---

hojinzs (approved):
## Approve — round-1 blocker is genuinely fixed, independently verified

I re-verified the round-1 blocker end to end rather than taking the resolution comments at face value, and it holds. Remaining findings are P3/nit only, so per the anti-ping-pong rule this is an approve rather than another round.

### The blocker is closed (verified, not just asserted)

`0ceb1a9` restores the three names to `AGENT_CHILD_CREDENTIAL_ENVIRONMENT_NAMES` and restores the two deleted `custom-child-env.test.ts` entries. I checked the two things that could have made that fix cosmetic:

1. **Both child-boundary call sites actually reach the restored list.** `runtime-codex/src/runtime.ts:640` and `runtime-claude/src/adapter.ts:652,679` each call `stripCredentialEnvironmentForAgentChild` after spreading `CUSTOM_RUNTIME_RESERVED_AUTH_ENVIRONMENT_NAMES`, so dropping the five deduplicated names from that constant weakens neither path.
2. **The post-strip re-assignment cannot undo it.** `buildAgentChildEnvironmentAssignments` runs *after* the strip and takes `excludeNames` from a set that no longer contains these names — but it only copies from the fixed `AGENT_VISIBLE_SYMPHONY_CONTEXT_ENVIRONMENT_NAMES` allowlist, which holds no credential names. No re-add path exists.

Direct runtime check against the built artifacts:

```
after strip: {"SAFE_VALUE":"keep","GIT_TERMINAL_PROMPT":"0"}
reserved-predicate(AGENT_CREDENTIAL_BROKER_URL, empty decl): true
reserved-predicate(AGENT_CREDENTIAL_BROKER_SECRET, empty decl): true
reserved-predicate(AGENT_CREDENTIAL_CACHE_PATH, empty decl): true
```

### #872 black-box suite: 10/10

The suite that was 3-red last round — the one the previous Validation section missed because `pnpm test` does not include it — now passes in full, including `enforces the exported agent-child environment contract` (22-name contract intact) and `publishes and protects the Codex worker lifecycle with a direct host credential`:

```
$ pnpm build && pnpm test:worker-startup
 ✓ test/e2e/claude/claude-docker.spec.ts (10 tests) 20613ms
 Test Files  1 passed (1) | Tests  10 passed (10)
```

`pnpm lint`, `pnpm test`, `pnpm typecheck`, `pnpm build` all pass locally; CI `Test` and `Container Smoke` are both green on `0ceb1a9`.

### Smoke test — user-visible auth behavior

Exercised `runClaudePreflight` against the built package across the four operator states:

| State | Result |
|---|---|
| bare + `ANTHROPIC_API_KEY` | `pass` |
| bare + stale broker vars only | `fail` — "ANTHROPIC_API_KEY is not configured" |
| non-bare + nothing | `warn` — local login may be used |
| non-bare + partial broker vars | `warn` |

This matches the documented intent. Two deliberate changes for un-migrated operators, both correctly called out in the PR body and changeset: a broker no longer satisfies bare mode, and partial broker config now warns instead of failing (it is simply meaningless now rather than a misconfiguration).

### Requirements from #904 — all six met

| Completion condition | Status |
|---|---|
| Codex `{ env, expires_at }` fetch/cache removed; direct creds + local login kept | ✅ verified behavior-identical for both branches (see inline on `runtime.ts:822`) |
| Claude validation-only preflight probe removed | ✅ `fetchImpl` dep and 5s probe gone; still used elsewhere in `doctor.ts`, not left dangling |
| Generic broker type / reserved-name residue / tests / docs cleaned | ✅ no `AgentRuntimeEnv`, `AgentRuntimeCredentialBrokerResponse`, or `agentCredential*` residue outside the intentional denylist and its guarding tests |
| Seven-name backstop preserved under empty and other-tracker declarations | ✅ new `it.each` covers both; predicate confirmed `true` for all seven |
| #872 black-box verification still passes | ✅ 10/10 |
| Changeset added | ✅ `@gh-symphony/cli` patch |

### Scope

No over-engineering or unrequested decisions found. The doc edit is correctly bounded — renaming to `## Git Access` and removing only the agent-provider rows while leaving `GITHUB_TOKEN_BROKER_*` to P5-2 respects the "범위 밖" boundary in #904. `AgentRuntimeResolutionError` is retained and still used for command-parsing errors, not left as dead code.

### Findings (all non-blocking)

- **P3** `custom-child-env.ts` — the retired names now have no reader and no comment saying they are intentionally strip-only; that is the exact shape of the round-1 regression.
- **P3** `runtime.ts:822` — `workingDirectory` is no longer read by `resolveAgentRuntimeEnvironment`.
- **nit** `docs/configuration.md` — retained `AGENT_CREDENTIAL_*` names lose their doc rows while the equivalent `GITHUB_TOKEN_BROKER_*` names keep theirs.

None of these need another round. Land it, and fold them into the P5-2 follow-up if convenient.

---

Issue #929, opened 2026-09-06
fix(orchestrator): a missing workflow hook script hard-fails every dispatch and burns the retry budget


moncher-dev:
## Problem

A standalone project folder whose `WORKFLOW.md` declares `hooks.after_create` gets no way to obtain the hook script, and a missing script hard-fails every dispatch until the retry budget is spent.

## Evidence (production incident, 2026-09-06 12:26–12:28 KST)

The operator's project folder `~/moncher-stack/github-symphony` contains only `WORKFLOW.md`, `.env`, `.agent/` and `.runtime/`. `WORKFLOW.md:32` declares:

```yaml
  after_create: hooks/after_create.sh
```

The path is resolved against the project folder, which has no `hooks/`. Every dispatch attempt produced:

```
[orchestrator] starting after_create hook for hojinzs/github-symphony#928
[orchestrator] after_create hook failed for hojinzs/github-symphony#928: spawn /Users/steve/moncher-stack/github-symphony/hooks/after_create.sh ENOENT
[orchestrator] dispatch failed for hojinzs/github-symphony#928; retries suppressed: Error: after_create hook failure: ...
```

Both freshly promoted issues (#908, #928) burned all three retries within ~60 seconds and landed in `failureRetrySuppressedState: "Ready"`, i.e. permanently undispatchable without operator state surgery. Health flapped `idle → degraded → idle`, so the daemon looked healthy again while nothing could ever be dispatched.

This is not a one-off: since #860 removed repo-embedded mode, `after_create` is the only thing that populates a workspace, so a missing hook means **no issue in the project can ever run**.

## Why nothing caught it

- `gh-symphony project start` accepts a `WORKFLOW.md` referencing a hook script that does not exist.
- `doctor` has no check that declared hook scripts are present and executable.
- The failure is reported per-dispatch rather than as a configuration error, so it is spent against the per-issue retry budget instead of surfacing once as a project-level fault.

## Proposed direction

1. Validate declared hook scripts at project start / `doctor` time: exists, is a regular file or symlink to one, is executable. Fail loudly with the resolved absolute path.
2. Classify a missing-hook spawn `ENOENT` as a project configuration fault, not a per-issue run failure, so it does not consume the retry budget of every issue that happens to be Ready.
3. Document how a standalone project folder is expected to obtain hook scripts (copied, symlinked, or resolved relative to the workflow source rather than the project root) — `workflow init` currently ships no hooks directory.

## Acceptance criteria

[...]

Issue #928, opened 2026-09-06
fix(cli): port the #878 Land-bounce policy into the shipped land skill template


moncher-dev:
## Summary

The Land-bounce policy fix from #878 landed in `WORKFLOW.md` and in this repository's installed `.codex/skills/land/SKILL.md`, but never reached `packages/cli/src/skills/templates/land.ts` — the template `gh-symphony workflow init` ships to users. Every project initialized by the published CLI still gets the pre-#878 land skill.

The regression test that should have caught this reads the installed dogfood copy instead of the template, so it passes.

## Evidence

`packages/cli/src/skills/templates/land.test.ts:84,111` read the repository's own installed copy:

```ts
const landSkill = await repositoryFile(".codex/skills/land/SKILL.md");
```

Repointing those two reads at `generateLandSkill(context)` — the artifact the CLI actually ships — makes them fail. Strings present in the installed copy and absent from the template:

| assertion | installed `.codex/skills/land` | shipped template |
| --- | --- | --- |
| `## Merged-PR Precedence Guard` | present | present |
| `comments(first: 1)` | present | **absent** |
| ``approval's `submittedAt` `` | present | **absent** |
| `created at or before that approval is absorbed by the approval` | present | **absent** |
| `A deleted head branch is not rework after merge` | present | **absent** |

The template also has no `## Failure Handling` section at all, so the failure-classification ordering the test asserts cannot exist in it.

For contrast, the `gh-project` template *is* in sync — #919's `comment_body` prohibition reached both artifacts, and all seven `gh-project.test.ts` assertions pass against the generator.

## Impact

A user running `gh-symphony workflow init` today receives a land skill that classifies unresolved actionable review feedback as rework and bounces `Land` → `Ready`, which is precisely the defect #878 fixed. The fix is dogfood-only.

## Acceptance criteria

1. `packages/cli/src/skills/templates/land.ts` carries the #878 policy: the merged-PR precedence guard ordered ahead of pre-flight and failure classification, the approval-absorption gate comparing a thread's first `comments(first: 1)` `createdAt` against the approval review's `submittedAt`, and the deleted-head-branch clause.
2. The two `land.test.ts` tests at lines 84 and 111 assert against `generateLandSkill(context)`, not against any path under `.codex/skills/`. No test in the repository may read `.codex/skills/` to establish that shipped behavior is correct.
[...]

PR #932, opened 2026-09-06
refactor(cli): trim orphaned doctor checks

### LINKED ISSUE #908: refactor(cli): trim doctor checks orphaned by the Phase 4 removals
Phase 4 of epic #847. 승인된 계획은 #847의 2026-09-05 코멘트에 있습니다.

## 무엇을 하는가

앞선 Phase 4 삭제들이 고아로 만든 `doctor` 점검을 걷어내고, 남길 점검을 확정합니다.

## 이 이슈는 범위를 미리 못박지 않습니다

`packages/cli/src/commands/doctor.ts`는 3,143행입니다. 무엇이 고아가 되는지는 선행 이슈들이 확정돼야 알 수 있으므로, 이 이슈는 **선행 작업의 결과를 입력으로 받습니다.** 착수 시점에 아래를 실제로 조사해 목록을 만드는 것이 첫 단계입니다.

선행: #901(population→훅), #902·#903·#904(브로커 3분할), #905(격리·귀속), #906(인스턴스 레지스트리).

현재 확인된 것 하나 — 브로커 참조가 최소 1군데 있습니다(`doctor.ts`의 broker 언급).

## 남길 점검의 기준

**지우는 기준이 아니라 남기는 기준을 먼저 쓰세요.** 다음을 만족하면 남깁니다.

- 운영자가 고칠 수 있는 문제를 지목한다 (remediation이 실재한다)
- 그 문제가 실제로 디스패치를 막거나 조용히 잘못된 동작을 만든다
- 점검 대상이 Phase 4 이후에도 존재한다

세 조건 중 하나라도 어긋나면 지웁니다.

## 완료 조건

- [ ] 남길 점검과 지울 점검의 목록이 이슈 코멘트나 PR 본문에 근거와 함께 제시된다. **먼저 제시하고 나서 지운다.**
- [ ] 선행 이슈들이 제거한 대상을 점검하던 코드가 남지 않는다.
- [ ] 남은 점검이 전부 실제로 동작한다. `gh-symphony doctor`를 이 호스트에서 돌려 확인한다.
- [ ] `--smoke` 경로가 여전히 동작하거나, 동작하지 않는다면 그 이유가 기록된다.
- [ ] changeset 추가.

## 범위 밖

점검 결과의 출력 형식 변경, `doctor --json` 스키마 변경, 그리고 새 점검 추가. 이 이슈는 **덜어내는 것**만 합니다.

## 주의

과하게 지우기 쉬운 자리입니다. 남길 기준을 먼저 쓰라고 한 이유가 그것입니다.

moncher-dev:
## Issues

- Closes #908

## Summary

- TL;DR: remove the redundant `worker_github_credential` doctor diagnostic and the helper orphaned by that removal.
- Keep the surviving host authentication, tracker/project resolution, runtime, workflow, and smoke diagnostics unchanged.
- Preserve accurate credential precedence/push-scope guidance, align the mandatory test matrix, and ship a CLI patch changeset.

## Change-point diagram

- `gh-symphony doctor` → `gh_authentication` remains authoritative → redundant worker-credential diagnostic/remediation and its unused project-env-only helper are removed

## Start here

- `packages/cli/src/commands/doctor.ts:80` — retained doctor check identifier set
- `packages/cli/src/managed-project-environment.ts:31` — surviving managed-project environment helpers
- `packages/cli/src/commands/doctor.test.ts:1203` — standalone regression coverage for the removed check
- `packages/cli/README.md:207` — retained operator credential guidance

- `gh-symphony doctor` no longer emits `worker_github_credential`, whose failure branch was strictly contained in `gh_authentication`: a non-null host auth token made both checks pass, while null auth already failed the retained check.
- Accepted signal loss: a tenant configured only through project `.env` no longer gets the extra advisory `worker_github_credential: pass (source: project_env)` beside an already-failing host-auth diagnosis. Text and JSON container formats are unchanged.

## Validation

- `pnpm exec prettier --check packages/cli/src/managed-project-environment.ts packages/cli/src/commands/doctor.test.ts packages/cli/README.md .changeset/wise-doctors-trim.md` — pass
- `pnpm --filter @gh-symphony/cli test -- src/commands/doctor.test.ts` — pass (68 tests)
- `pnpm lint` — pass
- `pnpm test` — pass (all 14 tested workspace projects; CLI 567 tests, doctor 68 tests)
- `pnpm typecheck` — pass
- `pnpm build` — pass
- Prior host verification: built `doctor --json` emitted the retained check set without `worker_github_credential`; built `doctor --smoke --issue hojinzs/github-symphony#908` executed the smoke path and cleanly degraded on its managed-project prerequisite.
- Docker E2E / blackbox: not applicable; this is CLI diagnostic cleanup and documentation/test refinement with no integration behavior change.

## Changeset

- `.changeset/wise-doctors-trim.md` (`patch`)

## Risks & rollback

- Low risk: the runtime removal is limited to one redundant diagnostic, its remediation branch, its now-unused helper, tests, and documentation. Roll back by reverting the issue commits if the retained host-auth diagnosis proves insufficient.

## Changed files

- `packages/cli/src/commands/doctor.ts` — remove the redundant check and remediation path.
- `packages/cli/src/managed-project-environment.ts` — remove the helper orphaned by the check deletion.
- `packages/cli/src/commands/doctor.test.ts` — remove obsolete cases and add a standalone absence assertion.
- `packages/cli/README.md` — remove only the obsolete diagnostic claim while preserving credential precedence and push-scope facts.
- `AGENT_TEST.md` — align the mandatory matrix with surviving host-auth and tracker-resolution coverage.
- `.changeset/wise-doctors-trim.md` — record the CLI patch and accurate redundancy rationale.

- [ ] None beyond normal review; automated and host CLI validation is completed before handoff.

## Security


moncher-dev:
## Steward review — PR #932 (issue #908): removal verified correct, one doc row blocks

Reproduced at head `2fe33ef3` in a clean worktree: `pnpm install --frozen-lockfile && pnpm build && pnpm test` — all green (14 projects; CLI 566 tests).

### Verified by exercising the real command, not by reading the diff

I ran the built `doctor` from this branch and from `main` against the operator's live project folder and diffed the emitted check sets:

```
main (66eaf9b5)   15 checks, all pass
this branch       14 checks, all pass
delta             -worker_github_credential
```

Every other retained check — `node_runtime`, `git_installation`, `gh_installation`, `managed_project`, `github_graphql_endpoint`, `gh_authentication`, `gh_scopes`, `github_project_resolution`, `config_directory`, `runtime_root`, `workspace_root`, `workflow_file`, `runtime_command`, `priority_mapping` — still executes and still passes. The removal is surgical; nothing was over-deleted, which was the stated risk in the issue.

The retain/remove inventory was published in the issue before deletion (comment of 2026-09-06T03:37:49Z), satisfying the "먼저 제시하고 나서 지운다" criterion. The retention rationale for `worker_github_credential` is sound: it predicted credentials by re-resolving project/daemon environment state, which is no longer the authoritative delivery path after the Phase 4 direct-host-auth move, so it could only produce false failures.

### Blocking: the mandatory test matrix still points at deleted coverage

`AGENT_TEST.md:197`, row *Tenant-scoped worker tracker credentials*, still reads:

> `packages/cli/src/commands/doctor.test.ts` covers the worker-effective GitHub check.

That coverage no longer exists — this PR deletes all three `worker_github_credential` assertions (`doctor.test.ts:1224,1251,1291` on `main`) and the check itself (`doctor.ts:91,468`). `grep -rn "worker_github_credential" packages/` on this branch returns nothing, and `AGENT_TEST.md` is untouched by the diff.

`AGENTS.md:34` requires living documents — `AGENT_TEST.md` named explicitly — to be updated in the same change when the CLI surface changes. Leaving the row as-is means the mandatory matrix asserts coverage that was intentionally removed, which is the same defect shape as #928: a guard pointed at something that is no longer what it claims to be.

**Fix:** edit that one row to describe the surviving host-authentication coverage (`gh_authentication` / `gh_scopes` / tracker-project resolution), or drop the obsolete clause. Nothing else in the PR needs to change. Codex raised this as P2 and it is correct.

### Incidental finding, not for this PR

While checking retained checks I confirmed `workflow_hooks` (`doctor.ts:1280`, `buildHookChecks`) exists and validates that declared `WORKFLOW.md` hook paths resolve on disk — but it is only reached from the `--smoke` path (`doctor.ts:1854`), so a plain `gh-symphony doctor` never runs it. That is why the operator's missing `hooks/after_create.sh` was not caught. Recorded on #929; out of scope here.

Issue #908 is returned to `Ready`.

packages/cli/src/commands/doctor.test.ts
```diff
@@ -1334,6 +1239,9 @@ Prompt body`
       })
     );
     expect(report.ok).toBe(true);
+    expect(
+      report.checks.some((check) => check.title === "Worker GitHub credential")
+    ).toBe(false);
```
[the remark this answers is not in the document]
moncher-dev: Updated `AGENT_TEST.md` so the matrix now describes the surviving host-authentication and tracker-resolution coverage instead of the removed worker-effective credential check. The focused Prettier check and the full lint, test, typecheck, and build chain pass at `9930bef7`.

moncher-dev:
## Steward verification — PASS

The documentation blocker from last cycle is corrected at the current head.

`AGENT_TEST.md:197` ("Tenant-scoped worker tracker credentials") previously claimed `packages/cli/src/commands/doctor.test.ts` covers the worker-effective GitHub check that this PR deletes outright. It now reads "covers host auth plus tracker resolution" (commit `9930bef7`) — corrected rather than deleted, which is right, since the rest of the row is still true. That satisfies the `AGENTS.md` mandate to update living documents in the same change.

`packages/cli/README.md` was updated too: the paragraph advertising a required `Worker GitHub credential` doctor check is gone. That was a second latent doc blocker and it is likewise resolved.

Swept the branch for other stale assertions of the deleted check: the only remaining mentions are `.changeset/wise-doctors-trim.md` (describing the removal) and an immutable `packages/cli/CHANGELOG.md` history entry. Every live assertion present on main (`doctor.ts` 91/468/482/484/2977/2985, `doctor.test.ts` 1203/1224/1232/1251/1291) is gone.

Diff scope is clean: 5 files, +16/−202, nothing beyond the check removal and its docs. Changeset present and correctly scoped to `@gh-symphony/cli` (patch), as issue #908 requires.

Code behavior was verified surgical last cycle (built `doctor` 15 → 14 checks, delta exactly `worker_github_credential`, all 14 retained checks pass); not re-run.

Note for the merge decision: #909 is blocked on this landing.

Requesting @hojinzs.

packages/cli/src/commands/doctor.ts
```diff
@@ -52,14 +52,9 @@ import {
...
 } from "../github/gh-auth.js";
-import type { CliProjectConfig } from "../config.js";
 import type { GlobalOptions } from "../index.js";
 import { resolveRuntimeRoot } from "../orchestrator-runtime.js";
-import {
-  resolveManagedProjectDirectory,
-  resolveManagedProjectEnvironment,
-  resolveManagedProjectFileEnvironment,
-} from "../managed-project-environment.js";
+import { resolveManagedProjectEnvironment } from "../managed-project-environment.js";
```
hojinzs, line 57: **P2 — this import narrowing orphans `resolveManagedProjectFileEnvironment`**

Dropping `resolveManagedProjectFileEnvironment` from this import leaves it with zero callers anywhere in the workspace. Verified on this head:

```
$ grep -rn "resolveManagedProjectFileEnvironment" packages --include=*.ts
packages/cli/src/managed-project-environment.ts:42:export function resolveManagedProjectFileEnvironment(
```

Definition only — no import, no test. `buildWorkerGitHubCredentialCheck` was its sole consumer, and it has no `.test.ts` of its own. Neither ESLint nor `tsc` flags an unused *export*, which is why `pnpm lint` and `pnpm typecheck` stayed green over it.

This matters because it is the exact defect class #908 exists to clear: "선행 이슈들이 제거한 대상을 점검하던 코드가 남지 않는다". The PR removes the check but leaves behind the helper written for it, so the next reader inherits a new orphan in place of the old one.

Please delete `resolveManagedProjectFileEnvironment` (`packages/cli/src/managed-project-environment.ts:40-51`). `resolveManagedProjectEnvironment` and `resolveManagedProjectDirectory` both stay — they are still used by `doctor.ts:57` and `start.ts:60`.

---

packages/cli/README.md
```diff
@@ -205,16 +205,12 @@ login material into a private, workspace-contained child home. It does not
...
-`gh-symphony doctor` reports a required `Worker GitHub credential` check for
-the selected tenant using that same precedence. Without a usable GitHub or
-Linear worker credential, startup fails before the agent launches; known-empty
-dispatches are skipped and surfaced in status snapshot warnings.
+The orchestrator retains tracker and Git publication credentials at the worker
+host boundary, enabling polling and host-side tools while the agent child
+remains credential-free. `gh-symphony doctor` validates the operator-facing
+host authentication and tracker/project resolution. Without usable host
+credentials, startup fails before the agent launches; known-empty dispatches
+are skipped and surfaced in status snapshot warnings.
```
hojinzs: **P3 — the rewrite drops two operator facts that are still true**

Removing the `Worker GitHub credential` doctor sentence is correct and required. But the paragraph was rewritten rather than trimmed, and two still-accurate statements went with it:

1. *"Project `.env` credentials take precedence over the daemon's resolved credential."* Still live — `packages/tracker-github/src/orchestrator-adapter.ts:55-60` iterates `[environments.project, environments.daemon]` and returns the first non-empty `GITHUB_GRAPHQL_TOKEN`, and the orchestrator calls it that way at dispatch (`packages/orchestrator/src/service.ts:3135-3139`). Precedence is the one thing an operator running more than one tenant has to know, and it now appears nowhere in this README.

2. *"the required direct token … must also permit repository pushes."* Still true — the same `GITHUB_GRAPHQL_TOKEN` is what the worker hands to Git as a credential (`packages/worker/src/git-transport.ts:231-233`). "Tracker and Git publication credentials" in the new text does not tell the reader they are one token, so an operator can reasonably provision a read-scoped token and only discover the problem on a failed push.

#908 scopes this change to 덜어내기 of doctor checks; deleting accurate credential documentation is outside it. Suggest keeping the two sentences and only striking the `Worker GitHub credential` claim. The `The Git publication broker branch remains in code for later removal` clause is fine to drop.

---

packages/cli/src/commands/doctor.test.ts
hojinzs: **nit — the regression guard is a passenger in an unrelated test**

This is the only assertion holding the removal in place, and it lives inside `"accepts env-token auth when gh CLI is unavailable"`, which is about `gh_installation`. Matching on `title` rather than `id` is the right call here (the id is gone from the `DoctorCheckId` union, so `check.id === "worker_github_credential"` would not typecheck) — no change needed there.

The placement is what I would change: a standalone `it("no longer emits the worker GitHub credential check", …)` states the intent, and survives someone retitling or re-scoping the host test. Non-blocking.

---

hojinzs (changes requested):
## Review — one P2 blocks; the removal itself is correct but the stated reason is not

Reproduced at head `9930bef7`: `pnpm install --frozen-lockfile && pnpm build` clean, `pnpm test` green (14 projects; CLI 566, doctor 67, orchestrator 340), `pnpm lint` and `pnpm typecheck` clean, `pnpm exec prettier --check` clean on all five changed files. CI is green on both checks.

### Smoke test — the retained checks actually run

Ran the built CLI from this branch on this host:

```
$ node packages/cli/dist/index.js doctor --project-dir "$PWD" --json
13 checks emitted; no worker_github_credential
node_runtime · git_installation · gh_installation · managed_project · gh_authentication
gh_scopes · github_project_resolution · config_directory · runtime_root · workspace_root
workflow_file · runtime_command · priority_mapping

$ node packages/cli/dist/index.js doctor --project-dir "$PWD" --smoke --issue hojinzs/github-symphony#908
smoke path executed; smoke_issue degraded cleanly on its managed_project prerequisite; exit 1
```

Both #908 completion conditions about live execution hold: every retained check executes, and `--smoke` still works. The removal is surgical — nothing outside `worker_github_credential` moved, and no leftovers remain in `doctor.ts` (the `DoctorCheckId` union, the emit site, and the `runDoctorFixes` remediation branch are all consistently gone). The retain/remove inventory was published before deletion, as the issue required.

### P2 — a new orphan replaces the old one

`resolveManagedProjectFileEnvironment` (`packages/cli/src/managed-project-environment.ts:40-51`) now has zero callers workspace-wide; the deleted check was its only consumer, and it has no test of its own. `tsc` and ESLint do not flag an unused export, so the green Completion Bar did not catch it. Details inline on `doctor.ts:57`. This is the one thing I need before approving — #908's completion condition is precisely that no code checking a removed target survives.

### P3 — the removal rationale in the PR body and the issue inventory is factually wrong

Both say the check "re-resolves project/daemon environment state, but that is no longer the authoritative host delivery path and can report a false failure." It **is** still the authoritative path, and the check mirrored it faithfully:

- `packages/tracker-github/src/orchestrator-adapter.ts:54-62` still implements the project→daemon precedence.
- `packages/orchestrator/src/service.ts:3135-3139` still calls `resolveWorkerCredentials` at dispatch with exactly `{ project: projectEnvironment, daemon: process.env }` — the same shape the check built.
- `packages/worker/src/tracker-credential-preflight.ts:42-50`, reached from `packages/worker/src/index.ts:543-546`, still hard-fails worker startup on a missing `GITHUB_GRAPHQL_TOKEN`, emitting the *same remediation string* the deleted check emitted.

So all three of #908's retention criteria read as satisfied on their face — remediation exists, the failure blocks dispatch, the target survived Phase 4. If that were the whole picture, this check should have been retained.

It is still right to delete it, for a different reason the inventory does not give: **the check could never fail independently.** `auth` is non-null iff `gh_authentication` passes (`doctor.ts:2075-2086`), and the deleted `buildWorkerGitHubCredentialCheck` injected that same `auth.token` into its simulated daemon environment before resolving. So `auth` non-null ⇒ both checks pass; `auth` null ⇒ `gh_authentication` already fails as required. The fail branch was strictly contained in `gh_authentication`'s. That is a clean, defensible removal — it just is not the reason on record.

Worth recording the one thing that is genuinely lost, so it is a decision and not an accident: for a tenant whose credential lives only in the project `.env` with no host `gh` auth, doctor used to emit `worker_github_credential: pass (source: project_env)`. That PASS signal is gone, and `gh_authentication` will still show red for that operator. I judge that acceptable — the signal was advisory and the setup already produced a red doctor — but it should be stated rather than described as removing a false failure.

Please correct the PR body (and ideally the #908 inventory comment). #909 and the rest of Phase 4 will cite this reasoning, and "the target no longer exists" is the wrong precedent to leave behind for a target that demonstrably does.

### Scope and over-engineering

Clean. 5 files, +16/−202, nothing beyond the check, its remediation branch, its tests, and the two docs. No new checks, no output-format or `--json` schema change — all three of #908's 범위 밖 constraints respected. Changeset present and correctly scoped to `@gh-symphony/cli` (patch). The prior review's `AGENT_TEST.md:197` blocker is genuinely fixed at `9930bef7`, and `packages/cli/README.md` was corrected in the same pass.

### To close this out

Fix the P2. The P3 items are the doc restoration (inline on `README.md`) and the PR-body correction — both cheap, and I would like them, but neither is a merge blocker on its own. The test-placement note is a nit; take it or leave it. Push those and I will approve on the next pass; I do not intend another round of findings on this change.

---

packages/cli/src/commands/doctor.ts
moncher-dev: Resolved in `947653be`: deleted the now-unused `resolveManagedProjectFileEnvironment` export while retaining `resolveManagedProjectEnvironment` and `resolveManagedProjectDirectory`. A workspace-wide symbol search now finds no remaining definition or caller.

packages/cli/README.md
moncher-dev: Resolved in `947653be`: restored both surviving operator facts—project `.env` credentials take precedence over the daemon credential, and the required direct token must permit repository pushes—while keeping only the obsolete doctor-check claim removed.

packages/cli/src/commands/doctor.test.ts
moncher-dev: Applied the suggestion in `947653be`: the negative assertion now lives in a standalone `no longer emits the worker GitHub credential check` test. The focused doctor suite passes with 68 tests.

moncher-dev:
## Steward re-check — P2 addressed, re-requesting review

@hojinzs the blocking item is fixed at head `947653be`.

- **P2 (blocking) — the orphan is gone.** `packages/cli/src/managed-project-environment.ts` now exports only `resolveManagedProjectEnvironment` and `resolveManagedProjectDirectory`; `resolveManagedProjectFileEnvironment` is deleted. Both survivors remain in use (`doctor.ts:57`, `start.ts:60`).

CI on `947653be`: `Test` and `Container Smoke` both green.

Please check the two P3s yourself, since they are judgement calls on wording rather than things I can verify mechanically:

1. whether the corrected removal rationale now states the real reason — that the check could never fail independently, because `auth` non-null ⇒ both checks pass and `auth` null ⇒ `gh_authentication` already fails — rather than the incorrect "no longer the authoritative host delivery path";
2. whether the restored `packages/cli/README.md` paragraph carries back both operator facts (project `.env` precedence over the daemon credential, and that the same token must permit repository pushes) while still striking the `Worker GitHub credential` claim.

**Scope of this pass:** I verified the P2 by inspecting the exported symbols at the new head, plus full CI. I did not re-run the built CLI against this head; my `doctor --json` / `--smoke` black-box run was on the previous one.

packages/cli/src/managed-project-environment.ts
```diff
@@ -37,17 +37,3 @@ export function resolveManagedProjectDirectory(
     join(runtimeRoot, "projects", encodeURIComponent(projectConfig.projectId))
   );
 }
```
hojinzs, line 39: **nit — `resolveManagedProjectDirectory` is now an export with no external caller**

Not a blocker, and not the same defect as the P2 you just fixed: this function is still live, called on line 13 by `resolveManagedProjectEnvironment`. Only the `export` keyword is now surplus.

Correcting myself from last round — I wrote that it "stays, still used by `doctor.ts:57` and `start.ts:60`." The `start.ts` half was wrong; it never imported this. Verified on this head:

```
$ grep -rn "resolveManagedProjectDirectory" --include=*.ts . --exclude-dir=node_modules --exclude-dir=dist
packages/cli/src/managed-project-environment.ts:13:  const projectDirectory = resolveManagedProjectDirectory(
packages/cli/src/managed-project-environment.ts:31:export function resolveManagedProjectDirectory(
```

`doctor.ts` was its last cross-module consumer, and this PR dropped it from that import. Narrowing it to a module-local `function` would keep the file honest about its surface. Entirely optional — take it or leave it, and it is not worth another push on its own.

---

hojinzs (approved):
## Approve — all three findings from the last round are fixed and verified

Re-verified at head `947653be` in a clean checkout: `pnpm install --frozen-lockfile`, `pnpm build`, `pnpm lint`, `pnpm typecheck` all clean; `pnpm test` green (14 projects; CLI 567 tests, doctor 68). `pnpm exec prettier --check` clean on all six changed files. CI green on both checks (`Test`, `Container Smoke`).

One note on the repo-wide `pnpm format`: it fails with 65 files, but it fails identically on `main` at `66eaf9b5`. Pre-existing, not this PR's.

### P2 (blocking) — fixed

`resolveManagedProjectFileEnvironment` is gone. A workspace-wide sweep finds no definition and no caller, and `managed-project-environment.ts` has no orphaned imports left behind — `readEnvFile` is still consumed by `resolveManagedProjectEnvironment`. I also swept the adjacent surface for orphans of the same class that the removal could have created: `deps.resolveTrackerAdapter` is **not** one — `doctor.ts:1501` still uses it for another check. The one thing left is a nit, inline.

### P3 — both fixed

**PR body rationale.** Now states the containment argument, and I confirmed it against this head rather than taking it on faith: `auth` non-null ⇒ `gh_authentication` passes (`doctor.ts:2077-2087`); `auth` null ⇒ it fails, on both branches (`doctor.ts:2088-2107`). The deleted check injected that same `auth.token` into its simulated daemon environment before resolving, so its fail branch was strictly contained. The body also now records the accepted signal loss (`worker_github_credential: pass (source: project_env)` for a project-`.env`-only tenant) as a decision rather than describing a false failure. That is the right precedent for #909 and the rest of Phase 4.

**README.** Both operator facts are restored and only the obsolete doctor-check claim is gone — precedence (`packages/cli/README.md:209-211`) and the push-scope requirement on the same token.

### nit — taken

The negative assertion is now a standalone `it("no longer emits the worker GitHub credential check")` at `doctor.test.ts:1203`.

### The guard is real, not vacuous

A negative assertion is only worth its line if the check would otherwise have been emitted under that exact fixture, so I checked rather than assumed. Built `main` in a separate worktree and ran its two `worker_github_credential` tests — both pass there on the same `createProjectConfig` + `authDependencies()` fixture the new test uses. Same fixture, check present on `main`, absent here. The guard bites.

### Smoke test

Ran the built CLI from this branch:

```
$ node packages/cli/dist/index.js doctor --project-dir "$PWD" --json
13 checks emitted, no worker_github_credential; every check produced a verdict
node_runtime · git_installation · gh_installation · managed_project · gh_authentication
gh_scopes · github_project_resolution · config_directory · runtime_root · workspace_root
workflow_file · runtime_command · priority_mapping

$ node packages/cli/dist/index.js doctor --project-dir "$PWD" --smoke --issue hojinzs/github-symphony#908
smoke path executed; smoke_issue degraded cleanly on its managed_project prerequisite; exit 1
```

Both #908 live-execution conditions hold. I also diffed the emitted check set against a built `main`: identical here, because this sandbox has no `gh` auth and no managed project, so the removed check's `kind === "resolved" && adapter === "github-project"` guard never fired on either side. That comparison is therefore uninformative in this environment — the 15 → 14 delta from the operator host and the unit coverage above are what carry it, and I am relying on those rather than claiming a delta I did not observe. Top-level `--json` keys are byte-identical between `main` and this branch, so the 범위 밖 constraint on schema changes holds.

### Requirements and scope

All five #908 completion conditions are met: the retain/remove inventory was published before deletion, no code checking a removed target survives, the retained checks execute on a real host, `--smoke` works, and the changeset is present and correctly scoped to `@gh-symphony/cli` (patch). 6 files, +18/−196, nothing beyond the check, its remediation branch, its tests, and the docs. No new checks, no output-format or `--json` schema change. No over-engineering and no unrequested decisions.

Approving. The inline nit is optional and explicitly not worth another push.

---

PR #935, opened 2026-09-06
fix(orchestrator): recover clean foreign-branch workspaces

### LINKED ISSUE #927: fix(orchestrator): recover clean foreign-branch workspaces before worker preflight
## Problem

An existing issue workspace can be clean but checked out on a branch that belongs to another issue. The orchestrator currently skips foreign-branch recovery unless the configured workspace is dirty, then launches the worker with that foreign branch. Worker identity preflight rejects startup fail-closed.

## Expected behavior

Detect a clean foreign-issue branch before worker startup and recover without weakening worker identity protection or risking cross-issue contamination.

## Context

Discovered during review of #905 / PR #915. This is pre-existing behavior: the removed quarantine path also only handled dirty workspaces, so it is intentionally out of scope for #905.

## Suggested coverage

- Configure an existing clean workspace on `fix/2-foreign` for issue #1.
- Run a real dispatch through worker identity preflight.
- Assert the assigned issue starts safely and no user work is deleted.

moncher-dev:
## Issues

- Closes #927

## Summary

- Detect persisted issue workspaces checked out on a foreign issue branch even when their worktree is clean.
- Preserve the original checkout and committed user work while dispatching from the stable recovery workspace on the assigned branch.
- Emit an operator-visible relocation event for clean recovery, while retaining the existing dirty-recovery event without duplication.

## Change-point diagram

- Persisted workspace → inspect current branch → foreign issue detected → preserve original checkout → emit `workspace-root-relocated` → populate stable `-recovery` checkout on assigned branch → unchanged fail-closed worker identity preflight

## Start here

- `packages/orchestrator/src/service.ts:2877` — persisted checkout inspection and stable recovery-path selection
- `packages/orchestrator/src/service.ts:2913` — clean relocation observability without duplicating dirty-recovery reporting
- `packages/orchestrator/src/service.test.ts:2914` — real-Git recovery, preservation, assigned-branch, and event regression

- A clean workspace accidentally left on another issue's branch recovers before worker startup without deleting committed work.
- The original checkout remains untouched and discoverable through `workspace-root-relocated`; the persisted active workspace moves to the stable `-recovery` path.
- Dirty foreign-branch recovery continues to emit `recovery-dirty-workspace` with its retained and recovery paths.

## Validation

- `GIT_CONFIG_GLOBAL=/dev/null GIT_CONFIG_SYSTEM=/dev/null npx vitest run packages/orchestrator/src/service.test.ts -t "recovers a clean foreign-branch workspace"` — pass (1 test; ambient identity disabled)
- `npx vitest run packages/orchestrator/src/service.test.ts` — pass (220 tests)
- `bash -n e2e/run-standalone-project-e2e.sh` — pass
- `pnpm lint` — pass
- `pnpm test` — pass (all workspace packages)
- `pnpm typecheck` — pass
- `pnpm build` — pass
- `./e2e/run-standalone-project-e2e.sh` — not executed beyond preflight: attempted twice; both exited 69 because the Docker daemon is unreachable (tracked in #937)

## Changeset

- Not needed under workflow policy because issue #927 has no changeset label.

## Risks & rollback

- Branch identity continues to use the same numeric issue parser as worker preflight; a numeric non-issue branch segment can conservatively trigger recovery, but the relocation is now diagnosable through both paths in the event log.
- The stable recovery path intentionally diverges from upstream workspace-path reuse to preserve a foreign-issue checkout.
- Revert `9f9178d`, `e41c9bc`, `8e012cd`, and `ad97c63` to restore prior dispatch behavior and coverage.

## Changed files

- `packages/orchestrator/src/service.ts` — inspect clean persisted workspaces and report clean recovery relocation
- `packages/orchestrator/src/service.test.ts` — prove safe dispatch, preservation, assigned-branch startup, portable Git identity, and relocation event
- `e2e/run-standalone-project-e2e.sh` — seed a clean foreign branch, cross real worker preflight, and fail closed on missing logs
- `e2e/scenarios/13-standalone-project-model.md` — document the black-box assertion
- `docs/architecture.md` — document behavior and the intentional upstream path-reuse divergence
- `AGENT_TEST.md` — map unit and Docker coverage

- [ ] Run `./e2e/run-standalone-project-e2e.sh` on a host with a reachable Docker daemon (tracked by #937)

## Security


moncher-dev:
## Steward review — blocking: the new test cannot pass on CI

Reviewed at head `8e012cd3`. **CI `Test` is red on this head**, and the only failure is this PR's own new test:

```
FAIL packages/orchestrator/src/service.test.ts > OrchestratorService
     > recovers a clean foreign-branch workspace before worker preflight
Test Files  1 failed | 131 passed (132)

Error: Command failed: git -C ".../acme_platform_1-.../repository" add foreign-issue.txt
  && git -C ".../repository" commit -m 'fix: preserve foreign issue work'
  status: 128
  stderr: Author identity unknown
    *** Please tell me who you are.
    fatal: empty ident name (for <runner@runnervm....internal.cloudapp.net>) not allowed
```

(run [34012303014](https://github.com/hojinzs/github-symphony/actions/runs/34012303014); the `Coverage Report` ENOENT that follows is a cascade of this, not a second fault.)

### Root cause — confirmed, not inferred

`service.test.ts:2914` commits into `repositoryDirectory`, the checkout returned by `ensurePopulatedIssueWorkspaceRepository`. That directory is a **clone**, and `createRepositoryFixture` (`service.test.ts:17533-17535`) sets `user.email`/`user.name` on the **seed** repository only. Local git config is not propagated to a clone, so the new checkout has no committer identity.

I reproduced the mechanism with plain git, no repo code, mirroring the fixture exactly:

```
seed identity: tester@example.com
clone identity: []          <- not inherited
```

On a developer host the commit still succeeds because git silently synthesizes an identity from `username@hostname` (mine emitted `Committer: Steve Lee <steve@Steveui-MacBookPro.local>` with the "configured automatically" warning). On a GitHub runner that synthesized name is empty, so git hard-fails with the exact `empty ident name` error above. **This test passes on any dev host and can never pass in CI** — which is why it was not caught before push.

### Fix

Apply the pattern this repository already uses everywhere else, to the **clone** rather than the seed — `dispatch.test.ts:822-824`, `skills.test.ts:149-150`, `headless-verification.test.ts:114-116`, `default-hooks.test.ts:29-30` all do this:

```ts
execSync(`git -C ${shell(repositoryDirectory)} config user.email tester@example.com`);
execSync(`git -C ${shell(repositoryDirectory)} config user.name tester`);
```

immediately after the `switch -c fix/2-foreign`, before the `add`/`commit`. Verified: with those two lines the commit succeeds under a neutralized global identity; without them it fails.

Please **verify the fix under a neutralized identity**, not just on the host — otherwise the same false green recurs:

```bash
GIT_CONFIG_GLOBAL=/dev/null GIT_CONFIG_SYSTEM=/dev/null \
  npx vitest run packages/orchestrator/src/service.test.ts -t "recovers a clean foreign-branch workspace"
```

Worth considering as a follow-up (not this PR): any fixture that commits into a populated workspace has this hazard, so setting the identity inside `ensurePopulatedIssueWorkspaceRepository`'s test fixture helper would stop it recurring.

### The production change itself looks sound — no changes requested there

`service.ts:2877` widens `retainedWorkspaceBranch` from a gated read (`recovery?.kind === "incomplete-turn-dirty-workspace" || existingWorkspaceUsesStableRecoveryPath`) to reading the branch for **any** existing workspace record. That is the right shape for #927: a clean reused workspace parked on another issue's branch previously bypassed the foreign-branch check entirely.

Two things I checked specifically, both fine:

- **No new throw path.** `readGitCurrentBranch` (`packages/orchestrator/src/git.ts:99-115`) wraps `rev-parse` in `try/catch` and returns `null` on any failure. Widening it cannot reintroduce the deleted-branch throw that has previously aborted whole ticks — worth stating explicitly given that history.
- **No false positives.** `retainedWorkspaceHasForeignBranch` still requires the parsed branch issue number to differ from this issue's, so a workspace on its own branch is unaffected.

The cost is one extra `git rev-parse` per dispatch that has an existing workspace record. Acceptable.

### Verdict

**Changes requested — one fix.** Add the two identity lines, confirm green under a neutralized global identity, and push. Nothing else in this PR needs to change; the behavior fix and its comments are correct as written.

Issue #927 returned to `Ready`.

docs/architecture.md
```diff
@@ -129,7 +129,7 @@ the tracker adapter:
...
... clean or dirty checkout names another issue, it leaves that checkout untouched and persists a stable fresh recovery workspace as the issue's active workspace for this and later dispatches, so worker identity preflight remains fail-closed without returning to the foreign branch or risking committed user work. Dirty-workspace recovery additionally emits a warning-level `recovery-dirty-workspace` event containing the retained path and branch. If the orchestrator-owned recovery checkout is later left on another issue's branch, it is repopulated in place; the operator-owned original remains untouched and recovery paths never chain. Dirty recovery inspection follows the persisted run workspace, while project status keeps warning about the original path and branch until it is clean. The prompt distinguishes that operator-visible retained workspace from the active checkout. The orchestrator does not attribute, quarantine, rename, delete, or fail the run because of the retained original's dirty state. Terminal cleanup removes the persisted active recovery workspace; the original dirty workspace deliberately remains operator-owned and requires manual cleanup after its contents are reviewed.
```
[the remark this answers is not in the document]
moncher-dev: Resolved in e41c9bc: the architecture slice now explicitly identifies stable `<workspaceKey>-recovery` routing as an intentional repository-local divergence from upstream §§9.1–9.2 and explains that preserving the foreign-issue checkout takes precedence over path reuse.

moncher-dev:
## Steward review — passing, handing to @hojinzs

Issue #927. I sent this back earlier this cycle: CI was red at head `8e012cd3`, failing on the PR's own new test with `fatal: empty ident name`. That is fixed at head `e41c9bca`.

**What I found and how it was fixed.** `recovers a clean foreign-branch workspace before worker preflight` commits into the populated workspace **clone**, but `createRepositoryFixture` sets `user.email`/`user.name` on the **seed** repository only, and git does not propagate local config into a clone. I reproduced the mechanism with plain git — `clone identity: []`. On a dev host git silently synthesizes `username@hostname`, so it passes locally; on a runner that name is empty and git hard-fails. The fix sets the identity on `repositoryDirectory` after the `switch -c`, matching `dispatch.test.ts:822-824`, `skills.test.ts:149-150`, `headless-verification.test.ts:114-116`, `default-hooks.test.ts:29-30`. CI on `e41c9bca` is now green on both `Test` and `Container Smoke`.

**The production change, which I reviewed in full at the previous head and which is unchanged.** `service.ts:2877` widens `retainedWorkspaceBranch` from a gated read (`recovery?.kind === "incomplete-turn-dirty-workspace" || existingWorkspaceUsesStableRecoveryPath`) to reading the branch for any existing workspace record. That is the right fix for #927 — a clean reused workspace parked on another issue's branch previously bypassed the foreign-branch check entirely. Two things I checked specifically:

- **No new throw path.** `readGitCurrentBranch` (`packages/orchestrator/src/git.ts:99-115`) wraps `rev-parse` in try/catch and returns `null` on any failure, so widening it cannot reintroduce the deleted-branch throw that has previously aborted whole ticks.
- **No false positives.** `retainedWorkspaceHasForeignBranch` still requires the parsed branch issue number to differ from this issue's.

Cost is one extra `git rev-parse` per dispatch that has an existing workspace record.

**Boundary of my evidence:** the identity repro and the production-change analysis are mine and independent. I did not run a local build of the new head — the delta since my review is the two `git config` lines, and CI covers them. The behavior this PR fixes (a clean workspace on a foreign branch) is not something I exercised against a live orchestrator; the coverage for it is the new unit test.

packages/orchestrator/src/service.ts
```diff
@@ -2877,23 +2877,25 @@ export class OrchestratorService {
...
-    // on later dispatches. Falling back to the configured key path would
-    // repopulate (and therefore delete) the retained dirty workspace.
+    // Once a foreign workspace has forced the run onto the stable recovery
+    // path, that persisted path remains the issue's active workspace on later
+    // dispatches. Falling back to the configured key path would repopulate
+    // (and therefore delete) the retained workspace, including user commits in
+    // an otherwise clean checkout.
     const reuseStableRecoveryWorkspace =
       existingWorkspaceUsesStableRecoveryPath;
     const useFreshRecoveryWorkspace =
```
hojinzs, line 2901: **P1 — the clean foreign-branch relocation leaves no operator-visible trace anywhere.**

When `useFreshRecoveryWorkspace` is true the issue's active workspace is permanently repointed to `<workspaceKey>-recovery`, and the original checkout — which by definition holds committed user work on another issue's branch — is orphaned on disk. The dirty path announces that. The new clean path announces nothing:

1. `workspace-root-relocated` (line 2913) is explicitly suppressed by `!useFreshRecoveryWorkspace`. That guard was written when `useFreshRecoveryWorkspace` implied the dirty path, which always emitted `recovery-dirty-workspace` carrying `recoveryWorkspacePath`. This PR breaks that implication and the guard was not revisited.
2. `recovery-dirty-workspace` (line 2943) requires `recovery?.kind === "incomplete-turn-dirty-workspace"`, which is `null` on a clean reuse.
3. `resolveRetainedWorkspaceWarnings` (`service.ts:2259-2284`) `continue`s unless the retained checkout is dirty, so `project status` never warns about it either.
4. Terminal cleanup removes the persisted active recovery workspace only; the orphan is never removed and never mentioned.

Confirmed, not inferred — I instrumented the new test to dump the run's `events.ndjson`:

```
PROBE-EVENTS tenant-1-acme-platform-1-mmgzqfhc  tracker.list,run-dispatched
```

`run-dispatched` (`service.ts:1930`) carries no workspace path, so nothing in the event log or the status snapshot records that the workspace moved or that an orphaned clone containing user commits now exists.

The net effect is the opposite of the PR's stated intent: the commits are preserved on disk but become undiscoverable. Observability is one of the six layers work must be classified against, and this adds a workspace-lifecycle transition with zero events.

Smallest fix — narrow the suppression to the case that already reports itself:

```ts
if (
  existingWorkspaceRecord &&
  !existingWorkspaceAtConfiguredRoot &&
  !(useFreshRecoveryWorkspace &&
    recovery?.kind === "incomplete-turn-dirty-workspace")
) {
```

so the clean path emits `workspace-root-relocated` with previous/configured paths (it is already formatted for operators in `packages/core/src/observability/event-formatter.ts:51`). Extending `resolveRetainedWorkspaceWarnings` to also warn on a *clean* retained checkout whose branch is foreign would close the `project status` half — I'd treat the event as the blocking piece and the status warning as optional follow-up.

---

packages/orchestrator/src/service.ts
```diff
@@ -2877,23 +2877,25 @@ export class OrchestratorService {
...
-      existingWorkspaceRecord &&
-      (recovery?.kind === "incomplete-turn-dirty-workspace" ||
-        existingWorkspaceUsesStableRecoveryPath)
-        ? await readGitCurrentBranch(existingWorkspaceRecord.repositoryPath)
-        : null;
+    // Inspect every persisted checkout before worker startup. Dirty recovery
+    // already reaches this branch check, but a clean reused workspace can also
+    // have been left on another issue's branch and must not bypass worker
+    // identity preflight on that foreign branch.
+    const retainedWorkspaceBranch = existingWorkspaceRecord
```
hojinzs, line 2884: **P2 — widening this read also widens a heuristic's false-positive surface.**

`extractIssueNumbersFromBranch` (`packages/core/src/workflow/issue-identity.ts:50`) matches any digits that begin a slash-delimited segment, so `release/2024-q1` → `[2024]` and `chore/3-cleanup` → `[3]`. Before this change that heuristic only ran for a dirty workspace or one already on the recovery path — situations that are already exceptional and already reported. It now runs for **every** clean reused workspace on every dispatch, and any parsed number ≠ the issue number permanently repoints the workspace to `-recovery` and pays a full fresh clone.

I checked the obvious false positive and it is safe: the assigned branch `symphony/tenant-1/acme-platform-1` yields `[]`, because no segment starts with a digit. So this only bites a checkout that something else parked on a numeric-prefixed branch — a narrow window. But the consequence is heavy (permanent path divergence, orphaned clone) and, per my other comment, currently silent.

Not blocking on its own. Worth considering gating the foreign check on branches that actually match the project's `branch_template` shape rather than any leading-digit segment. At minimum, emitting the relocation event makes a false positive diagnosable instead of a mystery re-clone.

---

e2e/run-standalone-project-e2e.sh
```diff
@@ -79,6 +79,27 @@ cat > "$FIXTURE" <<EOF
...
+mkdir -p "$alpha_workspace" "$CONFIG_DIR/projects/$alpha_id/$alpha_workspace_key"
+git clone /e2e/repos/test-owner/test-repo "$alpha_original_repo" >/dev/null
+git -C "$alpha_original_repo" config user.email "e2e@test.local"
+git -C "$alpha_original_repo" config user.name "E2E Test"
+git -C "$alpha_original_repo" switch -c fix/2-foreign
+printf "foreign issue committed work\n" > "$alpha_original_repo/foreign-issue.txt"
+git -C "$alpha_original_repo" add foreign-issue.txt
+git -C "$alpha_original_repo" commit -m "fix: preserve foreign issue work" >/dev/null
+alpha_foreign_head=$(git -C "$alpha_original_repo" rev-parse HEAD)
+cat > "$CONFIG_DIR/projects/$alpha_id/$alpha_workspace_key/workspace.json" <<EOF
```
hojinzs, line 99: **P2 — this is the only black-box coverage of the change, and it has never been executed.**

The PR body says Docker E2E was not run (tracked in #937), so this seeding block and the four assertions at lines 165-169 are written-but-unverified. I checked the two things I could settle statically and both hold:

- `$CONFIG_DIR/projects/<id>/<workspaceKey>/workspace.json` matches `OrchestratorFsStore.issueWorkspaceDir` (`packages/orchestrator/src/fs-store.ts:354`, `join(projectDir(projectId), workspaceKey)`).
- `<workspaceKey>-recovery` matches `resolveIssueWorkspaceDirectory(root, \`${workspaceKey}-recovery\`)` in `service.ts:2872`.

The remaining unknown is that this seeds `workspace.json` into the project's config dir *before* `project start` first creates `project.json` there. If startup ever rewrites rather than merges into that directory the record vanishes — the failure would be loud (`test -d "$repo/.git"`), so it is a correctness risk for the test rather than a false green.

Ask: either run the script once on a Docker-capable host before merge, or adjust the Validation section so it does not read as if E2E coverage is in place — right now "Docker E2E — Cycle 1 runner coverage retained" understates that the assertions added by *this* PR have not run.

---

e2e/run-standalone-project-e2e.sh
```diff
@@ -124,6 +143,9 @@ for project in project-alpha project-beta; do
   # not under the runtime state directory (spec 9.1).
   test -z "$(find "$CONFIG_DIR/projects/$project_id" -path "*/repository" -type d)"
   repo=$(find "$PROJECT_ROOT/$project/.runtime/workspaces" -path "*/repository" -type d | head -1)
```
hojinzs: **nit** — the `find` result is computed and then immediately discarded for alpha. Worth making that an `if/else` so the intent is explicit:

```sh
if [ "$label" = alpha ]; then
  repo="$PROJECT_ROOT/project-alpha/.runtime/workspaces/$alpha_workspace_key-recovery/repository"
else
  repo=$(find "$PROJECT_ROOT/$project/.runtime/workspaces" -path "*/repository" -type d | head -1)
fi
```

This also documents *why* the override exists: alpha now has two repository directories under `workspaces/`, so `head -1` is no longer deterministic there.

---

e2e/run-standalone-project-e2e.sh
```diff
@@ -135,11 +157,16 @@ for project in project-alpha project-beta; do
...
+alpha_repo="$PROJECT_ROOT/project-alpha/.runtime/workspaces/$alpha_workspace_key-recovery/repository"
 beta_repo=$(find "$PROJECT_ROOT/project-beta/.runtime/workspaces" -path "*/repository" -type d | head -1)
 alpha_branch=$(git -C "$alpha_repo" branch --show-current)
 beta_branch=$(git -C "$beta_repo" branch --show-current)
 test "$alpha_branch" != "$beta_branch"
+test "$(git -C "$alpha_original_repo" branch --show-current)" = "fix/2-foreign"
+test "$(git -C "$alpha_original_repo" rev-parse HEAD)" = "$alpha_foreign_head"
+test "$(cat "$alpha_original_repo/foreign-issue.txt")" = "foreign issue committed work"
+test -z "$(git -C "$alpha_original_repo" status --porcelain)"
+! grep -q "Issue identity preflight failed" $(find "$CONFIG_DIR/projects/$alpha_id" -path "*/runs/*/worker.log" -type f)
```
hojinzs: **nit** — if the `$(find ...)` substitution is ever empty, `grep -q` falls back to reading stdin, and the `!` then makes this assertion pass vacuously rather than failing. Same shape as the guarded pattern already used at lines 155-156, so it is easy to make robust:

```sh
alpha_logs=$(find "$CONFIG_DIR/projects/$alpha_id" -path "*/runs/*/worker.log" -type f -print)
test -n "$alpha_logs"
! grep -q "Issue identity preflight failed" $alpha_logs
```

The `test -n` is the part that matters: it makes "no worker log at all" a failure instead of a pass. That is worth having, because a missing worker log is exactly what a regression in this code path would produce.

---

hojinzs (changes requested):
## Review at `e41c9bc` — one P1, otherwise sound

Reviewed against issue #927. The production change is the right fix and I verified the previously-blocking CI failure is genuinely gone. One blocking finding, and it is a one-line fix.

### Requirements traceability (#927)

| Acceptance condition | Status |
| --- | --- |
| Detect a clean foreign-issue branch before worker startup | ✅ `service.ts:2884` now reads the branch for any persisted workspace record, not just dirty/recovery ones |
| Recover without weakening worker identity preflight | ✅ preflight is untouched and still fail-closed; recovery moves the dispatch onto the assigned branch instead of relaxing the check |
| No cross-issue contamination | ✅ `retainedWorkspaceHasForeignBranch` still requires the parsed branch number to differ from the issue number |
| No user work deleted | ✅ unit-verified — original branch, HEAD, file contents and clean status all asserted |
| Assigned issue starts safely | ✅ `WORKING_DIRECTORY` and `SYMPHONY_ASSIGNED_BRANCH` asserted on the spawn env |
| Real dispatch through worker identity preflight | ⚠️ the deterministic test stubs `spawn`; the real-preflight assertion lives only in the Docker E2E, which has not been executed (see inline comment) |

### Smoke test — run locally, not taken on trust

```
pnpm install && pnpm build
GIT_CONFIG_GLOBAL=/dev/null GIT_CONFIG_SYSTEM=/dev/null \
  npx vitest run packages/orchestrator/src/service.test.ts -t "foreign-branch"
→ 1 passed (306ms)
```

The earlier `fatal: empty ident name` blocker is fixed: the two `git config` lines are set on the clone, and the test is green with ambient identity fully neutralized. CI on this head is green on both `Test` and `Container Smoke`.

### Design question I checked so it does not get raised as "simplify this"

The obvious objection to this PR is that a *clean* checkout does not need a whole second clone — just switch it back to the assigned branch and the foreign branch's commits survive untouched. That objection is wrong here, and it is worth recording why:

- A reused workspace (`needsPopulation === false`) never re-checks-out anything. The orchestrator only *reads* the current branch (`service.ts:3046`), and the `assignedBranch !== expectedAssignedBranch` guard is enforced only when populating. There is no in-place branch-switch capability to reach for.
- Branch checkout is owned by the `after_create` hook, and re-running it does `rm -rf` on the repository directory first (`service.ts:3015-3017`). Repopulating in place would destroy exactly the committed work this PR is protecting.

So the stable `-recovery` path is a consequence of the existing hook-owned population contract, not over-engineering. The divergence is now documented properly in `docs/architecture.md`, which resolves the earlier Codex P1.

### Follow-through on prior review rounds

- Codex P1 (declare the recovery-path routing as a spec divergence) — addressed in `e41c9bc`; the architecture slice now names it as an intentional repository-local divergence from §§9.1–9.2 with its rationale. Thread resolved.
- Steward blocker (CI-only `empty ident name` failure) — fixed and independently re-verified above.

### Findings

| Sev | Finding |
| --- | --- |
| **P1** | Clean foreign-branch relocation emits no event and produces no status warning — the orphaned checkout holding user commits is undiscoverable (`service.ts:2901`) |
| P2 | The foreign-branch heuristic now runs against every clean workspace, widening its false-positive surface (`service.ts:2884`) |
| P2 | The new E2E assertions are the only black-box coverage of this change and have never been executed (`e2e/run-standalone-project-e2e.sh:99`) |
| nit | Discarded `find` result for alpha (`e2e/...:145`) |
| nit | `grep -q` over a possibly-empty file list can pass vacuously (`e2e/...:169`) |

### Verdict

**Changes requested — the P1 only.** The observability gap is the one thing I consider blocking: this PR's whole purpose is to stop destroying user work, and as written it preserves the work while making it invisible to the operator, in a repository where Observability is a first-class layer. The suppression guard at `service.ts:2913` simply was not revisited when `useFreshRecoveryWorkspace` stopped implying the dirty path; narrowing it is a one-line change.

The two P2s are judgement calls, not gates — address or push back with reasoning and I will not hold the PR on them. The production logic itself needs no other change.

---

packages/orchestrator/src/service.ts
moncher-dev: Fixed in `9f9178d`. Clean foreign-branch recovery now emits `workspace-root-relocated` with the preserved original workspace path and the active `-recovery` path. The focused service regression reads `events.ndjson` and asserts both paths, while dirty recovery still relies on its existing `recovery-dirty-workspace` event without a duplicate relocation event.

packages/orchestrator/src/service.ts
moncher-dev: I’m keeping the branch-number check aligned with the existing worker identity parser in this issue. Gating it on the current project’s `branch_template` could miss a genuine foreign issue branch created under another project/template—the exact cross-issue condition this recovery protects. The narrow numeric-segment false-positive remains possible, but `9f9178d` makes any resulting relocation operator-visible with both paths. A stricter identity model would be a separate design change rather than part of #927.

e2e/run-standalone-project-e2e.sh
moncher-dev: Confirmed and now stated precisely in Cycle 3 validation: the Docker runner was attempted twice, but preflight exited 69 because the daemon is unreachable, so the new black-box assertions did not execute on this host. The fixture retains unit coverage plus shell syntax validation, and the external Docker execution remains tracked by #937.

e2e/run-standalone-project-e2e.sh
moncher-dev: Fixed in `9f9178d`. Repository discovery is now an explicit `if/else`: alpha selects the deterministic `-recovery` path, while beta uses `find`. This documents why alpha cannot use `head -1` after the preserved original checkout creates a second repository directory.

e2e/run-standalone-project-e2e.sh
moncher-dev: Fixed in `9f9178d`. The alpha worker-log paths are captured first and guarded with `test -n "$alpha_logs"` before `grep`, so an absent worker log fails instead of passing vacuously.

docs/architecture.md
```diff
@@ -129,7 +129,7 @@ the tracker adapter:
...
... branch or risking committed user work. Routing the same `workspaceKey` to a stable `<workspaceKey>-recovery` path is an intentional repository-local divergence from the upstream §§9.1–9.2 per-issue path and reuse contract: preserving a foreign-issue checkout takes precedence over reusing its configured path. Dirty-workspace recovery additionally emits a warning-level `recovery-dirty-workspace` event containing the retained path and branch. If the orchestrator-owned recovery checkout is later left on another issue's branch, it is repopulated in place; the operator-owned original remains untouched and recovery paths never chain. Dirty recovery inspection follows the persisted run workspace, while project status keeps warning about the original path and branch until it is clean. The prompt distinguishes that operator-visible retained workspace from the active checkout. The orchestrator does not attribute, quarantine, rename, delete, or fail the run because of the retained original's dirty state. Terminal cleanup removes the persisted active recovery workspace; the original dirty workspace deliberately remains operator-owned and requires manual cleanup after its contents are reviewed.
```
hojinzs, line 132: **nit** — the closing cleanup sentence still says only *dirty*, but this PR makes a clean original equally permanent.

> Terminal cleanup removes the persisted active recovery workspace; **the original dirty workspace** deliberately remains operator-owned and requires manual cleanup after its contents are reviewed.

I traced what happens to a clean preserved original after the run finishes, and it never comes back:

- `cleanupTerminalIssueWorkspace` (`service.ts:6057`) removes `workspaceRecord.workspacePath` — the `-recovery` path — and sets the record to `removed` rather than deleting it.
- `loadWorkspaceForIssue` (`service.ts:4570-4597`) does not filter on `status`, so the next dispatch still sees the record, `existingWorkspaceUsesStableRecoveryPath` is still true, and the issue stays pinned to `-recovery` forever.

That is the right outcome — it is precisely why the preserved original is never repopulated out from under the operator — but it means a clean foreign-branch checkout accumulates on disk with exactly the same manual-cleanup obligation as a dirty one, and the paragraph currently only tells the operator about the dirty case. Dropping the word "dirty" (or writing "the original retained workspace") covers both.

Non-blocking; docs-only.

---

hojinzs (approved):
## Approve — requirements met, all prior findings closed

Reviewed at head `9f9178d`. Cycle 3. **CI is green** on both `Test` and `Container Smoke`.

### Verification I ran myself

| Check | Result |
| --- | --- |
| `GIT_CONFIG_GLOBAL=/dev/null GIT_CONFIG_SYSTEM=/dev/null npx vitest run packages/orchestrator packages/core` | **672 passed / 38 files** |
| `pnpm lint` | pass |
| `pnpm typecheck` | pass |
| `pnpm build` | pass |
| `bash -n e2e/run-standalone-project-e2e.sh` | pass |
| `./e2e/run-standalone-project-e2e.sh` | **not run** — no Docker daemon on this host either, same limitation as the author's (#937) |

Neutralizing the ambient git identity is the condition that made CI red at `8e012cd`; the suite is green under it now, so the `empty ident name` regression is genuinely fixed rather than masked by a host identity.

### Issue #927 acceptance conditions

| #927 asks for | Where |
| --- | --- |
| Detect a **clean** foreign-issue branch before worker startup | `service.ts:2884` — `retainedWorkspaceBranch` widened from a gated read to every existing workspace record |
| Recover **without weakening** identity protection | Dispatch runs from the `-recovery` checkout on the assigned branch; preflight stays fail-closed. Asserted on `SYMPHONY_ASSIGNED_BRANCH` and the real `rev-parse --abbrev-ref HEAD` |
| **No user work deleted** | The original stays on `fix/2-foreign` at its commit, file intact, `status --porcelain` empty — all four asserted against real Git |
| Suggested coverage: clean workspace on `fix/2-foreign`, real dispatch through preflight | `service.test.ts:2914` (unit, real Git) + `e2e/run-standalone-project-e2e.sh` (black-box, unexecuted) |

No scope creep: the diff is one widened read, one narrowed event guard, and their coverage. No new options, no new config surface, no refactor riding along.

### Prior cycle findings

1. **P1 clean relocation left no operator-visible trace** — fixed exactly as proposed. `workspace-root-relocated` now fires on the clean path with both paths, and dirty recovery still reports through `recovery-dirty-workspace` without a duplicate. I confirmed the guard is correctly narrow: a plain `workspace.root` change still emits as before, and the dirty+foreign case is still the only suppressed one.
2. **P2 heuristic false-positive surface** — reasoned non-adoption, and I accept it. Keeping the branch parser aligned with worker preflight is the safer default; gating on the project's own `branch_template` really would miss a foreign branch created under a different project. The relocation is now diagnosable from the event log, which was my actual ask. A stricter identity model is a separate design change, not #927.
3. **P2 E2E never executed** — the Validation section now states this precisely (attempted twice, exit 69, daemon unreachable, tracked in #937) instead of reading as coverage in place. I could not run it either. I did settle the one thing I flagged as unknown last cycle: seeding `workspace.json` into the project config dir before first start is safe — `ensureProjectDir` (`fs-store.ts:59-60`) is `mkdir -p` and `saveProjectConfig` writes a single file into it, so nothing rewrites the directory. The fixture shape is sound; only its execution is outstanding.
4. **nit `find` discarded for alpha** — fixed, now an explicit `if/else`.
5. **nit vacuous `grep`** — fixed, `test -n "$alpha_logs"` guards it.

Every thread was either fixed or answered with a substantive rationale. Nothing was silently dropped.

### One non-blocking nit

Left inline on `docs/architecture.md` — the cleanup sentence still says "the original **dirty** workspace" remains operator-owned, but I traced the record lifecycle and a clean original is orphaned just as permanently. Wording only; do not respin CI for it.

### Boundary of my evidence

The unit test and the toolchain checks above are mine and ran green on this head. The Docker E2E assertions added by this PR remain unexecuted anywhere — that is the one gap I am knowingly approving over, on the strength of the real-Git unit coverage and the post-merge checkbox already in the PR body.

Approving. Please land the doc nit if you touch the branch for another reason; otherwise it can ride a follow-up.

---

Issue #900, opened 2026-09-05
bug(workflow): a steward review with no inline threads does not trigger the rework guard


moncher-dev:
## Observed behaviour

A blocking review posted by the board steward can be invisible to the worker's rework guard, so the worker re-affirms the PR and returns it to `In review` without doing any work. One full dispatch is spent producing nothing.

## Reproduction, as it happened on 2026-09-05

| time (UTC) | event |
| --- | --- |
| 13:54:00 | steward posts a blocking review on PR #899 — a top-level `COMMENTED` review, body only, **zero inline threads**, authored by `moncher-dev` |
| ~13:55 | operator sets #872 `In review` → `Ready` |
| 13:55:31 | worker opens cycle 2: *"PR #899 returned to Ready without actionable review feedback; re-validation and handoff restoration are starting"* |
| 13:57:34 | worker closes cycle 2: *"PR #899 remains ready; the unchanged published head satisfies the Completion Bar"* — `In progress` → `In review`, **no commit, head still `7d485848`** |

The run consumed 529,524 tokens and changed nothing. The review it should have acted on was 90 seconds old at dispatch.

## Why the guard missed it

`WORKFLOW.md` Step 0 item 4 opens a rework cycle when a linked PR has `CHANGES_REQUESTED`, **unresolved actionable review threads**, a human instruction indicating rework, or a Land-return rework marker.

The steward's review satisfies none of those:

- its state is `COMMENTED`, not `CHANGES_REQUESTED` — the steward is not a repository reviewer and posts findings as comments;
- `reviewThreads` on that PR is **empty**, because the findings were written in the review body rather than anchored to lines. Verified: `reviewThreads(first: 20)` returns zero nodes;
- the author is `moncher-dev`, which is the worker's own account, so it does not read as a human instruction.

So the worker's conclusion was correct under the rules as written. The rules are what is wrong.

## Why this is worth fixing rather than working around

The steward reviews every PR on this board, and its findings are the main source of rework instruction — it caught a real credential leak on #885 and a test that could not fail on #899. If the delivery channel is invisible half the time, the board silently loses those. The operator tried reposting the review body as an issue comment, but that channel had the same author-identity defect: a steward comment from the shared `moncher-dev` account did not read as a human instruction. #947 tracks making actionable steward issue comments reachable as well.

The cost is a full dispatch per occurrence, and it is silent: nothing errors, the run reports success, and the issue lands back in `In review` looking healthy.

## Directions, not a decision

[...]

PR #936, opened 2026-09-06
fix(orchestrator): preserve retries on hook faults

### LINKED ISSUE #929: fix(orchestrator): a missing workflow hook script hard-fails every dispatch and burns the retry budget

moncher-dev:
## Issues

- Closes #929

## Summary

- Validate enabled workflow hook script paths before project startup, dispatch, and due-retry restart so project configuration faults do not consume issue retry budgets.
- Run the existing `workflow_hooks` diagnostic during default `doctor`, with accurate immediate/deferred validation for each hook kind.
- Document and exercise standalone hook provisioning; the standalone E2E fixture now creates every project's label-specific skill directory before writing `SKILL.md`.

## Change-point diagram

- `WORKFLOW.md hooks` → shared core resolution/validation → default doctor + project-start preflight + reconciliation gates → project fault before issue retry accounting
- `workflow init` → executable `hooks/after_create.sh` → standalone project provisioning → TC-24 missing-hook assertion

## Start here

- `packages/core/src/workspace/hooks.ts` — shared command classification, per-kind resolution, and filesystem validation
- `packages/orchestrator/src/service.ts` — trust-aware pre-dispatch and due-retry gates
- `e2e/run-standalone-project-e2e.sh` — standalone fixture creation and missing-hook black-box assertion

- Enabled absent, non-file, non-executable, or syntactically invalid hook scripts now fail startup/reconciliation as project configuration faults with resolved paths where available.
- Default `doctor` reports `workflow_hooks`; workspace-relative run/remove hooks are deferred until an issue workspace exists.
- Bare command names are treated as relative script paths, matching execution behavior. Disabled hooks remain non-blocking while `doctor` still diagnoses their paths.

## Validation

- `bash -n e2e/run-standalone-project-e2e.sh` — pass
- `pnpm e2e:compose-project-test` — pass
- `git diff --check` — pass
- `pnpm lint` — pass
- `pnpm test` — pass
- `pnpm typecheck` — pass
- `pnpm build` — pass
- `pnpm e2e:standalone-project` — environment exception #938; two attempts exited 69 because the Docker daemon is unavailable
- Docker scenario: `e2e/scenarios/24-hook-configuration-fault.md`

## Changeset

- `.changeset/cool-hooks-wait.md` (`@gh-symphony/cli` patch)

## Risks & rollback

- Risk is limited to hook classification, per-kind base resolution, trust gating, startup/reconciliation ordering, and standalone fixture setup. Tests cover absent/non-executable hooks and dispatch/due-retry budget preservation.
- Roll back the hook-validation and E2E fixture commits plus changeset if startup compatibility issues appear; #941's merged project-lock changes remain independently intact.

## Changed files

- `packages/core/src/workspace/hooks.ts` and tests — shared hook resolution and validation
- `packages/cli/src/commands/doctor.ts`, `start.ts`, and tests — default diagnostics and startup preflight
- `packages/orchestrator/src/service.ts` and tests — project-fault gates before dispatch/retry accounting
- `README.md`, `packages/cli/README.md`, `docs/configuration.md`, `docs/architecture.md` — living documentation
- `AGENT_TEST.md`, `e2e/run-standalone-project-e2e.sh`, `e2e/scenarios/24-hook-configuration-fault.md` — black-box scenario and corrected fixture coverage
- `.changeset/cool-hooks-wait.md` — patch release note

- [ ] Re-run the standalone Docker E2E on a host with a reachable Docker daemon (tracked by #938).

## Security


moncher-dev:
## Steward review — the core fix is real and well-tested, but the gate has a hole that reproduces the original incident

Reviewed at head `1a488042` in a clean worktree: `pnpm install --frozen-lockfile && pnpm build` clean, CI green on both `Test` and `Container Smoke`. I ran the built CLI rather than reading the diff.

### What I verified works — do not rework any of this

**It does not duplicate the existing `--smoke` check; it promotes it.** `buildHookChecks` is removed from `buildDoctorSmokeChecks` and added to `runDoctorDiagnostics`. Confirmed black-box on the same fault project:

```
main  build: doctor --json → workflow_hooks NOT emitted   (13 checks)
#936  build: doctor --json → workflow_hooks emitted, fail (14 checks)
```

That is the actual gap that let the incident through, and it is genuinely closed.

**All three failure reasons are distinguished** — exercised for real, not asserted:

| filesystem state | result |
|---|---|
| hook absent | `fail` … `(missing)` |
| present, `chmod 644` | `fail` … `(not executable)` |
| present, `chmod 755` | `pass` |
| directory in place of file | `fail` … `(not file)` |

**The `project start` refusal is real.** Black-box, same fault project, against both builds:

- **#936**: exits 1 before the orchestrator launches — `Workflow preflight failed …: Project configuration fault: invalid WORKFLOW.md hook path: after_create=…/hooks/after_create.sh (missing).`
- **main**: daemon starts, connects, ticks (`tick — 0 active, degraded`) and stays up. That is the incident shape.

**Both new tests are load-bearing** (mutation-checked by me, not taken on trust):

- Removing the orchestrator validation block turns `reports a missing hook as a project fault without consuming issue retries` red (`expected 'idle' to be 'degraded'`).
- Removing the `start.ts` preflight turns both `rejects an %s configured hook before acquiring the project lock` cases red.
- Restored: `service.test.ts`, `start.test.ts`, `doctor.test.ts`, `hooks.test.ts` → **4 files, 346 tests, all passing.**

The orchestrator test is well constructed — it seeds `failureRetryCount: 2` against `maxFailureRetries: 3` and asserts it is still 2 after two ticks with `spawnImpl` never called. That encodes the incident semantics precisely.

---

### 🔴 P1 — BLOCKING: a bare hook name bypasses the gate entirely and reproduces the incident

Codex raised this on `hooks.ts:62`; I verified it independently and it is correct. `isWorkflowHookPathLike` requires a `/` or `\`, so a bare `setup.sh` is bucketed as "inline" and never validated. But `resolveHookCommandSpawn` (`hooks.ts:404-430`) has **no inline branch at all** — line 426 and line 429 both `return { ok: true, command: \`./${trimmed}\` }`. Every relative command is executed as a path; anything with whitespace or a shell metacharacter is *rejected* at line 413. So the validator's `inline` category does not correspond to anything the executor supports.

Demonstrated black-box on the #936 build, two projects differing only by a cosmetic `./`:

```
after_create: setup.sh     (missing) → workflow_hooks PASS
                                       "No hook paths required filesystem validation.
                                        Treated 1 hook as inline command."
after_create: ./setup.sh   (missing) → workflow_hooks FAIL  … (missing)
```

Both execute as `./setup.sh`. A project written in the bare form passes `doctor`, passes the `start` preflight, passes the per-tick gate, then hard-fails on dispatch and burns each issue's retry budget — **the exact failure this PR exists to prevent**.

The predicate was moved verbatim from `doctor.ts:1333`, so the flaw is inherited rather than introduced. It was tolerable when this was an advisory `--smoke` check; it is not tolerable now that the same predicate gates dispatch and daemon startup. Fix: classify every argument-free command the executor accepts as path-like, so validation and execution share one definition.

### 🔴 P1 — BLOCKING: validation resolves all four hooks against the workflow directory, but only `after_create` is executed that way

Codex raised this as P2 on `hooks.ts:97`; I verified it and I rate it P1 because it can refuse to start a project that works today.

`service.ts:4938-4944` resolves a relative hook against `hookBaseDirectory` **only when `kind === "after_create"`**; the other three keep the configured relative string and are spawned with `cwd = repositoryPath` (the populated checkout, `service.ts:4973`). But `validateWorkflowHookPaths` resolves all four against the single `baseDirectory`. In standalone project mode, where WORKFLOW.md lives outside the checkout, that splits two ways:

- a repo-relative `before_run` that works today is **rejected at startup** — a false positive that now blocks the daemon entirely rather than warning;
- a project-folder `before_run` **passes** validation and then ENOENTs at execution — the very false negative this PR removes for `after_create`.

Validation must use the same per-kind base directory the executor uses. This does not bite the live project because it configures only `after_create`, which is why CI and my smoke run are both green on it — but the gate is now a hard failure, so a wrong answer here takes a project down.

### 🟡 P2 — the due-retry restart path still runs ahead of the gate

Codex's `service.ts:1392` finding. Confirmed by ordering: `reconcileRun` (`service.ts:1349`) and `handleRestartRunFailure` (`:1359`) both execute **before** `validateWorkflowHookPaths` (`:1401`). A run sitting in `retrying` whose retry becomes due reaches `restartRun` first, so a `before_run` removed mid-flight is still handled as a restart failure and still increments `failureRetryCount` — potentially suppressing an issue already at `max_failure_retries - 1`.

The new test covers the dispatch path only, so this gap is invisible to it. Validate before the due-retry restart path, and extend the test to cover a run in `retrying` state.

### 🟡 P2 — no changeset

`.changeset/` is untouched, but this changes released `@gh-symphony/cli` behavior twice over: `doctor` emits a new `workflow_hooks` entry in its default (non-`--smoke`) run and in `--json` output, and `project start` now refuses to start on an invalid hook path. `CLAUDE.md` requires a changeset for behavior-changing PRs. Add one (patch or minor as you judge the `start` refusal).

---

### Verdict

**Changes requested.** The diagnosis, the promotion out of `--smoke`, the shared core helper, the executable/not-file distinctions, the `start` refusal and both new tests are all correct and should stay exactly as they are. What is missing is that the gate's notion of "a hook path" does not match the executor's in two independent ways, and both holes let the original incident through. Fixing the two P1s means one shared definition of how a hook command resolves — used by `validateWorkflowHookPaths`, `resolveHookCommandSpawn` and the orchestrator's per-kind base directory alike — which is the right shape for this change anyway.

Issue #929 returned to `Ready`. Credit to Codex for first flagging all three code findings; I verified each independently and the black-box `setup.sh` / `./setup.sh` contrast above is the reproduction.

packages/orchestrator/src/service.ts
```diff
@@ -1387,6 +1389,26 @@ export class OrchestratorService {
         tenant,
         tenant.repository
       );
+      if (isUsableWorkflowResolution(workflowResolution)) {
```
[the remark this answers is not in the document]
moncher-dev: Fixed in 14dd42f. Due retries now validate hook paths immediately after the retry becomes due and before tracker refresh, capacity checks, or restartRun, so a missing repository hook leaves failureRetryCount unchanged.

packages/core/src/workspace/hooks.ts
```diff
@@ -34,6 +36,98 @@ export type HookExecutionOptions = {
...
+  const trimmed = command.trim();
+  return (
+    trimmed.length > 0 &&
+    !trimmed.includes("\n") &&
+    !/\s/.test(trimmed) &&
+    (trimmed.startsWith("/") ||
+      trimmed.startsWith("./") ||
+      trimmed.startsWith("../") ||
+      trimmed.includes("/") ||
+      trimmed.includes("\\"))
```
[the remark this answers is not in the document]
moncher-dev: Fixed in 14dd42f. Bare argument-free names now use the same path predicate as the executor and are resolved/validated as script paths; core coverage exercises a missing bare setup.sh.

packages/core/src/workspace/hooks.ts
```diff
@@ -34,6 +36,98 @@ export type HookExecutionOptions = {
...
+  for (const [hook, command] of configured) {
+    if (!command) continue;
+    if (!isWorkflowHookPathLike(command)) {
+      result.inline += 1;
+      continue;
+    }
+
+    const path = isAbsolute(command)
+      ? command
+      : resolve(baseDirectory, command);
```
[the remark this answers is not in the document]
moncher-dev: Fixed in 14dd42f. Shared hook resolution now selects the workflow directory for after_create and the populated repository directory for before_run, after_run, and before_remove; validation and execution use that same resolver.

moncher-dev (commented):
Steward review: **blocking**. `pnpm install --frozen-lockfile && pnpm build && pnpm test` are green in a scratch worktree at `7f33e2c3` (orchestrator 342/342, cli 570/570), and the `AGENT_TEST.md` conflict against `cbca5c3c` is resolved cleanly with both sides' provisioning text intact. The last cycle's three defects stay fixed; I am not restating them.

I also checked the containment of the new poll-loop `throw` before anything else, given this repo's history with tick-aborting exceptions: it is raised inside the per-tenant `try` at `service.ts:1385` and caught at `2194`, which sets `lastError`/`trackerError` and still saves state and builds the project snapshot. One project with a bad hook path degrades only itself. That part is right.

The blocking problem is that the new documentation states a resolution rule the code does not implement, and the new gate is blind to exactly the cases where the two diverge.

**1. `docs/configuration.md` is false for three of the four hook kinds.** The added text says flatly:

> Relative hook paths resolve from the directory containing `WORKFLOW.md`.

`resolveWorkflowHookPath` picks the base directory per kind:

```ts
const baseDirectory =
  hook === "after_create"
    ? directories.workflowDirectory
    : directories.repositoryDirectory;
```

So only `after_create` resolves from the `WORKFLOW.md` directory. `before_run`, `after_run` and `before_remove` resolve from `repositoryDirectory` — the per-issue workspace. Probe against the shipped functions on this branch:

```
PROBE after_create -> /…/hookprobe-WMqRrx/project/hooks/after_create.sh
PROBE before_run   -> /…/hookprobe-WMqRrx/workspace/hooks/before_run.sh
```

An operator who follows the new sentence and puts `hooks/before_run.sh` beside `WORKFLOW.md` gets `<issue-workspace>/hooks/before_run.sh` instead — a path that generally does not exist.

**2. The gate cannot see that mistake, so it lands in the retry budget anyway.** The `project start` / `doctor` / poll-tick call site passes only `{ workflowDirectory }` — no `repositoryDirectory`, correctly, since no workspace exists yet. `validateWorkflowHookPaths` therefore returns `null` from `resolveWorkflowHookPath` for those three kinds and counts them as `deferred`. Same probe, same run:

```
PROBE gate checked  : [{"hook":"after_create", …/project/hooks/after_create.sh"}]
PROBE gate deferred : 1
PROBE gate problems : []
```

A misconfigured `before_run` passes validation with zero problems, the issue is dispatched, and the hook fails at execution — consuming retry budget. That is the failure mode #929 exists to prevent, surviving for three of the four hook kinds. Only after the run reaches `retrying` does the second gate (`reconcileRun`, which *does* pass `repositoryDirectory: run.workingDirectory`) stop further consumption. So the first retry is still burned.

Deferring is the right call — you cannot stat a workspace that has not been created. The defect is that nothing tells the reader, and the surrounding prose promises the opposite:

- `README.md`: "Both `project start` and the default `doctor` run reject absent, non-file, or non-executable hook paths … **before any issue dispatch can consume retry budget**" — unqualified, but true only for `after_create` and absolute paths.
- `docs/architecture.md`: "Path-like workflow hooks are validated as executable regular files (following symlinks) **relative to the workflow source**" — again describes only the `after_create` rule as if it were general.
- `.changeset/cool-hooks-wait.md`: "Validate configured workflow hook scripts before project dispatch" — same overclaim.

What I'd want before this lands: make the per-kind base directory explicit in `docs/configuration.md`, qualify the three claims above to say which kinds are checked pre-dispatch and which are deferred to reconciliation, and let `validateWorkflowHookPaths` surface the deferred count to the operator rather than dropping it silently — it is already computed and returned, and nothing reads it. Right now `deferred` is a field no caller consumes, which is why the gap is invisible.

**3. Minor — `TC-23` collides with PR #941.** #941 also adds `e2e/scenarios/23-cross-runtime-root-project-lock.md` as TC-23 plus its own `AGENT_TEST.md` index row. Different filenames, so both merge cleanly and leave two TC-23s in one table. `e2e/scenarios/` already has duplicate 10/13/15/17/19 pairs so the pattern predates you — but whichever lands second should renumber.

Moving #929 back to Ready with the above.

moncher-dev:
Rework completed in `7ef00586`.

- `project start` now warns when repository-relative run/remove hooks are deferred until an issue workspace exists; default `doctor` uses the same diagnostic, including alongside other hook failures.
- Documentation and the changeset now state the actual per-kind base directories and explicitly distinguish pre-dispatch validation from the initial workspace-relative execution case.
- The standalone scenario is renumbered from TC-23 to TC-24 to avoid the collision with #941.
- Focused suites pass (346 tests), and the full lint/test/typecheck/build gate passes. Docker E2E was attempted twice but the host daemon remains unavailable, tracked in #938.

moncher-dev (commented):
## Steward review — BLOCK (returning to `Ready`)

Build, lint, typecheck and all 12 package suites pass on `7ef00586` (1946 tests, 0 failures) in a scratch worktree.

### Both of last cycle's blocking findings are fixed — verified by probe, not by reading the diff

I ran the shipped `packages/core/dist/index.js` against real scripts on disk, executing each hook kind with `executeWorkspaceHook` and having the script report its own `$0`/`pwd`. With the script present **only** in the workflow directory:

```
after_create   resolved=<ROOT>/workflowdir/hooks/h.sh   success
before_run     resolved=<ROOT>/workspacedir/hooks/h.sh  failure ENOENT
after_run      resolved=<ROOT>/workspacedir/hooks/h.sh  failure ENOENT
before_remove  resolved=<ROOT>/workspacedir/hooks/h.sh  failure ENOENT
```
Inverted (script only in the workspace dir), `after_create` fails and the other three succeed. The per-kind split is real, and both the executor (`service.ts:4970-4977`) and every gate now call the same `resolveWorkflowHookPath` (`hooks.ts:64-79`), so their bases agree by construction. The reconciliation gate supplies the missing half (`service.ts:3676-3682`, `repositoryDirectory: run.workingDirectory`, which is exactly the executor's value via `service.ts:3001` → `:3254`); with both dirs supplied all three workspace kinds move from `deferred` to `checked`. The docs were corrected to match and their stated rationale is now **true**.

`deferred` also has real readers now: `start.ts:263-266` (stderr warning, asserted in `start.test.ts`) and `doctor.ts:1299-1301, 1310, 1327` (both fail and pass messages, plus JSON details).

The new tests are not happy-path only — `start.test.ts:428-470` covers absent *and* non-executable `after_create`; `service.test.ts:1095-1219` and `:16149-16205` assert `failureRetryCount` stays put across both gates. Good.

### Blocking — the `deferred` bucket also swallows hooks that can *never* run, and tells the operator something false about them

`validateWorkflowHookPaths` increments `deferred` in two disjoint cases (`hooks.ts:107-116`):
- (a) path-like but no base directory yet — the legitimate case this PR is about;
- (b) **`!isWorkflowHookPathLike(command)`** — any hook value containing whitespace or shell metacharacters.

Case (b) is not deferrable, it is permanently invalid. `isWorkflowHookPathLike` (`hooks.ts:54-62`) and the executor's `resolveHookCommandSpawn` (`hooks.ts:429-444`) use the **identical predicate**, so `!isWorkflowHookPathLike(cmd)` ⟺ the executor hard-fails. `docs/configuration.md:91-93` and `README.md:874-876` both state this class is rejected — but no gate rejects it, and the message the gates print about it is untrue.

**Reproduction** — calls exactly what `start.ts:255` / `doctor.ts:1293` call, then what the orchestrator executes, with config `hooks: { before_run: "hooks/setup.sh --verbose" }` and the script present and executable:

```
GATE  problems = 0  deferred = 1
GATE  message  = Deferred 1 repository-relative workflow hook until an issue workspace is available;
                 these hooks resolve from that workspace and are validated during run reconciliation.
EXEC  outcome  = failure | Workflow hooks must reference a script path without shell syntax or arguments.
```

Every clause of that message is wrong for this input: it is not repository-relative, it does not resolve from a workspace, and reconciliation will never validate it — `hooks.ts:107-110` defers *before* any directory is consulted, so the same text is printed even for `after_create`, which is not workspace-relative at all.

**Why this blocks rather than being a nit:** `parser.ts:366-371` accepts any string (`readOptionalString`), so such a config parses, passes `project start` preflight with only a warning, passes `doctor` as a **PASS**, and gets dispatched. At runtime `service.ts:3149-3150` throws, the `startRun` catch at `service.ts:1822-1866` computes `failureRetryCount = prior + 1`, and the retries burn to suppression — which is precisely the #929 symptom this PR is titled to prevent, for a config class the docs already declare invalid. The PR's own test cements the misclassification: `hooks.test.ts:226-256` passes `afterRun: "echo ready"` and asserts `deferred: 1`.

**Fix shape:** classify `!isWorkflowHookPathLike(command)` as a third outcome (e.g. `problems` with `reason: "not_a_path"`) so `start.ts:257-261`, `doctor.ts:1303-1320` and both orchestrator gates fail it as a project configuration fault. Keep `deferred` for the genuine "no workspace yet" case only — one bucket, one meaning.

### Non-blocking

1. **Undocumented behavior change.** Bare command names now hard-fail `project start`: `hooks: { after_create: "make" }` → `problems=[["missing","<workflowdir>/make"]]`. The old `doctor.ts` `isHookPathLike` required a `/` and treated `make` as an inline command (PASS). The new behavior matches the executor (which spawns `./make`), so such a config was already broken — but any existing project with one now cannot start at all. Worth a line in `.changeset/cool-hooks-wait.md`.
2. **Overstated doc line.** `packages/cli/README.md:430-436` says start/doctor "validate every path-like hook command and report its resolved absolute path". Workspace-relative path-like hooks are *not* validated there and no resolved path is printed for them. `README.md` and `docs/configuration.md` state it correctly; the CLI README does not.
3. **A third gate is still absent.** `gh-symphony workflow validate` (`packages/cli/src/commands/workflow.ts:895-942`) only echoes hook values, never validates them — and it is the command users run first. Pre-existing and out of scope for this PR, but it is the same absence class, so worth a follow-up issue.
4. **A configuration fault is routed through the tracker-error channel.** The pre-dispatch throw (`service.ts:1404-1409`) lands in the generic tick catch at `service.ts:2194-2201`, which sets `trackerError` and feeds `resolveDispatchSuppressedUntil(trackerError, …)`. It does reach a named destination (snapshot `lastError` + `degraded`, asserted by tests) and consumes no retry, so this is not blocking — but classifying a config fault as a tracker fault for backoff purposes will mislead whoever reads the snapshot.
5. **Stale prose from a deleted concept.** `docs/reports/2026-08-28-upstream-spec-drift-research.md:143` still cites `README.md:873-885` for inline-hook drift; those line numbers have shifted. Cosmetic.

packages/core/src/workspace/hooks.ts
```diff
@@ -34,6 +36,125 @@ export type HookExecutionOptions = {
...
+    ["before_remove", hooks.beforeRemove],
+  ] as const;
+  const result: WorkflowHookPathValidation = {
+    checked: [],
+    deferred: 0,
+    problems: [],
+  };
+
+  for (const [hook, command] of configured) {
+    if (!command) continue;
```
moncher-dev, line 108: **Blocking — the `deferred` bucket also swallows hooks that can *never* run, and prints something false about them.**

`validateWorkflowHookPaths` increments `deferred` in two disjoint cases: (a) path-like but no base directory yet — the legitimate case this PR is about; and (b) `!isWorkflowHookPathLike(command)`, i.e. any hook value with whitespace or shell metacharacters.

Case (b) is not deferrable, it is permanently invalid. `isWorkflowHookPathLike` and the executor's `resolveHookCommandSpawn` use the **identical predicate**, so `!isWorkflowHookPathLike(cmd)` ⟺ the executor hard-fails. `docs/configuration.md:91-93` and `README.md:874-876` both state this class is rejected — but no gate rejects it.

**Reproduction** — calls exactly what `start.ts:255` / `doctor.ts:1293` call, then what the orchestrator executes, with `hooks: { before_run: "hooks/setup.sh --verbose" }`, script present and executable:
```
GATE  problems = 0  deferred = 1
GATE  message  = Deferred 1 repository-relative workflow hook until an issue workspace is available;
                 these hooks resolve from that workspace and are validated during run reconciliation.
EXEC  outcome  = failure | Workflow hooks must reference a script path without shell syntax or arguments.
```
Every clause of that message is wrong for this input: it is not repository-relative, it does not resolve from a workspace, and reconciliation will never validate it — the deferral happens *before* any directory is consulted, so the same text prints even for `after_create`, which is not workspace-relative at all.

**Why it blocks:** `parser.ts:366-371` accepts any string, so this config parses, passes `project start` preflight with only a warning, passes `doctor` as a **PASS**, and gets dispatched. At runtime `service.ts:3149-3150` throws, `startRun`'s catch at `service.ts:1822-1866` computes `failureRetryCount = prior + 1`, and retries burn to suppression — precisely the #929 symptom this PR is titled to prevent, for a config class the docs already declare invalid.

**Fix shape:** classify `!isWorkflowHookPathLike(command)` as a third outcome (e.g. `problems` with `reason: "not_a_path"`) so `start.ts:257-261`, `doctor.ts:1303-1320` and both orchestrator gates fail it as a project configuration fault. Keep `deferred` for the genuine "no workspace yet" case only — one bucket, one meaning.

Note the PR's own test cements the misclassification: `hooks.test.ts:226-256` passes `afterRun: "echo ready"` and asserts `deferred: 1`. That assertion needs to flip with the fix.

moncher-dev (commented):
## Steward review — re-posted as **anchored inline threads**

My earlier top-level `COMMENTED` review on this PR had no inline threads. I have since confirmed on #909/#943 that the Step 0 rework guard does not see that channel — nor the issue-comment workaround, when the steward posts under `moncher-dev` (the worker's own account). That is issue #900. Same finding below, anchored so `reviewThreads` is non-empty.

**Both of last cycle's blocking findings are fixed, and I verified the fix by probe rather than by reading the diff.** Running the shipped `packages/core/dist/index.js` against real scripts on disk, with the script present only in the workflow dir: `after_create` succeeds while `before_run` / `after_run` / `before_remove` fail ENOENT against `<workspacedir>/hooks/h.sh`; inverted, the result inverts. Both the executor and every gate now call the same `resolveWorkflowHookPath`, so their bases agree by construction, and with both dirs supplied all three workspace kinds move from `deferred` to `checked`. `deferred` now has real readers in `start.ts:263-266` and `doctor.ts:1299-1327`. The docs were corrected and their rationale is now true. The new tests exercise failing cases, not just passing ones.

One blocking finding remains, plus non-blocking notes in the issue-comment copy on #929.

packages/cli/README.md
```diff
@@ -427,6 +427,13 @@ the status surface never expands it into environment names or values.
 
 Use a project folder as an orchestration instance decoupled from the repository it targets. The folder owns `WORKFLOW.md` (with `repository.slug: owner/name`), plus its `hooks/after_create.sh`, optional `.mcp.json`, `.env`, and `.agent/skills/`. The orchestrator creates issue directories and invokes `after_create`; the shipped default hook fully clones the target and checks out the project-scoped issue branch so the host publication transport can fetch the branch from the workspace. `start` derives and caches configuration from the folder on every run; `status` and `stop` address the same runtime by folder without reading `WORKFLOW.md`.
 
+Create the folder with `gh-symphony workflow init` to scaffold
```
moncher-dev, line 432: **Non-blocking — overstated.** This says start/doctor "validate every path-like hook command and report its resolved absolute path". Workspace-relative path-like hooks are *not* validated there and no resolved path is printed for them — they are deferred. `README.md` and `docs/configuration.md` state it correctly; this one does not.

docs/configuration.md
```diff
@@ -288,6 +288,22 @@ retained in plaintext cache metadata. An explicit `--config <dir>` is exported
 to `GH_SYMPHONY_CONFIG_DIR` for the process so spawned workers use the same
```
moncher-dev, line 290: **Non-blocking — undocumented behavior change worth a changeset line.** Bare command names now hard-fail `project start`: `hooks: { after_create: "make" }` → `problems=[["missing","<workflowdir>/make"]]`. The old `doctor.ts` `isHookPathLike` required a `/` and treated `make` as an inline command (PASS).

The new behavior matches the executor (which spawns `./make`), so such a config was already broken — but any existing project holding one now cannot start at all. Please add a line to `.changeset/cool-hooks-wait.md`.

Separately, and out of scope for this PR but the same absence class: `gh-symphony workflow validate` (`packages/cli/src/commands/workflow.ts:895-942`) only *echoes* hook values and never validates them — and it is the command users reach for first. Worth a follow-up issue.

moncher-dev:
Rework completed in `d109029f`.

- Hook strings containing whitespace, arguments, or shell metacharacters now produce an explicit `not_a_path` configuration problem; they are no longer counted as deferred.
- The deferred bucket now has one meaning only: a valid repository-relative run/remove hook cannot yet be resolved because no issue workspace exists.
- Core coverage asserts both argument-bearing and shell-syntax commands fail validation, and the existing doctor fixture now proves a genuine workspace-relative hook remains deferred.
- The CLI README now distinguishes immediate project-root/absolute validation from deferred workspace-relative validation, and the changeset documents bare command-name behavior.
- Focused suites pass (126 tests), and the full lint/test/typecheck/build gate passes.

The `workflow validate` gap remains out of scope for #929 as noted in the review.

moncher-dev (commented):
Passing. Both of last round's blocking findings are now closed, and I verified this round by probe against the built output rather than by reading the diff.

## The remaining blocker is fixed

Last round `!isWorkflowHookPathLike` commands were absorbed into `deferred`, so a hook that could never run passed `doctor` as PASS, dispatched, then threw and burned retries to suppression — the #929 symptom itself. They now raise an explicit `not_a_path` problem. Probing the built `packages/core/dist/index.js`:

```
bare name, file exists in workflow dir   deferred=0 checked=1 problems=[]
bare name, missing                       deferred=0 checked=0 [["after_create","missing","<WF>/absent.sh"]]
shell syntax                             deferred=0 checked=0 [["after_create","not_a_path","hooks/x.sh; echo hi"]]
args                                     deferred=0 checked=0 [["after_create","not_a_path","hooks/x.sh --v"]]
relative before_run, no repoDir          deferred=1 checked=0 problems=[]
```

`deferred` is now reached only by a genuinely path-like hook whose base directory does not exist yet, which is what the name claims.

## End-to-end, with the built CLI on a real standalone project folder

Three fault classes, each stopping `project start` before any dispatch and naming the resolved absolute path:

```
A: after_create missing          exit=1  after_create=<proj>/hooks/after_create.sh (missing)
B: after_create chmod 644        exit=1  after_create=<proj>/hooks/after_create.sh (not executable)
C: before_run "hooks/br.sh; echo pwned"  exit=1  before_run="hooks/br.sh; echo pwned" (not a path)
```

Case C is the one that mattered: on `main` that configuration starts cleanly and fails later against the issue's retry budget. It now cannot start.

## Absence checks I ran

- **Where else does this pattern occur?** All three gates — `start.ts:255`, `doctor.ts:1293`, and both reconciliation sites in `service.ts` (1402, 3676) — call the same `validateWorkflowHookPaths` and format through the same helper, so a new reason cannot be handled in one place and dropped in another. `not_a_path` is handled uniformly.
- **Is the stated reason true?** The validator and the executor now share one `SHELL_METACHARACTER_PATTERN` constant, and `isWorkflowHookPathLike` applies exactly the conditions `resolveHookCommandSpawn` rejects. They agree by construction rather than by matching literals, which is what I asked for last round. Bare-name resolution also matches: the executor spawns `./<name>` from the execution directory, and the validator resolves `<baseDirectory>/<name>`.
- **Does every branch have a named destination?** `missing`, `not_file`, `not_executable`, `not_a_path` all render; `deferred` has real readers in `start.ts:263` and `doctor.ts:1299`.
- **Residual gap, honestly disclosed.** A workspace-relative hook's first execution is still the first point at which its path is observable and can consume one retry. The README, `docs/configuration.md`, and `docs/architecture.md` now all say so in the same terms rather than implying full coverage, and reconciliation validates the retained workspace before any restart. I am satisfied with this as a documented bound rather than a hidden one.

## Tests

`pnpm install --frozen-lockfile` and `pnpm build` clean. `packages/core/src/workspace/hooks.test.ts`, `packages/cli/src/commands/{start,doctor}.test.ts`, `packages/orchestrator/src/service.test.ts` — **347 passed**. The new hook tests assert failing cases, not only passing ones.

Handing to @hojinzs. The judgement call worth your eye is the changeset's semantic change — bare command names are now script paths relative to the execution directory, so a missing bare-name hook fails validation instead of being treated as an inline command. That matches the existing README rule that Symphony executes no implicit shell, and `resolveHookCommandSpawn` already spawned `./<name>`, so I read it as making the validator agree with long-standing executor behaviour rather than as a new restriction. It ships as a `patch`.

packages/cli/src/commands/start.ts
```diff
@@ -246,6 +252,19 @@ async function preflightWorkflowStart(
         workflowPath,
       }
     );
+    const hookValidation = await validateWorkflowHookPaths(workflow.hooks, {
```
hojinzs: **P1 (blocking) — the gate ignores `SYMPHONY_ALLOW_WORKFLOW_HOOKS`, so it hard-fails projects whose hooks can never run.**

Hook execution is opt-in. `executeWorkspaceHook` (`packages/core/src/workspace/hooks.ts:388-396`) returns `skipped` unless `trusted`, and `trusted` comes from `isWorkflowHookExecutionAllowed` (`service.ts:6291-6295`) reading `SYMPHONY_ALLOW_WORKFLOW_HOOKS`. Without that flag a declared hook is *never spawned*, cannot fail, and cannot consume any retry budget — the entire #929 failure mode is unreachable.

`validateWorkflowHookPaths` takes no trust input and no call site consults the flag, so this preflight rejects those projects anyway.

**Reproduction with the built CLI at `d109029f`** — two identical standalone folders, both declaring `after_create: hooks/after_create.sh` with the script absent; the only difference is `.env`:

```
A: .env has SYMPHONY_ALLOW_WORKFLOW_HOOKS=1
   exit=1  Project configuration fault: invalid WORKFLOW.md hook path:
           after_create=<proj>/hooks/after_create.sh (missing).

B: .env has no approval flag  (default posture)
   exit=1  Project configuration fault: invalid WORKFLOW.md hook path:
           after_create=<proj>/hooks/after_create.sh (missing).
```

A is the fix #929 asked for. B is a regression: `git show origin/main:packages/cli/src/commands/start.ts` contains no hook reference at all, so that folder starts today and cannot start after this PR — for a hook the runtime would have reported as `skipped`. Confirmed directly against the built core:

```
executeWorkspaceHook({ kind: "after_create", afterCreate: "hooks/absent.sh", trusted: false })
  → outcome: skipped
     error:  Workflow hook "after_create" skipped because WORKFLOW.md hooks
             require explicit trust approval.
```

**Why this blocks rather than being a nit:** the same unguarded validation is the pre-dispatch tick gate (`service.ts:1402`). For an already-running daemon on an unapproved project, deleting or mistyping a hook path now throws every tick, sets `lastError`, degrades health, and suppresses all dispatch — converting a runtime no-op into a project that can never dispatch. That is the #929 symptom class this PR exists to remove, widened to a new population. This repo's own `WORKFLOW.md:31` documents skip-by-default as normal operation (`# must export SYMPHONY_ALLOW_WORKFLOW_HOOKS=1; otherwise hooks are skipped`), so the configuration shape is sanctioned, not hypothetical.

Note the E2E added here writes `SYMPHONY_ALLOW_WORKFLOW_HOOKS=1` into every fixture `.env` (`e2e/run-standalone-project-e2e.sh:67`), and no unit test sets or clears the flag — the untrusted path is uncovered end to end, which is why this survived five review cycles.

**Fix shape (small, no redesign):** thread the resolved approval state into the three enforcing call sites (`start.ts:255`, `service.ts:1402`, `service.ts:3676`) and fault only when hooks will actually execute; when they will not, degrade to the same stderr/`doctor` warning channel `deferred` already uses. `doctor` may keep reporting it as a finding — a diagnostic that names a broken path is useful — but it should say the hook is currently skipped rather than presenting it as a dispatch-blocking fault. Please add a test that fixes the flag off and asserts start succeeds.

---

packages/orchestrator/src/service.ts
```diff
@@ -1387,6 +1390,26 @@ export class OrchestratorService {
...
+        const workflowBaseDirectory =
+          this.workflowHookBaseDirectories.get(
+            this.workflowCacheKey(tenant.repository)
+          ) ??
+          (workflowResolution.workflowPath
+            ? dirname(workflowResolution.workflowPath)
+            : null);
+        if (workflowBaseDirectory) {
+          const hookValidation = await validateWorkflowHookPaths(
```
hojinzs, line 1405: **P1 (blocking) — same trust-flag gap as `start.ts:255`, and this is the damaging half of it.**

Full analysis on that thread; the part specific to this gate:

This runs on **every tick**, and the `throw` is unconditional on whether hooks will execute. For a project without `SYMPHONY_ALLOW_WORKFLOW_HOOKS`, a hook path that goes missing after the daemon is already up now raises a project configuration fault each tick. Reading the surrounding flow, that lands in the tick catch at `service.ts:2194-2201`, which sets `trackerError`/`lastError` and feeds `resolveDispatchSuppressedUntil(trackerError, …)` — so dispatch stops for a hook the executor would have reported as `skipped`. On `main` the same edit is a no-op.

Two smaller notes on this block, both non-blocking:

1. The `throw` happens before `pollIntervalMs = await this.loadProjectPollInterval(tenant)` (line 1410), so a faulted project silently reverts from its configured interval to `DEFAULT_POLL_INTERVAL_MS` for the duration of the fault. Not a hot loop, but the snapshot's `effectivePollIntervalMs` will not match the project's configuration and nothing explains why.
2. Routing a configuration fault through the tracker-error channel means `trackerError` and the tracker backoff describe something that is not a tracker problem. The steward already flagged this as non-blocking and I agree it is — but if you are touching this block for the P1 anyway, a distinct fault classification would be cheap here.

The `not_a_path` fix from last cycle is confirmed present, by probe against `packages/core/dist/index.js` rather than by reading:

```
"hooks/x.sh; echo hi"   → problems=[["after_create","not_a_path","hooks/x.sh; echo hi"]]
"hooks/x.sh --verbose"  → problems=[["after_create","not_a_path","hooks/x.sh --verbose"]]
"make" (absent)         → problems=[["after_create","missing","<WF>/make"]]
before_run "hooks/br.sh"→ deferred=1, problems=[]
```

`deferred` is now reached only by a genuinely path-like hook lacking its base directory.

---

packages/core/src/workspace/hooks.ts
```diff
@@ -34,6 +36,130 @@ export type HookExecutionOptions = {
...
+export function resolveWorkflowHookPath(
+  command: string,
+  hook: HookKind,
+  directories: {
+    workflowDirectory: string;
+    repositoryDirectory?: string;
+  }
+): string | null {
+  if (!isWorkflowHookPathLike(command)) return null;
+  if (isAbsolute(command)) return command;
```
hojinzs: **P3 — validator and executor disagree on whitespace, because only one of them trims.**

`isWorkflowHookPathLike` decides on `command.trim()` (line 55), and `resolveHookCommandSpawn` also trims before `isAbsolute` (line 439-451). This function alone uses the raw `command` for both `isAbsolute` and `resolve`, so a value that survives the predicate with surrounding whitespace resolves somewhere the executor would never look.

Probed against the built `packages/core/dist/index.js` with `after_create: " <WF>/hooks/after_create.sh"` (leading space, script present and executable):

```
problems = [["after_create","missing","<WF>/ <WF>/hooks/after_create.sh"]]
```

The base directory gets prepended to an absolute path. With the P1 above unfixed this fails `project start` outright, and the message names a path the operator never wrote.

`readOptionalString` (`packages/core/src/workflow/parser.ts:1216-1229`) does not trim, so the value reaches here as written — a quoted YAML scalar is enough to hit it. Narrow, but the fix is one line: resolve from `command.trim()`, matching the two predicates that already do.

---

AGENT_TEST.md
```diff
@@ -367,6 +367,7 @@ idle → [inject issue + refresh]
 | `e2e/scenarios/16-packaged-runtime-entrypoints.md`      | Verify the built CLI's MCP dispatcher and Git credential helper subprocesses inside Docker                                |
 | `e2e/scenarios/19-required-label-routability.md`        | Verify required-label filtering cancels active runs without workspace cleanup and exposes the reason                      |
 | `e2e/scenarios/20-agent-child-isolation.md`             | Verify unconditional child credential/config isolation, host-only MCP tools, and worker-exit Git publication              |
+| `e2e/scenarios/24-hook-configuration-fault.md`           | Verify a missing standalone-project hook is rejected before dispatch and its resolved path is reported                    |
```
hojinzs: **nit** — TC-24 is inserted between the TC-20 and TC-17 rows. The table is already unordered (…19, 20, 17), so this is pre-existing and harmless; if you touch this file again for the P1, appending at the end costs nothing.

Renumbering away from the TC-23 collision with #941 is done and correct — no note needed there.

---

hojinzs (changes requested):
## Review — REQUEST CHANGES (one P1; everything else is sound)

Reviewed at `d109029f`. This is the requested reviewer pass following the steward hand-off. I verified by running the code, not by reading the diff.

### Verification performed

- `pnpm install --frozen-lockfile` and `pnpm build` — clean.
- `packages/core/src/workspace/hooks.test.ts`, `packages/cli/src/commands/{start,doctor}.test.ts` — **126 passed**.
- `packages/orchestrator/src/service.test.ts` — **221 passed**.
- Probed `validateWorkflowHookPaths` / `resolveWorkflowHookPath` / `executeWorkspaceHook` against the built `packages/core/dist/index.js` with real scripts on disk.
- **Smoke test** against the built CLI (`packages/cli/dist/index.js`): `project start` on real standalone folders, file tracker, missing / present / non-executable / shell-syntax hooks.
- Docker E2E was **not** run — the daemon is unavailable in this environment too, matching the exception recorded in #938. TC-24 is reviewed as written, not executed.

### Requirements vs. issue #929

| --- | --- |
| Absent / non-executable hook rejected at start with the resolved path named; `doctor` reports the same | ✅ verified by smoke test for `after_create` and absolute paths; workspace-relative kinds deferred and documented |
| `workflow_hooks` runs in the **default** `doctor` pass, not only `--smoke` | ✅ moved out of `buildDoctorSmokeChecks` |
| Missing hook does not decrement any issue's failure retry budget | ⚠️ met for `after_create` and for all kinds at reconciliation; a workspace-relative hook's first execution can still spend exactly one retry. Documented in three places in the same terms, never reaches suppression. I accept this as a stated bound — you cannot stat a workspace that does not exist yet |
| Tests cover absent, non-executable, and retry non-consumption | ✅ `hooks.test.ts` (4 new cases incl. symlink and bare-name), `service.test.ts` (both gates assert `failureRetryCount` unchanged) |
| Documented provisioning path exists and is exercised | ✅ `docs/configuration.md`, both READMEs, TC-24 + runner assertion |

No over-engineering found. The two judgement calls both narrow toward existing behavior rather than inventing policy: `not_a_path` rejects a class `docs/configuration.md` already declared invalid, and bare-name-as-path matches what `resolveHookCommandSpawn` has always spawned (`./<name>`). Both are disclosed in the changeset.

### Prior review threads

All three open threads are resolved in code — verified independently:

- `hooks.ts:107` `not_a_path` misclassification — **fixed**; probe output on the `service.ts:1402` thread.
- `packages/cli/README.md:430` overstated coverage — **fixed**; now qualifies project-root/absolute vs. deferred.
- `docs/configuration.md:288` bare-name behavior change absent from the changeset — **fixed**; `.changeset/cool-hooks-wait.md` carries it.

Please resolve those threads.

### Blocking finding

**P1 — every gate ignores `SYMPHONY_ALLOW_WORKFLOW_HOOKS`.** Hook execution is opt-in; without the flag `executeWorkspaceHook` returns `skipped`, so the hook is never spawned, never fails, and never touches the retry budget. The new gates validate regardless. Two identical folders differing only in that flag both fail `project start` at `exit=1`, and `main`'s `start.ts` contains no hook reference at all — so the unapproved one starts today and cannot after this PR. Worse at `service.ts:1402`: a running daemon on an unapproved project now throws every tick and suppresses all dispatch for a hook the runtime would have skipped. That is the #929 symptom class, widened to a new population.

Details, reproduction, and a small fix shape are on the `start.ts:255` and `service.ts:1402` threads. Neither the unit tests nor the E2E exercise the untrusted path (`run-standalone-project-e2e.sh:67` writes the flag into every fixture), which is why it survived five cycles.

### Non-blocking

- **P3** — `resolveWorkflowHookPath` does not trim while both predicates do, so a whitespace-padded absolute path resolves under the base directory. One-line fix; thread on `hooks.ts:73`.
- **nit** — poll interval silently reverts to the default while a project is faulted (`service.ts:1402` thread).
- **nit** — TC-24 row ordering in `AGENT_TEST.md`.

### Verdict

The design is right and I want it merged. Fix the P1 — bound the fault to projects where hooks will actually execute, and add a test with the flag off — and this is an approve on the next pass. **This is the final review round**: with that one change in, ship it; I will not open new lines of feedback.

---

packages/core/src/workspace/hooks.ts
moncher-dev: Resolved in `d109029f`: invalid hook strings now produce `not_a_path` problems, while `deferred` is reserved for valid workspace-relative paths whose repository base is not yet available. The focused and full test suites continue to pass at `579d5ff0`.

packages/cli/README.md
moncher-dev: Resolved in `d109029f` and clarified again in `579d5ff0`: the CLI README distinguishes immediately validated project-root/absolute paths from deferred workspace-relative hooks, and now also states that enforcing startup validation follows the hook trust flag.

docs/configuration.md
moncher-dev: Resolved: `.changeset/cool-hooks-wait.md` documents bare command names as relative script paths and now also records that disabled hooks remain non-blocking while doctor stays diagnostic. The separate `workflow validate` enhancement remains outside #929.

packages/cli/src/commands/start.ts
moncher-dev: Resolved in `579d5ff0`: `project start` reads the managed project environment and enforces filesystem hook validation only when `SYMPHONY_ALLOW_WORKFLOW_HOOKS` is `1` or `true`. A regression test fixes the flag off, declares a missing `after_create`, and verifies startup reaches lock acquisition and the orchestrator run.

packages/orchestrator/src/service.ts
moncher-dev: Resolved in `579d5ff0`: both the pre-candidate tick gate and due-retry reconciliation gate now use the same resolved trust decision as hook execution, so disabled hooks cannot become project-configuration faults. The existing runtime behavior for a configured but untrusted population hook remains unchanged and is asserted separately.

packages/core/src/workspace/hooks.ts
moncher-dev: Resolved in `579d5ff0`: `resolveWorkflowHookPath` trims once before both `isAbsolute` and `resolve`, matching the predicate and executor. The new core regression uses a whitespace-padded absolute executable path and verifies the canonical path is checked.

AGENT_TEST.md
moncher-dev: Resolved in `579d5ff0`: TC-24 now appears after the TC-17 row at the end of the scenario table.

packages/cli/src/commands/start.ts
hojinzs, line 1: **P1 from last round — verified fixed. Not by reading the diff; by running the built CLI on real standalone folders.**

Two folders, identical except `.env`, both declaring `after_create: hooks/after_create.sh` with the script absent:

```
trusted   (.env: SYMPHONY_ALLOW_WORKFLOW_HOOKS=1)
  exit=1  Project configuration fault: invalid WORKFLOW.md hook path:
          after_create=<proj>/hooks/after_create.sh (missing).

untrusted (.env: no approval flag)
  exit=0* starts, connects the repository, reaches the poll loop,
          "tick — 0 active, idle", no fault, no degrade
          (*exit=124 under `timeout`, i.e. still running)
```

That is exactly the split I asked for: #929's fault survives where hooks execute, and the regression against unapproved projects is gone. The tick gate is covered by the same change — the untrusted daemon ticked cleanly rather than throwing per-tick.

**The trust resolution agrees with the executor's, including via exported shell env.** This was my main worry about the fix shape, since `start.ts:70` introduces a *third* copy of the predicate rather than reusing `isWorkflowHookExecutionAllowed`. It happens to be equivalent, and not by luck: `resolveManagedProjectEnvironment` returns `{...readEnvFile(envPath), ...process.env}`, the same precedence as the daemon's `resolveProjectEnvironment`, so `environment[FLAG]` already carries the `process.env` fallback that `isWorkflowHookExecutionAllowed` applies explicitly. Confirmed against the sanctioned form in this repo's own `WORKFLOW.md:31` (`export`, not `.env`):

```
untrusted folder + SYMPHONY_ALLOW_WORKFLOW_HOOKS=1 exported in the shell
  → exit=1, fault raised with the resolved absolute path
```

Both orchestrator gates (`service.ts:1395`, `:3672`) call `isWorkflowHookExecutionAllowed` directly, so those agree by construction. Only this call site restates the rule.

**nit (do not block on it):** fold `WORKFLOW_HOOK_APPROVAL_ENV` and the `=== "1" || toLowerCase() === "true"` test into one exported core helper and have all three sites call it. Today the agreement between this line and `hooks.ts`'s executor is a coincidence of two independently-written expressions — the same shape of coupling that produced the resolver and predicate defects in the earlier rounds of this PR, both of which were fixed by making the two sides share one function. Worth doing whenever this file is next touched.

---

packages/core/src/workspace/hooks.ts
hojinzs, line 1: **P3 whitespace divergence — verified fixed.** `resolveWorkflowHookPath` now trims once before both `isAbsolute` and `resolve`, matching `isWorkflowHookPathLike` and `resolveHookCommandSpawn`. Probed against the built `packages/core/dist/index.js`:

```
padded absolute  "  <WF>/h.sh  "   checked=1 deferred=0 problems=[]
padded relative  "  h.sh  "        checked=1 deferred=0 problems=[]
shell syntax     "h.sh; echo hi"   checked=0 deferred=0 [["after_create","not_a_path"]]
with args        "h.sh --verbose"  checked=0 deferred=0 [["after_create","not_a_path"]]
before_run "h.sh" (no repoDir)     checked=0 deferred=1 problems=[]
```

Previously the first case produced `[["after_create","missing","<WF>/ <WF>/h.sh"]]` — a base directory prepended to an absolute path, naming a path the operator never wrote. Now it resolves canonically and passes.

The last two rows also re-confirm last round's `not_a_path` fix holds at this head, and that `deferred` is still reached only by a genuinely path-like hook whose base directory does not exist yet — one bucket, one meaning.

---

packages/cli/src/commands/doctor.ts
hojinzs, line 1: **nit — non-blocking, and I am explicitly not asking for a change before merge.**

`buildWorkflowHookChecks` calls `validateWorkflowHookPaths` unconditionally, so for a project without `SYMPHONY_ALLOW_WORKFLOW_HOOKS` a broken hook path is reported as a `failCheck`. That is the behavior I said was acceptable last round — the path genuinely *is* invalid, `doctor` blocks nothing, consumes no retry, and the remediation text ("Create the referenced hook script(s), make them executable, or fix the hook path(s)") is true regardless of trust. Diagnosing a latent fault before the operator turns hooks on is the useful direction to err in.

The only imprecision left: the message does not mention that the hook is currently **skipped** and would not run even if the path were fixed. An operator running `doctor` on an unapproved project sees a hard FAIL for something inert. A clause like "hook execution is not approved for this project; this path is diagnosed but not currently executed" would close it. `.changeset/cool-hooks-wait.md` already records the intended split ("Disabled hooks remain non-blocking, while doctor still diagnoses their paths"), so the behavior is documented — it is just not self-describing in the output.

Follow-up material, not merge material.

---

hojinzs (approved):
## Review — APPROVE (final round, as announced)

Reviewed at `579d5ff0`. The single P1 from my last pass is fixed and I verified it by running the code, not by reading the diff. Per my previous review this was the last round; I am opening no new lines of feedback and the two notes below are follow-up material.

### Verification performed

- `pnpm install --frozen-lockfile`, `pnpm build`, `pnpm lint`, `pnpm typecheck` — all clean.
- `packages/core/src/workspace/hooks.test.ts`, `packages/cli/src/commands/{start,doctor}.test.ts` — **128 passed**.
- `packages/orchestrator/src/service.test.ts` — **222 passed**.
- Probed `validateWorkflowHookPaths` / `resolveWorkflowHookPath` against the built `packages/core/dist/index.js`.
- **Smoke test** with the built CLI (`packages/cli/dist/index.js`): `project start` on real standalone folders, file tracker, missing `after_create`, trust flag on vs. off, and via exported shell env vs. `.env`.
- Docker E2E **not** run — daemon unavailable here too, matching the exception recorded in #938. TC-24 reviewed as written, not executed.

### The P1 is closed

Two folders differing only in the approval flag now behave differently, which is the whole point:

```
trusted    exit=1  Project configuration fault: invalid WORKFLOW.md hook path:
                   after_create=<proj>/hooks/after_create.sh (missing).
untrusted  starts, connects, "tick — 0 active, idle", no fault, no degrade
```

The damaging half — the per-tick `service.ts:1395` throw suppressing dispatch on unapproved projects — is gone with it; both orchestrator gates now call `isWorkflowHookExecutionAllowed`, the same function hook execution uses, so they agree by construction. The trust decision also resolves identically whether the flag comes from `.env` or an exported shell variable, which I checked explicitly because this repo's own `WORKFLOW.md:31` documents the `export` form. `start.test.ts` now fixes the flag off and asserts startup proceeds — the coverage gap that let this survive five cycles is closed.

The P3 whitespace divergence is fixed too (details on the `hooks.ts` thread).

### Requirements vs. issue #929

| --- | --- |
| Absent / non-executable hook rejected at start with the resolved path named; `doctor` reports the same | ✅ smoke-verified for `after_create` and absolute paths, now correctly scoped to projects where hooks execute |
| `workflow_hooks` runs in the **default** `doctor` pass | ✅ |
| Missing hook does not decrement any issue's failure retry budget | ✅ within the documented bound — a workspace-relative hook's first execution can still spend exactly one retry, stated identically in `README.md`, `docs/configuration.md` and `docs/architecture.md`, and never reaching suppression |
| Tests cover absent, non-executable, retry non-consumption | ✅ plus the untrusted path and a whitespace-padded absolute path |
| Documented provisioning path exists and is exercised | ✅ TC-24 + runner assertion |

No over-engineering. Every judgement call in this PR narrowed toward behavior that already existed rather than inventing policy — `not_a_path` rejects a class the docs already declared invalid, bare-name-as-path matches what `resolveHookCommandSpawn` has always spawned, and the trust gate matches what `executeWorkspaceHook` already did. All three are disclosed in `.changeset/cool-hooks-wait.md`, which is accurate as written.

All ten prior review threads are resolved, and I spot-checked the substantive ones against the code rather than trusting the replies.

### Before merge — mechanical, not a review objection

`mergeable_state` is `dirty`: **`packages/cli/README.md` conflicts with `main`**. It is the only conflicting file and it is documentation. Please merge `main` in and resolve it, keeping both sides' text.

### Follow-ups (do not block)

- **nit** — `start.ts:70` introduces a third hand-written copy of the approval predicate. It is equivalent today, but the resolver and predicate defects earlier in this PR were both caused by two sides restating one rule; fold it into a shared core helper when this file is next touched. (`start.ts` thread)
- **nit** — `doctor` reports a broken hook path on an unapproved project as a hard FAIL without noting the hook is currently skipped. Correct to diagnose, just not self-describing. (`doctor.ts` thread)
- Pre-existing, out of scope: `gh-symphony workflow validate` still only echoes hook values and never validates them — worth its own issue.

### Verdict

Six rounds in, this landed in the right place: one shared resolver, one shared predicate, one shared trust decision, a `deferred` bucket that means exactly one thing, and an honest documented bound where a real one exists. Resolve the README conflict and ship it.

---

moncher-dev (commented):
## PR #936 is `CONFLICTING` against the new `main` — and the conflict is outside the trivial set

`#941` merged as `43f7bddb` at 07:20Z. GitHub recomputed `#936` immediately after and it is now `mergeable: CONFLICTING` (recorded 07:22:53Z, head `579d5ff0`).

Overlapping paths between `43f7bddb` and `#936` — `comm -12` of the two file lists:

```
AGENT_TEST.md
README.md
docs/architecture.md
e2e/run-standalone-project-e2e.sh
packages/cli/README.md
packages/cli/src/commands/start.ts
```

`43f7bddb` rewrote `packages/cli/src/commands/start.ts` (+27/-15 across that file and `packages/orchestrator/src/lock.ts`), which `#936` also modifies.

**This is not a trivial conflict.** The trivial set in `WORKFLOW.md:279` is `.changeset/*`, `docs/**`, `CHANGELOG.md`, and the lockfile. Five of the six overlapping paths — `packages/cli/src/commands/start.ts`, `README.md`, `AGENT_TEST.md`, `e2e/run-standalone-project-e2e.sh`, `packages/cli/README.md` — are outside it, and one of them is a source file. So the in-cycle `Land` → `Land` recovery does not apply; under the Land failure taxonomy this classifies as a **rework failure** and would bounce `Land` → `Ready` on the first Land attempt.

Returning `#929` to `Ready` now rather than spending a Land cycle to discover this. **The rework is the conflict resolution only** — hojinzs's `CHANGES_REQUESTED` points from 07:05:59Z were already addressed in `579d5ff0`; do not redo that work:

1. `/pull` to merge `origin/main` into the branch (never rebase — `WORKFLOW.md:454`).
2. Resolve `packages/cli/src/commands/start.ts` against `43f7bddb`'s new `project-start-lock` module, keeping both changes. `43f7bddb` added `packages/cli/src/project-start-lock.ts` and reworked `packages/orchestrator/src/lock.ts`; check whether the hook-fault retry path in this PR still holds against that lock's new shape rather than only resolving the textual conflict.
3. Keep both changesets.
4. Re-run the Completion Bar and hand back to `In review`.

No new review findings from me at `579d5ff0` beyond this — the conflict is the blocker.

e2e/run-standalone-project-e2e.sh
```diff
@@ -71,6 +71,18 @@ EOF
 
 write_project "$PROJECT_ROOT/project-alpha" alpha
 write_project "$PROJECT_ROOT/project-beta" beta
+write_project "$PROJECT_ROOT/project-broken" broken
```
hojinzs, line 73: **P1 — this line aborts the whole standalone E2E run before any assertion, TC-24's included.**

`write_project` ends at line 69 with

```sh
printf "%s\n" "---" "name: $label" "---" "$label skill" > "$project_dir/.agent/skills/$label/SKILL.md"
```

and it never creates that directory. It works for `alpha` and `beta` only because line 30 pre-creates exactly those two:

```sh
mkdir -p "$PROJECT_ROOT/project-alpha/.agent/skills/alpha" "$PROJECT_ROOT/project-beta/.agent/skills/beta"
```

`project-broken/.agent/skills/broken` is never created, so the redirection fails. The container script runs under `set -euo pipefail` (line 22), so the run dies here — before the missing-hook assertion below it, and before the pre-existing alpha/beta dispatch assertions further down.

Reproduced with the real `write_project` body, no Docker needed:

```
write_project alpha OK
write_project beta OK
repro.sh: line 14: /tmp/tmp.oj1Lc94Vgl/project-broken/.agent/skills/broken/SKILL.md: No such file or directory
exit=1
```

So `pnpm e2e:standalone-project` cannot pass on this branch, and TC-24 has never executed. That matters beyond the new scenario: the acceptance criterion "the documented provisioning path exists **and is exercised**" currently rests on a scenario that aborts at setup. It survived because the Docker daemon was unavailable in every validation environment (#938) — the exception is doing real damage here, not just deferring confirmation.

**Fix (one line)** — move the directory creation into the function so it covers every project:

```sh
 write_project() {
   project_dir=$1
   label=$2
-  mkdir -p "$project_dir/hooks"
+  mkdir -p "$project_dir/hooks" "$project_dir/.agent/skills/$label"
```

Line 30's `mkdir -p` then becomes redundant and can go.

The runtime behaviour this scenario is meant to prove is genuinely correct — I verified it directly against the built CLI on real standalone folders, so this is a harness bug, not a product bug:

```
A: after_create missing, flag=1    exit=1  Project configuration fault: invalid WORKFLOW.md hook path:
                                           after_create=<proj>/hooks/after_create.sh (missing).
B: after_create missing, flag=0    starts, connects, "idle", no fault   (exit=124 on my timeout)
C: after_create chmod 644, flag=1  exit=1  ... after_create=<proj>/hooks/after_create.sh (not executable).
```

Case A also exits non-zero, which is what the `if (...); then echo "unexpectedly passed"` guard below depends on — that part is right.

---

hojinzs (changes requested):
## Review — REQUEST CHANGES (one P1, one line to fix; then it ships)

Reviewed at `bcf64d4c`. I said last round was final and I meant it — I am opening no new lines of feedback on the design or the hook code. But the merge round surfaced a defect I should have caught at `579d5ff0`, and it is not one I can wave through: **the standalone E2E runner aborts at setup, so `pnpm e2e:standalone-project` cannot pass on this branch and TC-24 has never run.** One line fixes it. Details and reproduction on the `e2e/run-standalone-project-e2e.sh:74` thread.

### The conflict resolution itself is correct — verified, not assumed

`bcf64d4c` is a true merge of `579d5ff0` and `43f7bddb` (#941), and both sides survive:

- **`packages/cli/src/commands/start.ts`** — the only conflicting source file. Merged cleanly: `preflightWorkflowStart` (with the hook gate and the trust check) runs at `:1096`, and `acquireProjectStartLocks` at `:1135`. Ordering preserved — a faulted project is rejected **before** any lock is taken, which is what you want; a project that cannot start should not first claim the runtime root.
- **`packages/orchestrator/src/service.ts`** carries #935's changes plus this PR's gates; `project-start-lock.{ts,test.ts}` and `lock.ts` come across intact.
- **Docs and harness** — `README.md`, `packages/cli/README.md`, `docs/architecture.md`, `AGENT_TEST.md`, `e2e/run-standalone-project-e2e.sh` all keep both sides' text. Both changesets are present (`cool-hooks-wait.md` and #941's `tidy-project-start-locks.md`). Both TC rows survive, and the TC-23/TC-24 collision the steward flagged resolved itself correctly: #941 landed TC-23 first, this PR is TC-24.

`git diff 43f7bddb bcf64d4c` is exactly this PR's own change and nothing else — no main-side work was dropped in the resolution.

### Verification performed

- `pnpm install --frozen-lockfile`, `pnpm build` — clean.
- `hooks.test.ts`, `start.test.ts`, `doctor.test.ts`, `service.test.ts`, `project-start-lock.test.ts`, `lock.test.ts` — **372 passed, 0 failures**.
- Smoke test against the built CLI (`packages/cli/dist/index.js`) on real standalone project folders with the file tracker: missing hook / non-executable hook / trust flag on vs. off. All three behave as documented, output on the inline thread. The post-merge code is right.
- CI on `bcf64d4c`: Test ✅, Container Smoke ✅.
- Docker E2E not runnable here either (#938) — which is exactly how the P1 stayed invisible.

### Requirements vs. #929

Unchanged from my approve at `579d5ff0`, with one downgrade:

| --- | --- |
| Absent / non-executable hook rejected at start with the resolved path named; `doctor` reports the same | ✅ smoke-verified post-merge |
| `workflow_hooks` runs in the default `doctor` pass | ✅ |
| Missing hook does not decrement any issue's retry budget | ✅ within the documented bound |
| Tests cover absent, non-executable, retry non-consumption | ✅ unit level, 372 green |
| Documented provisioning path exists **and is exercised** | ⚠️ the exercise aborts at setup — this is the P1 |

No over-engineering in the merge; the resolution is minimal and mechanical.

### Verdict

Everything I asked for across six rounds is in and holding. Fix the one-line `mkdir` in the E2E runner and this is an approve on sight — I will not raise anything further, and there is no need to re-verify the hook work, which I have now checked against the built output twice.

---

packages/cli/src/commands/start.ts
moncher-dev: Acknowledged as non-blocking follow-up material. The shared trust-helper cleanup remains outside this narrowly scoped harness rework; no code change was made here.

packages/core/src/workspace/hooks.ts
moncher-dev: Thank you for verifying the whitespace normalization against the built core. No further change is needed in this cycle; the existing implementation and regression coverage remain intact.

packages/cli/src/commands/doctor.ts
moncher-dev: Acknowledged as explicitly non-blocking follow-up material. Doctor's current diagnostic behavior and the existing changeset wording are unchanged in this narrowly scoped rework.

e2e/run-standalone-project-e2e.sh
moncher-dev: Fixed in `1dc9b34`. `write_project` now creates both `$project_dir/hooks` and `$project_dir/.agent/skills/$label`, so the broken-project fixture reaches the missing-hook assertion. The redundant alpha/beta-only precreation was removed. `bash -n`, the Compose helper test, and the full lint/test/typecheck/build gate pass; two Docker runs remain blocked before execution by the unavailable host daemon tracked in #938.

hojinzs (approved):
## Review — APPROVE

Reviewed at `b90dcd40`. The one-line P1 from my last pass is fixed, and I verified it by running the thing that was broken rather than by reading the diff. As promised, no new lines of feedback.

### The P1 is closed

`write_project` now creates `$project_dir/.agent/skills/$label` alongside `hooks`, and the alpha/beta-only precreation at line 30 is gone. I re-ran my earlier reproduction — the real `write_project` body, no Docker needed — which previously died at `write_project broken`:

```
write_project alpha OK
write_project beta OK
write_project broken OK
--- broken project tree ---
/project-broken/.env
/project-broken/hooks                     <- present, after_create.sh removed
/project-broken/WORKFLOW.md
/project-broken/.agent/skills/broken/SKILL.md
setup reached the missing-hook assertion point: exit=0
```

Setup now runs to completion for all three fixtures, so TC-24's missing-hook assertion is reachable for the first time. `bash -n` clean. The fix generalises rather than special-casing `broken` — any future fixture gets its skill directory from the same line.

### The merge is clean and drops nothing

`b90dcd40` is a true merge of `1dc9b341` and `2eeda4d9`. The merge commit touches exactly one file, `docs/architecture.md`, and keeps both sides: `main`'s spec-mandate column from #943 is present, and this PR's hook rows at `:137` and `:145` survive intact. `git diff main..b90dcd40` is 16 files, +926/−107 — exactly this PR's own change and nothing else. Both changesets present (`cool-hooks-wait.md`, `tidy-project-start-locks.md`).

### Verification performed

- `pnpm install --frozen-lockfile`, `pnpm build`, `pnpm lint`, `pnpm typecheck` — all clean.
- `hooks.test.ts`, `start.test.ts`, `doctor.test.ts` — **128 passed**.
- `service.test.ts`, `project-start-lock.test.ts`, `lock.test.ts` — **244 passed**. 372 total, 0 failures.
- **Smoke test** against the built CLI (`packages/cli/dist/index.js`) on real standalone folders with the file tracker, post-merge. All four classes behave as documented:

```
trusted + missing          exit=1  Project configuration fault: invalid WORKFLOW.md hook path:
                                   after_create=<proj>/hooks/after_create.sh (missing).
trusted + non-executable   exit=1  ... after_create=<proj>/hooks/after_create.sh (not executable).
trusted + shell syntax     exit=1  ... after_create="hooks/after_create.sh; echo pwned" (not a path).
UNTRUSTED + missing        starts, connects, reaches the poll loop (exit=124 under timeout)
                                   0 configuration faults, 0 hook mentions in the log
```

The untrusted run is the one I checked hardest, since it was the P1 two rounds ago: grepping its log gives zero occurrences of both "configuration fault" and "hook". The trust split holds after the merge.
- CI on `b90dcd40`: Test ✅, Container Smoke ✅. `mergeable_state` is now `clean` — the `packages/cli/README.md` conflict is resolved.
- Docker E2E still not run here — daemon unavailable, matching #938. TC-24 is reviewed as written and now *provably reachable*, but has still never executed; that remains the post-merge checkbox already on this PR.

### Requirements vs. #929

| --- | --- |
| Absent / non-executable hook rejected at start with the resolved path named; `doctor` reports the same | ✅ smoke-verified post-merge |
| `workflow_hooks` runs in the default `doctor` pass | ✅ |
| Missing hook does not decrement any issue's retry budget | ✅ within the documented bound — a workspace-relative hook's first execution can still spend exactly one retry, stated in the same terms in `README.md`, `docs/configuration.md` and `docs/architecture.md`, never reaching suppression |
| Tests cover absent, non-executable, retry non-consumption | ✅ plus the untrusted path and whitespace-padded absolute paths |
| Documented provisioning path exists **and is exercised** | ✅ setup no longer aborts; the assertion is reachable |

No over-engineering. Every judgement call in this PR narrowed toward behaviour that already existed rather than inventing policy — `not_a_path` rejects a class the docs already declared invalid, bare-name-as-path matches what `resolveHookCommandSpawn` has always spawned, and the trust gate matches what `executeWorkspaceHook` already did. All three are disclosed in the changeset.

### Follow-ups (do not block, already agreed as out of scope)

- Fold the third copy of the approval predicate at `start.ts:70` into a shared core helper when that file is next touched.
- `doctor` could note that a diagnosed hook is currently skipped on an unapproved project.
- `gh-symphony workflow validate` still only echoes hook values — pre-existing, worth its own issue.

### Verdict

Ship it. Seven rounds in, this landed where it should: one shared resolver, one shared predicate, one shared trust decision, a `deferred` bucket that means exactly one thing, an honest documented bound where a real one exists, and an E2E scenario that can now actually run. The remaining Docker E2E execution is tracked by #938 and is the right thing to leave as a post-merge check rather than another round.

---

PR #941, opened 2026-09-06
fix(cli): prevent duplicate project starts across runtime roots

### LINKED ISSUE #921: fix(cli): prevent duplicate project starts across runtime roots
## Problem

Removing the host-global orchestrator instance registry in #906 also removes the only guard that detected the same project being started under different runtime roots.

Project locks are scoped to `runtimeRoot + projectId`, so two configurations with different runtime roots can start orchestrators for the same project folder and tracker binding concurrently. The project overlap check does not cover this case.

## Proposed direction

Add a registry-independent, lock-based guard that identifies the project folder across runtime roots and prevents a duplicate start unless an explicit, deliberately designed override exists.

## Acceptance criteria

- Starting the same project folder under two different runtime roots cannot attach two orchestrators concurrently.
- Existing per-runtime-root project locks and owner identity validation remain intact.
- Stale owners are handled safely.
- Focused unit tests and an integration scenario cover the cross-runtime-root case.

Discovered during review of #906. This follow-up intentionally stays outside that registry-removal PR.

moncher-dev:
## Issues

- Closes #921

## Summary

- Prevent duplicate orchestrators from owning the same canonical project folder through different runtime roots.
- Anchor the registry-independent identity lease in the project folder, avoiding shared temporary-directory permissions and spoofable host-global paths.
- Preserve the existing runtime-root lock plus heartbeat and process-identity validation for stale-owner recovery.

## Change-point diagram

- `project start` → canonicalize folder → folder-local identity lease → existing runtime-root lease → orchestrator
- startup failure / shutdown → release runtime-root lease → release identity lease

## Start here

- `packages/cli/src/project-start-lock.ts:1` — composes the folder-local identity lease and runtime lock
- `packages/orchestrator/src/lock.ts:44` — supports a caller-selected lock path and diagnostic label while preserving lease validation
- `packages/cli/src/project-start-lock.test.ts:10` — focused cross-root, canonical-path, stale-owner, and diagnostic coverage

- A live canonical project folder cannot be started through another `--config` root, caller temporary directory, user, or process mount namespace that sees the same accessible folder.
- Duplicate-start errors name the canonical project folder rather than an opaque digest.
- Different folders remain independent, and legacy cached projects retain their existing runtime-root lock.

## Validation

- `pnpm exec vitest run packages/cli/src/project-start-lock.test.ts packages/orchestrator/src/lock.test.ts packages/cli/src/commands/start.test.ts --reporter=dot` — pass (65 tests)
- `pnpm lint` — pass
- `pnpm test` — pass (all 14 tested workspace packages; CLI 576 tests; orchestrator 341 tests)
- `pnpm typecheck` — pass
- `pnpm build` — pass
- `git diff --check` — pass
- `./e2e/run-standalone-project-e2e.sh` — retried twice; documented environment exception #942 because Docker preflight exits 69 with the daemon unavailable
- Docker scenario: `e2e/scenarios/23-cross-runtime-root-project-lock.md`

## Changeset

- `.changeset/tidy-project-start-locks.md` (patch)

## Risks & rollback

- The project folder must permit the launching account to create its transient `.gh-symphony-start.lock`; permission errors fail startup closed.
- Reviewer focus: folder permissions, custom lock-path lease semantics, and stale-owner takeover.
- Roll back by reverting the identity-lock composition; the per-runtime-root lock format remains unchanged.

## Changed files

- `packages/cli/src/project-start-lock.ts` — folder-local canonical identity-lock composition and legacy fallback
- `packages/cli/src/commands/start.ts` — acquire and release both leases
- `packages/orchestrator/src/lock.ts` — custom lock path and caller-facing diagnostic label
- `packages/cli/src/project-start-lock.test.ts` and `packages/orchestrator/src/lock.test.ts` — live, stale, canonical-path, diagnostic, per-root, and legacy tests
- `e2e/run-standalone-project-e2e.sh` and `e2e/scenarios/23-cross-runtime-root-project-lock.md` — bounded cross-root/TMPDIR black-box regression
- `README.md`, `packages/cli/README.md`, `docs/architecture.md`, and `AGENT_TEST.md` — living documentation
- `.changeset/tidy-project-start-locks.md` — CLI patch release note

- [ ] Restore Docker daemon access tracked by #942 and rerun TC-23.

## Security


moncher-dev (commented):
Steward review: **blocking**. `pnpm install --frozen-lockfile && pnpm build && pnpm test` are green in a scratch worktree at `60ff2fc1` (cli 574/574). The direction is right — same-`TMPDIR`, different `--config` roots is now correctly rejected, which is the #921 scenario. But the guarantee the docs claim is not the guarantee the code provides, and I confirmed the gap by running it rather than reading it.

**Confirmed: the folder-identity lock is defeated by `TMPDIR` alone.** `resolveProjectIdentityLock` roots the lock at `os.tmpdir()`, and Node reads `$TMPDIR`/`$TMP`/`$TEMP` on every call. Probe against the shipped function in this branch:

```
process.env.TMPDIR = tmpA;  await acquireProjectStartLocks({runtimeRoot: A, projectId: "project-a", projectDir});
process.env.TMPDIR = tmpB;  await acquireProjectStartLocks({runtimeRoot: B, projectId: "project-b", projectDir});
// second call resolves .../gh-symphony-project-locks-501 under tmpB and SUCCEEDS
// PROBE second start acquired a lock: true
```

Two live owners for one canonical project folder. This is not hypothetical for this repo's own deployment shape: a daemon started from an interactive shell on macOS gets `TMPDIR=/var/folders/<...>/T`, while the same `project start` launched from cron/launchd/systemd typically gets `/tmp` or an unset `TMPDIR`. Those are different lock namespaces, so the shell-started daemon and the service-started daemon would both dispatch against the same folder — the failure #921 was filed to stop. The `userNamespace` suffix adds a second axis of the same problem: `sudo gh-symphony project start` is uid 0 and cannot see the uid-501 lock.

This is the same point Codex raised as P1; I'm blocking on it because it is reproducible and because the docs assert the strong form.

**The stated rationale is false as written.** Three documents now claim a host-wide property the implementation does not have:

- `README.md`: "Consequently, pointing separate `--config` runtime roots at the same project folder **cannot** start duplicate orchestrators."
- `packages/cli/README.md`: "**Only one orchestrator may own a canonical project folder on a host**, even when separate `--config` runtime roots are used."
- `packages/cli/README.md`: "both use lease heartbeats and process-owner identity validation so stale owners can be reclaimed safely" — true, but it is the *namespace selection*, not the lease, that fails here.

`docs/architecture.md` is the one honest description ("A second lock under the OS temporary directory…"), and it quietly contradicts the other two. Either root the identity lock somewhere stable and host-wide that does not depend on caller environment, or downgrade all three claims to what is actually delivered and name the residual gap.

**TC-23 is structurally blind to this.** `e2e/run-standalone-project-e2e.sh` spawns the duplicate start as a child of the same shell, so it inherits the same `TMPDIR` and the same uid. The scenario can only ever exercise the case that already works. If the host-wide claim stays, the scenario needs a differing-`TMPDIR` start; if the claim is narrowed, the scenario should say which case it covers.

**Secondary — `/tmp` is world-writable and the path is predictable.** The identity root is `<tmp>/gh-symphony-project-locks-<uid>` with a name derivable from the target uid. `ensureSecureDirectory` does `mkdir(recursive, 0o700)` then `chmod(0o700)` on the leaf. On a shared host, another local user can pre-create that directory; the subsequent `chmod` then fails `EPERM` and `project start` dies. The previous design kept locks under the runtime root (`~/.gh-symphony`), which is not world-writable, so this is a new denial-of-service surface introduced by moving the lock to `/tmp`. Worth a deliberate decision even if you judge the risk acceptable.

**Minor — the `projectDir`-absent branch silently opts out.** `OrchestratorProjectConfig.projectDir` is optional (`packages/core/src/contracts/status-surface.ts`), and `acquireProjectStartLocks` falls back to runtime-root-only locking with no warning when it is missing. A persisted `project.json` predating the folder-addressed migration would therefore get none of this protection while the docs promise it unconditionally. Either assert `projectDir` is present for `project start` or log the downgrade.

**Cross-PR — TC-23 number collision.** PR #936 also adds a scenario numbered `TC-23` (`e2e/scenarios/23-hook-configuration-fault.md`) and its own row to the `AGENT_TEST.md` scenario index. The filenames differ, so both will merge cleanly and leave two different TC-23s referenced from the same table. `e2e/scenarios/` already carries duplicate 10/13/15/17/19 pairs, so this is a pre-existing pattern rather than a defect you introduced — but whichever of the two lands second should renumber.

Moving #921 back to Ready with the above.

packages/cli/src/project-start-lock.ts
```diff
@@ -0,0 +1,86 @@
...
+}> {
+  const canonicalProjectDir = await realpath(input.projectDir);
+  const digest = createHash("sha256").update(canonicalProjectDir).digest("hex");
+  const userNamespace =
+    typeof process.getuid === "function" ? String(process.getuid()) : "user";
+  return {
+    runtimeRoot: join(
+      input.temporaryDirectory ?? tmpdir(),
+      `${PROJECT_IDENTITY_LOCK_NAMESPACE}-${userNamespace}`
+    ),
```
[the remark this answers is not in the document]
moncher-dev: Resolved in `8ceb9c3`. The identity lock now uses a fixed host-wide namespace (`/var/tmp/gh-symphony-project-locks` on POSIX and the system-drive `ProgramData` directory on Windows), with no dependency on `TMPDIR`/`TMP`/`TEMP` or the effective UID. A focused test changes `TMPDIR` between resolutions and asserts the exact stable namespace. Existing per-runtime locks and lease/process-identity validation remain unchanged; inability to access the shared namespace fails startup closed.

moncher-dev (commented):
## Steward review — BLOCK (returning to `Ready`)

Full `pnpm install --frozen-lockfile && pnpm build && pnpm test && pnpm lint && pnpm typecheck` all green in a scratch worktree at `8ceb9c38` (cli 575/575). Changeset present. Lock ordering in `acquireProjectStartLocks` (`project-start-lock.ts:64-73`) is correct — identity lock first, released if the per-runtime lock then fails.

### The reported regression is genuinely fixed — verified by re-running last cycle's probe

The anchor moved off `os.tmpdir()` to a hardcoded `/var/tmp/gh-symphony-project-locks` (`packages/cli/src/project-start-lock.ts:10-18`, shipped at `dist/chunk-ZXO2HVI7.js:926`). Re-running the exact two-TMPDIR probe against the shipped dist function:

```
P1: TMPDIR=/tmp/probe-a → ACQUIRED /var/tmp/gh-symphony-project-locks/projects/ac715e2b…/.lock (pid 6511, alive)
P2: TMPDIR=/tmp/probe-b → REFUSED: Project "ac715e2b79e4…" is already running (PID 6511).  exit=3
```

The cross-TMPDIR duplicate is now correctly refused. Last cycle's README-vs-architecture contradiction is also gone — the three documents now agree with each other.

**But moving the anchor to a world-writable shared path introduced two new failure modes, and the claim is still stronger than the guarantee.**

### B1 (blocking) — the shared root is created `0700`, so a second UID is locked out of *every* project, permanently

`ensureSecureDirectory` (`packages/orchestrator/src/lock.ts:281-284`) does `mkdir(recursive, 0o700)` + `chmod(0o700)`. That mode was written for the per-user runtime root `~/.gh-symphony`. Applied to a host-shared path it is self-contradictory: the namespace cannot be both host-wide and owner-private.

Already true on this host, created by this branch's own runs:
```
$ ls -ld /var/tmp/gh-symphony-project-locks
drwx------@ 3 steve  wheel  /var/tmp/gh-symphony-project-locks
```
What any other UID gets (reproduced by making the dir inaccessible to the caller — i.e. exactly a `0700` dir owned by someone else):
```
$ node hold.mjs <freshProjectDir> <root>
REFUSED: EACCES: permission denied, mkdir '/var/tmp/gh-symphony-project-locks/projects/cfce0b9d…'
```
Concretely: one `sudo gh-symphony project start` on a fresh host creates a **root-owned `0700`** directory, after which the normal user can never start *any* project — and `/var/tmp`'s sticky bit means they cannot remove it without root. Note `sudo` is one of the two scenarios issue #921 names, so this is squarely in scope.

The EACCES/EPERM branch has **no named destination**: `packages/cli/src/commands/start.ts:1107-1113` wraps the acquire in `try/finally` with no `catch`, so a raw `EACCES` reaches `formatErrorForTerminal` with no remediation text.

### B2 (blocking) — local unprivileged denial of service

`/var/tmp` is world-writable and sticky, so any local user can pre-create the namespace directory or plant a lock record:
```
$ mkdir -p /var/tmp/gh-symphony-project-locks/projects/<digest>
$ echo '{"ownerToken":"planted","pid":1,"startedAt":"…","processIdentity":null,"cwd":"/"}' > .../.lock
$ node hold.mjs <projectDir> <root>
REFUSED: Project "fcb5cc1e…" is already running (PID 1).
```
`isActiveLock` (`lock.ts:249-267`) accepts `processIdentity: null` and any live PID, so pid 1 is permanently "active". The victim cannot delete the planted file — sticky bit, attacker owns it.

**Recommended fix for B1 and B2 together:** anchor the folder-identity lock in the folder whose identity it represents — `<canonicalProjectDir>/.gh-symphony/start.lock`. That is namespace-proof, mount-proof, permission-natural, self-cleaning, and diagnosable. **Nothing in the code, the changeset, or the docs states why `/var/tmp` was chosen over the project folder** — if there is a reason, please write it down; if there isn't, the project folder is the better anchor.

### B3 (blocking) — the docs still overclaim, in a narrower form than last cycle

- `README.md:965-966` — "a **stable host-wide namespace independent of caller temporary-directory and user settings**"
- `packages/cli/README.md:432` — "**Only one orchestrator may own a canonical project folder on a host**"
- `docs/architecture.md:82` — "…preventing the same folder from being started through different runtime roots, caller temporary-directory settings, **or user namespaces**"

False two ways. (a) **Mount namespaces:** `systemd.exec`'s `PrivateTmp=` gives a unit a private `/tmp/` *and `/var/tmp/`* — so the shell-daemon-vs-service-daemon case named in #921 still yields two live locks under any hardened unit; same for two containers bind-mounting the folder. (b) "independent of user settings" is not mutual exclusion across users — across users it is the hard `EACCES` of B1, which is a lockout, not a lock.

Either qualify all three sentences (single UID, single mount namespace) or move the anchor per B1, which would make the strong claim true.

### B4 (blocking) — TC-23 hangs instead of failing on the regression it exists to guard

`e2e/run-standalone-project-e2e.sh:110-117` runs the duplicate start as a foreground long-lived poller — no `--once`, no `timeout`:
```sh
(cd "$PROJECT_ROOT/project-alpha" && … node …/index.js --config /tmp/alternate-runtime project start > /tmp/duplicate-project.log 2>&1)
duplicate_status=$?
```
If the identity lock ever stops refusing, this blocks forever and the suite **hangs rather than fails** — the failure mode is indistinguishable from a slow run. Add `--once` or `timeout 60`.

### Non-blocking, but please read

- **The unit test still cannot fail on the reported scenario.** `project-start-lock.test.ts:56-84` mutates `process.env.TMPDIR` in-process, but `HOST_PROJECT_IDENTITY_LOCK_ROOT` is a module-level const, so `second === first` is tautological. The literal-path assertion at :68-76 *is* a real guard against reverting to `tmpdir()`, so this is not blocking — but adding `TMPDIR=/tmp/alt` to the B4 e2e command is a one-token change that covers the reported scenario end-to-end across processes.
- **Unit tests now write real host state.** With the `temporaryDirectory` seam deleted, `project-start-lock.test.ts:12` and `:86` take real locks under `/var/tmp/gh-symphony-project-locks`; one `pnpm test` grew `projects/` from 11 to 14 dirs. `releaseProjectLock` only unlinks `.lock`, never the `projects/<digest>` dir — unbounded litter, and the suite fails outright for any user hitting B1.
- **The duplicate-start message names a digest, not the folder** (`lock.ts:133-135`): `Project "ac715e2b79e4…" is already running (PID 6511)`. The PR's stated purpose is a *clear* duplicate-start failure; a sha256 prefix is not actionable. Print the canonical project path.
- **The daemon path — the documented deployment — is untested and opaque.** `--daemon` surfaces only `Daemon exited before acquiring the project lock … See <logPath>` (`start.ts:1555-1559`); the real reason stays in the log. TC-23 covers foreground only.
- **Dead branch kept alive by a test.** `start.ts:1386-1391`'s `input.projectLock` / `input.releaseLock` path is unreachable in production (its sole caller at :1188 always passes `projectLocks`), yet `start.test.ts:295-304` asserts it. Meanwhile the new `projectLocks` branch has **no** unit coverage — dropping the identity-lock release on shutdown would fail nothing.
- **Silent degradation when `projectDir` is absent.** `OrchestratorProjectConfig.projectDir` is optional (`packages/core/src/contracts/status-surface.ts:42`); `project-start-lock.ts:46-53` then skips the identity lock entirely **with no warning**, so "only one orchestrator per folder on a host" quietly does not apply to any legacy `project.json` lacking the field. That branch needs a named destination too.

moncher-dev (commented):
## Steward review — re-posted as **anchored inline threads**

My earlier top-level `COMMENTED` review on this PR had no inline threads, and I have just confirmed on #909/#943 that the Step 0 rework guard does not see that channel — nor the issue-comment workaround, when the steward posts under `moncher-dev` (the worker's own account). That is issue #900. Same findings below, anchored so `reviewThreads` is non-empty.

**The reported regression is genuinely fixed** — I re-ran last cycle's two-TMPDIR probe against the shipped dist and the cross-TMPDIR duplicate is now correctly refused (`TMPDIR=/tmp/probe-a` acquires, `TMPDIR=/tmp/probe-b` gets `REFUSED ... already running (PID 6511)`, exit 3). Build/test/lint/typecheck all green, changeset present, lock ordering correct.

But moving the anchor to a world-writable shared path introduced two new failure modes, and the docs still overclaim. Full write-up is in the issue-comment copy on #921.

packages/cli/src/project-start-lock.ts
```diff
@@ -0,0 +1,87 @@
+import { createHash } from "node:crypto";
+import { realpath } from "node:fs/promises";
+import { join, parse } from "node:path";
+import {
+  acquireProjectLock,
+  releaseProjectLock,
+  type ProjectLockHandle,
+} from "@gh-symphony/orchestrator";
+
+const PROJECT_IDENTITY_LOCK_NAMESPACE = "gh-symphony-project-locks";
```
moncher-dev: **B1 (blocking) — this shared root is created `0700`, which locks a second UID out of *every* project, permanently.**

`ensureSecureDirectory` (`packages/orchestrator/src/lock.ts:281-284`) does `mkdir(recursive, 0o700)` + `chmod(0o700)`. That mode was designed for the per-user runtime root `~/.gh-symphony`. Applied to a host-shared path it is self-contradictory: the namespace cannot be both host-wide and owner-private.

Already true on this host, created by this branch's own runs:
```
$ ls -ld /var/tmp/gh-symphony-project-locks
drwx------@ 3 steve  wheel  /var/tmp/gh-symphony-project-locks
```
What another UID gets:
```
REFUSED: EACCES: permission denied, mkdir '/var/tmp/gh-symphony-project-locks/projects/cfce0b9d…'
```
One `sudo gh-symphony project start` on a fresh host creates a **root-owned `0700`** dir, after which the normal user can never start *any* project — and `/var/tmp`'s sticky bit means they cannot remove it without root. `sudo` is one of the two scenarios #921 names, so this is in scope.

**B2 (blocking) — local unprivileged DoS.** `/var/tmp` is world-writable + sticky, so any local user can plant a lock record:
```
$ mkdir -p /var/tmp/gh-symphony-project-locks/projects/<digest>
$ echo '{"ownerToken":"planted","pid":1,"processIdentity":null,…}' > .../.lock
REFUSED: Project "fcb5cc1e…" is already running (PID 1).
```
`isActiveLock` (`lock.ts:249-267`) accepts `processIdentity: null` and any live PID, so pid 1 is permanently "active", and the victim cannot delete the planted file.

**Recommended fix for both:** anchor the folder-identity lock in the folder whose identity it represents — `<canonicalProjectDir>/.gh-symphony/start.lock`. Namespace-proof, mount-proof, permission-natural, self-cleaning, diagnosable. **Nothing in the code, the changeset, or the docs says why `/var/tmp` was chosen over the project folder** — if there is a reason, please write it down.

packages/cli/src/project-start-lock.ts
```diff
@@ -0,0 +1,87 @@
...
+    canonicalProjectDir,
+  };
+}
+
+export async function acquireProjectStartLocks(input: {
+  runtimeRoot: string;
+  projectId: string;
+  projectDir?: string;
+}): Promise<ProjectStartLocks> {
+  if (!input.projectDir) {
```
moncher-dev, line 39: **Silent degradation — this branch has no named destination.**

`OrchestratorProjectConfig.projectDir` is optional (`packages/core/src/contracts/status-surface.ts:42`). When it is absent the identity lock is skipped entirely **with no warning**, so the advertised "only one orchestrator may own a canonical project folder on a host" quietly does not apply to any legacy `project.json` lacking the field. Emit a warning at minimum.

README.md
```diff
@@ -962,6 +962,13 @@ Read orchestration state via the status API (`/api/v1/state`) rather than readin
 
 Run `gh-symphony doctor --smoke` before the first `start --once` when you want a safe pre-dispatch readiness check. `gh-symphony project start --project-dir <path> --once` is the first production-like run: it validates the real GitHub Project binding, project `WORKFLOW.md`, and dispatch eligibility, then performs one poll/reconcile/dispatch tick instead of starting a long-lived poller. Add `--port [port]` when you want the JSON status API available; `--http [port]` remains a supported alias. With `--once --port`, the one-shot tick still completes, but the HTTP server stays up afterward and the process keeps the project lock until you stop it with `Ctrl+C`. `server.port` in `WORKFLOW.md` enables the same API when no CLI port is supplied. Add `--web` instead when you want the browser dashboard at `/` plus the JSON API.
 
+Project start also holds a lock keyed by the canonical project-folder path in a
```
moncher-dev: **B3 (blocking) — the claim is still stronger than the guarantee.**

"a stable host-wide namespace **independent of caller temporary-directory and user settings**" is false two ways:

1. **Mount namespaces.** `systemd.exec`'s `PrivateTmp=` gives a unit a private `/tmp/` *and `/var/tmp/`* — so the shell-daemon-vs-service-daemon case named in #921 still yields two live locks under any hardened unit. Same for two containers bind-mounting the folder.
2. **"independent of user settings"** is not mutual exclusion across users — across users it is the hard `EACCES` of B1, which is a lockout, not a lock.

The three documents are now mutually consistent (last cycle's contradiction is fixed) — they are consistently overclaiming. Either qualify all three (single UID, single mount namespace) or move the anchor, which would make the strong claim true.

Same sentence at `packages/cli/README.md:432` and `docs/architecture.md:82`.

e2e/run-standalone-project-e2e.sh
```diff
@@ -101,6 +101,23 @@ done
 test -f "$CONFIG_DIR/projects/$alpha_id/project.json"
 test -f "$CONFIG_DIR/projects/$beta_id/project.json"
 
+# A second runtime root must not start another orchestrator for the same
+# canonical project folder while the first owner is live.
+for _ in $(seq 1 20); do
+  test -f "$CONFIG_DIR/projects/$alpha_id/.lock" && break
+  sleep 1
+done
+test -f "$CONFIG_DIR/projects/$alpha_id/.lock"
```
moncher-dev, line 129: **B4 (blocking) — TC-23 hangs instead of failing on the regression it exists to guard.**

The duplicate start runs as a foreground long-lived poller — no `--once`, no `timeout`. If the identity lock ever stops refusing, this blocks forever and the suite **hangs rather than fails**, which is indistinguishable from a slow run.

Add `--once` or `timeout 60`. While you are here, adding `TMPDIR=/tmp/alt` to this command is a one-token change that would cover the reported cross-TMPDIR scenario end-to-end across processes — see the note on `project-start-lock.test.ts`.

packages/cli/src/project-start-lock.test.ts
```diff
@@ -0,0 +1,151 @@
...
+    const projectLink = join(root, "project-link");
+    await symlink(projectDir, projectLink);
+    const viaSymlink = await resolveProjectIdentityLock({
+      projectDir: projectLink,
+    });
+
+    expect(viaSymlink).toEqual(direct);
+  });
+
+  it("uses one host-wide namespace regardless of the caller temporary directory", async () => {
```
moncher-dev: **Non-blocking, but this test cannot fail on the reported scenario.**

It mutates `process.env.TMPDIR` in-process, but `HOST_PROJECT_IDENTITY_LOCK_ROOT` is a module-level const evaluated at import — so `second === first` is tautological and would hold no matter what the function did with TMPDIR.

The literal-path assertion below *is* a real guard against reverting to `tmpdir()`, and my cross-process probe covers the real case, so this is not blocking. The e2e `TMPDIR` change noted on `run-standalone-project-e2e.sh` is the cheap way to cover it properly.

Also: with the `temporaryDirectory` seam deleted, this suite now takes **real locks under `/var/tmp/gh-symphony-project-locks`**. One `pnpm test` grew `projects/` from 11 to 14 dirs, and `releaseProjectLock` only unlinks `.lock`, never the `projects/<digest>` dir — unbounded host litter, and the suite fails outright for any user hitting B1.

moncher-dev (commented):
Passing. All four of last round's blocking findings are closed, and the design change that closed them is the right one. I verified by probe against the built CLI, not by reading the diff.

## The anchor moved, and that dissolves three findings at once

Last round the folder-identity lock lived in a world-writable `/var/tmp/gh-symphony-project-locks` created `0700`, which meant one `sudo` start locked every other UID out of *every* project — and `sudo` is a scenario #921 names. It is now a `.gh-symphony-start.lock` lease inside the canonical project folder.

- **`sudo` lockout — gone.** There is no shared host-wide directory left to be captured, so the failure mode has no surface, rather than being mitigated.
- **Planted-lock DoS — contained.** Planting a lock now requires write access to the project folder, and anyone with that could equally rewrite `WORKFLOW.md` or `hooks/after_create.sh`. It is no longer a privilege boundary crossing.
- **Docs overclaiming — now actually true.** I pushed back last round because `PrivateTmp=` covers `/var/tmp` too, so the "stable host-wide namespace" claim was false under systemd. The in-folder lease genuinely is shared across mount namespaces that bind-mount the same folder, and genuinely is not shared between separate copies. The README, `packages/cli/README.md`, and `docs/architecture.md` now say the same thing and it holds.

`acquireProjectLock` correctly skips `ensureSecureDirectory` when a `lockPath` is supplied, so it does not chmod a user's project folder to `0700` on the way in — which is exactly the mistake being undone.

## Probes

Duplicate exclusion, with **both** escapes applied at once — separate `--config` runtime root and separate `TMPDIR`:

```
start A (daemon, --config cfgA, TMPDIR=tmpA)   exit=0, lock written
start B (--once, --config cfgB, TMPDIR=tmpB)   exit=1
  Project "<canonical folder>" is already running (PID 60437).
```

The diagnostic names the canonical folder rather than the opaque sha256 digest it used to print. Full lifecycle:

```
project stop        → SIGTERM sent, daemon stopped, .gh-symphony-start.lock removed
restart via cfgB    → exit=0, starts cleanly
```

No stale lease is left behind, so the operator restart procedure is unaffected.

**TC-23 no longer hangs on regression.** The runner now starts the duplicate with `--once` and asserts `test "$duplicate_status" -ne 0` plus the message grep. Under a duplicate-exclusion regression the second start completes one tick and exits 0, so the scenario fails on status instead of hanging the runner. That is what I asked for.

Tests: `pnpm install --frozen-lockfile` and `pnpm build` clean; `project-start-lock.test.ts` + `orchestrator/src/lock.test.ts` — **21 passed**.

## Two items for @hojinzs, neither blocking

**1. A read-only project folder can no longer start, and says so badly.** On `main` the lease lived outside the folder, so a read-only project folder still started. It now cannot, and the failure is a bare Node error:

```
exit=1
EACCES: permission denied, open '<project>/.gh-symphony-start.lock'
```

No mention of the lock, of the folder needing to be writable, or of what to do — compare the sibling hook work on #936, which prints `Project configuration fault: ... (not executable)`. The new docs say "Only one orchestrator may own an **accessible** canonical project folder", and "accessible" is carrying this whole change without being defined anywhere. Most deployments keep the folder writable (`workspace.root` defaults inside it), so I do not think this blocks, but it is an undocumented behaviour change with an unactionable message, and this PR is on its fourth round — I would rather name it for you than send it back again.

**2. Nothing ignores the new lock file.** `.gh-symphony-start.lock` is now written into a folder that teams plausibly version-control — it holds `WORKFLOW.md`, `hooks/`, `.mcp.json`, `.agent/skills/`. `workflow init` scaffolds no `.gitignore`, and no doc tells users to ignore it. A committed lock would be reclaimed as stale rather than wedging anything, so this is untidiness rather than breakage.

The lock file is written `0644` and carries `ownerToken`, `pid`, `processIdentity`, and `cwd`. Nothing there is secret, and taking the lease over requires write access to the file, so I do not think the mode needs tightening — flagging it only so the choice is visible rather than incidental.

packages/orchestrator/src/lock.ts
```diff
@@ -131,7 +138,7 @@ export async function acquireProjectLock(input: {
       })
     ) {
       throw new Error(
-        `Project "${input.projectId}" is already running (PID ${existing.record.pid}).`
+        `Project "${projectLabel}" is already running (PID ${existing.record.pid}).`
```
hojinzs, line 141: **P2 — the docs claim exclusion across process/mount namespaces; the liveness check cannot deliver it.**

Moving the lease into the folder does make the *file* visible across bind mounts. But whether that file counts as held is decided by `isActiveLock` (`lock.ts:256-267`), which validates with `isProcessRunning(pid)` and `getProcessIdentity(pid)` — and `getProcessIdentity` shells out to `ps -p <pid>` (`lock.ts:406-412`). Both are **PID-namespace-local**. A starter in a second namespace resolves the recorded pid to a different process (or to nothing), scores the lease stale, `rm`s it at :145, and takes over.

Confirmed by probe against this branch, modelling container B's `ps` resolving the same numeric pid to a different process:

```
NS owner A token: 3765:b2c62e07-811d-4bae-873d-2ef1d0b19ba4
NS B identity-mismatch RESULT: TOOK OVER, token: 3765:ae337a21-...
NS lock now owned by: 3765:ae337a21-... (A was 3765:b2c62e07-...)
```

Two live orchestrators on one canonical folder, for up to `leaseTtlMs / 3` (~20s) until A's heartbeat notices the changed `ownerToken` and fail-closes itself. That is the #921 failure mode, just narrower than before.

This is the same class of overclaim that blocked rounds 1 and 3 — round 4 resolved it for `TMPDIR`/uid but the namespace half of the sentence survived:

- `README.md:966-967` — "pointing separate `--config` runtime roots, temporary directories, users, **or mount namespaces** at the same accessible…"
- `docs/architecture.md:82` — "preventing starts through different runtime roots **or process namespaces**…"

**I am not asking for namespace-safe process identity** — that is a real design problem (boot id + `/proc` start-time, or an `flock`-based lease) and well outside what #921 asked for. Just drop the two namespace clauses and let the sentences claim what the code actually proves: one owner per canonical folder within a single PID namespace, regardless of runtime root, `TMPDIR`, or user. Doc-only, no code change.

---

packages/cli/src/project-start-lock.ts
```diff
@@ -0,0 +1,82 @@
...
+    return {
+      projectLock: await acquireProjectLock({
+        runtimeRoot: input.runtimeRoot,
+        projectId: input.projectId,
+      }),
+    };
```
hojinzs, line 45: **P2 — the `projectDir`-absent branch silently opts out of the whole guarantee.** Raised in rounds 2 and 3 and still unaddressed, so recording it once more with a reproduction and a concrete destination.

`OrchestratorProjectConfig.projectDir` is optional and `undefined` is explicitly tolerated — `assertAbsoluteProjectDir` returns early for it (`packages/core/src/contracts/status-surface.ts:85-88`). So `start.ts:1113` can pass `undefined`, this branch returns a runtime-root-only lock, and nothing warns. Probed against this branch:

```
P6  no projectDir identityLock:                    SKIPPED silently
P6b second start, same folder, no projectDir:      ACQUIRED (no cross-root guard)
```

Meanwhile `packages/cli/README.md:432` states the guarantee unconditionally. Any persisted `project.json` predating the folder-addressed migration gets none of it and no signal.

The repo already has the canonical resolver for exactly this question: `resolveManagedProjectDirectory(projectConfig, runtimeRoot)` (`packages/cli/src/managed-project-environment.ts:31-39`) falls back to `<runtimeRoot>/projects/<projectId>` when `projectDir` is absent. Using it here would make the branch disappear and the guarantee unconditional. One wrinkle if you take that route: the identity lock is acquired **before** the per-root lock, and `acquireProjectLock` skips `ensureSecureDirectory` when `lockPath` is supplied — so that fallback directory may not exist yet and you would get a raw `ENOENT`. A warning line on the downgrade is the cheaper fix and closes the "silent" half either way.

---

packages/orchestrator/src/lock.ts
```diff
@@ -72,7 +77,9 @@ export async function acquireProjectLock(input: {
 
   for (;;) {
     try {
-      await ensureSecureDirectory(dirname(lockPath));
+      if (!input.lockPath) {
+        await ensureSecureDirectory(dirname(lockPath));
+      }
```
hojinzs, line 82: **P3 — skipping `ensureSecureDirectory` is right, but it leaves the caller-supplied-path failures with no named destination.**

Not chmod'ing a user's project folder to `0700` is exactly correct, and it is what undid round 3's `sudo` lockout. The consequence is that every filesystem failure on the caller-supplied path now surfaces raw, because `start.ts:1107-1113` wraps the acquire in `try/finally` with no `catch`:

- read-only project folder → `EACCES: permission denied, open '<project>/.gh-symphony-start.lock'`
- parent missing → bare `ENOENT`

Neither mentions the lock, that the folder must be writable, or what to do. The new docs hang the whole behaviour change on the word "**accessible**" canonical project folder, and "accessible" is defined nowhere. Compare the sibling hook work on #936, which prints `Project configuration fault: … (not executable)`.

I could not reproduce this here — the container runs as uid 0, which walks straight through a `0555` folder (`P5 read-only: ACQUIRED`) — so I am taking the round-4 reproduction at face value rather than claiming my own. Non-blocking, but this is a behaviour change from `main` (where the lease lived outside the folder and a read-only folder still started) with an unactionable message. A `catch` that maps `EACCES`/`EPERM`/`ENOENT` on `identityLock.lockPath` to a sentence naming the folder and the write requirement would close it.

---

packages/cli/src/project-start-lock.ts
```diff
@@ -0,0 +1,82 @@
+import { realpath } from "node:fs/promises";
+import { join } from "node:path";
+import {
+  acquireProjectLock,
+  releaseProjectLock,
+  type ProjectLockHandle,
+} from "@gh-symphony/orchestrator";
+
+const PROJECT_IDENTITY_LOCK_ID = "start";
+const PROJECT_IDENTITY_LOCK_FILE = ".gh-symphony-start.lock";
```
hojinzs, line 10: **nit — nothing ignores this file.** `.gh-symphony-start.lock` now lands in a folder teams plausibly version-control (it holds `WORKFLOW.md`, `hooks/`, `.mcp.json`, `.agent/skills/`). `workflow init` scaffolds no `.gitignore` and no doc mentions it. A committed lock is reclaimed as stale rather than wedging anything, so this is tidiness — but a one-line mention in `packages/cli/README.md` next to the lock description, or a scaffolded ignore entry, would save people a confusing diff.

---

e2e/run-standalone-project-e2e.sh
```diff
@@ -101,6 +101,24 @@ done
...
```
hojinzs, line 129: **nit — the readiness gate waits on the wrong lock.** TC-23 exists to prove the *identity* lease refuses the duplicate, but it gates on `$CONFIG_DIR/projects/$alpha_id/.lock`, the per-runtime-root lock. It happens to be sound because `acquireProjectStartLocks` takes the identity lock first, so the runtime lock existing implies the identity lock exists — but that is an implicit dependency on acquisition order inside the code under test. Waiting on `$PROJECT_ROOT/project-alpha/.gh-symphony-start.lock` would assert the thing the scenario is actually about, and would fail loudly if the ordering ever flipped.

The `--once` + `test "$duplicate_status" -ne 0` + message grep below is the right shape — that closes round 3's hang-instead-of-fail correctly.

---

hojinzs (approved):
## Approve — round 5, closing the loop

Reviewed at `41809f9` on behalf of @hojinzs. I re-verified by probe against the built code rather than by reading the diff, and I am approving rather than opening a sixth round: every acceptance criterion in #921 is met, and everything I found left is a doc trim or a diagnostic, none of which changes the fix.

### #921 acceptance criteria — all met, verified

| Criterion | Result |
|---|---|
| Same folder under two runtime roots cannot attach two orchestrators | ✅ refused, **with `--config` root and `TMPDIR` both differing** |
| Per-runtime-root locks and owner identity validation intact | ✅ unchanged format, lease + pid + `ps` identity preserved |
| Stale owners handled safely | ✅ release removes the lease; heartbeat loss fails closed |
| Focused unit tests + integration scenario | ✅ 65 targeted tests; TC-23 added |

```
P1 acquired: /tmp/probe-fi62sx/proj/.gh-symphony-start.lock
P2 (rootB, TMPDIR=tmpB) refused -> Project "/tmp/probe-fi62sx/proj" is already running (PID 3592).
P4 (symlinked alias to same folder) refused -> Project "/tmp/probe-fi62sx/proj" is already running (PID 3592).
P3 after release exists: false
```

The `realpath` canonicalization is doing real work — a symlinked alias to the same folder is correctly refused (P4), which is not something earlier rounds probed. And the diagnostic now names the canonical folder rather than a sha256 prefix, which was the point of `projectLabel`.

Every validation claim in the PR body reproduces: `pnpm lint`, `pnpm typecheck`, `pnpm build` clean; `pnpm test` **576/576** in cli, all packages green; the three targeted suites **65 passed**. Changeset present and correctly scoped to `@gh-symphony/cli` as a patch.

### Scope and design

No over-engineering, and the fix is smaller than the two designs that preceded it. The identity lease reuses `acquireProjectLock` through two additive optional parameters (`lockPath`, `projectLabel`) rather than introducing a second lock implementation — that is the right call, and `projectLabel` defaulting to `projectId` keeps every existing caller byte-identical. Lock ordering in `acquireProjectStartLocks` is correct: identity first, released if the per-root lock then fails, so a failed start leaves nothing behind. Layering is clean — the composition sits in `cli`, the mechanism in `orchestrator`, and nothing tracker-specific leaked into either.

Round 3's `/var/tmp` anchor was a genuine design mistake and moving into the folder dissolved the `sudo` lockout and the planted-lock DoS rather than mitigating them. Skipping `ensureSecureDirectory` for caller-supplied paths is the detail that makes that safe, and it is correct.

### Prior review findings

All four of round 3's blocking findings are closed, and I confirmed the two I could reach: the cross-`TMPDIR` duplicate is refused (B1/B3 mechanism), and TC-23 now uses `--once` with an explicit non-zero-status assertion so a duplicate-exclusion regression fails the runner instead of hanging it (B4). The three documents no longer contradict each other.

### What I found — none of it blocking

One item is new to this round; the rest are inline.

**P2 — the namespace half of the docs claim is still not delivered.** Round 4 fixed the `TMPDIR`/uid overclaim but `README.md:966-967` still says "or mount namespaces" and `docs/architecture.md:82` says "or process namespaces". The lease *file* is shared across a bind mount, but `isActiveLock` decides whether it is held via `ps -p <pid>`, which is PID-namespace-local — so a starter in a second namespace scores the live lease stale and takes it over. Probed and confirmed. Detail and the exact reproduction are on `lock.ts:141`.

I am explicitly **not** asking for namespace-safe process identity — that is a real design problem, outside what #921 asked for, and chasing it here would be the over-engineering this PR has already been sent back for twice. Deleting the two namespace clauses is a doc-only edit that makes the sentences true.

**P2** — the `projectDir`-absent branch silently skips the identity lock (`project-start-lock.ts:39`), raised in rounds 2 and 3 and still open; reproduced, with a concrete destination.
**P3** — `EACCES`/`ENOENT` on the folder-anchored lock surface raw, and "accessible" carries the behaviour change undefined (`lock.ts:80`).
**nit** — nothing ignores `.gh-symphony-start.lock` (`project-start-lock.ts:10`); TC-23 gates on the runtime-root lock instead of the identity lease (`run-standalone-project-e2e.sh:106`).

### Merging

Approving. I would land the `README.md` / `docs/architecture.md` namespace trim in this PR since it is two clauses and needs no re-review; the P2 downgrade warning, the P3 diagnostic, and the nits are fine as follow-ups. The unchecked box for TC-23 under a restored Docker daemon (#942) still needs a human to close out post-merge.

---

PR #945, opened 2026-09-06
docs(workflow): complete Land failure transitions

moncher-dev:
## Issues

- Closes #939

## Summary

- Completes the Workpad Lifecycle table with every policy-defined Land exit and cycle effect.
- Documents trivial-conflict resolution as an in-cycle recovery that remains in `Land`.
- Keeps installed and generated Land guidance aligned, with focused mutation-checked coverage for pull-skill recovery, conflict classification, and portable lockfile handling.

## Change-point diagram

- `WORKFLOW.md` lifecycle audit → complete Land cycle outcomes
- installed Land skill ↔ generated CLI Land skill → consistent failure destinations and in-cycle recovery
- generated recovery → pull skill fetch/merge → re-run full pre-flight in `Land`
- lifecycle regression test → pins recovery routing, trivial-set classification, and package-manager portability

## Start here

- `WORKFLOW.md:297` — Workpad Lifecycle table
- `packages/cli/src/skills/templates/land.ts:90` — shipped conflict-recovery and rework classification
- `packages/cli/src/skills/templates/land.test.ts:59` — focused lifecycle and recovery coverage

## User-Visible Behavior / Operational Impact

Generated workflow operators can audit every Land failure outcome and its cycle effect. Generated Land recovery routes through the repository pull skill, distinguishes trivial conflicts from rework mechanically, and remains portable across package managers.

## Validation

- `pnpm --filter @gh-symphony/cli test -- src/skills/templates/land.test.ts` — pass (6 tests)
- M1 mutation: replace pull-skill recovery with bare `git merge origin/<base>` — expected fail on `through the pull skill`
- M2 mutation: remove the `outside the trivial set` qualifier — expected fail on `outside the trivial set defined in item 4`
- `pnpm lint` — pass
- `pnpm test` — pass
- `pnpm typecheck` — pass
- `pnpm build` — pass
- `pnpm exec prettier --check .changeset/calm-lockfiles-land.md WORKFLOW.md .codex/skills/land/SKILL.md packages/cli/src/skills/templates/land.ts packages/cli/src/skills/templates/land.test.ts` — pass
- Docker E2E — not applicable: documentation/template guidance only; no integration behavior changed

## Changeset

- `.changeset/calm-lockfiles-land.md` (`@gh-symphony/cli` patch)

## Risks & rollback

- Low risk: policy text, generated guidance, tests, and release metadata only. Revert commits `8e5de88`, `71fe448`, `5460ccf`, `50ac0b1`, and `60f8933` to roll back.
[...]

PR #943, opened 2026-09-06
docs(architecture): add spec-mandate column

moncher-dev:
## Issues

- Closes #909

## Summary

- Adds spec-mandate evidence to every row in the architecture component, layer, and conformance maps.
- Distinguishes RFC 2119 requirements, implementation-defined choices, repository extensions, and documented divergences.
- Corrects workflow, agent-runtime, tracker-tool, and credential citations and preserves retention rationale for repository additions.

## Change-point diagram

- `docs/symphony-spec.md` (read-only source) → mandate classification → `docs/architecture.md` component, layer, and conformance maps

## Start here

- `docs/architecture.md:15` — mandate terminology and component-to-implementation mapping
- `docs/architecture.md:36` — six-layer mandate mapping, including existing Coordination divergences
- `docs/architecture.md:201` — conformance-test mandate mapping

## User-Visible Behavior / Operational Impact

- None. This is a documentation-only clarification of existing architecture and conformance responsibilities.

## Validation

- `pnpm format` — fails on 66 pre-existing files outside this issue; `docs/architecture.md` is not listed
- `pnpm exec prettier --check docs/architecture.md` — pass
- `pnpm lint` — pass
- `git diff --check` — pass
- Unit tests, typecheck, build, and Docker E2E: not applicable under the docs-only validation policy because implementation and integration behavior did not change

## Changeset

- Not needed because this is a documentation-only change and the issue has no changeset label.

## Risks & rollback

- Risk is limited to an inaccurate mandate classification; reviewer focus should be the cited upstream sections and the distinction between implementation-defined behavior, extensions, and documented divergences.
- Rollback by reverting the documentation commits.

## Changed files

- `docs/architecture.md` — adds verifiable mandate classifications to all three requested maps and incorporates all review corrections.

## Post-merge / human validation

- [ ] None beyond reviewing the cited spec sections.

## Security

- [ ] No real tokens, private keys, `.env` files, or generated installation tokens are committed

PR #946, opened 2026-09-06
fix(workflow): detect actionable body-only reviews

### LINKED ISSUE #900: bug(workflow): a steward review with no inline threads does not trigger the rework guard

moncher-dev:
## Issues

- Closes #900

## Summary

- Treat recent body-only `COMMENTED` reviews as rework only when a non-Bot author requests changes or reports findings.
- Preserve same-account steward feedback while excluding automated boilerplate and passing steward reviews.
- Define the status-comment handoff boundary and its missing-boundary fallback explicitly.

## Change-point diagram

- GitHub review → `prReviewState` exposes body and author type → Ready-return guard applies handoff time + non-Bot + actionable-content filters → qualifying body opens rework

## Start here

- `WORKFLOW.md:167` — review data and handoff-boundary definition
- `WORKFLOW.md:168` — actionable body-only review classification
- `packages/cli/src/skills/templates/land.test.ts:144` — regression coverage

- Blocking steward findings delivered only in a top-level review body trigger rework even from the worker account; passing reviews and Bot boilerplate do not.

## Validation

- `pnpm exec vitest run packages/cli/src/skills/templates/land.test.ts` — pass (6 tests)
- `pnpm lint` — pass
- `pnpm test` — pass
- `pnpm typecheck` — pass
- `pnpm build` — pass
- `git diff --check` — pass
- Docker E2E / blackbox — not applicable: repository policy text only; no runtime integration behavior changed

## Changeset

- Not needed because issue #900 has no changeset label and changes repository workflow policy only.

## Risks & rollback

- Risk: actionability is interpreted from natural-language review content; the policy now names the required “requests changes or reports findings” signal explicitly.
- Rollback: revert `84d8441`, `ad36c7a`, and `a2f0113`.

## Changed files

- `WORKFLOW.md` — recognize actionable body-only User reviews with explicit bot, pass-review, state, and boundary exclusions.
- `packages/cli/src/skills/templates/land.test.ts` — pin the actionability, same-account, bot, fallback, state, and query requirements.

- [ ] None required beyond normal review of the policy wording.

## Security


moncher-dev (commented):
Blocking, but the core of the fix is right and I verified it against the real incident rather than reading it.

## What I confirmed works

**The data channel carries the finding.** I ran the amended `prReviewState` selection against PR #899 — the PR from the #900 repro — with the `body` field this PR adds:

```
reviewDecision: APPROVED  threads: 4
2026-09-05T13:54:00Z  moncher-dev  COMMENTED  bodylen=3889  | Blocking. The new assertion is green in CI, but mutation testing shows...
2026-09-05T15:07:34Z  hojinzs      APPROVED   bodylen=4032  | Approving. The round-1 blocking finding is genuinely fixed...
```

The 3889-character steward body is retrievable, and it is retrievable for `moncher-dev` — the worker's own account. A fix that keyed on a third-party author would not have caught this, and this one does not.

**The boundary rule is correct on the real timeline.** Issue #872's last `🔁 Status: In progress → In review` comment is `2026-09-05T13:01:34Z`; the steward review is `13:54:00Z`. After the boundary, so the guard fires. The 13:55 dispatch that produced no commit would have been classified as rework.

**The test can fail.** Mutating `regardless of author identity` to `for reviews authored by a human other than the worker` turns it red. It is a string pin on a prompt document, which is this file's convention, and it pins the clause that actually matters.

## Blocking: no bot exclusion, so the guard fires on every `Ready` return

`chatgpt-codex-connector[bot]` posts a top-level `COMMENTED` review with a non-empty body on every PR in this repository. Verified — each of the four other open PRs (#936, #941, #943, #945) carries exactly one. Its body is boilerplate only:

```
### 💡 Codex Review
Here are some automated review suggestions for this pull request.
**Reviewed commit:** `60f893394b`
<details> <summary>ℹ️ About Codex in GitHub</summary> ... </details>
```

The bot's actual findings are inline thread comments; the body never carries them. And by its own description the review is triggered when you "mark a draft as ready" — which is Step 2.8. So it always lands *after* the most recent `→ In review` status comment.

That means it satisfies every clause of the new condition, on every PR, always: non-empty, top-level, `COMMENTED`, after the boundary, and now explicitly `regardless of author identity`.

Consequence: once this merges, no `Ready` entry on an issue with an open linked PR can classify as fresh pickup or resume again. Every one becomes a rework cycle. That includes:

- the retry-budget re-arm path, where a suppressed issue is moved `Ready` ↔ `In progress` specifically to re-arm — the board depends on this;
- plain board drift, which item 6 exists to route.

The worker then opens a rework cycle and goes looking for feedback that does not exist, producing a dispatch with no commit. That is the exact cost #900 was filed to eliminate, reintroduced from the other side.

**The discriminator is already in the query you amended.** `reviews.nodes.author.login` is `chatgpt-codex-connector[bot]`, and adding `author { __typename }` yields `Bot` versus `User`. Excluding bot-authored review bodies leaves the entire steward case intact, because the steward posts as the `User` `moncher-dev` — which is the whole point of the author-agnostic clause. "Regardless of author identity" is broader than the problem requires; what #900 needs is "not restricted to humans other than the worker", not "including machines".

## Secondary — two branches without a named destination

1. **No boundary comment.** "Find the most recent issue status comment recording a transition to `In review`; its comment `createdAt` is the handoff boundary." The text does not say what happens when no such comment exists — an issue that has never reached `In review`, or one where the worker died between the board write and publishing the status comment. Both readings are available from the text and they behave oppositely: vacuously false (no review body ever qualifies) or unbounded (every review body qualifies, including ones addressed cycles ago). Name the fallback explicitly.

2. **Only `COMMENTED` is read.** A `DISMISSED` review body, and an `APPROVED` body carrying conditions ("approving once X is fixed"), are both silently excluded. #899's approval body is 4032 characters. Excluding them is probably correct, but given that this issue is precisely about a delivery channel being invisible, the exclusion should be stated rather than left to inference.

Item 1 in this section should be fixed alongside the blocker. Item 2 is a judgement call — say which way you went and why, and I will not push on it again.

moncher-dev (commented):
Passing at `ad36c7a8`. All three findings from my earlier review this cycle are fixed, and the fix for the blocker is better than what I proposed.

## The bot exclusion

I asked for bot-authored review bodies to be excluded, because `chatgpt-codex-connector` posts a non-empty boilerplate `COMMENTED` body on every PR after every "mark ready" — which is Step 2.8 — so it would have satisfied the new condition on every single `Ready` return and turned every fresh pickup and every retry re-arm into an empty rework cycle.

The rule now reads "a non-empty top-level `COMMENTED` review body **from a non-`Bot` author**", and `prReviewState` selects `author{login __typename}`.

**I verified the discriminator against live data on PR #945**, which carries both kinds of review:

```
chatgpt-codex-connector    __typename=Bot    COMMENTED  bodylen=621
moncher-dev                __typename=User   COMMENTED  bodylen=4075
moncher-dev                __typename=User   COMMENTED  bodylen=0
moncher-dev                __typename=User   COMMENTED  bodylen=3766
```

The boilerplate is excluded and the steward body is kept. Note the detail that makes `__typename` the right choice rather than a nicety: GraphQL returns the bot's login as `chatgpt-codex-connector`, **without** the `[bot]` suffix that REST shows. A login-suffix heuristic would have silently failed here. Choosing the type discriminator over the login string is correct, and I would not have caught the difference from the diff alone.

The steward case is also preserved with the right phrasing — "triggers rework **even when its author login matches the worker account**" — which states the actual #900 condition rather than the over-broad "regardless of author identity" I objected to. The empty-bodied `moncher-dev` review above is excluded by "non-empty", as intended.

## The two secondary items

Both now have named destinations rather than being left to inference:

- **No boundary comment:** "If no such status comment exists, no top-level review body qualifies through this condition; use the other rework signals below." Vacuously false, stated explicitly, so the unbounded reading is closed off.
- **Only `COMMENTED`:** "`APPROVED` and `DISMISSED` bodies retain their separate review-state semantics." The exclusion is now a decision on the page instead of a gap.

## Still true from my first pass

I re-verified rather than assuming these survived the rework:

- `reviews { body }` returns the 3889-character steward review on PR #899 authored by `moncher-dev` — the exact review the #900 incident lost.
- The boundary holds on the real timeline: issue #872's last `🔁 Status: In progress → In review` is `13:01:34Z`, the steward review is `13:54:00Z`, so the guard fires and the 13:55 no-commit dispatch would have been classified as rework.

## Tests

`packages/cli/src/skills/templates/land.test.ts` — 6 passed. I mutation-tested both new clauses rather than trusting a green suite:

- deleting `from a non-`Bot` author` → **fails**
- deleting the "If no such status comment exists" sentence → **fails**

So neither of the two fixes can be silently reverted. The test also pins the `__typename` selection in the query table, which is what makes the rule executable rather than aspirational.

Handing to @hojinzs. One thing worth your eye, though I think it is right: excluding `Bot` authors means that if Codex ever moves its findings from inline threads into the review body, those findings stop reaching the guard through this path. Today they arrive as `reviewThreads`, which the guard already reads, so nothing is lost now — but the exclusion is a bet on that staying true.

WORKFLOW.md
```diff
@@ -164,8 +164,8 @@ When entering `Ready`, before treating it as a fresh pickup, board drift, or res
...
...a non-`Bot` author submitted after the most recent transition to `In review`, an unresolved actionable review thread, a human instruction indicating rework, or a recent `Land` → `Ready` transition recorded as a **Land-return rework**, this `Ready` state means **review rework return** — not a fresh pickup and not drift. A qualifying top-level `COMMENTED` review body triggers rework even when its author login matches the worker account and does not require an inline thread; this covers steward feedback posted through a shared worker account without treating automated review boilerplate as actionable. Only `COMMENTED` review bodies qualify through this condition; `APPROVED` and `DISMISSED` bodies retain their separate review-state semantics. When a latest qualifying human approval exists on the current head, narrow the thread condition to unresolved actionable threads created after that approval (compare the thread's first comment `createdAt` with the approval review's `submittedAt`); a thread created at or before that approval is absorbed by the approval and does not trigger rework. When no qualifying approval exists, any unresolved actionable review thread triggers rework as before.
```
hojinzs: **P2 — the new clause has no actionability qualifier, so a *passing* steward review triggers rework**

The `Bot` exclusion closes the boilerplate case, but the residual false positive is the same shape as the round-1 blocker, one step narrower: **any** non-empty `COMMENTED` body from a `User` after the boundary qualifies, whether or not it asks for anything.

That is not hypothetical — **this PR's own review history is a live instance.** Review [`5124519335`](https://github.com/hojinzs/github-symphony/pull/946#pullrequestreview-5124519335) is:

| clause | value |
| --- | --- |
| state | `COMMENTED` |
| body | non-empty (~2.8 KB) |
| author `__typename` | `User` (`moncher-dev`) |
| `submittedAt` | `2026-09-06T07:07:24Z` |
| boundary (`In progress` → `In review` status comment on #900) | `2026-09-06T07:01:41Z` |

Every clause of item 4 is satisfied. Its content is *"Passing at `ad36c7a8`. All three findings from my earlier review this cycle are fixed... Handing to @hojinzs."*

This is structural, not incidental: #900 records that the steward **is not a repository reviewer and cannot submit `APPROVED`**, so *every* steward pass lands as a `COMMENTED` review with a long body. Under this rule a pass is byte-for-byte indistinguishable from a block.

Consequence on any subsequent `Ready` entry — the retry-budget re-arm path and plain board drift, i.e. the two paths you named as blocking for the bot case:

1. Guard classifies rework and opens a cycle (item 5).
2. Step 2.2 rework preamble distills "the main merge blockers" — there are none.
3. Step 2.5 resting-state rule requires an unchecked in-scope Plan item to end the turn in `In progress`, and there is nothing to plan.

A dispatch that ends where it started with no commit — the exact cost #900 was filed to eliminate. There is no escape hatch back to item 6 once the guard has classified.

**The two sibling conditions in this same sentence both carry the qualifier this one is missing:** "an unresolved **actionable** review thread", "a human instruction **indicating rework**". Adding it here restores consistency and closes the case, e.g.:

> a non-empty top-level `COMMENTED` review body from a non-`Bot` author **that requests changes or reports findings** submitted after the most recent transition to `In review`

Pin it the way you pinned the other two clauses and this is done. To be explicit about scope: this one qualifier is the whole ask — please do not widen the rule further.

---

WORKFLOW.md
```diff
@@ -164,8 +164,8 @@ When entering `Ready`, before treating it as a fresh pickup, board drift, or res
...
+3. For each remaining open linked PR, read `reviewDecision`, reviews (including each review's top-level `body`, `state`, `submittedAt`, and author `__typename`), review threads (`prReviewState` query), top-level PR comments, and recent issue comments. Find the most recent issue status comment recording a transition to `In review`; its comment `createdAt` is the handoff boundary for top-level review bodies. If no such status comment exists, no top-level review body qualifies through this condition; use the other rework signals below.
```
hojinzs, line 167: **P3 — item 4 does not point back at the boundary item 3 defines**

Item 3 defines the boundary precisely and correctly: *the most recent **issue status comment** recording a transition to `In review`*, using that **comment's `createdAt`**. That is the right choice, because it is the only one of the two candidates the worker can actually read.

Item 4 then says "submitted after the most recent **transition** to `In review`". Read on its own, that phrase names the Project field change — which has its own timestamp, is not the status comment's `createdAt`, and is not in any query in the Data Access table. The two timestamps diverge whenever the board write and the comment publish straddle a retry, which is precisely the crash window your new fallback sentence was written for.

Same fix as the fallback: name the destination. "submitted after the handoff boundary defined in item 3" removes the second reading at no cost.

Not blocking on its own — but #900 is a case study in an under-specified clause costing a full dispatch, so it is worth spending one phrase here.

---

WORKFLOW.md
```diff
@@ -421,19 +421,19 @@ None
...
+| Create Draft PR                       | `createPullRequest(input:{repositoryId, baseRefName, headRefName, title, body, draft:true}){pullRequest{id number url}}`                                                                                                                                                                                                                     |
+| Refresh PR body / mark ready          | `updatePullRequest(input:{pullRequestId, title, body})` · `markPullRequestReadyForReview(input:{pullRequestId}){pullRequest{isDraft}}` · `requestReviews(input:{pullRequestId, userIds, union:true})`                                                                                                                                        |
+| Review state                          | `prReviewState`: `pullRequest(number){ id headRefOid reviewDecision reviews(last:30){nodes{state body author{login __typename} submittedAt commit{oid}}} reviewThreads(first:100){nodes{id isResolved isOutdated path comments(first:30){nodes{id body author{login} createdAt}}}} comments(last:30){nodes{body author{login} createdAt}} }` |
```
hojinzs, line 431: **P3 — `body` on `reviews(last:30)` is a per-dispatch context cost worth bounding**

`__typename` is free and is the right discriminator (your note that GraphQL returns the login as `chatgpt-codex-connector` **without** the `[bot]` suffix is the detail that makes it right rather than merely tidy — a login heuristic would have failed silently here).

`body` is not free. This selection now pulls up to 30 full review bodies, and on this board a steward body runs 3–4 KB — the #900 repro cites 3,889 and 4,032 characters on a single PR. `prReviewState` is read in Step 0 item 3, again in the Step 2.2 rework preamble, again in Step 3, and again in the Land pre-flight, so the worst case is roughly 100 KB of review text pulled into worker context up to four times in one dispatch.

For a guard whose stated cost of failure is measured in tokens (529,524 for one wasted dispatch), that is worth bounding. The rule only ever consults bodies newer than the handoff boundary, and reviews come back newest-last, so `reviews(last:10)` covers every case the guard can act on. If you would rather keep 30 for the approval-history lookups elsewhere, say so in the reply and I will not raise it again.

---

packages/cli/src/skills/templates/land.test.ts
```diff
@@ -141,6 +141,33 @@ describe("merged-PR lifecycle guards", () => {
...
+    const ready = section(
+      workflow,
+      "##### Ready-return rework guard",
+      "##### Stalled-handoff safety net"
+    );
+
+    expect(ready).toContain(
+      "a non-empty top-level `COMMENTED` review body from a non-`Bot` author submitted after the most recent transition to `In review`"
+    );
+    expect(ready).toContain("from a non-`Bot` author");
```
hojinzs: **nit — this assertion is subsumed by the one above it**

`"from a non-\`Bot\` author"` is a strict substring of the string asserted on lines 149–151, so it cannot fail unless that one already has. I confirmed the coverage claim by mutation rather than by reading: deleting ` from a non-\`Bot\` author ` from `WORKFLOW.md:168` turns the suite red (1 failed / 5 passed), so the pin is real — but it is line 149 doing the work, and this line adds nothing.

Worth deleting, or worth changing into the assertion it looks like it wants to be — a pin on the actionability qualifier, if the P2 on `WORKFLOW.md:168` lands.

The rest of this test is the right shape for the file's convention: it slices the guard section rather than searching the whole document, so a matching string appearing elsewhere in `WORKFLOW.md` cannot produce a false green, and the `__typename` pin is correctly asserted against the full document because the query table lives outside the section.

---

hojinzs (changes requested):
One P2 blocker, three non-blocking notes. The direction is right and the round-1 rework genuinely improved it — I am blocking on a single missing qualifier, and that qualifier is the whole ask.

## Requirement coverage against #900 — complete

I checked the issue body and all 10 issue comments requirement-by-requirement:

| #900 requirement | status |
| --- | --- |
| Guard-side fix (the direction the issue's controlled comparison settled on) | ✅ `WORKFLOW.md:168` |
| Fires on a body-only review with **zero inline threads** | ✅ "does not require an inline thread" |
| Fires when the author **is** the worker account (the load-bearing constraint) | ✅ "even when its author login matches the worker account" |
| Boundary anchored to the last `→ In review` handoff | ✅ `WORKFLOW.md:167` |
| Out of scope: steward auth, #878 approval re-entry | ✅ untouched |

No over-engineering: two policy sentences, one query field, one test. Nothing here that #900 did not ask for.

## Smoke test — run locally at `ad36c7a8`

This PR changes prompt policy, so there is no API or screen to exercise; the executable surface is the test pin and the GraphQL selection. I verified both rather than reading them.

- `pnpm exec vitest run packages/cli/src/skills/templates/land.test.ts` — **6 passed**
- **Mutation test** (the assertion is the deliverable here, so a green suite proves nothing on its own): deleting ` from a non-\`Bot\` author ` from `WORKFLOW.md:168` → **1 failed / 5 passed**. The bot exclusion cannot be silently reverted.
- `npx prettier --check WORKFLOW.md packages/cli/src/skills/templates/land.test.ts` — clean, so neither touched file joins the 67-file pre-existing baseline
- `pnpm lint` — pass · `pnpm build` — pass · `pnpm typecheck` — pass
  - Note for the record: `pnpm typecheck` fails on a **clean checkout of `main`** too, with `TS2307: Cannot find module '@gh-symphony/tool-linear-graphql'` in `packages/tracker-linear`, until `pnpm build` has populated the workspace `dist/`. Not this PR's; the ordering is just implicit in the documented command list.
- Delivery-path check: the Ready-return guard exists **only** in `WORKFLOW.md`. `packages/cli/src/skills/templates/land.ts` carries the Land-side rework classification but not this guard, so there is no second copy to keep in sync and no generated-skill path that would ship the old rule. Editing `WORKFLOW.md` alone does deliver the fix.

## Round-1 findings — all three fixed, verified independently

I re-derived these rather than taking the round-2 review's word:

1. **Bot exclusion** — present, and `__typename` is the correct discriminator over a login-suffix heuristic.
2. **No-boundary fallback** — named explicitly, and mutation-tested by your round 2.
3. **`COMMENTED`-only scope** — now a stated decision instead of a gap.

## Blocking — `WORKFLOW.md:168`

The `Bot` exclusion closes the boilerplate case but leaves the same defect class one step narrower: the clause has **no actionability qualifier**, so a steward review that says *"Passing"* satisfies it exactly as well as one that blocks. Since #900 records that the steward cannot submit `APPROVED`, every steward pass is a `COMMENTED` review with a long body — and **this PR's own review `5124519335` satisfies every clause of the new rule right now**. Full reasoning and the timestamps are in the inline comment.

Both sibling conditions in that same sentence already carry the qualifier ("**actionable** review thread", "human instruction **indicating rework**"). Adding it to this one is a single phrase.

## Non-blocking

- `WORKFLOW.md:167` (P3) — item 4 says "after the most recent transition to `In review`" where item 3 defines the boundary as the status **comment's** `createdAt`; point item 4 at item 3.
- `WORKFLOW.md:431` (P3) — `body` on `reviews(last:30)` can pull ~100 KB of review text into worker context, up to four times per dispatch.
- `land.test.ts:155` (nit) — assertion subsumed by the one above it.

## Follow-up, deliberately not in this PR

Issue comment [`5557533795`](https://github.com/hojinzs/github-symphony/issues/900#issuecomment-5557533795) asked for a second correction: Step 0 item 3's *"a human instruction indicating rework"* is still unreachable for the steward, so the issue body's claim that the issue-comment workaround "Step 0 item 3 does read" remains wrong on the page — demonstrated failing twice from `moncher-dev`. This PR closes #900 and would close that with it. Please **open a follow-up issue** rather than widening this one; keeping this PR to the review-body channel is the right call.

Fix the one qualifier and I will approve on the next pass.

---

WORKFLOW.md
moncher-dev: Resolved in `84d8441`: item 4 now compares `submittedAt` against “the handoff boundary defined in item 3,” removing the ambiguous Project-transition timestamp reading.

WORKFLOW.md
moncher-dev: Resolved in `84d8441`: the guard now requires the review body to request changes or report findings, and explicitly excludes passing review bodies from actionable feedback. The regression test pins both the actionability clause and the passing-review exclusion.

WORKFLOW.md
moncher-dev: Keeping `reviews(last:30)` intentionally because `prReviewState` also supports latest qualifying approval and approval-history checks in Ready-return and Land flows. Reducing the shared window to 10 could omit the approval that bounds actionable threads; the guard still applies the handoff-boundary and actionability filters before classifying review bodies.

packages/cli/src/skills/templates/land.test.ts
moncher-dev: Resolved in `84d8441`: removed the subsumed author-only assertion and replaced it with a regression pin for the new passing-review/actionability distinction.

hojinzs (approved):
Approving at `84d8441`. The blocking qualifier landed, I mutation-tested it rather than reading it, and all four round-2 findings are closed.

## The blocker is fixed — verified by mutation, not by reading

`WORKFLOW.md:168` now reads:

> a non-empty top-level `COMMENTED` review body from a non-`Bot` author **that requests changes or reports findings** and was submitted after the handoff boundary defined in item 3

plus the explicit exclusion: *"without treating automated review boilerplate or **passing review bodies** as actionable"*.

Two independent mutations, each run against the pinned suite:

| mutation on `WORKFLOW.md:168` | result |
| --- | --- |
| drop `that requests changes or reports findings and ` | **1 failed / 5 passed** |
| drop ` without treating automated review boilerplate or passing review bodies as actionable` | **1 failed / 5 passed** |

Neither clause can be silently reverted. That matters more than usual here, because on a prompt document the assertion *is* the deliverable.

**The live instance I blocked on is now correctly excluded.** Review [`5124519335`](https://github.com/hojinzs/github-symphony/pull/946#pullrequestreview-5124519335) — the "Passing at `ad36c7a8`… Handing to @hojinzs" body — still satisfies non-empty, `COMMENTED`, `User`, and after-boundary, but no longer satisfies the rule, because a passing body is now named as not actionable on the page. The structural problem I raised (the steward cannot submit `APPROVED`, so every steward pass is a long `COMMENTED` body indistinguishable from a block) is closed at its source rather than worked around.

Note that you fixed it with the phrase I proposed *and* the exclusion sentence, which is the stronger of the two: the qualifier alone would have left "reports findings" to do a lot of unaided work against a body that says *"All three findings from my earlier review are fixed."* Naming passing bodies explicitly removes that reading.

## Round-2 findings — all four closed

1. **P2 actionability qualifier** — fixed, mutation-verified above.
2. **P3 `WORKFLOW.md:167` boundary reference** — item 4 now says "the handoff boundary defined in item 3". The Project-field-timestamp reading is gone.
3. **P3 `reviews(last:30)` context cost** — declined with a reason I accept and will not raise again: the same selection backs the latest-qualifying-approval lookup in Ready-return and Land, and narrowing to 10 could drop the approval that bounds actionable threads. Correct call; the window is load-bearing for a second consumer I had not weighed.
4. **nit `land.test.ts:155`** — subsumed assertion replaced with a real pin on the new actionability/passing-review distinction. Better than the deletion I suggested.

## Smoke test at `84d8441`

Policy-only change, so the executable surface is the test pin and the GraphQL selection; I exercised both.

- `npx vitest run packages/cli/src/skills/templates/land.test.ts` — **6 passed**
- `pnpm test` — **917 passed** (342 orchestrator + 575 cli), no failures
- `pnpm lint` · `pnpm typecheck` · `pnpm build` — pass
- `npx prettier --check WORKFLOW.md packages/cli/src/skills/templates/land.test.ts` — clean; neither file joins the pre-existing baseline. The Data Access table reflow in the diff is Prettier re-padding columns after `body`/`__typename` widened the cell — no semantic change.
- CI on this head: **Test** ✅ · **Container Smoke** ✅

## Requirement coverage against #900 — complete, and still no over-engineering

Guard-side fix; fires on a zero-thread body; fires when the author is the worker account; boundary anchored to the last `→ In review` handoff with a named fallback; steward auth and #878 left untouched. Final shape is two policy sentences, one query field, one test — nothing #900 did not ask for, and the round-2 rework did not widen it.

## One item for the board, not for this PR

The follow-up I flagged last round still stands and is correctly not in scope here: Step 0 item 3's *"a human instruction indicating rework"* remains unreachable for the steward, so the #900 issue body's claim that the issue-comment workaround "Step 0 item 3 does read" is still wrong on the page. This PR closes #900 and takes that with it — please open a follow-up issue after merge.

Good convergence. Three rounds, each one narrower than the last.

---

PR #949, opened 2026-09-06
fix(core): preserve legacy runtime timeout fallback

### LINKED ISSUE #888: fix(core): runtime block without timeouts silently discards legacy codex timeout declarations
## Problem

`resolveWorkflowRuntimeTimeouts` (`packages/core/src/workflow/config.ts:285`) is written as a fallback:

```ts
export function resolveWorkflowRuntimeTimeouts(
  workflow: Pick<WorkflowDefinition, "runtime" | "codex">
): WorkflowRuntimeTimeoutsConfig {
  return workflow.runtime?.timeouts ?? workflow.codex;
}
```

The `?? workflow.codex` arm is unreachable for any workflow that declares a `runtime:` block, because the parser materializes `runtime.timeouts` with hardcoded defaults whenever a `runtime:` block is present. A legacy top-level `codex:` timeout declaration is then silently discarded.

The orchestrator forwards this resolver to the runtime, so this is a behavior bug, not only a reporting one.

## Reproduction

Front-matter containing **both** a `runtime:` block with no `timeouts:` sub-block and a legacy top-level `codex:` block:

```yaml
runtime:
  kind: codex-app-server
  command: codex
  args:
    - app-server
codex:
  read_timeout_ms: 7777
  stall_timeout_ms: 88888
  turn_timeout_ms: 999999
```

```
parseWorkflowMarkdown(...)
  workflow.codex          -> {readTimeoutMs: 7777,  stallTimeoutMs: 88888,  turnTimeoutMs: 999999}
  workflow.runtime.timeouts -> {readTimeoutMs: 5000, stallTimeoutMs: 300000, turnTimeoutMs: 3600000}  # defaults, never declared
  resolveWorkflowRuntimeTimeouts(workflow) -> 5000 / 300000 / 3600000
```

The declared 7777 / 88888 is dropped without a warning. Removing the `runtime:` block resolves correctly to 7777, which confirms the trigger is the presence of the `runtime:` block rather than the `codex:` block being malformed.

## Why it matters

This is the mirror of #882. There the CLI reported a value the runtime did not use; here the runtime uses a value the operator did not declare. A repository migrating from the legacy `codex:` timeouts to `runtime.timeouts` — the exact state where both blocks coexist — silently reverts to a 5s read timeout and a 300s stall timeout. On this repository those defaults are 6x and 3x tighter than the declared values, which is the failure mode #882 was chasing.

Found while verifying #884 (PR for #882). Not caused by that PR: `main` and the PR branch resolve identically here. #884 is correct to report the effective value; this issue is about the effective value itself being wrong.

## Suggested fix

Distinguish "declared" from "defaulted" for `runtime.timeouts` so the fallback can actually fire — e.g. keep `runtime.timeouts` optional through parsing and apply defaults only in `resolveWorkflowRuntimeTimeouts` after the `?? workflow.codex` arm, or track which timeout fields were explicitly present.

Per-field merge is worth considering: today a `runtime.timeouts` block that sets only `read_timeout_ms` also silently defaults the other two rather than inheriting them from `codex:`.

Either way, a workflow declaring both blocks should not lose the declared values in silence — a validation warning naming the shadowed keys would help operators mid-migration.

## Acceptance

- A workflow with a `runtime:` block, no `runtime.timeouts:`, and a legacy `codex:` timeout block resolves to the `codex:` values.
- A workflow with an explicit `runtime.timeouts:` continues to resolve to those values (#882 behavior preserved).
- A workflow with neither still resolves to the documented defaults.
- Unit tests in `packages/core` cover all three, plus the partial-`runtime.timeouts` case.

moncher-dev:
## Issues

- Closes #888

## Summary

- Fix runtime blocks without explicit timeouts so they retain declared legacy `codex.*_timeout_ms` values.
- Resolve partial runtime timeout blocks per field, falling back to legacy values and then documented defaults.
- Report each effective timeout's actual source in text and JSON validation output, including mixed-source blocks.
- Ship the behavior change as an `@gh-symphony/cli` patch changeset.

## Change-point diagram

- `WORKFLOW.md` front matter → core parser resolves value and source per timeout field → orchestrator consumes effective values / CLI reports matching provenance → patch release metadata documents the change

## Start here

- `packages/core/src/workflow/parser.ts` — per-field timeout precedence and provenance
- `packages/cli/src/commands/workflow.ts` — text and JSON validation reporting
- `packages/core/src/workflow-loader.test.ts` and `packages/cli/src/commands/workflow.test.ts` — legacy, explicit, default, and partial cases
- `.changeset/tidy-wolves-inherit.md` — patch release note

- A declared legacy Codex timeout is no longer silently replaced by a default merely because a `runtime` block exists.
- `workflow validate` identifies each field as sourced from `runtime.timeouts` or `codex/defaults`; JSON uses `summary.runtimeTimeoutSources`.

## Validation

- `pnpm --filter @gh-symphony/core test -- workflow-loader.test.ts` — pass (106 tests)
- `pnpm --filter @gh-symphony/cli test -- workflow.test.ts` — pass (47 tests)
- `pnpm lint` — pass
- `pnpm test` — pass
- `pnpm typecheck` — pass
- `pnpm build` — pass
- `pnpm exec prettier --check .changeset/tidy-wolves-inherit.md` — pass
- `git diff --check` — pass
- Docker E2E — cycle 1 environment exception tracked in #950; cycle 3 changes release metadata only

## Changeset

- `.changeset/tidy-wolves-inherit.md` (`patch`, `@gh-symphony/cli`)

## Risks & rollback

- Risk is limited to timeout precedence and validation-report metadata. Revert commits `e75b85a`, `1d8f386`, `48738f9`, and `9c1be9a` to restore the former behavior and release metadata.

## Changed files

- `packages/core/src/workflow/parser.ts`, `packages/core/src/workflow/config.ts` — resolve values and retain per-field provenance
- `packages/core/src/workflow-loader.test.ts` — cover requested precedence and source cases
- `packages/cli/src/commands/workflow.ts`, `packages/cli/src/commands/workflow.test.ts` — report and test accurate per-field sources
- `README.md`, `packages/cli/README.md`, `docs/configuration.md` — document inheritance and report metadata
- `docs/architecture.md` — map timeout resolution to the Configuration Layer
- `.changeset/tidy-wolves-inherit.md` — declare the shipped CLI patch change

- [ ] None required beyond automated validation.

## Security


moncher-dev (commented):
## Steward review — CHANGES NEEDED at `48738f9`

Scratch worktree at the PR head: `pnpm install --frozen-lockfile && pnpm build && pnpm test` all green. The **behavioural** half of #888 is genuinely fixed — `parseRuntimeConfig` now takes `codexConfig` and falls back to it per timeout field, and the `codexConfig` declaration was correctly hoisted above `runtime` so the new argument is in scope.

But the PR leaves the **reporting** half broken, and the test change codifies the wrong answer. #888's own body says the resolver bug is "a behavior bug, **not only** a reporting one" — the reporting one still has to go.

### Reproduced against this PR's build, not asserted

I built `wt-949` from `48738f9` and ran the shipped CLI against a probe workflow with a `runtime:` block that declares **no** `timeouts:` sub-block and a legacy top-level `codex:` block:

```yaml
runtime:
  kind: codex-app-server
  command: codex
  args:
    - app-server
codex:
  read_timeout_ms: 7000
  stall_timeout_ms: 60000
  turn_timeout_ms: 120000
```

```
$ node packages/cli/dist/index.js workflow validate --file probe/WORKFLOW.md
  runtime.timeouts.read_timeout_ms=7000 (source: runtime.timeouts)
  runtime.timeouts.stall_timeout_ms=60000 (source: runtime.timeouts)
  runtime.timeouts.turn_timeout_ms=120000 (source: runtime.timeouts)
```

There is no `runtime.timeouts` in that file. All three values came from `codex:`. `workflow validate` names the wrong config block.

The mixed case is worse. Adding **only** `read_timeout_ms: 11111` under `runtime.timeouts`:

```
  runtime.timeouts.read_timeout_ms=11111 (source: runtime.timeouts)   <- true
  runtime.timeouts.stall_timeout_ms=60000 (source: runtime.timeouts)  <- false, from codex
  runtime.timeouts.turn_timeout_ms=120000 (source: runtime.timeouts)  <- false, from codex
```

One label, three different provenances.

### Cause

`packages/cli/src/commands/workflow.ts:892` is unchanged by this PR:

```ts
runtimeTimeoutSource: workflow.runtime ? "runtime.timeouts" : "codex/defaults",
```

It selects the label from the mere *existence* of `workflow.runtime`. That was defensible before this PR, when a `runtime:` block always materialised its own timeouts. This PR makes the block's presence stop implying the values' origin, so the condition is now false in exactly the case the PR was written to introduce. This is the "is the stated reason actually true?" class — the change made a previously-true label false and left it in place.

### And the test locks the wrong answer in

`packages/cli/src/commands/workflow.test.ts` was renamed to "legacy codex timeout fallback with a runtime block" and its `expected` correctly updated to `[7000, 60000, 120000]`, but `expectedSource` was left as `"runtime.timeouts"`. So the suite now *asserts* the misreport. Whoever fixes the label later will read this as intended behaviour. (Same point the Codex bot raised inline at line 162 — it is correct.)

This matters beyond tidiness: `workflow validate` is the diagnostic an operator reaches for when debugging a timeout, and this repository has already paid for a read-timeout incident (#814). A provenance report that points at the wrong config block sends that operator to edit a section that has no effect.

### Asked for

1. Carry per-field provenance through the report — a `runtimeTimeouts` source per field (`runtime.timeouts` | `codex` | `defaults`), or at minimum a `mixed` label — instead of one flag derived from block existence.
2. Update `expectedSource` in the fallback case to the accurate source, and add a mixed-source case (one field from `runtime.timeouts`, two inherited from `codex`) so the mixed report is pinned.
3. Add a test that fails if the provenance logic is reverted to `workflow.runtime ? … : …`. Two of three fixes in the sibling PR #945 shipped unpinned and had to be reworked; please do not repeat that here.

### Checked and clean

- `resolveWorkflowRuntimeTimeouts` has no other callers whose behaviour changes unexpectedly; the orchestrator forwarding path is the intended fix target.
- The `codexConfig` hoist introduces no use-before-declaration — nothing between the old and new positions reads `runtime`, and `pnpm typecheck` would have caught it.
- Only `timeouts` inherit from `codex`. `model`, `sandbox`, and `approval_policy` are untouched, which matches #888's scope — flagging it only so the narrowness is a recorded decision rather than an oversight.
- Docs updates (`README.md`, `docs/configuration.md`, `docs/architecture.md`, `packages/cli/README.md`) describe the new fallback correctly.

Returning #888 to `Ready`.

packages/cli/src/commands/workflow.test.ts
```diff
@@ -159,7 +159,7 @@ codex:
   read_timeout_ms: 7000
   stall_timeout_ms: 60000
   turn_timeout_ms: 120000`,
-      expected: [5000, 300000, 3600000],
+      expected: [7000, 60000, 120000],
```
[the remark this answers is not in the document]
moncher-dev: Resolved in 1d8f386. The parser now preserves a source for each timeout field. `workflow validate` prints that per-field source and JSON exposes `summary.runtimeTimeoutSources`, so runtime-less fallback and partial runtime blocks report `codex/defaults` only for the inherited fields. Core and CLI regression tests cover both cases.

moncher-dev:
@codex review

moncher-dev (commented):
Functional verification passes. Blocking on one repo-rule gap only — do not redo the fix.

**Verified for real, not by reading the diff.** Built the branch at `1d8f386a` in a clean worktree (`pnpm install --frozen-lockfile && pnpm build`), then ran the built CLI against three real workflow files:

| workflow | `read` | `stall` | `turn` |
| --- | --- | --- | --- |
| `runtime:` block, no `timeouts:`, legacy `codex: 7000/60000/120000` | `7000 (source: codex/defaults)` | `60000 (codex/defaults)` | `120000 (codex/defaults)` |
| `runtime.timeouts.read_timeout_ms: 30000` + legacy `codex` stall/turn | `30000 (source: runtime.timeouts)` | `60000 (codex/defaults)` | `120000 (codex/defaults)` |
| full `runtime.timeouts` block + legacy `codex` | `30000 (runtime.timeouts)` | `900000 (runtime.timeouts)` | `1800000 (runtime.timeouts)` |

Row 1 is the exact repro I blocked on last cycle: it printed `read_timeout_ms=7000 (source: runtime.timeouts)` before, and now labels it `codex/defaults` correctly. Row 2 confirms mixed blocks no longer label every field identically. `--json` exposes `summary.runtimeTimeoutSources` per field and the misleading singular `runtimeTimeoutSource` is gone with no stale references left in `packages/`, `docs/`, or `README.md`.

I also checked the things a source-label fix could quietly get wrong:
- **The merge happens at parse time, not only in the resolver.** `workflow.runtime.timeouts` itself carries the merged values, so `packages/worker/src/non-codex-runtime.ts:235-237`, which reads `runtime.timeouts.*` directly rather than through `resolveWorkflowRuntimeTimeouts`, gets the same numbers `validate` prints. The claude-print path and the codex path agree.
- **Would a test fail if the fix were removed?** Yes. The new `partial runtime timeout precedence` case asserts per-field `(source: …)` strings and the JSON map, and reverting to the block-existence label fails it. The previously-wrong `expectedSource` that made the suite assert the misreport is corrected, and the old case was renamed rather than deleted.
- `pnpm test` (all packages), `pnpm typecheck`, and `pnpm lint` all pass on the branch.

**Blocking item: no changeset.** This PR changes shipped behavior, not just a label. A workflow with a `runtime:` block and no `timeouts:` alongside legacy `codex.*_timeout_ms` previously resolved to runtime defaults (`5000/300000/3600000`) and now inherits the legacy values — the test expectation moved from `[5000, 300000, 3600000]` to `[7000, 60000, 120000]`, and `packages/orchestrator/src/service.ts:3135` feeds that resolver straight into dispatch. Existing users' effective worker timeouts change on upgrade.

`CLAUDE.md` requires a changeset for behavior-changing PRs, and the convention is being kept — #936 and #945 both carry one; #946 and this PR's sibling #953 skip it correctly because they touch only policy and docs. This one ships source.

Add `.changeset/*.md` for `"@gh-symphony/cli"` (patch) naming the precedence change and the `summary.runtimeTimeoutSource` → `runtimeTimeoutSources` JSON rename. Nothing else needs to change.

moncher-dev:
Addressed the repository-rule gap in `e75b85a`: `.changeset/tidy-wolves-inherit.md` now declares an `@gh-symphony/cli` patch and names both the legacy timeout precedence fix and the `runtimeTimeoutSources` validation summary. No functional code was changed in this cycle. The full `pnpm lint`, `pnpm test`, `pnpm typecheck`, and `pnpm build` gate passes.

moncher-dev:
## Steward re-review — changeset present and correct, unblocking

My only blocker last cycle was the missing changeset for a behavior change that alters effective worker timeouts on upgrade (`[5000,300000,3600000]` → `[7000,60000,120000]`, consumed at `orchestrator/src/service.ts:3135`). That is now resolved by `e75b85a9`:

`.changeset/tidy-wolves-inherit.md`
```
"@gh-symphony/cli": patch
```
> Preserve legacy `codex` timeout declarations when a runtime block omits all or part of `runtime.timeouts`, and report each effective timeout's source through the `runtimeTimeoutSources` validation summary (#888).

Checks:
- Correct package — `@gh-symphony/cli` is the single publish unit.
- The note names both halves of the change: the fallback preservation *and* the per-field `runtimeTimeoutSources` reporting. It does not describe only the first commit.
- `patch` is right for a regression fix that restores previously-declared timeouts rather than introducing new configuration surface.
- `e75b85a9` adds the changeset file and nothing else (`1 file changed, 5 insertions(+)`), so the functional code is byte-identical to `1d8f386a` — the head I verified for real last cycle (built the CLI, ran it against three workflow files; the `runtime:`-without-`timeouts:` case printed `7000 (source: codex/defaults)`, mixed blocks labelled per field, and `worker/src/non-codex-runtime.ts:235-237` agreed with `validate`). I have not re-derived that verification and it does not need re-deriving.

No remaining blockers from me. Handing to @hojinzs for approval.

packages/cli/src/commands/workflow.ts
```diff
@@ -889,9 +893,11 @@ function validateWorkflow(
         stallTimeoutMs: effectiveTimeouts.stallTimeoutMs,
         turnTimeoutMs: effectiveTimeouts.turnTimeoutMs,
       },
-      runtimeTimeoutSource: workflow.runtime
-        ? "runtime.timeouts"
-        : "codex/defaults",
+      runtimeTimeoutSources: workflow.runtime?.timeoutSources ?? {
```
hojinzs, line 896: **P3 — `codex/defaults` fuses two sources the operator needs to tell apart.**

Per-field provenance is correct and I verified it against the built CLI; this is about the label's granularity, not its correctness. Two probe files produce the same label for different origins:

```
# runtime: block, no timeouts:, codex: 7777/88888/999999
read_timeout_ms=7777   (source: codex/defaults)   <- declared in codex:
# runtime: block, no timeouts:, no codex: block
read_timeout_ms=5000   (source: codex/defaults)   <- nothing declared anywhere
```

#888's framing is "distinguish *declared* from *defaulted*". The behavior half now does exactly that — `parseRuntimeConfig` tracks whether the field was present. The report collapses it back. The audience #888 names is the operator mid-migration asking "is my `codex:` block still live or is it dead weight?", and `codex/defaults` doesn't answer that; they'd have to know the defaults by heart (5000/300000/3600000) and compare.

`WorkflowRuntimeTimeoutSource` is already a union, so splitting it into `"runtime.timeouts" | "codex" | "defaults"` is a one-line type change plus a third branch in `parseRuntimeConfig` (the value is `null` in `timeouts` **and** absent from the `codex` node). Non-blocking — the acceptance criteria don't require it, the label is honest as far as it goes, and it matches the name used before this PR. Worth a follow-up issue rather than another cycle here.

---

.changeset/tidy-wolves-inherit.md
```diff
@@ -0,0 +1,5 @@
+---
+"@gh-symphony/cli": patch
+---
+
+Preserve legacy `codex` timeout declarations when a runtime block omits all or part of `runtime.timeouts`, and report each effective timeout's source through the `runtimeTimeoutSources` validation summary (#888).
```
hojinzs, line 5: **nit — the changeset names the new JSON key but not the removed one.**

Last cycle's blocking item asked for a changeset "naming the precedence change and the `summary.runtimeTimeoutSource` → `runtimeTimeoutSources` JSON rename". The precedence change is covered, and `runtimeTimeoutSources` is named, but the note reads as an addition when it is a rename: `summary.runtimeTimeoutSource` is gone from `--json` (I confirmed the key is absent, not aliased — consistent with `packages/cli/README.md`'s "no compatibility aliases are emitted" policy for the earlier `summary.codex.*TimeoutMs` removal).

Anyone scripting against `--json` reads `undefined` after this patch. One clause would cover it:

> …and replaces the singular `summary.runtimeTimeoutSource` JSON field with a per-field `summary.runtimeTimeoutSources` map.

Not worth holding the PR for, and fine to fold into the next changeset touching this area.

---

packages/core/src/workflow/config.ts
```diff
@@ -122,13 +122,24 @@ export type WorkflowRuntimeTimeoutsConfig = {
...
+};
+
 export type WorkflowRuntimeConfig = {
   kind: WorkflowRuntimeKind;
   command: string;
   args: readonly string[];
   isolation: WorkflowRuntimeIsolationConfig;
   auth: WorkflowRuntimeAuthConfig;
   timeouts: WorkflowRuntimeTimeoutsConfig;
+  timeoutSources: WorkflowRuntimeTimeoutSourcesConfig;
```
hojinzs, line 142: **nit — provenance metadata on the domain config type; checked for side effects, found none.**

`timeoutSources` is reporting metadata riding on `WorkflowRuntimeConfig`, the type the orchestrator and worker hand to runtime adapters. Carrying it on the parse result / diagnostics instead would keep the dispatch-facing config to values that actually affect execution. Recording it as a deliberate trade rather than asking for a change — the parser is where the presence/absence information exists, and threading a parallel structure through would cost more than it buys.

I chased the one place this could have bitten: `calculateWorkflowVersionHash` (`workflow/loader.ts:124`) hashes the whole `ParsedWorkflow` minus diagnostics, so the new field shifts every workflow's revision hash on upgrade — including workflows whose effective timeouts are unchanged (a fully explicit `runtime.timeouts` block). Traced the consumers: `createWorkflowRevision` → `service.ts:1935`, where it lands in the `run-dispatched` observability event and nothing else. No cache invalidation, no re-dispatch, no restart keyed on it. So the churn is a one-time cosmetic change in event logs, not operational impact. Flagging it only so a future reader who wires that hash into anything load-bearing knows this field is in it.

---

hojinzs (approved):
## Approve at `e75b85a` — cycle 3, blocking item cleared

Last cycle's only blocker was the missing changeset. `.changeset/tidy-wolves-inherit.md` (`patch`, `@gh-symphony/cli`) is present and names the precedence change. Nothing else was asked for, and nothing else changed — cycle 3 touches release metadata only.

### Smoke test — ran the built CLI, did not read the diff and assume

Built clean at the PR head (`pnpm install --frozen-lockfile && pnpm build`) and ran `workflow validate` against five probe workflow files, one per acceptance case in #888 plus the #882 regression:

| probe front matter | read | stall | turn |
| --- | --- | --- | --- |
| `runtime:`, no `timeouts:`, legacy `codex: 7777/88888/999999` | `7777 (codex/defaults)` | `88888 (codex/defaults)` | `999999 (codex/defaults)` |
| `runtime.timeouts.read_timeout_ms: 11111` + same legacy `codex:` | `11111 (runtime.timeouts)` | `88888 (codex/defaults)` | `999999 (codex/defaults)` |
| no `runtime:` block, legacy `codex:` only | `7777 (codex/defaults)` | `88888 (codex/defaults)` | `999999 (codex/defaults)` |
| `runtime:` block, nothing declared anywhere | `5000 (codex/defaults)` | `300000 (codex/defaults)` | `3600000 (codex/defaults)` |
| full `runtime.timeouts` + legacy `codex:` | `11111 (runtime.timeouts)` | `22222 (runtime.timeouts)` | `333333 (runtime.timeouts)` |

Row 1 is #888's exact repro — it resolved to `5000/300000/3600000` on `main` and now inherits the declared values. Row 5 confirms #882 behavior is preserved. Rows 3 and 4 cover the remaining two acceptance bullets.

**The report is not the only thing I checked.** #888 stresses this is a behavior bug, so I called `resolveWorkflowRuntimeTimeouts` directly — the resolver `service.ts:3135` feeds into dispatch — on all five parsed workflows. It returns exactly what `validate` prints in every case. The merge lands in `runtime.timeouts` at parse time, so `worker/src/non-codex-runtime.ts:234-236`, which reads `runtime.timeouts.*` without going through the resolver, gets the same numbers; the claude-print and codex paths agree.

`--json` exposes `summary.runtimeTimeoutSources` as a per-field map; the singular `runtimeTimeoutSource` is gone.

### Also verified

- `pnpm lint`, `pnpm typecheck`, `pnpm test` (13 packages, all suites) green at the head commit.
- **Would a test fail if the fix were reverted?** Yes. `partial runtime timeout precedence` in both suites pins per-field values *and* sources; restoring the `workflow.runtime ? … : …` label fails the CLI case. The previously-wrong `expectedSource` that asserted the misreport is corrected. This was the specific ask after #945 shipped unpinned fixes.
- The `codexConfig` hoist above `runtime` in `parseWorkflowConfig` introduces no use-before-declaration — nothing between the old and new positions reads `runtime`, and TS block-scope checking would reject it anyway.
- `readOptionalIntegerLike` returns `number | null`, so the `=== null` source discrimination is sound — no `undefined` path that would silently mislabel a field as `runtime.timeouts`.
- `parseRuntimeConfig` is the sole constructor of `WorkflowRuntimeConfig` in non-test code, so the new required field can't be silently omitted somewhere.
- Docs (`README.md`, `packages/cli/README.md`, `docs/configuration.md`) describe per-field inheritance and the sources map accurately against the observed output.

### Scope

No over-engineering found. Only `timeouts` inherit from `codex:` — `model`, `sandbox`, `approval_policy` are untouched, matching #888's scope. #888 floats a validation warning naming shadowed keys; it is not in Acceptance and correctly wasn't built.

The `docs/architecture.md` edit is the one bit of scope stretch: the Config Layer cell reword forced a whitespace reflow of all eight table rows, and no code moved across packages or layers here — which is what `CLAUDE.md` asks that map to track. Harmless, not worth a cycle.

### Non-blocking, left inline

Three comments, all P3/nit, none requiring a change before merge:
1. `codex/defaults` fuses "declared in `codex:`" with "defaulted" — the distinction #888 names, resolved in behavior but collapsed in the report.
2. The changeset names the new JSON key but not the removed `summary.runtimeTimeoutSource`.
3. `timeoutSources` rides on the dispatch-facing config type and lands in `calculateWorkflowVersionHash`; I traced it to an observability field only, no operational impact.

Item 1 is the only one worth a follow-up issue. Approving — this is cycle 3 on a fix that is verified correct end to end, and further rounds would buy label polish at the cost of a shipped bug fix.

---

PR #953, opened 2026-09-06
fix(workflow): detect steward issue comments

### LINKED ISSUE #947: bug(workflow): steward issue comments are unreachable for the rework guard
## Context

Split out of #900 at hojinzs's request on PR #946 ([review](https://github.com/hojinzs/github-symphony/pull/946#pullrequestreview-5124612731)), to keep #946 scoped to the review-body channel.

#900 fixed one of the two unreachable steward channels: a top-level `COMMENTED` review body now triggers rework even when its author login matches the worker account. The **issue-comment** channel still has the same defect.

## Problem

The Ready-return rework guard (`WORKFLOW.md`, "Ready-return rework guard", item 4) lists as a trigger:

> a human instruction indicating rework

The steward authenticates as `moncher-dev` — the worker's own account. An issue comment carrying findings from the steward is therefore not read as "a human instruction", and the clause never fires. This was demonstrated failing twice from `moncher-dev` (#900 comment [5557533795](https://github.com/hojinzs/github-symphony/issues/900#issuecomment-5557533795)); it worked once on 2026-09-05 only because **hojinzs** posted the comment.

Consequence: #900's issue body states that the operator's issue-comment workaround "Step 0 item 3 does read". That claim is wrong for the steward and remains wrong on the page after #946 merges.

## Scope

- Correct the guard clause so a steward issue comment is reachable, using the same discriminator #946 settled on for reviews (author `__typename`, not a login-string heuristic — GraphQL returns bot logins without the `[bot]` suffix REST shows).
- Carry over #946's blocking lesson: any clause added here needs an **actionability qualifier**, so a steward comment reporting a pass is not classified as rework. Both sibling conditions in that sentence carry one.
- Correct the stale claim in #900's issue body.

## Out of scope

- Steward review authentication / making the steward a repository reviewer.
- #878 approval re-entry.

moncher-dev:
## Issues

- Closes #947

## Summary

- Treat actionable issue comments from non-bot authors as Ready-return rework even when the steward shares the worker login.
- Bound both issue comments and top-level review bodies to the latest `In review` handoff, preventing stale or worker-authored pre-handoff comments from repeatedly reopening rework.
- Query issue-comment author `__typename` and preserve an actionability qualifier so bot boilerplate and passing reports remain non-actionable.

## Change-point diagram

- `issueContext` → actor `__typename` + `createdAt` → latest In-review handoff cutoff → actionable non-Bot steward comment → rework cycle

## Start here

- `WORKFLOW.md:167` — shared handoff boundary for review bodies and issue comments
- `WORKFLOW.md:168` — actor-type, actionability, and timestamp-qualified issue-comment signal
- `packages/cli/src/skills/templates/land.test.ts:173` — focused policy regression coverage

- A post-handoff steward issue comment from the shared worker account triggers rework when it requests changes or reports findings. Pre-handoff/stale comments, bot comments, and passing reports do not.

## Validation

- `pnpm gh-symphony workflow validate --file WORKFLOW.md` — pass
- `pnpm exec vitest run packages/cli/src/skills/templates/land.test.ts` — pass (7 tests)
- `pnpm exec prettier --check WORKFLOW.md packages/cli/src/skills/templates/land.test.ts` — pass
- `pnpm lint` — pass
- `pnpm test` — pass
- Typecheck/build — not applicable under the docs-only validation posture
- Docker E2E / blackbox — not applicable: repository workflow-policy text only; no runtime integration code changed

## Changeset

- Not needed because this is a workflow-policy documentation correction and the issue has no changeset label.

## Risks & rollback

- Risk is limited to natural-language actionability classification. Roll back commits `438ad6b` and `77436ec` to restore the prior guard.

## Changed files

- `WORKFLOW.md` — makes actionable steward issue comments reachable, filters by actor type, and applies the handoff cutoff
- `packages/cli/src/skills/templates/land.test.ts` — asserts actor type, shared-login, actionability, and boundary semantics

- [ ] None required beyond normal review of the policy wording.

## Security


moncher-dev (commented):
Blocking. Verified against the branch at `77436ecb`; the Codex P1 at WORKFLOW.md:168 is correct and I can restate it precisely.

**The issue-comment qualifier has no handoff boundary.**

Item 3 defines the boundary and scopes it explicitly: "its comment `createdAt` is the handoff boundary **for top-level review bodies**." Item 4 then adds `a recent issue comment from a non-`Bot` author that requests changes or reports findings` with no `createdAt` comparison at all. "Recent" resolves only to `comments(last:50)` from `issueContext` — a window of 50 comments, not a window in time.

The consequence is not hypothetical. This is the exact defect PR #946 had to fix for review bodies one commit ago (merged as `9a440201`), and this PR reintroduces it on the sibling signal:

1. Steward posts an actionable issue comment on #N. Worker reworks, pushes, moves to `In review`. Fine.
2. Anything later returns #N to `Ready` — a resume after a worker crash, board-drift repair, a fresh follow-up cycle. The same steward comment is still within the last 50.
3. It re-qualifies. The guard opens a *new* work cycle against feedback already addressed, and step 3's boundary does not stop it because the boundary does not apply to this signal.
4. Repeat at every subsequent `Ready` entry. There is no state that ever retires the comment.

**A second failure mode the review-body qualifier never had.** Review bodies only exist on PRs. Issue comments include the worker's *own* output: `## Workpad` comments and status-transition comments, all posted by `moncher-dev`, a non-`Bot` author. The PR then states outright that a matching worker login does not disqualify a comment. So a workpad that records a blocker or a failed Completion Bar — i.e. one that "reports findings" — can qualify the worker's own writing as rework feedback against itself. The `__typename` check filters bots; it does not filter the worker.

**Fix:** bound the issue-comment qualifier to the same handoff boundary as review bodies. Widen item 3 from "for top-level review bodies" to cover both signals, and in item 4 add `and was created after the handoff boundary defined in item 3` to the issue-comment clause, mirroring the review-body clause verbatim.

**Test gap.** `land.test.ts` asserts only that three sentences are present in `WORKFLOW.md`. No test distinguishes a comment before the boundary from one after it, so the boundary could be added, omitted, or later deleted with the suite green either way. Add a guard that asserts the issue-comment clause carries the boundary phrase.

Everything else in the PR is sound: the `__typename` addition to `issueContext` is valid on the `Actor` interface, and the `APPROVED`/`DISMISSED` carve-out is preserved.

WORKFLOW.md
```diff
@@ -164,8 +164,8 @@ When entering `Ready`, before treating it as a fresh pickup, board drift, or res
...
...return rework**, this `Ready` state means **review rework return** — not a fresh pickup and not drift. A qualifying top-level `COMMENTED` review body or issue comment triggers rework even when its author login matches the worker account; this covers steward feedback posted through a shared worker account. A qualifying review body does not require an inline thread. The review-body qualifier does so without treating automated review boilerplate or passing review bodies as actionable. The issue-comment qualifier likewise operates without treating automated boilerplate or passing reports as actionable. Only `COMMENTED` review bodies qualify through this condition; `APPROVED` and `DISMISSED` bodies retain their separate review-state semantics. When a latest qualifying human approval exists on the current head, narrow the thread condition to unresolved actionable threads created after that approval (compare the thread's first comment `createdAt` with the approval review's `submittedAt`); a thread created at or before that approval is absorbed by the approval and does not trigger rework. When no qualifying approval exists, any unresolved actionable review thread triggers rework as before.
```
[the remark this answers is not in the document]
moncher-dev: Resolved in 438ad6b. Item 3 now defines the latest `In review` status comment as the handoff boundary for both top-level review bodies and issue comments; item 4 requires an actionable non-`Bot` issue comment's `createdAt` to be after that boundary. The focused policy test now asserts both the widened boundary and the issue-comment clause's explicit boundary phrase.

moncher-dev:
## Steward re-review — all three blocking findings resolved, mutation-verified

Verified at head `438ad6b3` in a clean worktree (`pnpm install --frozen-lockfile && pnpm build && npx vitest run`): **133 files, 1936 tests, all passing.**

### 1. Issue-comment handoff boundary — fixed
Item 3 now extends the boundary to issue comments ("handoff boundary for top-level review bodies **and issue comments**") and closes the no-boundary case ("neither a top-level review body **nor an issue comment** qualifies"). Item 4's issue-comment qualifier now carries the explicit `createdAt` comparison ("created after the handoff boundary defined in item 3"). This retires the defect I blocked on: an addressed steward comment no longer re-qualifies at every later `Ready` entry, because each `In review` entry resets the boundary. Same shape as the fix #946 applied to review bodies.

### 2. Bot filtering — `__typename` now requested for issue comments
The `issueContext` query was changed to `comments(last:50){nodes{id body author{login __typename} createdAt}}`, so the non-`Bot` qualifier in item 4 has the field it needs. Previously the qualifier named `__typename` for review authors only and the issue-comment path had no way to evaluate it.

### 3. Tests now bind the boundary — I mutated it to confirm
This was my concern that the boundary could be deleted with the suite green. I applied three separate mutations to `WORKFLOW.md` and re-ran `packages/cli/src/skills/templates/land.test.ts` each time:

| Mutation | Result |
| --- | --- |
| Removed `and was created after the handoff boundary defined in item 3` from item 4's issue-comment qualifier | **FAIL** |
| Reverted item 3 to `handoff boundary for top-level review bodies` (dropping issue comments) | **FAIL** |
| Reverted the query to `author{login}` (dropping `__typename`) | **FAIL** |

All three are caught by `treats an actionable steward issue comment as Ready-return rework`. The boundary can no longer be removed silently.

### Worker self-trigger — bounded, no longer a blocker
My other finding was that issue comments include the worker's own `## Workpad` and status comments posted as `moncher-dev`, and the rule deliberately does not disqualify a matching worker login (correctly so — the steward posts through that same account). The boundary now bounds this: the workpad is created before the `In review` transition that sets the boundary, so it does not qualify on the next `Ready` entry. One narrow window remains — a worker that dies between creating a new workpad and reaching `In review` leaves that workpad after the *previous* boundary, so a recovery `Ready` entry could classify a resume as rework. The outcome there is a new cycle instead of a resume: a degradation, not corruption, and unbounded no longer. Not blocking; worth a follow-up issue if it is ever observed.

### Observation, not blocking
`packages/cli/src/skills/templates/land.ts:90` and `.codex/skills/land/SKILL.md:180` classify a Land-time **Rework failure** from CI, conflicts, changesets and unresolved review threads — they do not consider issue comments. That path is guarded by the Land pre-flight approval requirement, and the steward's block path routes through `Ready` rather than `Land`, so I do not think it is reachable today. Noting it as the other place this pattern lives.

Merge check: `git merge-tree pr945 pr953` is clean, so this does not conflict with #945 despite both touching `WORKFLOW.md` and `land.test.ts`.

Handing to @hojinzs for approval.

WORKFLOW.md
```diff
@@ -164,8 +164,8 @@ When entering `Ready`, before treating it as a fresh pickup, board drift, or res
...
```
hojinzs, line 168: **P2 — the issue-comment signal captures the worker's own `⛔ Blocker` comment on the documented `Backlog` → `Ready` recovery path**

The steward's re-review assessed the worker-self-trigger risk and bounded it to a narrow crash window ("a worker that dies between creating a new workpad and reaching `In review`"). That under-scopes it. There is a second instance on a path this document defines as the *standard* recovery route, so it recurs systematically rather than rarely:

1. A cycle reaches `In review` — that status comment is now the boundary.
2. A later `Ready` entry is classified fresh pickup / resume (board drift, for example) → `In progress`.
3. The worker hits a code-blocker. Per Posture 3 (`WORKFLOW.md:115`) it posts a `⛔ Blocker` issue comment — "what · why · how to unblock" — and moves to `Backlog`.
4. Per the Backlog lane definition (`WORKFLOW.md:59`): "human resolves and moves back to `Ready`". This is the documented unblock route, not an edge case.
5. The guard re-runs. The `⛔ Blocker` comment is authored by a non-`Bot` (the worker itself), was created **after** the boundary — the boundary only advances at the next `In review`, which never happened — and it reports findings. It qualifies.

The entry is classified **review rework return**. The worker opens a new cycle and runs the Step 2.2 rework preamble looking for review feedback that does not exist: the same "one dispatch spent producing nothing" failure #900 documented, arriving from the other direction.

Neither existing filter helps. `__typename` cannot discriminate — the worker is not a Bot — and item 4 deliberately declines to filter on login, correctly, since the steward shares that account. That leaves comment *shape* as the only available discriminator, and the policy currently gives a reading agent no deterministic answer as to whether a `⛔ Blocker` comment "reports findings". That ambiguity is the same class #900 and #946 were opened to remove.

**Suggested fix** — one clause on this line, which also subsumes the workpad case the steward flagged for follow-up:

> a recent issue comment from a non-`Bot` author **that is not one of the worker's own `## Workpad`, `🔁 Status:`, or `⛔ Blocker` comments** and that requests changes or reports findings and was created after the handoff boundary defined in item 3

Plus one assertion in `land.test.ts` so the exclusion is bound the way the boundary now is.

Not blocking the approval — the misclassification costs an extra cycle and a wasted rework preamble, it does not corrupt state, and this PR is a clear net improvement over a feedback channel that was entirely unreachable. But it is one clause and one assertion, so it is worth landing here rather than deferring. If you would rather keep this PR at its current scope, a follow-up issue covering both this and the workpad window would be the right home.

---

hojinzs, line 168: **P3 — dangling "does so" left behind by the sentence split**

Reading the three sentences in sequence:

> A qualifying review body does not require an inline thread. **The review-body qualifier does so without** treating automated review boilerplate or passing review bodies as actionable.

Before this PR these were one clause, and "without treating…" attached to "this covers steward feedback posted through a shared worker account". Splitting them stranded the anaphor: "does so" now has "does not require an inline thread" as its nearest antecedent, which makes the sentence mean nothing.

This file is a prompt an agent reads as normative policy, so an unresolvable referent in the qualifier sentence is worth a line. Merging the two restores it:

> A qualifying review body does not require an inline thread, and qualifies without treating automated review boilerplate or passing review bodies as actionable.

The sibling issue-comment sentence that follows is fine — "likewise operates" carries its own verb.

---

hojinzs (approved):
## Approve — requirements met, boundary independently mutation-verified

Reviewed at head `438ad6b3`. I re-ran the verification myself rather than taking the steward's re-review at face value.

### Requirement coverage against #947

All three scope items are satisfied:

| #947 scope item | Status |
| --- | --- |
| Guard clause reachable for steward issue comments, via author `__typename` rather than a login-string heuristic | ✅ item 4 + `issueContext` query now selects `author{login __typename}` |
| Actionability qualifier so a steward pass report is not classified as rework | ✅ "requests changes or reports findings", mirrored for both channels |
| Correct the stale claim in #900's issue body | ✅ already corrected (#900 updated 2026-09-06); the body now points at #947 instead of asserting the workaround reads |

Out-of-scope items (steward review authentication, #878 approval re-entry) stayed out. No overengineering: +29/-4 across two files, and the handoff boundary added in `438ad6b` is a correctness necessity raised in review, not scope creep — it mirrors what #946 settled for review bodies one commit ago.

### Prior-round findings

The single review thread (Codex P1 at `WORKFLOW.md:168`, restated as blocking by the steward) is genuinely resolved, not just claimed resolved. Item 3 widens the boundary to both channels and closes the no-boundary case; item 4's issue-comment qualifier carries the explicit `createdAt` comparison.

### Verification performed

Clean worktree at `438ad6b3`, `pnpm install --frozen-lockfile && pnpm build`:

- `npx vitest run` — **133 files, 1936 tests, all passing**
- `node packages/cli/dist/index.js workflow validate --file WORKFLOW.md` — `WORKFLOW.md validation passed`, exit 0
- `npx prettier --check WORKFLOW.md packages/cli/src/skills/templates/land.test.ts` — clean
- `pnpm lint` — clean

**Mutation testing — reproduced independently.** The steward's central claim was that the new test binds the boundary so it cannot be silently deleted. I applied the three mutations to `WORKFLOW.md` myself and re-ran `land.test.ts` after each, restoring between runs:

| Mutation | Result |
| --- | --- |
| Drop `and was created after the handoff boundary defined in item 3` from item 4's issue-comment qualifier | **1 failed / 6 passed** ✅ caught |
| Revert item 3 to `handoff boundary for top-level review bodies` | **1 failed / 6 passed** ✅ caught |
| Revert the query to `author{login}` | **1 failed / 6 passed** ✅ caught |

All three are caught by `treats an actionable steward issue comment as Ready-return rework`. The claim holds.

I also checked that the Ready-return rework guard is not duplicated into a generated artifact — `packages/cli/src/skills/templates/land.ts` and `.codex/skills/land/SKILL.md` only reference the guard by name from the Land-time *Rework failure* path, so there is no copy left out of sync by this change.

### Findings

Two inline comments, neither blocking:

- **P2** at `WORKFLOW.md:168` — the issue-comment signal also captures the worker's own `⛔ Blocker` comment on the `Backlog` → `Ready` unblock route, which this document defines as the standard recovery path. That is broader than the narrow crash window the steward's re-review scoped it to, so it recurs systematically. One clause fixes it and subsumes the workpad case already flagged for follow-up.
- **P3** at `WORKFLOW.md:168` — the sentence split left "The review-body qualifier does so without…" with no valid antecedent.

Approving because the PR does what #947 asked, the regression risk it introduces is bounded degradation (an extra cycle, no state corruption) and is strictly smaller than the fully-unreachable feedback channel it fixes. Landing the P2 clause here would be my preference since it is one line plus one assertion; a follow-up issue covering it together with the workpad window is an acceptable alternative.

---

PR #955, opened 2026-09-08
fix(cli): detect stale project workflow copies

### LINKED ISSUE #952: fix(cli): a stale project-dir WORKFLOW.md copy is invisible to status and doctor
## Problem

In standalone project mode the project directory holds its own `WORKFLOW.md`. When that file is a **copy** of the repository's `WORKFLOW.md` rather than a link to it, every merged `WORKFLOW.md` change is silently absent from the running daemon, and nothing in `project status`, `doctor`, or the status snapshot says so.

The daemon reports the workflow it loaded as healthy in every case:

```json
"workflow": { "revision": "sha256:e23770e692cf", "isValid": true, "usedLastKnownGood": false }
```

`isValid: true` means "this file parses", not "this file is the policy that was merged". There is no signal that distinguishes a current policy from a stale one.

## Observed

Today at 08:21:46Z, #946 (`fix(workflow): detect actionable body-only reviews`, closing #900) merged a `WORKFLOW.md` change. The daemon for this repository kept serving `sha256:e23770e692cf`, which corresponds to the `WORKFLOW.md` at 66eaf9b5 (#915) — the state immediately *before* #946. Workers dispatched in that window ran the old rework guard. Health stayed `idle` / `running` with zero warnings the entire time.

The project directory's `WORKFLOW.md` was a regular file, not the symlink the deployment is documented to use, so it had to be refreshed by hand after each merge and this time was not. Restoring the symlink fixed it and the daemon reloaded to `sha256:9ced963bad2f` on the next tick. But the failure was silent, and a copy is a supported layout — a user who intends a copy gets no warning either.

## Impact

A merged policy fix can be inert for an unbounded period with no operator-visible signal. The failure is invisible precisely when it matters most: right after a policy fix lands, which is when the next worker most needs it. This is the same defect shape as the daemon-vs-installed-binary version skew already recorded in `CLAUDE.md` — a running component silently older than `main`.

## Acceptance criteria

1. `project status` (and the status snapshot consumed by the dashboard/control plane) surfaces the loaded workflow's content hash alongside enough identity to compare it against a source of truth.
2. `doctor` warns when the project directory's `WORKFLOW.md` resolves to a regular file whose content differs from the configured repository's committed `WORKFLOW.md` at the current base branch, naming both revisions.
3. The warning distinguishes "diverged copy" from "deliberately independent policy" — a copy that never matched any repository revision should not warn on every tick.
4. Tests cover: symlink layout (no warning), copy identical to the repository (no warning), copy stale relative to the repository (warning), and copy with no relationship to the repository (no warning, or a distinct informational note).

## Notes

Do not make the check fetch from the network on every tick; compare against the local checkout the project is configured against.

Filed by the board steward after the incident above. The symlink is restored on the affected host; this issue is about detection, not that one host.

moncher-dev:
## Issues

- Closes #952

## Summary

- Project status and doctor expose the loaded workflow hash alongside committed local repository identity and warn on historically related stale copies.
- Repositories that intentionally have no committed `WORKFLOW.md` now report `no-repository-policy` without a permanent warning.
- Local checkout selection skips unusable persisted paths; comparisons normalize CRLF/LF and never fetch from the network.

## Change-point diagram

- project `WORKFLOW.md` + configured repository + persisted workspace/run state → shared validated checkout resolver → Git identity inspector → status snapshot and doctor
- readable base ref → committed policy present → linked / synchronized-copy / stale-copy / independent
- readable base ref → committed policy absent → no-repository-policy (informational)
- unreadable checkout/ref → unavailable (warning)

## Start here

- `packages/orchestrator/src/workflow-source-identity.ts:111` — resolves source and repository identities, including the no-policy boundary
- `packages/orchestrator/src/service.ts:2820` — attaches identity and warnings to status snapshots
- `packages/cli/src/commands/doctor.ts:2319` — emits relationship-aware doctor diagnostics

- `project status`, JSON snapshots, and the dashboard show loaded and committed workflow identities.
- Stale historical copies warn with both revisions; symlinks, synchronized copies, never-related independent policies, and repositories with no committed policy remain distinct non-warning states.
- The broad legacy “external workflow shadows repository WORKFLOW.md” warning is removed because supported synchronized and deliberately independent policies should not warn; no known consumer relies on that prose string.
- CRLF working copies compare consistently with LF committed content.
- HTTPS clone configurations reuse usable local project checkouts and comparisons never fetch during a tick.

## Validation

- `pnpm lint` — pass
- `pnpm test` — pass (full workspace rerun; the first run hit a timing-sensitive untouched hook test, which passed in isolation and on rerun)
- `pnpm typecheck` — pass
- `pnpm build` — pass
- focused workflow identity, doctor, and service regressions — pass
- mutation check disabling the no-policy split — expected fail; restored implementation — pass
- `./e2e/run-standalone-project-e2e.sh` — documented environment exception #956; two attempts stopped at Docker preflight because `/var/run/docker.sock` is unreachable

## Changeset

- `.changeset/tame-workflows-identify.md` (`patch`, `@gh-symphony/cli`)

## Risks & rollback

- Identity reflects locally available base-ref state by design; it never fetches during status or doctor.
- Independent policies still scan reachable local policy history; this remains bounded but could be optimized separately.
- Revert the issue commits to restore the prior snapshot and doctor behavior.

## Changed files

- `packages/orchestrator/src/workflow-source-identity.ts` and tests — repository selection, normalized comparison, and relationship classification
- `packages/orchestrator/src/service.ts` and tests — status identity and warning behavior
- `packages/core/src/observability/` and contracts — snapshot identity fields
- `packages/cli/src/commands/`, dashboard renderer, and tests — operator-visible status and doctor output
- `.changeset/tame-workflows-identify.md` — CLI patch release note
- `e2e/` and living documentation — TC-25 and operational contract

- [ ] Run TC-25 on a host with a reachable Docker daemon (tracked by #956 for this worker environment).

## Security


moncher-dev:
## Steward review — blocking: the check cannot fire on the deployment that produced the incident

Verified at head `69e9851a` in a clean worktree: `pnpm install --frozen-lockfile && pnpm build && npx vitest run` → **134 files, 1952 tests, all passing.** The detector itself is good work, and I exercised it for real rather than trusting the unit tests. But it is wired behind a guard that is false on the live standalone project, so the incident it is meant to catch would still go undetected.

### What I ran

I built five fixtures against the **real** repository and its **real** git history and called the built `inspectWorkflowSourceIdentity` from `packages/orchestrator/dist`:

| Fixture | Result | Warns? |
| --- | --- | --- |
| A — regular-file copy of `WORKFLOW.md` at `66eaf9b5`, i.e. the exact copy that went stale when #946 merged | `stale-copy`, matched `66eaf9b5` | **YES** |
| B — symlink to the repository `WORKFLOW.md` (the supported layout) | `linked` | no |
| C — regular-file copy identical to `HEAD` | `synchronized-copy` | no |
| D — `HEAD` + a local team rule (a legitimately customized copy) | `independent` | no |
| E — no file at all | `unavailable` | no |
| G — fixture A with `base_branch: main` (the value this repo actually declares) | `stale-copy`, matched `66eaf9b5` | **YES** |

So the classification boundary you were asked for holds: a diverged copy warns and names the commit it is stuck on, while a customized copy is `independent` and neither warns nor fails. Every relationship also has a named destination on both surfaces (`status.ts:165-175`, `doctor.ts:2289-2338`). I mutation-tested the fix as well — flipping `stale-copy` to `independent` in the detector, and neutering the warning in `resolveWorkflowWarnings`, each fail a test. The fix is bound.

### The blocker

`resolveWorkflowSourceIdentity` (`service.ts:2815-2830`) returns `null` unless `resolveLocalRepositoryDirectory` yields a path, and that is

```
service.ts:2844  repository.path ?? this.resolveLocalCloneUrlPath(repository.cloneUrl)
service.ts:5862  new URL(cloneUrl).protocol === "file:" ? fileURLToPath(url) : null
```

The live standalone project declares no `repository.path` and an `https:` clone URL:

```
repository.path    : (absent)
cloneUrl           : https://github.com/hojinzs/github-symphony.git
localRepositoryDir : null
=> identity computed: NO  (no source in the snapshot, no warning, ever)
```

That is the configuration the incident happened in. `doctor.ts:2292-2299` reproduces the same guard independently, so both surfaces skip the check. The result is exactly the symptom this PR exists to remove: `isValid: true`, zero warnings, while the project runs stale policy.

This is invisible to the suite because every test supplies a local repository directory — `workflow-source-identity.test.ts` passes `repositoryDirectory` directly, the new service test uses `repository.path` from a fixture, and `e2e/scenarios/25-workflow-copy-identity.md` seeds a local repository. There is no case with an `https:` clone URL, which is the only shape standalone projects have in practice.

A repository directory **is** available in standalone mode — the orchestrator's own issue workspaces are full checkouts. I confirmed one on the live host resolves `main` and carries all 38 `WORKFLOW.md` revisions. So this is reachable, not a missing capability.

### Second, smaller: the check resolves to silence when it cannot run

`unavailable` covers three unrelated causes — the project file is missing or unreadable, the directory is not a git repository, and the declared `base_branch` cannot be resolved locally. I reproduced the third: fixture A with `baseRef: refs/heads/no-such-branch` returns `unavailable` and **no warning**, even though the copy is provably stale. Worse, `doctor.ts:2323` reports `unavailable` through `passCheck` — a check that could not run is reported as passing. Please make "could not determine" a warn (or a distinct skipped status) rather than a pass. A green surface hiding an unverified condition is the failure mode this issue is about.

### Non-blocking notes

- For a diverged copy the detector walks `rev-list` and spawns a `git show` per revision — 38 today, and this runs on every tick's snapshot build. Fixture D measured 323 ms versus 1-10 ms for the linked and synchronized paths. The live symlink layout stays cheap, so this is not urgent, but consider stopping at the first match from the newest commit backwards, or bounding the scan.
- The old `External workflow source … shadows repository WORKFLOW.md` warning is fully retired, including its test. Removing it is right — it fired on the supported symlink layout, since `resolve()` does not follow symlinks — and the only remaining reference is a historical `CHANGELOG.md` entry, which is correct to leave. Calling it out so the deletion is on the record.

### Ask

Reach the check on an `https:`-clone standalone project — the shape the incident occurred in — and add a test at that shape so the guard cannot silently regress. Then make the unrunnable case visible rather than passing. Moving #952 back to `Ready`.

packages/orchestrator/src/workflow-source-identity.ts
```diff
@@ -0,0 +1,150 @@
...
+        realpath(input.workflowPath),
+        realpath(repositoryWorkflowPath),
+      ]);
+      if (sourceTarget === repositoryTarget) {
+        const revision = contentRevision(sourceContent);
+        return {
+          ...unavailable(revision),
+          relationship: "linked",
+          repositoryRevision: revision,
+        };
```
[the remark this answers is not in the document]
moncher-dev: Addressed in a412ba0. Linked workflows now retain the `linked` relationship but continue through `git show <base-ref>:WORKFLOW.md` and `rev-parse`, so status reports the committed repository revision and commit even when the symlink target has uncommitted edits. Added a dirty-symlink regression test.

packages/orchestrator/src/workflow-source-identity.ts
```diff
@@ -0,0 +1,150 @@
...
+        matchedRepositoryCommit: repositoryCommit.trim(),
+      };
+    }
+
+    const revisions = (
+      await execGit(input.repositoryDirectory, [
+        "rev-list",
+        repositoryRef,
+        "--",
+        "WORKFLOW.md",
```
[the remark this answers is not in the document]
moncher-dev: Addressed in a412ba0. The path-limited history query now uses `git rev-list --full-history`, with a regression assertion covering the exact Git invocation so reachable side-branch policy revisions are not pruned.

packages/cli/src/commands/doctor.ts
```diff
@@ -2283,6 +2286,58 @@ export async function runDoctorDiagnostics(
...
+    resolvedProjectConfig.kind === "resolved"
+  ) {
+    const repository = resolvedProjectConfig.projectConfig.repository;
+    const repositoryDirectory =
+      repository?.path ??
+      (repository && isAbsolute(repository.cloneUrl)
+        ? repository.cloneUrl
+        : repository?.cloneUrl.startsWith("file:")
+          ? fileURLToPath(repository.cloneUrl)
+          : null);
```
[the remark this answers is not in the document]
moncher-dev: Addressed in a412ba0. Doctor now resolves dot-prefixed clone paths against its captured repository root, matching the orchestrator's supported local-path shape. The stale-copy diagnostic test now uses `cloneUrl: "."`, and also verifies that an unresolvable base ref is a warning rather than a pass.

moncher-dev (commented):
Re-reviewed at `a412ba0c`. The wiring gap I blocked on last time is genuinely fixed on the orchestrator side, and I verified it the way I said I would — against the live project shape, not a fixture.

**What now works.** `resolveWorkflowSourceIdentity` falls back from `resolveLocalRepositoryDirectory` to the most recently updated issue workspace's `repositoryPath`, then to a run's `workingDirectory` (`service.ts:2825`). On this host that resolves to a real worktree, and the classifier returns a correct answer:

```
repositoryDirectory = .../symphony-workspaces/hojinzs_github-symphony_948-.../repository
-> relationship "synchronized-copy", repositoryCommit 71479b17, no warning
```

The new test `detects a stale copy from an HTTPS-clone project's local issue workspace` uses the real shape (`cloneUrl: "https://…"`, no `path`). I confirmed it is load-bearing: deleting the two fallback clauses and re-running `packages/orchestrator/src/service.test.ts` fails exactly that test (1 failed | 222 passed). The three Codex P2 findings are addressed as described.

**Blocking: `doctor` does not use that resolution, and falls back to `process.cwd()`.**

`doctor.ts:2295-2303` ends its chain with `: repoRoot`, and `repoRoot = process.cwd()` (`doctor.ts:1845`). For any remote clone URL — which is this deployment — doctor does not compare against the configured repository at all. It compares against whatever directory the operator is standing in. Two observed failures, both with the real project dir:

*1. False warning in the documented deployment.* Run from the project directory, as the runbook says:

```
$ cd ~/moncher-stack/github-symphony && gh-symphony doctor --json --project-dir ~/moncher-stack/github-symphony
workflow_source_identity: WARN (required)
"Project WORKFLOW.md source identity could not be determined (sha256:587555c199f3); committed main was not readable."
relationship: "unavailable", repositoryCommit: null
```

`WORKFLOW.md` there is a symlink to the repository policy and is exactly current. The orchestrator says `synchronized-copy` at `71479b17`; doctor says it cannot tell. The two halves disagree about the same file on the same host in the same minute. A required check that is permanently wrong on the canonical deployment trains operators to ignore it, which is the same end state as the silence this issue was filed about.

*2. Worse — a confident PASS against an unrelated repository.* I made a throwaway git repo with a one-line `WORKFLOW.md` and ran doctor from inside it, still pointing at the live project:

```
$ mkdir fakerepo && cd fakerepo && git init -b main && echo 'totally unrelated policy' > WORKFLOW.md && git commit -am init
$ gh-symphony doctor --json --project-dir ~/moncher-stack/github-symphony
workflow_source_identity: PASS (required)
"Project WORKFLOW.md is an independent policy (sha256:587555c199f3); it does not match reachable repository WORKFLOW.md history."
repositoryPath: ".../scratchpad/fakerepo/WORKFLOW.md"
repositoryCommit: d0576b26
```

Doctor compared the live policy against a scratch directory created seconds earlier and returned green. The verdict is a function of the operator's cwd, not of the project. AC 3 asks the check to distinguish a diverged copy from a deliberately independent policy; `independent` is being reported here on no evidence, and reported as a pass.

**What I'd expect instead.** Doctor should resolve the repository directory the same way the orchestrator now does — the project's own state store, via issue workspace `repositoryPath` / run `workingDirectory` — rather than `process.cwd()`. Where no configured repository can be identified, the honest outcome is a check that says so and names why, not a verdict computed against an arbitrary directory; `independent` in particular should not be a pass unless the compared directory is actually the configured repository.

Two smaller notes, non-blocking:

- The two resolution chains are now written twice and have drifted (the orchestrator gained the workspace/run fallback; doctor gained dot-prefixed-path handling and cwd). Sharing one resolver would keep them from disagreeing again.
- The orchestrator fallback picks whichever workspace was updated most recently, and reads `main` from that worker worktree. If the shared clone cache has not fetched, that `main` can lag `origin/main`, so a project copy matching a stale `main` classifies as `synchronized-copy`. Local-only comparison is the issue's explicit constraint, so this is a documentation point rather than a defect — but the revision reported to the operator is "main as this worktree last saw it", and the status output does not say so.

Verified on this host at `cfa88cde`: `pnpm install --frozen-lockfile` and `pnpm build` pass; `packages/orchestrator/src/service.test.ts` passes 223/223 unmodified.

moncher-dev:
Addressed in `de67d3a`. Doctor now loads project-scoped persisted issue workspaces and runs, then uses the same exported repository resolver as the orchestrator. HTTPS clone URLs therefore resolve to the newest local project checkout; the operator cwd is no longer accepted as repository evidence and cannot produce a false green `independent` result.

The new doctor regression creates a stale project workflow, registers its repository under persisted project workspace state, runs doctor from an unrelated directory, and requires a `stale-copy` warning pointing at the persisted checkout. Resolver unit coverage separately requires remote repositories with no persisted evidence to return `null`. Full lint, unit tests, typecheck, and build pass; Docker preflight remains unavailable under #956.

moncher-dev (commented):
## Blocking — the detector still cannot fire in this deployment. The cause moved from `process.cwd()` to unfiltered workspace records.

I re-ran both probes from the last pass against `de67d3aa`.

### Probe B — fixed

From a throwaway `git init` scratch repo, the check no longer fabricates a confident `independent policy` verdict from a directory created seconds earlier. It is simply not emitted. That was the worse of the two findings, and it is properly addressed. `resolveConfiguredRepositoryPath`'s `process.cwd()` default is also no longer reachable from `doctor`, which now passes `baseDirectory: repoRoot`.

### Probe A — still wrong, and now points at a directory that does not exist

From the documented project directory, with `WORKFLOW.md` a correct current symlink to `/Users/steve/Projects/github-symphony/WORKFLOW.md`:

```
WARN Project workflow source identity
  Project WORKFLOW.md source identity could not be determined (revision unavailable); committed main was not readable.
  Fix: Ensure .../symphony-workspaces/hojinzs_github-symphony_888-cc2478b4bbb34ee7/repository/WORKFLOW.md
       is a local Git checkout with main and a committed WORKFLOW.md.
```

`#888` merged earlier today and its workspace was torn down. That path **does not exist**:

```
$ ls .../hojinzs_github-symphony_888-cc2478b4bbb34ee7/repository/WORKFLOW.md
ls: No such file or directory

$ ls .../symphony-workspaces/
hojinzs_github-symphony_844-...   hojinzs_github-symphony_948-...
hojinzs_github-symphony_952-...   hojinzs_github-symphony_954-...
```

Four live checkouts are present. The selector picked the one dead record ahead of all of them, and told the operator to go repair a deleted directory.

### Mechanism

`resolveWorkflowRepositoryDirectory` (`workflow-source-identity.ts:49-52`):

```ts
const latestWorkspace = [...(input.issueWorkspaces ?? [])]
  .sort((left, right) => right.updatedAt.localeCompare(left.updatedAt))
  .find((workspace) => workspace.repositoryPath);
if (latestWorkspace) return latestWorkspace.repositoryPath;
```

No `status` filter, and no existence check on the path it returns. A workspace record survives teardown with `status: "removed"` and its `repositoryPath` intact — and because teardown *updates* `updatedAt`, the workspace that was just deleted reliably sorts **first**. The selector is biased toward the one candidate guaranteed to be gone.

This is not an edge case on this host. From live `status.json`: **37 of 41** workspace records are `removed`. The selector treats all 37 as candidates.

### Consequence: the stale-copy branch is unreachable here

I ran `inspectWorkflowSourceIdentity` directly against both directories, same `workflowPath`, same `baseRef`:

| repositoryDirectory | relationship | contentRevision | repositoryRevision |
|---|---|---|---|
| `_888_` (what doctor selects) | `unavailable` | `null` | `null` |
| `_954_` (a live workspace) | `synchronized-copy` | `sha256:03bfa1095860` | `sha256:03bfa1095860` |

The detector itself is correct — given a real checkout it produces a real verdict. But in this deployment the selector hands it a deleted path, so **every** invocation returns `unavailable`, and the `stale-copy` branch — the entire reason #952 exists — can never be reached. That is the same block as last time: the verdict is a function of incidental host state, only the state variable changed from the operator's cwd to whichever workspace was most recently destroyed.

The advisory severity (doctor still exits 0) makes this worse rather than better: a detector that cries `unavailable` on the healthy case teaches operators to ignore the one line that will eventually say `stale-copy`.

### Why no test caught it

`workflow-source-identity.test.ts:160,173` — the only workspace fixtures in the suite are `status: "active"`. There is no fixture for a `removed` record, none for a `repositoryPath` that no longer exists, and none asserting which of several candidates wins. Deleting the `status`/existence handling cannot turn anything red, because that handling was never written.

### Fix direction

1. Filter candidates to live workspaces (`status !== "removed"`) before sorting.
2. Verify the selected `repositoryPath` still exists and is a Git checkout; if it does not, fall through to the next candidate rather than reporting `unavailable`.
3. Consider resolving the symlink target of the project `WORKFLOW.md` as a first-class source — on this deployment it points straight at the real checkout and would answer correctly with no workspace involved at all.
4. Add fixtures for: a `removed` record newer than a live one; a live record whose path was deleted out from under it; and a multi-candidate ordering assertion. Each should go red if the corresponding guard is removed.

### Verification run

`pnpm install --frozen-lockfile && pnpm build && pnpm test` — all pass (cli 581, orchestrator 353, worker 242). Green suite, broken behavior; both probes were run against the built CLI on this branch.

### Note on sequencing (non-blocking)

This branch is based on `a737db16` and does not contain `#945`, `#953`, or `#949`. GitHub reports it MERGEABLE, but it touches `land.test.ts` and `WORKFLOW.md`, which `#958` also rewrites. Worth rebasing before landing so the semantic overlap is visible rather than silently merged.

moncher-dev:
Fixed in `b32ec7d` and verified again on merged head `72d24cb`.

`resolveWorkflowRepositoryDirectory` now excludes `status: "removed"` workspace records before ordering and validates every configured, workspace, and run candidate with `git rev-parse --is-inside-work-tree`. A missing or non-Git path falls through to the next candidate instead of forcing an `unavailable` result.

The resolver regression now supplies a newer removed checkout, a newer active-but-missing checkout, and an older available Git checkout, and requires the available candidate to win. The existing multi-live-candidate test still requires newest-first ordering. The full lint, unit suite, typecheck, and build pass after merging current `main`; Docker preflight remains unavailable under #956.

moncher-dev (commented):
## Blocking finding is resolved — verified by live probe, not by reading the diff. Passing to hojinzs.

Head `72d24cb6` (`b32ec7d3 fix(cli): skip stale checkout records` + a clean merge of `main` at `4917816f`).

### The probe that failed last pass now passes

Same command, same project directory, built from this branch:

```
$ node <branch>/packages/cli/dist/index.js doctor --project-dir ~/moncher-stack/github-symphony
PASS Project workflow source identity
  Project WORKFLOW.md source is synchronized-copy (sha256:03bfa1095860).
```

Previously this warned `unavailable` and pointed the operator at `…_888_…/repository`, a directory deleted hours earlier. `sha256:03bfa1095860` is the same revision I measured last cycle against a live workspace, so the check now reports the answer I had to compute by hand.

### The selector picks a live checkout — and the `stale-copy` branch is now reachable

I ran `resolveWorkflowRepositoryDirectory` against the **real** state store for this project, then fed its answer into `inspectWorkflowSourceIdentity`:

```
workspace records: 41 (removed 37, live 4); runs 180
SELECTED:  …/symphony-workspaces/hojinzs_github-symphony_954-0124018612c64967/repository   exists=true
STALE probe   (WORKFLOW.md from HEAD~30) -> relationship=stale-copy       content=sha256:fc9dbba89418  repo=sha256:03bfa1095860
CURRENT probe (WORKFLOW.md from HEAD)    -> relationship=synchronized-copy
```

That is the thing #952 exists for and could not previously do: with a genuinely diverged copy, `stale-copy` fires, with both revisions filled in. 37 of 41 records are still `removed`; they are now filtered out rather than sorted to the front.

One observation worth recording, because it shows why the fixture matters more than the probe: the newest record on this host *today* happens to be an `active` workspace that exists, so the original symptom would not reproduce right now even on the old code. The bug was never fixed by the host state changing — it is fixed by the filter and the existence check, and those are what I mutated.

### Would a test fail if the fix were removed? Yes — both halves

`skips newer removed and missing workspace records` builds exactly the adversarial ordering: a `removed` record at `03:00`, an `active` record whose path does not exist at `02:00`, and the live one at `01:00`, asserting the live one wins.

| mutation to `workflow-source-identity.ts` | result |
|---|---|
| drop `.filter((workspace) => workspace.status !== "removed")` | **1 failed** — `skips newer removed and missing workspace records` |
| replace the `rev-parse --is-inside-work-tree` guard with `return candidate` | **1 failed** — same test |
| unmodified | 9 passed |

Each guard is independently load-bearing on that fixture. Fix directions 1, 2 and 4 from the last pass are covered. Direction 3 (treat the symlink target as a first-class source) was a "consider" and declining it is fine — the workspace path now answers correctly without it.

`resolveConfiguredRepositoryPath` is also now validated before being returned rather than trusted, which closes the same class of hole on the configured-path branch.

### Async conversion — callers checked

The function became `async`. Both call sites `await` it (`doctor.ts:2302`, `service.ts:2825`); the `service.ts:2810` hit is a same-named private method, not this one. No un-awaited call that would make a Promise truthy at every site.

### Validation

`pnpm install --frozen-lockfile && pnpm build && pnpm test && pnpm lint && pnpm typecheck` — all pass on `72d24cb6` (orchestrator 354, cli 584). CI on the head: `Test: SUCCESS`, `Container Smoke: SUCCESS`. `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN` against current `main`.

My sequencing note from last pass is resolved by the merge: the branch now contains `#945`, `#953` and `#949`, and its file list touches neither `WORKFLOW.md` nor `land.test.ts`, so the overlap with #958 I was worried about does not exist.

### One open item for @hojinzs — no changeset

`git diff --name-only origin/main...72d24cb6 -- .changeset/` is empty. `#952` carries no `changeset:*` label, so `WORKFLOW.md` Step 8 did not require one — but this PR adds a user-visible `doctor` check and a new status-surface field, and `CLAUDE.md` states that behavior-changing PRs must add a changeset. As it stands the next `@gh-symphony/cli` release ships this with no release note. For contrast, #957 shipped `.changeset/curvy-hooks-validate.md` from an issue that is also unlabeled.

I deliberately did **not** add `changeset:patch` to #952: with the label present, Land pre-flight ("changeset present if labeled") would fail and force a full rework cycle for a one-file addition. Your call — wave it through, or label it and accept the extra cycle. That is the only thing I found open.

### Non-blocking

If no workspace qualifies, the loop falls through to `runs`, which is **not** status-filtered — 180 run records on this host. Each dead path fails cheaply at spawn, so it is bounded rather than harmful, and the existence check makes it correct either way. Worth a `status`/terminal filter on the run candidates for symmetry if this is ever touched again.

@hojinzs over to you.

packages/orchestrator/src/workflow-source-identity.ts
```diff
@@ -0,0 +1,215 @@
...
+        // A revision may precede the file; keep searching its reachable history.
+      }
+    }
+    return {
+      ...common,
+      relationship: "independent",
+      matchedRepositoryCommit: null,
+    };
+  } catch {
+    return unavailable(sourceRevision);
```
hojinzs: **P1 — a repository with no committed `WORKFLOW.md` now warns on every tick, forever.**

`git show <ref>:WORKFLOW.md` and `git rev-parse <ref>` share one `Promise.all` and one `catch`, so "the ref is unreadable" and "the ref is readable but has no `WORKFLOW.md`" collapse into the same `unavailable` verdict. The two are cleanly separable — I checked:

```
$ git rev-parse main
f7e204bfc8ce62916bf197054f8ec2d9bd97e0de     # succeeds
$ git show main:WORKFLOW.md
fatal: path 'WORKFLOW.md' does not exist in 'main'   # exit 128
```

I reproduced the consequence against this branch's source directly:

```
CASE A (repo has no committed WORKFLOW.md):
{
  "relationship": "unavailable",
  "contentRevision": "sha256:25fba518a3be",
  "repositoryCommit": null,
  "repositoryRevision": null
}
```

That verdict reaches the operator twice per the wiring in this PR: `service.ts:2846` emits a snapshot warning on **every status tick**, and `doctor.ts:2341` emits a required WARN whose remediation reads *"Ensure …/WORKFLOW.md is a local Git checkout with main and a committed WORKFLOW.md."*

This is not a corner case. `packages/cli/src/commands/project.ts:88` sets `workflowSource: { type: "external", path: <projectDir>/WORKFLOW.md }` for **every** standalone project, so the check runs for all of them — while the configured `repository` is the target code repository, which has no reason to carry Symphony's own `WORKFLOW.md`. For those deployments the warning is permanent and its remediation is actively wrong advice.

The pre-existing behavior handled this deliberately (`access(repositoryWorkflowPath)` → no warning on ENOENT), and the test that pinned it — `does not warn for a remote repository without a committed WORKFLOW.md` — was rewritten into the HTTPS-clone stale-copy test in `service.test.ts:15806`, so nothing goes red.

This is the same failure mode you blocked on twice already: a check that is permanently noisy on a healthy layout trains operators to ignore the one line that will eventually say `stale-copy`.

Suggested fix: resolve `rev-parse` first; if the ref resolves but `show` reports the path missing, return a distinct non-warning relationship (`no-repository-policy` or similar) that both surfaces report informationally, and restore a regression test for it.

---

packages/orchestrator/src/workflow-source-identity.ts
```diff
@@ -0,0 +1,215 @@
...
+          sourceContent ===
+          (await execGit(input.repositoryDirectory, [
+            "show",
+            `${revision}:WORKFLOW.md`,
+          ]))
+        ) {
+          return {
+            ...common,
+            relationship: "stale-copy",
+            matchedRepositoryCommit: revision,
```
hojinzs, line 234: **P2 — the `independent` layout pays a full history rescan, one `git show` subprocess per revision, on every status tick.**

The loop short-circuits on a match, so `stale-copy` is cheap. `independent` is the worst case: no revision ever matches, so every reachable `WORKFLOW.md` revision is spawned and read on each tick, forever. This repository is at 11 revisions today (`git rev-list --full-history HEAD -- WORKFLOW.md | wc -l`), and policy files only accumulate.

AC 3 explicitly calls `independent` the case that "should not warn on every tick" — it doesn't warn, but it is now the most expensive branch, permanently, and there is no memoization anywhere in `resolveWorkflowSourceIdentity` between ticks.

A single-subprocess form gets the same answer:

```ts
const blob = (await execGit(dir, ["hash-object", "--stdin"], sourceContent)).trim();
// git log --full-history --format=%H -- WORKFLOW.md  →  "<sha>:WORKFLOW.md" lines
// git cat-file --batch-check='%(objectname)'  →  compare blob ids, no content transfer
```

Not blocking on its own — batching it with the P1 fix would be reasonable.

---

packages/orchestrator/src/workflow-source-identity.ts
```diff
@@ -0,0 +1,215 @@
...
+    input.baseDirectory ?? process.cwd()
+  );
+  const workspacePaths = [...(input.issueWorkspaces ?? [])]
+    .filter((workspace) => workspace.status !== "removed")
+    .sort((left, right) => right.updatedAt.localeCompare(left.updatedAt))
+    .map((workspace) => workspace.repositoryPath);
+  const runPaths = [...(input.runs ?? [])]
+    .sort((left, right) => right.updatedAt.localeCompare(left.updatedAt))
+    .map((run) => run.workingDirectory);
+
```
hojinzs, line 57: **P2 — run candidates carry no status filter, so the fallback is unbounded in exactly the case it is reached.**

`workspacePaths` filters `status !== "removed"` (correct, and load-bearing per your own mutation test). `runPaths` does not. That only matters when no configured path and no live workspace qualifies — but that is precisely the state where every run record is dead, so all of them get probed with a `git rev-parse` subprocess, every tick. Your live host reports 180 run records.

I confirmed each candidate costs one spawn and there is no early bail beyond the first success:

```
CASE E selected: live repo   probe count: 5   (5 run records, live one last)
fatal: cannot change to '…/dead-run-4': No such file or directory
… ×4
```

You flagged this as non-blocking last pass and I agree it is bounded rather than harmful — but symmetry with the workspace filter is a two-line change, and it removes 180 spawns/tick from the one path that reaches it.

---

packages/orchestrator/src/workflow-source-identity.ts
```diff
@@ -0,0 +1,215 @@
...
+}
+
+export async function inspectWorkflowSourceIdentity(input: {
+  workflowPath: string;
+  repositoryDirectory: string;
+  baseRef?: string | null;
+  dependencies?: WorkflowSourceIdentityDependencies;
+}): Promise<WorkflowSourceIdentity> {
+  const execGit = input.dependencies?.execGit ?? defaultExecGit;
+  const repositoryWorkflowPath = `${input.repositoryDirectory}/WORKFLOW.md`;
```
hojinzs: **nit** — `${input.repositoryDirectory}/WORKFLOW.md` is the only path construction in the module that bypasses `node:path`. `join()` is already imported elsewhere in the package and this value is surfaced verbatim to operators as `repositoryPath` in both the snapshot and the doctor remediation, so a trailing slash on the resolved directory renders as `…/repository//WORKFLOW.md` in operator output.

---

packages/orchestrator/src/workflow-source-identity.ts
```diff
@@ -0,0 +1,215 @@
...
+    ]);
+    const repositoryRevision = contentRevision(repositoryContent);
+    const common = {
+      path: input.workflowPath,
+      contentRevision: sourceRevision,
+      repositoryPath: repositoryWorkflowPath,
+      repositoryRef,
+      repositoryCommit: repositoryCommit.trim(),
+      repositoryRevision,
+    };
```
hojinzs: **P3 — byte-exact comparison classifies a current copy as `independent` under line-ending normalization.**

`readFile(…, "utf8")` returns working-tree bytes; `git show` returns index bytes. With `core.autocrlf=true` (Windows default) a copy that is exactly current differs by CRLF and falls through to `independent` — reported as a **PASS**, silently. The detector then goes blind in precisely the way #952 describes, with no signal at all.

Confirmed against this branch:

```
CASE D (CRLF copy of an LF committed file): independent
```

Low priority given the documented deployment is macOS/Linux, but normalizing line endings before comparison is a one-liner and removes a silent-failure class from a feature whose whole purpose is not failing silently.

---

packages/orchestrator/src/service.ts
```diff
@@ -2803,39 +2815,43 @@ export class OrchestratorService {
...
+      return [
+        `Project WORKFLOW.md source identity could not be determined: loaded ${identity.contentRevision ?? "an unreadable revision"} from ${identity.path}, but committed ${identity.repositoryRef ?? "workflow ref"} could not be read from ${identity.repositoryPath ?? "the configured repository"}.`,
+      ];
     }
-
-    return [];
+    if (identity.relationship !== "stale-copy") return [];
+    return [
+      `Project WORKFLOW.md is a diverged copy: loaded ${identity.contentRevision} from ${identity.path}, but ${identity.repositoryRef} has ${identity.repositoryRevision} at ${identity.repositoryPath}.`,
+    ];
```
hojinzs, line 2854: **P2 — the "shadows repository WORKFLOW.md" warning is removed, and the PR body does not say so.**

The old `resolveWorkflowWarnings` warned whenever an external workflow source coexisted with a readable repository `WORKFLOW.md` at a different path. That warning is gone entirely; the only surviving conditions are `stale-copy` and `unavailable`.

Dropping it is defensible — AC 3 asks for exactly the opposite of warning on a deliberately independent policy, and `synchronized-copy` shadowing is genuinely not worth a warning. But it is a user-visible removal in a snapshot field operators may key on, and "Changed files" / "User-visible behavior" in the PR body describe only additions. Please state the removal explicitly in the body (and note whether anything consumed that string).

---

packages/cli/src/commands/doctor.ts
```diff
@@ -2283,6 +2287,94 @@ export async function runDoctorDiagnostics(
...
+          "Run the project so it creates a local issue workspace, or configure repository.path with a local Git checkout.",
+          {
+            relationship: "unavailable",
+            contentRevision: null,
+            repositoryPath: null,
+            repositoryRef: null,
+            repositoryCommit: null,
+            repositoryRevision: null,
+            matchedRepositoryCommit: null,
+          }
```
hojinzs, line 2374: **P3 — a project that has never dispatched WARNs on its first `doctor` run.**

With a remote clone URL and no persisted workspace or run yet, `resolveWorkflowRepositoryDirectory` returns `null` and this branch fires. That is the state of every freshly-created project, so the first thing a new operator sees from `doctor` is a required WARN. The remediation ("Run the project so it creates a local issue workspace") is at least correct here, unlike the `unavailable` branch above — but a check whose first observation on a healthy new install is a warning is worth demoting to informational until the project has actually run once.

Note this also interacts with the P1: the two branches produce near-identical operator-facing text for genuinely different conditions (no checkout yet vs. checkout present but no committed policy vs. ref unreadable). Distinct wording per condition would make the eventual real `stale-copy` line stand out.

---

hojinzs (changes requested):
## Review for @hojinzs — one P1, then this is done

Reviewed `72d24cb6`. I verified behavior by executing this branch's `workflow-source-identity.ts` against real git fixtures, not by reading the diff. The feature works: all four AC-4 classifications are correct, and the previous two blocking findings are genuinely closed. One new blocking issue, found by probing a layout the deleted test used to cover.

### Acceptance criteria — #952

| AC | Verdict | Evidence |
|---|---|---|
| 1. Status exposes content hash + comparable identity | ✅ | `workflow.source` on `ProjectStatusSnapshot`; 7 fields incl. `contentRevision`, `repositoryCommit`, `repositoryRevision` |
| 2. `doctor` warns on a diverged regular-file copy, naming both revisions | ✅ | `doctor.ts:2325`; message carries loaded + committed revisions |
| 3. Distinguishes diverged copy from deliberately independent policy | ✅ | `independent` passes and never warns |
| 4. Tests: symlink / identical / stale / unrelated | ✅ | 9/9 in `workflow-source-identity.test.ts`; independently reproduced below |
| Note: no network fetch per tick | ✅ | local git only — `show`, `rev-parse`, `rev-list` |

Independent classification probe against this branch:

```
CASE B symlink       -> linked              sha256:19667cd72438 == repo sha256:19667cd72438
CASE C1 identical    -> synchronized-copy
CASE C2 diverged     -> stale-copy          sha256:19667cd72438 -> sha256:34d9392a58d4  matched ee51cb38
CASE C3 unrelated    -> independent
```

Each is correct. The mechanism does what #952 asked for.

### Blocking: P1 — a repository with no committed `WORKFLOW.md` warns on every tick, permanently

Full detail inline on `workflow-source-identity.ts:145-213`. Short version:

`git show <ref>:WORKFLOW.md` and `git rev-parse <ref>` share one `Promise.all` and one `catch`, so **"ref unreadable"** and **"ref is fine, the repo simply has no `WORKFLOW.md`"** collapse into `unavailable`. Reproduced:

```
CASE A (repo has no committed WORKFLOW.md) -> relationship "unavailable", repositoryCommit null
```

`project.ts:88` sets an external `workflowSource` for **every** standalone project, so the check runs everywhere — while the configured `repository` is the target *code* repository, which has no reason to carry Symphony's own policy file. Those deployments get a per-tick snapshot warning plus a required doctor WARN telling them to *"Ensure …/WORKFLOW.md is a local Git checkout with main and a committed WORKFLOW.md"* — advice that is wrong for that layout.

The old code handled this on purpose (`access()` → ENOENT → no warning) and `does not warn for a remote repository without a committed WORKFLOW.md` pinned it. That test was rewritten into the HTTPS-clone stale-copy case, so nothing goes red.

It is the failure mode you blocked on twice: a check that is permanently noisy on a healthy layout trains operators past the one line that will eventually say `stale-copy`. `rev-parse` and `show` fail separably (`fatal: path 'WORKFLOW.md' does not exist in 'main'` with `rev-parse` succeeding), so this is a small fix: split them, give "no repository policy" its own non-warning relationship, restore the regression test.

### Non-blocking

- **P2** — `independent` rescans full `WORKFLOW.md` history with one `git show` subprocess per revision, every tick, with no memoization. It is the branch that never short-circuits. (11 revisions here and growing.)
- **P2** — run candidates aren't status-filtered, unlike workspaces. Only reached when every record is dead, which is exactly when all 180 get probed. Two-line symmetry fix.
- **P2** — the "shadows repository WORKFLOW.md" warning is removed and the PR body doesn't mention it. Removal is defensible under AC 3; please state it.
- **P3** — CRLF working-tree copies classify as `independent` and PASS silently (`CASE D` confirmed). Windows-only, but it is a silent-failure class in a feature built to stop silent failures.
- **P3** — first `doctor` on a never-run project WARNs by construction.
- **nit** — `${repositoryDirectory}/WORKFLOW.md` bypasses `path.join`, and that string is shown to operators.

### Over-engineering / scope

None found. The workspace/run fallback chain and the shared resolver exist because review demanded them; the classification set is the minimum AC 3 requires. The Docker E2E gap (TC-25, #956) is honestly disclosed rather than papered over.

### Changeset — please just add one

`git diff --name-only origin/main...72d24cb6 -- .changeset/` is empty. `CLAUDE.md` requires a changeset for behavior-changing PRs, and this adds a doctor check and a snapshot field. Your concern was that labeling #952 `changeset:patch` forces a rework cycle — but Land pre-flight only checks *"changeset present if labeled"*. Adding the file without the label costs nothing and fails nothing. Add `.changeset/*.md` for `@gh-symphony/cli`, leave #952 unlabeled.

### Validation on my side

`workflow-source-identity.test.ts` 9/9 pass unmodified. `doctor.test.ts` fails locally only on unbuilt workspace `dist` (`Failed to resolve entry for package "@gh-symphony/core"`) — not a code defect; CI on this head is green (`Test: success`, `Container Smoke: success`), `mergeable_state: clean`.

### Convergence

This PR is at five review rounds. **Fix the P1 and add the changeset and I approve on the next pass** — the P2/P3 items are yours to take or decline with a reason, and declining them will not hold this up. If the P1 is contested rather than fixed, argue it in a thread reply and I'll settle it that round rather than opening a sixth.

---

packages/orchestrator/src/workflow-source-identity.ts
moncher-dev: Fixed in `daeed62`. The detector now resolves the base ref first, uses `git ls-tree` to distinguish an absent committed `WORKFLOW.md`, and returns `no-repository-policy` with the repository commit preserved. Doctor reports that state as an informational PASS and the daemon emits no warning. Unit, doctor, and service regressions cover it; disabling the split makes the focused regression fail as requested.

packages/orchestrator/src/workflow-source-identity.ts
moncher-dev: I kept the full-history scan unchanged in this cycle. It is bounded by the local reachable policy history and short-circuits for stale copies, while replacing it safely with batched object plumbing would broaden this final correctness rework. The blocking no-policy case returns before history traversal. This remains a performance improvement rather than a correctness gate.

packages/orchestrator/src/workflow-source-identity.ts
moncher-dev: I did not filter run records by lifecycle status. Unlike an issue workspace's explicit `removed` state, terminal run status does not mean its working directory was removed; those valid retained checkouts are useful fallbacks. Every candidate is already validated with `git rev-parse --is-inside-work-tree`, which is the authoritative usability check and skips dead paths.

packages/orchestrator/src/workflow-source-identity.ts
moncher-dev: Fixed in `daeed62`: the operator-visible repository policy path now uses `node:path.join`, avoiding doubled separators.

packages/orchestrator/src/workflow-source-identity.ts
moncher-dev: Fixed in `daeed62`. Comparison and content revisions now normalize CRLF to LF, with a regression requiring a CRLF working copy of the current LF policy to classify as `synchronized-copy` and report matching revisions.

packages/orchestrator/src/service.ts
moncher-dev: Acknowledged. The refreshed PR body will explicitly call out removal of the broad legacy shadow warning: supported synchronized and independent external policies no longer warn; only stale historical copies and genuinely unavailable comparisons do.

packages/cli/src/commands/doctor.ts
moncher-dev: I kept the first-run/no-checkout result as WARN. Identity cannot be verified until a local checkout exists, and the diagnostic gives the correct action without fabricating a policy relationship. The new `no-repository-policy` state now makes this wording distinct from a repository checkout that exists but intentionally carries no committed policy.

packages/orchestrator/src/workflow-source-identity.ts
```diff
@@ -0,0 +1,249 @@
...
+    ).trim();
+    const repositoryPolicyPath = (
+      await execGit(input.repositoryDirectory, [
+        "ls-tree",
+        "--name-only",
+        repositoryRef,
+        "--",
+        "WORKFLOW.md",
+      ])
+    ).trim();
```
hojinzs, line 164: **P2 — `ls-tree` pathspec is cwd-relative, `git show ref:path` is root-relative; the two disagree when `repositoryDirectory` is a subdirectory.**

Not blocking — the P1 split itself is correct and I verified it. But the two probes that now decide the `no-repository-policy` branch use different path semantics, and only one of them is anchored to the repository root:

```
$ cd <repo>/packages/app
$ git ls-tree --name-only main -- WORKFLOW.md
                                    # empty — looked for packages/app/WORKFLOW.md
$ git show main:WORKFLOW.md
# policy                            # found — root-relative
$ git rev-parse --is-inside-work-tree
true                                # the usability guard still passes
```

So if `repositoryDirectory` ever resolves to a subdirectory of a checkout, `repositoryPolicyPath` is empty for a repository that *does* carry a committed policy, and the function returns `no-repository-policy` — a silent PASS in doctor and no snapshot warning. A genuinely stale copy in that layout would be classified as "repository has no policy" and never reach the `stale-copy` branch.

Reachability is narrow, which is why this is P2 and not a block: the resolver's own fallbacks (`repositoryPath`, `workingDirectory`) are worktree roots. The exposed input is operator-supplied `repository.path`, which is validated only by `--is-inside-work-tree` — and that returns `true` from a subdirectory. Worth noting this is a small regression in reach: before this commit, `git show` alone was root-relative, so a subdirectory-configured repository still classified correctly.

One flag fixes it, matching `show`'s semantics:

```ts
await execGit(input.repositoryDirectory, [
  "ls-tree",
  "--full-tree",
  "--name-only",
  repositoryRef,
  "--",
  "WORKFLOW.md",
])
```

Verified: `git ls-tree --full-tree --name-only main -- WORKFLOW.md` from `packages/app` returns `WORKFLOW.md`. Take it here or as a follow-up — it does not hold up this PR.

---

hojinzs (approved):
## Approved — P1 fixed and verified by execution, changeset added. Round 6, converging here.

Reviewed `daeed62b`. As last round, I verified behavior by running this branch's built code against real git fixtures rather than reading the diff.

### The P1 is genuinely fixed

`rev-parse` now resolves first and `git ls-tree` distinguishes "ref unreadable" from "ref fine, no committed policy". Probed against real repositories:

```
CASE A  repo with no committed WORKFLOW.md -> no-repository-policy  repositoryCommit 8b9d9350  repositoryRevision null
CASE A2 unreadable ref                     -> unavailable
```

The split is real, not a blanket swallow — A2 still warns, which is what makes A trustworthy. Downstream both surfaces agree: `service.ts:2851` returns no warning for `no-repository-policy`, and `doctor.ts` reports it as a PASS with wording distinct from both `independent` and the first-run case. The permanent per-tick warning on every standalone project whose target repository has no `WORKFLOW.md` is gone.

### All four AC-4 classifications still correct after the change

```
CASE B  symlink        -> linked             sha256:2b76021d41cf == sha256:2b76021d41cf
CASE C1 CRLF copy      -> synchronized-copy  sha256:7ed115a6fb72 == sha256:7ed115a6fb72
CASE C2 diverged       -> stale-copy         sha256:7ed115a6fb72 -> sha256:2b76021d41cf  matched c8f9f936
CASE C3 unrelated      -> independent
```

`stale-copy` — the branch #952 exists for — still fires with both revisions populated. The CRLF P3 is fixed: what was `independent` (silent PASS) last round is now `synchronized-copy` with matching revisions. The `path.join` nit is fixed; a trailing-slash directory renders `…/repoD/WORKFLOW.md`, not `…//WORKFLOW.md`.

### Would tests fail if the fixes were removed? Both, independently

| mutation to `workflow-source-identity.ts` | result |
|---|---|
| `if (!repositoryPolicyPath)` → `if (false)` | **1 failed** — `reports a repository with no committed policy without warning identity` |
| `normalizePolicyContent` → identity function | **1 failed** — `normalizes line endings when comparing a synchronized copy` |
| unmodified | 11 passed |

Coverage is at all three levels, not just the unit: `doctor.test.ts:740` asserts the PASS and the `no-repository-policy` details, and `service.test.ts:15869` asserts `snapshot.warnings` is empty with `repositoryRevision: null`. The regression the last round deleted is restored in substance.

### Declined items — all with reasons, all acceptable

The full-history rescan (P2), the run-record status filter (P2), and the first-run WARN (P3) were each declined with an argument rather than silence. The run-filter reply is the better analysis: terminal run status does not imply a removed working directory, and `rev-parse --is-inside-work-tree` is the authoritative usability check — I'll withdraw that suggestion. The shadow-warning removal is now stated explicitly in the PR body as asked.

### One new P2, inline, explicitly not blocking

`ls-tree`'s pathspec is cwd-relative while `git show ref:path` is root-relative, so a `repositoryDirectory` pointing at a subdirectory would report `no-repository-policy` for a repository that does have a committed policy. Narrow reach — the resolver's own fallbacks are worktree roots, and only operator-supplied `repository.path` exposes it — and `--full-tree` is a one-flag fix. Details and the probe are in the inline comment. Take it here or as a follow-up; it does not gate this.

### Validation on my side

`pnpm install --frozen-lockfile`, `pnpm build`, `pnpm lint`, `pnpm typecheck` — pass. `pnpm test` — full workspace green (exit 0; orchestrator 357, cli 585, worker 242, core 340). `pnpm format` reports 64 pre-existing files on `main` as well and every file this PR touches is Prettier-clean, so that is not this PR's. CI on `daeed62b`: `Test: success`, `Container Smoke: success`; `mergeable_state: clean`.

The TC-25 Docker E2E gap is honestly disclosed and tracked by #956 rather than papered over; the post-merge checkbox is the right place for it.

### Convergence

Six rounds, and the two conditions I set last time are met. Approving. Ship it — the remaining P2s are yours to take or leave.

---

moncher-dev:
## Verified at `daeed62b` — P1 and the changeset are both closed, and both are guarded. Over to @hojinzs.

Your `CHANGES_REQUESTED` binds to `72d24cb6`; head is now **`daeed62b`**. I verified by executing this branch against real git fixtures and by mutation, not by reading the diff.

### P1 — fixed

`rev-parse` and the policy-presence probe are now separate: `rev-parse <ref>` runs first, then `git ls-tree --name-only <ref> -- WORKFLOW.md`, and an empty result returns a new `no-repository-policy` relationship instead of collapsing into `unavailable`. Using `ls-tree` rather than a `show` failure is cleaner than what you proposed — it distinguishes by presence instead of by parsing an error.

Live probe on this branch:

```
CASE A  repo has commits, no committed WORKFLOW.md -> no-repository-policy   repositoryCommit d2d52161 (populated, not null)
CASE E  unreadable ref (baseRef "nope")            -> unavailable            (still warns, correctly)
```

The two conditions you identified as conflated are now genuinely separable: A no longer warns, E still does.

`doctor` gives it a named destination rather than silence — `doctor.ts:2352` routes it to **`passCheck`** with its own message: *"Project WORKFLOW.md is an independent policy (…); the configured repository has no committed WORKFLOW.md at `<ref>`."* That was my one open question from the issue thread, and it is answered.

### The four original classifications still hold

```
symlink                              -> linked
identical copy                       -> synchronized-copy
copy matching an older commit        -> stale-copy   matched d136ea76
content matching no commit           -> independent
```

### Regression guards — mutation-tested, all three fire

| Mutation | Result |
|---|---|
| Delete the `no-repository-policy` branch so it falls through to `show` → `unavailable` | **2 failed**: `OrchestratorService > does not warn when the configured repository has no committed workflow`, and `inspectWorkflowSourceIdentity > reports a repository with no committed policy without warning identity` |
| Route `no-repository-policy` to the WARN branch in `doctor.ts` | **1 failed**: `runDoctorDiagnostics > reports an independent project policy when the repository has no policy` |

So the regression test you asked to have restored exists at both the unit and service level, and the doctor's *non-warning* verdict is pinned too — reverting the split cannot pass green. That was the specific thing I asked to be checked before pushing, and it holds.

### Changeset — present

`.changeset/tame-workflows-identify.md`, `@gh-symphony/cli` patch. Unlabeled, per your point that Land pre-flight only checks "changeset present if labeled".

### Validation

Fresh worktree at `daeed62b`: `pnpm install --frozen-lockfile` → `pnpm build` → `pnpm test`. All packages green — orchestrator 357/357, cli 585/585 across 40 files, worker 242/242, tracker-github 123/123, runtime-claude 107/107, runtime-codex 68/68.

### One observation, not a request

`stale-copy` only fires when the project copy matches an *older commit exactly*; a hand-edited copy that matches no revision in history classifies `independent` and PASSes silently. That is the AC-3 design you already accepted (your `CASE C3 unrelated -> independent`), and it is unchanged by this PR — I raise it only so the guarantee is not over-read later: this detects *drifted-behind* copies, not *edited* ones. Your P2/P3 items (memoization, run-candidate status filter, stating the "shadows" removal in the body, CRLF) remain yours to take or decline.

Nothing blocking from me. This needs only your approval at `daeed62b`.

---

PR #957, opened 2026-09-08
feat(cli): validate hook paths in workflow validate

### LINKED ISSUE #948: feat(cli): validate hook paths in workflow validate
## Problem

`gh-symphony workflow validate` parses and echoes declared workflow hooks but does not perform the filesystem/path diagnostics used by `doctor` and trusted `project start` preflight.

## Context

Discovered while reviewing #929 / PR #936. Keep this enhancement separate from the retry-budget and startup fix in #929.

## Suggested direction

Reuse the shared hook validation and formatting helpers so `workflow validate` reports invalid path syntax and immediately resolvable missing/non-executable scripts without duplicating hook-resolution semantics.

moncher-dev:
## Issues

- Closes #948

## Summary

- TL;DR: `workflow validate` now rejects invalid hook path syntax and immediately resolvable missing, non-file, or non-executable scripts.
- Reuses the core hook path validation and formatting helpers already used by startup and doctor diagnostics.
- Review follow-up preserves structured JSON errors, gives multi-hook failures an honest parent path, adds remediation, documents trust behavior, and keeps hook-free human output quiet.

## Change-point diagram

- `workflow validate` → core `validateWorkflowHookPaths` → shared path resolution/filesystem diagnostics → typed CLI error or validation report

## Start here

- `packages/cli/src/commands/workflow.ts` — shared diagnostics integration, typed failure mapping, and human report formatting
- `packages/cli/src/commands/workflow.test.ts` — syntax, filesystem, deferred-resolution, JSON, remediation, and multi-hook coverage

- Local workflow validation now fails early for invalid hook commands and locally resolvable bad scripts, regardless of runtime hook trust. Repository-relative run hooks remain deferred until an issue workspace exists. JSON failures use `workflow_validation_error` with `hooks.<kind>` for one problem or `hooks` for multiple problems.

## Validation

- `pnpm exec vitest run packages/cli/src/commands/workflow.test.ts` — pass (31 tests; rerun after rebuilding merged core output)
- `pnpm lint` — pass
- `pnpm test` — pass
- `pnpm typecheck` — pass
- `pnpm build` — pass
- `pnpm exec prettier --check README.md docs/configuration.md packages/cli/src/commands/workflow.ts packages/cli/src/commands/workflow.test.ts` — pass
- `git diff --check` — pass
- Docker E2E — not applicable: deterministic local CLI validation with focused and full unit coverage

## Changeset

- `.changeset/curvy-hooks-validate.md` (`patch`)

## Risks & rollback

- Risk is limited to workflows whose hooks were previously echoed without diagnostics and scripts consuming typed validation failures; revert this PR to restore prior behavior.

## Changed files

- `packages/cli/src/commands/workflow.ts` — run shared hook diagnostics, preserve typed errors, add remediation, and suppress zero-count human output
- `packages/cli/src/commands/workflow.test.ts` — cover invalid, missing, non-file, non-executable, checked, deferred, JSON, multi-error, and hook-free behavior
- `README.md`, `packages/cli/README.md`, `docs/configuration.md`, `docs/architecture.md` — document behavior, trust semantics, and layer mapping
- `.changeset/curvy-hooks-validate.md` — publish a CLI patch

- [ ] None beyond CI and review

## Security


moncher-dev (commented):
Reviewed at `1b5c44a0` in a scratch worktree: `pnpm install --frozen-lockfile`, `pnpm build`, `pnpm lint`, `pnpm typecheck`, `pnpm test` all pass (584 CLI tests; full monorepo green). I then ran the built binary against nine hand-built WORKFLOW.md fixtures and mutated the source six ways to measure test strength.

**The core logic is right, including the part I most expected to be wrong.** All four hook kinds are enumerated (`core/src/workspace/hooks.ts:95-100`), and the base-directory rule is not reimplemented — `validateWorkflowHookPaths` calls the same `resolveWorkflowHookPath` the runtime calls at `orchestrator/src/service.ts:4981`, with the same argument shape, so `after_create` resolves from the WORKFLOW.md directory and the other three from the repository directory. The command deliberately passes only `workflowDirectory`, so relative `before_run`/`after_run`/`before_remove` come back `null` and are counted as deferred rather than falsely resolved. I verified both directions: four kinds pointing at one existing script gives `1 checked, 3 deferred` and exits 0, and a `before_run` pointing at a missing repo-relative path **passes** rather than failing. That is the false-positive trap avoided.

Also worth correcting from my own earlier note: doctor's `workflow_hooks` check is **not** gated behind `--smoke`. `buildHookChecks` is called at `doctor.ts:2278` under `if (workflow.status === "pass")`; the `--smoke` block only starts at `:2398`. There is genuine overlap with plain `doctor`, but `workflow validate` is an offline surface with no GitHub auth, no managed-project resolution, and an arbitrary `--file`, so the added coverage is justified.

**Blocking: `--json` failures have no destination.** `workflow.ts:861` throws a bare `Error`. The handler at `workflow.ts:1096-1108` converts only `WorkflowValidationError` into `writeCliError({code, path, message, json})`; everything else falls through to a plain stderr line. Observed:

```
$ node packages/cli/dist/index.js --json workflow validate --file fx/b/WORKFLOW.md
stdout: (empty)   exit: 1
stderr: Workflow command failed: Invalid WORKFLOW.md hook path: after_create=/…/hooks/nope.sh (missing).
```

while a schema failure on the same command emits the documented object:

```
{ "error": { "code": "workflow_front_matter_not_a_map", "path": "front_matter", "message": "…" } }
```

This contradicts the line this PR itself edits at `packages/cli/README.md:282` ("Strict front-matter failures use stable workflow error codes; `workflow validate --json` also emits the failing `error.path`"). A `--json` consumer gets empty stdout, a nonzero exit, and nothing parseable. Codex found the same thing independently in its inline comment on `workflow.ts:862-864`. Please throw a `WorkflowValidationError` with a stable code and a `hooks.<kind>` path, or route through `writeCliError`, and add a `--json` failure test — there is currently no `--json` test for any of the new code, which is exactly why this got through.

**Second finding, your call rather than blocking: a bad hook path suppresses every other diagnostic.** The throw happens inside `validateWorkflow` before the report is built, so the warnings computed at `workflow.ts:880-883` never reach the user. Same fixture, run twice:

```
### with a missing hook
Workflow command failed: Invalid WORKFLOW.md hook path: … (missing).   exit 1

### identical fixture, hook line removed — what was suppressed
Warnings
  Legacy priority mapping: tracker.priority_field is deprecated …
```

`doctor` accumulates rather than aborting. Either behaviour is defensible; the undocumented divergence is not.

**Test strength.** Most of the new behaviour is genuinely load-bearing — deleting the throw fails 3 tests, deleting the `path_checks=` line fails 2, deleting the deferred warning fails 1, and resolving all four kinds from the WORKFLOW.md directory fails 1. Two gaps:

- Dropping the `isFile()` check leaves **27/27 CLI and 12/12 core tests passing**. `not_file` has no test anywhere in the repo (`grep -rn "not_file"` over `*.test.ts` is empty), yet the branch is reachable — a directory as a hook path prints `(not file)`. This PR is what surfaces that reason to end users.
- Making every failure report `not_executable` still passes all 27 CLI tests; only core catches it. The missing-script test asserts just `after_create=` (`workflow.test.ts:130`) and never asserts `(missing)`.

Smaller notes: `checked` and `deferred` are reported as counts only, so a user with three deferred hooks cannot see which — doctor names them in `details.checked`. `after_run` and `before_remove` are never exercised in the new tests. And `start.ts:263` gates hook validation behind `hooksTrusted` while `validate` and `doctor` do not, so a workflow `project start` accepts can hard-fail `workflow validate`; strictness here is probably right, but it should be a conscious choice.

Item 1 is the only thing I consider blocking.

packages/cli/src/commands/workflow.ts
```diff
@@ -848,6 +855,14 @@ function validateWorkflow(
       })()
     : ("skip" as const);
   const effectiveTimeouts = resolveWorkflowRuntimeTimeouts(workflow);
+  const hookValidation = await validateWorkflowHookPaths(workflow.hooks, {
+    workflowDirectory: dirname(workflowPath),
+  });
+  if (hookValidation.problems.length > 0) {
+    throw new Error(
+      `Invalid WORKFLOW.md hook path${hookValidation.problems.length === 1 ? "" : "s"}: ${formatWorkflowHookPathProblems(hookValidation.problems)}.`
+    );
```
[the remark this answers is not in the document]
moncher-dev: Resolved in 5bb400d. Hook-path failures now throw `WorkflowValidationError` with the stable `workflow_validation_error` code and a field-aware `hooks.<kind>` path, so `workflow validate --json` emits a parseable error object. Added JSON-mode coverage plus explicit `missing` and `not_file` diagnostic assertions.

moncher-dev (commented):
## Verified — all three blocking items from the previous pass are resolved

**1. `--json` contract (the block).** Ran the built CLI against a `WORKFLOW.md` with `after_create: hooks/does_not_exist.sh`:

```
$ node packages/cli/dist/index.js workflow validate --file WORKFLOW.md --json
{
  "error": {
    "code": "workflow_validation_error",
    "path": "hooks.after_create",
    "message": "hooks.after_create: Invalid WORKFLOW.md hook path: after_create=.../hooks/does_not_exist.sh (missing)."
  }
}
$ echo $?
1
```

stdout parses as JSON, exit is 1, and the plain (non-`--json`) path still prints the one-line typed error. The README line this PR edits is now true.

**Mutation check** — replaced `WorkflowValidationError` with a bare `throw new Error(...)` and re-ran `packages/cli/src/commands/workflow.test.ts`:

```
× workflow command handler > prints typed hook path errors from workflow validate
Tests  1 failed | 28 passed (29)
```

The fix is load-bearing; it does not survive its own deletion.

**2. `not_file` coverage.** Now asserted at `packages/cli/src/commands/workflow.test.ts:228` (`"hook-directory (not file)"`), and I exercised the classification for real by pointing `after_create` at a directory — reported as `(not file)`.

**3. Suppressed diagnostics.** `formatWorkflowHookPathProblems(hookValidation.problems)` reports every hook problem in one message rather than only the first, so nothing inside the hook check is hidden. Non-hook warnings are still suppressed by the throw, which matches how `WorkflowValidationError` behaves elsewhere in this command — acceptable.

### Deferred-hook destination checked, and it is real

I probed the `1 checked, 3 deferred` message because a "deferred" branch that names a destination is exactly the kind of claim worth distrusting. It holds up: `resolveWorkflowHookPath` bases every non-`after_create` hook on `repositoryDirectory`, which `workflow validate` legitimately does not have, and `OrchestratorService` re-runs `validateWorkflowHookPaths` at `service.ts:1405` and — with `repositoryDirectory: run.workingDirectory` — at `service.ts:3687`, raising `Project configuration fault: invalid WORKFLOW.md hook path...`. So deferred hooks really are validated later, and the message is accurate rather than reassuring.

Confirmed live: `after_run: hooks/nope.sh` is reported as deferred, not silently passed, and `after_create` in the same block is still checked.

### Verification run

`pnpm install --frozen-lockfile && pnpm build && pnpm test` — all pass (cli 586, orchestrator 345). Plus the built-CLI probes and the mutation above.

Nothing blocking from me. Handing to @hojinzs.

packages/cli/src/commands/workflow.ts
```diff
@@ -848,6 +855,16 @@ function validateWorkflow(
       })()
     : ("skip" as const);
   const effectiveTimeouts = resolveWorkflowRuntimeTimeouts(workflow);
+  const hookValidation = await validateWorkflowHookPaths(workflow.hooks, {
+    workflowDirectory: dirname(workflowPath),
+  });
```
hojinzs, line 864: **P2 — this validates ungated, while every runtime hook site is gated behind hook trust.**

I traced all four call sites of `validateWorkflowHookPaths`:

| site | gate |
| --- | --- |
| `orchestrator/src/service.ts:1395` | `isWorkflowHookExecutionAllowed(this.resolveProjectEnvironment(tenant))` |
| `orchestrator/src/service.ts:3675` | same guard |
| `cli/src/commands/start.ts:263` | `hooksTrusted` (`SYMPHONY_ALLOW_WORKFLOW_HOOKS`) |
| `cli/src/commands/doctor.ts:1293` | ungated |
| **this line** | **ungated** |

The runtime rule is consistent: if hooks are not trusted they are never executed (`executeWorkspaceHook` returns `skipped`), so their paths are not validated. Trust is opt-in, so *untrusted is the default*. That means a workflow `project start` accepts and runs cleanly today will now exit 1 from `workflow validate` because of a stale hook path that never runs.

`doctor` is a precedent, but not an equivalent one: it emits one failed check among many, with a remediation string, and the rest of the report still prints. Here the throw aborts the whole validation.

I don't think this is obviously wrong — validating a declared-but-untrusted hook is defensible, and your own earlier review pass called it out as something that "should be a conscious choice." My ask is just that it *be* one: either gate on trust to match the runtime, downgrade to a warning alongside `deferred`, or keep the hard failure and say so in `docs/configuration.md` next to the paragraph this PR adds. Right now the divergence is silent, and the docs describe the check without mentioning it fires regardless of trust.

---

packages/cli/src/commands/workflow.ts
```diff
@@ -848,6 +855,16 @@ function validateWorkflow(
...
+  if (hookValidation.problems.length > 0) {
+    throw new WorkflowValidationError(
+      "workflow_validation_error",
+      `hooks.${hookValidation.problems[0].hook}`,
+      `Invalid WORKFLOW.md hook path${hookValidation.problems.length === 1 ? "" : "s"}: ${formatWorkflowHookPathProblems(hookValidation.problems)}.`
+    );
```
hojinzs, line 872: **P3 — `error.path` names only the first failing hook while `message` enumerates all of them.**

Reproduced against a fixture with two bad hooks (`after_create: ./scripts/nope.sh`, `before_run: /absolutely/not/here.sh`):

```json
{
  "error": {
    "code": "workflow_validation_error",
    "path": "hooks.after_create",
    "message": "hooks.after_create: Invalid WORKFLOW.md hook paths: after_create=.../scripts/nope.sh (missing), before_run=/absolutely/not/here.sh (missing)."
  }
}
```

`--json` consumers are pointed at `error.path` by `packages/cli/README.md:282`, so automation that keys off it fixes `after_create`, re-runs, and hits `before_run` on the next pass. The message is complete, so this is a papercut rather than a defect — but the field and the message disagree about scope.

Cheapest fix that keeps the single-path contract: when `problems.length > 1`, use the parent path `hooks` instead of `hooks.${problems[0].hook}` so the path is honest about covering more than one field. Your call whether it's worth the churn.

---

packages/cli/src/commands/workflow.ts
```diff
@@ -848,6 +855,16 @@ function validateWorkflow(
...
     : ("skip" as const);
   const effectiveTimeouts = resolveWorkflowRuntimeTimeouts(workflow);
+  const hookValidation = await validateWorkflowHookPaths(workflow.hooks, {
+    workflowDirectory: dirname(workflowPath),
+  });
+  if (hookValidation.problems.length > 0) {
+    throw new WorkflowValidationError(
+      "workflow_validation_error",
+      `hooks.${hookValidation.problems[0].hook}`,
+      `Invalid WORKFLOW.md hook path${hookValidation.problems.length === 1 ? "" : "s"}: ${formatWorkflowHookPathProblems(hookValidation.problems)}.`
```
hojinzs: **P3 — the failure carries a diagnosis but no remediation.**

`doctor` fails the same condition with a fix hint (`doctor.ts:1309`):

> Create the referenced hook script(s), make them executable, or fix the hook path(s).

Here the user gets only the diagnosis. Observed live:

```
$ gh-symphony workflow validate --file WORKFLOW.md
hooks.after_create: Invalid WORKFLOW.md hook path: after_create=/…/after-create.sh (not executable).
```

`(not executable)` is precisely the reason where a one-line hint (`chmod +x`) saves a round trip, and `(not a path)` is the one where the rule — a bare script path, no shell syntax or arguments — is not guessable from the message. Appending the same sentence `doctor` uses would make the two surfaces say the same thing about the same failure.

---

packages/cli/src/commands/workflow.ts
```diff
@@ -940,7 +961,13 @@ Hooks
   after_run=${report.summary.hooks.afterRun ?? "unset"}
   before_remove=${report.summary.hooks.beforeRemove ?? "unset"}
   hooks.timeout_ms=${report.summary.hooks.timeoutMs}
+  path_checks=${report.checks.hookPaths.checked} checked, ${report.checks.hookPaths.deferred} deferred
```
hojinzs: **nit — `path_checks=0 checked, 0 deferred` prints for workflows with no hooks at all.**

Verified on a hook-free `WORKFLOW.md`:

```
Hooks
  after_create=unset
  before_run=unset
  after_run=unset
  before_remove=unset
  hooks.timeout_ms=60000
  path_checks=0 checked, 0 deferred
```

Four `unset` lines already say it. `doctor` special-cases this (`doctor.ts:1282`: "No WORKFLOW.md hooks are configured."). Skipping the line when `checked + deferred === 0` would keep the common case quiet — the JSON report still carries the zeros for automation. Purely cosmetic; ignore if you prefer the line to be unconditional.

---

hojinzs (approved):
## Approve — requirements met, nothing blocking in the code. One mechanical blocker before merge.

Reviewed at `5bb400d` from a clean install: `pnpm install --frozen-lockfile`, `pnpm build`, `pnpm lint`, `pnpm typecheck` all pass, and `packages/cli/src/commands/workflow.test.ts` is 29/29. I then ran the **built binary** against eight hand-built `WORKFLOW.md` fixtures rather than trusting the unit tests.

### Blocking for merge (not code): `README.md` conflicts with `main`

`mergeable_state` is `dirty`. Confirmed the scope:

```
$ git merge-tree --write-tree HEAD origin/main
CONFLICT (content): Merge conflict in README.md
Auto-merging docs/architecture.md
Auto-merging docs/configuration.md
Auto-merging packages/cli/README.md
Auto-merging packages/cli/src/commands/workflow.ts
Auto-merging packages/cli/src/commands/workflow.test.ts
```

One docs file, and every code file merges cleanly — `main` moved under the `workflow validate` paragraph this PR also edits. Mechanical, but it has to be resolved before this can land.

### Requirements — #948 is fully covered

| #948 asks for | status |
| --- | --- |
| Report invalid path syntax | ✅ `bash ./scripts/setup.sh` → `after_create="bash ./scripts/setup.sh" (not a path)`, exit 1 |
| Report immediately resolvable missing scripts | ✅ exit 1, `(missing)` |
| Report non-executable scripts | ✅ mode `0644` → `(not executable)`, exit 1 |
| Reuse the shared helpers | ✅ `validateWorkflowHookPaths` / `formatWorkflowHookPathProblems` / `formatDeferredWorkflowHookPaths` imported from core |
| **Without duplicating hook-resolution semantics** | ✅ verified below |

That last row is the one worth checking rather than taking on trust, and it holds. `validateWorkflowHookPaths` delegates to the same `resolveWorkflowHookPath` the runtime calls at `service.ts:4983`, so `after_create` bases on the WORKFLOW.md directory and the other three on the repository directory. The command passes only `workflowDirectory`, so relative `before_run`/`after_run`/`before_remove` resolve to `null` and are **deferred, not failed** — which is the false-positive trap this design had to avoid:

```
$ gh-symphony workflow validate --file deferred/WORKFLOW.md
  path_checks=1 checked, 2 deferred
  Deferred 2 repository-relative workflow hooks until an issue workspace is available; …
$ echo $?   # 0
```

A repo-relative `before_run` pointing at a nonexistent path exits **0**, while `after_create` in the same block is still checked. Correct.

The `not_file` branch that had no coverage anywhere in the repo is now both tested and reachable for real — pointing `after_create` at a directory reports `(not file)`.

### The `--json` contract from the last round holds up

I re-verified the fix rather than reading the resolution note. Failure:

```json
{
  "error": {
    "code": "workflow_validation_error",
    "path": "hooks.after_create",
    "message": "hooks.after_create: Invalid WORKFLOW.md hook path: … (missing)."
  }
}
```

stdout parses, exit 1, and `workflow_validation_error` is an existing member of the `CliErrorCode` union (`cli-error.ts:7`) rather than a new string — so the stable-code claim in `packages/cli/README.md:282` is now true. Success mode carries the counts additively under `checks.hookPaths`, so no existing `--json` field changed shape.

### No over-engineering; no undeclared decisions

`checks.hookPaths` mirrors how `promptFresh`/`continuationGuidance` already report, and the deferred notice exists to explain why some hooks weren't checked — without it, `1 checked` on a four-hook block would read as a silent pass. The four doc edits track behavior this PR actually changes. Changeset is a correct `patch` on `@gh-symphony/cli`. The repo's own `WORKFLOW.md` (`after_create: hooks/after_create.sh`) still validates clean at `1 checked, 0 deferred`, so this doesn't regress the project on itself.

### Findings

Four inline, none blocking:

- **P2** — `workflow validate` validates hook paths ungated, while all three runtime sites gate on hook trust. Since trust is opt-in, a workflow `project start` accepts by default can now hard-fail `validate`. I think the strictness is probably right; the ask is to make it a stated decision in `docs/configuration.md` rather than a silent divergence.
- **P3** — `error.path` names only the first failing hook while `message` lists all of them.
- **P3** — the failure has no remediation hint, where `doctor` gives one for the identical condition.
- **nit** — `path_checks=0 checked, 0 deferred` prints even when no hooks are configured.

Approving. Fold in whichever of the four you agree with, resolve the `README.md` conflict, and this is good to land — please don't hold it for another full round on the P3s.

---

moncher-dev (commented):
## Not landing — approval binds, but the conflict falls outside the Land trivial set

**Approval is valid.** Head `5bb400d6`, and hojinzs's `APPROVED` review carries `commit_id=5bb400d6` (04:06:28Z). That gate passes.

**The merge does not.** `mergeable: CONFLICTING` — `4917816f` (#949) landed underneath this branch. I verified the scope myself rather than trusting the API field:

```
$ git merge-tree --write-tree --name-only origin/main 5bb400d6
README.md
Auto-merging README.md
CONFLICT (content): Merge conflict in README.md
Auto-merging docs/architecture.md
Auto-merging docs/configuration.md
Auto-merging packages/cli/README.md
Auto-merging packages/cli/src/commands/workflow.test.ts
Auto-merging packages/cli/src/commands/workflow.ts
```

One conflicted file: **`README.md` at the repository root**. Every other file — including both source files — auto-merges.

### Why this is rework and not in-cycle Land recovery

`WORKFLOW.md` enumerates the trivial set exhaustively: `.changeset/*`, `docs/**`, `CHANGELOG.md`, `pnpm-lock.yaml` (Land failure classification item 4), restated in the lifecycle table as *"Only base-branch merge commits and conflict resolution inside `.changeset/`, `docs/`, or the lockfile are permitted here."* The repository-root `README.md` is in none of those paths. `docs/**` does not cover it. So a `Land` cycle would classify this as a rework failure and route here anyway — moving to `Land` first would only burn a cycle.

### The resolution is semantic, not mechanical — do not pick a side

```
<<<<<<< origin/main
... each explicit runtime timeout takes precedence over its legacy `codex.*_timeout_ms`
counterpart, omitted runtime fields inherit legacy values, and documented defaults apply
when neither is configured. The source printed beside each field reflects that field's
precedence result, including mixed-source timeout blocks.
=======
... validates hook path syntax and any immediately resolvable hook scripts, ... Repository-
relative run hooks are reported as deferred until an issue workspace is available. Its
`runtime.timeouts.*` values are the effective runtime settings: `runtime.timeouts` takes
precedence over the legacy `codex.*_timeout_ms` fields, with documented defaults used when
neither is configured.
>>>>>>> 5bb400d6
```

Both sides rewrote the same paragraph. Taking **theirs** silently reverts #949's per-field fallback documentation; taking **ours** silently drops this PR's hook-validation sentence. The correct resolution keeps *both*: #949's per-field precedence wording **and** this PR's hook-validation sentences.

There is an external check on the right answer. The auto-merged result already carries #949's per-field wording at `docs/configuration.md:87` and `packages/cli/README.md:173` ("Each explicit runtime timeout takes precedence over its legacy `codex.*_timeout_ms` counterpart; omitted runtime fields inherit legacy…"). If `README.md` ends up saying something different, the repository will document two contradictory precedence rules in three files. **Make `README.md` agree with those two.**

### Scope for this cycle — the conflict only

1. `git merge origin/main` on the branch (never rebase, amend, or force-push).
2. Resolve `README.md` as above; touch no other file.
3. `/push`, then hand back to `In review`.

Note that the merge commit moves the head, so hojinzs's approval no longer binds and re-approval is required before Land. That is the expected cost of this route.

---
Verified this cycle: `merge-tree` conflict scope, the conflicted hunk, and the auto-merged content of `docs/configuration.md` / `packages/cli/README.md`.

moncher-dev (commented):
### Scope correction — read hojinzs's approval body before starting

I wrote "touch no other file" above. That is narrower than what the approving reviewer actually asked for, and the reviewer's instruction wins. From the `APPROVED` review at `5bb400d6`:

> Fold in whichever of the four you agree with, resolve the `README.md` conflict, and this is good to land — **please don't hold it for another full round on the P3s.**

So the cycle scope is:

1. **Required** — `git merge origin/main`, resolve `README.md` keeping *both* sides as described above. Do not rebase, amend, or force-push.
2. **Optional, your judgment** — the four non-blocking findings in hojinzs's approval body: the **P2** (state the `validate`-strictness vs. runtime-hook-trust divergence as a decision in `docs/configuration.md` — hojinzs leans toward the current strictness being right, so this is a documentation note, not a behavior change), and three P3/nit items (`error.path` naming only the first failing hook; no remediation hint on failure where `doctor` has one; `path_checks=0 checked, 0 deferred` printing with no hooks configured).

Take the P2 and any P3s you agree with in this same cycle. Do not open a new round for the ones you decline — say which you declined and why in the workpad.

Nothing else changes: no new behavior, no scope beyond the above. Read the four inline comments on this PR for the exact locations.

packages/cli/src/commands/workflow.ts
moncher-dev: Addressed in ebaf2a6. `docs/configuration.md` now explicitly states that `workflow validate` checks declared hook paths regardless of runtime hook trust; trust controls execution, while validation checks the declaration.

packages/cli/src/commands/workflow.ts
moncher-dev: Addressed in ebaf2a6. A single failing hook retains `hooks.<kind>`, while multiple hook failures now use the honest parent path `hooks`; focused JSON coverage verifies both diagnostics remain in the message.

packages/cli/src/commands/workflow.ts
moncher-dev: Addressed in ebaf2a6. Hook-path failures now append the same actionable remediation used by `doctor`: create the scripts, make them executable, or fix the paths. The CLI test asserts the hint.

packages/cli/src/commands/workflow.ts
moncher-dev: Addressed in ebaf2a6. Human-readable output omits `path_checks` when no hooks are configured, while the structured JSON report continues to expose zero checked/deferred counts. The hook-free test now asserts the quieter output.

hojinzs (approved):
## Approve — re-approving at `ebaf2a6`. Conflict resolved correctly, all four findings landed, nothing new. Ready to land.

Re-reviewed at `ebaf2a6` from a clean install (`pnpm install --frozen-lockfile`, `pnpm build`, `pnpm lint`, `pnpm typecheck` all pass; 589 CLI tests green). I ran the **built binary** against five fresh fixtures and mutated the three new code changes rather than reading the resolution notes.

### The `README.md` conflict — resolved the right way

This was the blocker last round, and the resolution had to be semantic rather than a side-pick. It is:

```
`gh-symphony workflow validate` … validates hook path syntax and any immediately resolvable
hook scripts, … Repository-relative run hooks are reported as deferred until an issue
workspace is available. Its `runtime.timeouts.*` values are the effective runtime settings:
each explicit runtime timeout takes precedence over its legacy `codex.*_timeout_ms`
counterpart, omitted runtime fields inherit legacy values, and documented defaults apply
when neither is configured. The source printed beside each field reflects that field's
precedence result, including mixed-source timeout blocks.
```

Both sides kept: #949's per-field precedence wording **and** this PR's two hook sentences. I checked this against the external constraint I named — `README.md` now agrees with the auto-merged `docs/configuration.md` and `packages/cli/README.md`, so the repo does not document two contradictory precedence rules. `git merge-tree --write-tree origin/main HEAD` produces a tree with no conflict, and CI is green on both checks (`Test`, `Container Smoke`).

### All four findings verified against the built binary

| finding | verified |
| --- | --- |
| **P2** — `validate` strictness vs. runtime hook trust | ✅ stated as a decision in `docs/configuration.md`: *"This declaration check runs regardless of whether workflow hook execution is trusted; trust controls whether runtime hooks execute, not whether an explicitly declared path is valid."* Behavior unchanged, divergence no longer silent — which is what I asked for. |
| **P3** — `error.path` named only the first failing hook | ✅ two bad hooks → `"path": "hooks"` with both `after_create=` and `before_run=` in `message`; one bad hook still → `"path": "hooks.after_create"` |
| **P3** — no remediation hint | ✅ `… (not executable). Create the referenced hook script(s), make them executable, or fix the hook path(s).` — same sentence `doctor` uses, present in both human and `--json` output |
| **nit** — `path_checks=0 checked, 0 deferred` on hook-free workflows | ✅ line suppressed; the four `unset` lines stand alone. `checks.hookPaths` still carries the zeros in `--json` |

### Mutation check — the fixes do not survive their own deletion

Each of the three code changes is load-bearing, which is what makes them safe to keep:

```
always use `hooks.${kind}` (kill parent-path fix)  → × uses the parent hooks path when multiple hook paths fail        (1 failed | 30 passed)
drop the remediation sentence                      → × rejects 'invalid path syntax' / × rejects 'missing script'      (2 failed | 29 passed)
print path_checks unconditionally (kill the nit)   → × validates a workflow file with strict prompt and continuation   (1 failed | 30 passed)
```

### The false-positive trap still holds after the merge

The one thing worth re-checking after pulling `main` in is that the deferred branch did not regress. It did not — a repo-relative `before_run` pointing at a nonexistent path exits **0** while `after_create` in the same block is still checked:

```
$ workflow validate --file deferred/WORKFLOW.md
  path_checks=1 checked, 1 deferred
  Deferred 1 repository-relative workflow hook until an issue workspace is available; …
$ echo $?   # 0
```

And the repo's own `WORKFLOW.md` still validates clean at `1 checked, 0 deferred`, so this does not regress the project on itself.

### Requirements (#948) — fully covered, no scope creep

Invalid path syntax, immediately resolvable missing, non-file, and non-executable scripts all report and exit 1; the shared `validateWorkflowHookPaths` / `formatWorkflowHookPathProblems` / `formatDeferredWorkflowHookPaths` helpers are reused, so hook-resolution semantics are not duplicated — the *"without duplicating"* clause is the load-bearing one and it holds. This round added only the four requested items plus docs; no new behavior, no undeclared decisions. Changeset is a correct `patch` on `@gh-symphony/cli`.

### Non-blocking, and not this PR's

`pnpm format` fails with *"Code style issues found in 64 files"* — I reproduced the **identical** failure on a clean `origin/main` worktree, and every file this PR touches passes `prettier --check`. Pre-existing repo hygiene, unrelated to this branch. Worth a separate issue; explicitly not a reason to hold this.

Nothing blocking. Approving at `ebaf2a6` — please land it rather than opening another round.

---

Issue #966, opened 2026-09-14
test(e2e): repair the standalone scenario runner end to end (jq, stale-policy doctor, broken-project overlap)


moncher-dev:
## Problem

`./e2e/run-standalone-project-e2e.sh` passes Docker preflight and builds the current E2E image, but the scenario exits before dispatch because the image does not contain `jq` while the pre-existing stale-policy assertions invoke it.

Reproduced twice in isolation on 2026-09-14:

```text
bash: line 73: jq: command not found
```

The runner exits 127 before reaching later scenario assertions, including the host-publish coverage added for #940.

## Expected

The E2E image supplies every executable used by the standalone runner, or the runner performs its JSON assertions with an already packaged runtime such as Node.js.

## Context

- Discovered while validating #940.
- Docker daemon was reachable; this is not the documented preflight status-69 exception.
- The failing `jq` calls predate #940's changed lines.

moncher-dev:
**Steward: scope expanded to "the standalone runner passes end to end". Adding `jq` alone would move the silent failure to line 101. Promoted to Ready. #940 is now blocked by this issue.**

On 2026-09-14 I ran `./e2e/run-standalone-project-e2e.sh` against PR #967's head (Docker 29.6.2 was reachable) and repaired the runner locally one defect at a time. Nothing was pushed. I found three independent defects, and each one hides the next:

1. **`jq` missing.** Exit 127 at outer line 93 (`bash: line 73`), the first stale-policy `jq -e` from #955. `docker/orchestrator-e2e.Dockerfile:22-23` installs only `git curl`. Reproduced.
2. **#955's doctor assertion (lines ~100-108) fails once it runs.** `node … --json doctor --project-dir "$policy_project"` **exits 1**, which kills the script under `set -e`. Inside the container, `gh_installation`, `gh_authentication`, `gh_scopes` and `workspace_root` fail, and `workflow_file` reports "WORKFLOW.md could not be parsed". **No `workflow_source_identity` check appears in the JSON at all**, so the `jq -e` predicate would fail even without `set -e`. The status assertion just before it passes: `relationship` is `stale-copy`, the revisions differ, and a "diverged copy" warning is present.
[...]

Issue #956, opened 2026-09-08
test(e2e): provide reachable Docker daemon for agent runs


moncher-dev:
## Problem

The issue #952 Completion Bar attempted `./e2e/run-standalone-project-e2e.sh` twice, but the run environment's Docker daemon was unreachable before the scenario started.

## Evidence

Both attempts reported:

```text
[e2e] Docker runtime unavailable: Docker Compose is installed, but the Docker daemon is not reachable.
Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
```

The preflight also reported an invalid staged `docker-init` CLI plugin under the run-scoped child home. The runner exited before building an image or executing TC-25.

## Expected

Agent runs that require Docker E2E should receive a reachable isolated Docker daemon, or the workflow should provide a supported external runner handoff for the Docker Completion Bar.

## Context

- Discovered while validating #952 / PR #955.
- Unit, lint, typecheck, and build gates completed independently.

moncher-dev:
Board intake: added to the project in **Backlog**, marked `blocked_by #914`, and assigned to @hojinzs.

This is the same condition as #914 (Docker environment decision), which is now the thirteenth report of an unreachable Docker daemon during a worker's Completion Bar. It is not independently actionable by a worker: the fix is an environment decision — give agent runs a reachable isolated Docker daemon, or define a supported external-runner handoff so the Docker Completion Bar can be satisfied (or explicitly waived) from inside a run. That decision is @hojinzs's and also gates #940.

One detail here is new and worth keeping rather than folding away: the preflight additionally reported **an invalid staged `docker-init` CLI plugin under the run-scoped child home**. That is adjacent to `f51cd3ea fix(runtime): stage Docker CLI plugins for workers`, so the plugin staging may be faulty independently of daemon reachability. Worth checking when #914 is decided.

Practical effect meanwhile: PR #955 could not run `./e2e/run-standalone-project-e2e.sh`, so `e2e/scenarios/25-workflow-copy-identity.md` is unexecuted. I reviewed #955 by exercising the built detector against the real repository and its real git history instead, and the blocking finding I filed there does not depend on the E2E.

moncher-dev:
[...]

Issue #938, opened 2026-09-06
fix(e2e): restore Docker daemon availability for worker validation


moncher-dev:
## Problem

The standalone-project Docker E2E runner cannot execute on the Symphony worker host because Docker Compose is installed but the Docker daemon is unreachable at `/var/run/docker.sock`.

Observed twice while validating #929 with `pnpm e2e:standalone-project`:

```text
[e2e] Docker runtime unavailable: Docker Compose is installed, but the Docker daemon is not reachable.
Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
```

Both attempts exited with code 69 before any scenario code ran.

## Expected

Workers that require Docker E2E validation should have a reachable Docker daemon, or the runner should provide an equivalent supported execution path.

## Context

Discovered as an unrelated environment exception during #929. The new scenario and unit coverage remain in that delivery PR.

moncher-dev:
Closing as a duplicate of #914.

This is the **tenth** worker report of the same condition — the Docker daemon is not reachable from the Symphony worker environment, so the E2E runner exits at its documented Docker preflight before any container is created. #934 and #937 were consolidated onto #914 for the same reason; this one was filed within ten minutes of #937.

The report is accurate and correctly classified as an external test-host condition, not a repository defect.

**#914 is assigned to @hojinzs and is awaiting his environment decision** — either make a Docker daemon reachable to validation workers, or have the orchestration environment explicitly classify Docker E2E as unavailable so workers stop rediscovering and re-filing it. Until then, a worker that hits this should note it on its own issue and continue rather than opening a new one.

Superseded by #914.

PR #967, opened 2026-09-14
test(e2e): assert hook workspace publication

### LINKED ISSUE #940: test(e2e): assert host publish from hook-created workspace
## Problem

The standalone Docker E2E flow installs `e2e/seed/hooks/after_create.sh` and verifies hook-driven workspace population, but it does not exercise the host-owned publish transport by fetching the worker branch out of that hook-created workspace.

Issue #930 added deterministic behavioral coverage around the shipped default hook and aligned the E2E seed hook. A dedicated Docker assertion is still valuable because divergence in the packaged/seeded runtime path could otherwise leave the integration scenario green while host publication fails.

## Scope

- Extend the standalone Docker E2E flow (or add a focused scenario) to create a worker branch in the real hook-created workspace.
- Perform the host-equivalent fetch from that workspace into an external host repository.
- Verify the fetched branch and its blob are readable.
- Ensure the scenario fails for a promisor clone missing a required blob.
- Update the E2E scenario table in `AGENT_TEST.md`.

## Acceptance criteria

- The Docker black-box suite proves host publication from a workspace produced by the installed `after_create` hook.
- The assertion fails with the `lazy fetching disabled` / `bad pack header` class of failure when the workspace is reverted to an unsafe partial clone.

Follow-up from #930 and PR #933 review.

moncher-dev:
## Issues

- Closes #940

## Summary

- TL;DR: Extend the standalone Docker black-box suite to publish a committed worker branch from the real workspace populated by the installed `after_create` hook.
- Fetch that branch into an external bare host repository and prove the fetched worker blob is readable.
- Make the fixture mutation-sensitive: its `file://` transport honors partial-clone filters, and an unsafe hook mutation fails with the production `lazy fetching disabled` / `bad pack header` signature.

## Change-point diagram

- filter-capable seed repository → installed `after_create` hook → beta workspace
- beta worker commit → host-equivalent fetch → external bare repository → readable blob
- unsafe `--filter=blob:none` mutation → missing historical policy blob → host fetch fails

## Start here

- `e2e/run-standalone-project-e2e.sh:59` — configures the filter-capable seed transport used by the installed hook.
- `e2e/run-standalone-project-e2e.sh:253` — commits, fetches, and reads the hook-workspace branch, then exercises the promisor failure control.
- `AGENT_TEST.md:251` — records the scenario in the E2E coverage matrix.

- No production runtime behavior changes. The standalone Docker E2E suite now detects regressions where hook-created workspaces cannot be published by the host transport.

## Validation

- `pnpm lint` — pass
- `pnpm test` — pass across all 14 public workspace projects
- `pnpm typecheck` — pass
- `pnpm build` — pass
- `./e2e/run-standalone-project-e2e.sh` — pass with the shipped safe hook
- Same Docker scenario with `e2e/seed/hooks/after_create.sh` mutated to `git clone --filter=blob:none` — expected failure at the real beta-workspace host fetch with `lazy fetching disabled` and `fatal: protocol error: bad pack header`

## Changeset

- Not needed because this test-only change has no changeset label and changes no released package behavior.

## Risks & rollback

- Risk is limited to the standalone Docker fixture using `file://` rather than a plain local path; this intentionally preserves local isolation while making Git filter semantics real.
- Roll back commits `7d6a1da` and `f64fd00` to remove the added scenario coverage.

## Changed files

- `e2e/run-standalone-project-e2e.sh` — adds filter-capable fixture transport, hook-workspace host publication, blob-read verification, clone-completeness assertions, and the promisor failure control.
- `e2e/scenarios/13-standalone-project-model.md` — documents the positive and negative scenario steps and outcomes.
- `AGENT_TEST.md` — maps hook-owned workspace publication to the standalone Docker scenario.

- [ ] None required beyond normal PR review.

## Security


moncher-dev (commented):
**Steward review at `f64fd00e`: changes needed. This PR cannot demonstrate its own coverage, and its negative control does not guard the regression #940 names. Issue #940 now depends on #966 and goes to Backlog.**

I ran `./e2e/run-standalone-project-e2e.sh` on this head. Docker 29.6.2 was reachable, so no preflight exception applied. I then repaired the runner locally in a scratch worktree, one defect at a time, to see what the new block actually does. None of those local patches were pushed.

### 1. The new assertions never execute, for three reasons, not one
| Run | Change (local only) | Result |
| --- | --- | --- |
| 1 | none | **exit 127**, `bash: line 73: jq: command not found`. That is outer line 93, the first #955 stale-policy `jq -e`. The image (`docker/orchestrator-e2e.Dockerfile:22-23`) installs only `git curl`. This PR's block starts at line 231. |
| 2 | add `jq` to the image | **exit 1, silent.** The status `jq -e` passes. Then `gh-symphony --json doctor --project-dir <stale-policy>` **exits 1** under `set -e`: in the container `gh_installation`, `gh_authentication`, `gh_scopes`, `workspace_root` and `workflow_file` ("WORKFLOW.md could not be parsed") all fail, and **no `workflow_source_identity` check is emitted at all**. #955's doctor assertion would fail even without `set -e`. |
| 3 | also bypass the doctor assertion | **exit 1, silent.** `project-beta`'s `project start` refuses with `Tracker mapping overlaps project(s): /tmp/standalone-projects/project-broken. Re-run interactively to confirm.` Beta's `project.json` never appears, and the 40-second wait fails. `project-broken` (added by #936) stays registered after its missing-hook assertion and collides with beta. |
| 4 | also remove `project-broken`'s registration after its assertion | **exit 0.** Your block runs and passes. The host fetch reads `host-publish.txt`, and the control fails with `lazy fetching disabled` … `fatal: protocol error: bad pack header`. |

So the new assertions are correct once they are reached. But nothing reaches them in any environment: CI does not run this script at all (`ci.yml` runs only `e2e:compose-project-test` and `e2e:fixture-replacement-test`), and locally it dies at run 1. #966 as filed covers only row 1. **I'm not accepting an E2E test because it would pass if it ran.**

### 2. Blocking: the scenario stays green when the hook-created workspace is made an unsafe partial clone (#940 AC 2)
AC 2 says the assertion must fail "when the workspace is reverted to an unsafe partial clone". The negative control builds its **own** clone, with `file://` and after setting `uploadpack.allowFilter true`. The hook-created workspace meets neither condition.

**The mutation, run for real.** On top of run 4, I changed `e2e/seed/hooks/after_create.sh` to `git clone --filter=blob:none "$SYMPHONY_REPOSITORY_CLONE_URL" …`, which is the regression #940 exists to catch. Result: **`standalone-project Docker E2E passed`, exit 0.** Diagnostics on the beta workspace showed `remote.origin.promisor=true`, `missing objects=0`. `SYMPHONY_REPOSITORY_CLONE_URL` is a plain path (`/e2e/repos/test-owner/test-repo`), and Git ignores `--filter` for local-path clones ("--filter is ignored in local clones; use file:// instead", reproduced on the host). The source repository does not allow filters either. In this fixture, a partial-clone hook can never produce a partial workspace, so the positive assertion cannot turn red. The control proves only that Git fails on a promisor clone. It does not prove that this scenario would notice one.

Two ways to make it mutation-sensitive (your choice):
- Make the fixture able to produce the regression: serve the seed repository with `uploadpack.allowFilter` and a `file://` clone URL **before** dispatch. The same hook mutation must then turn the positive fetch red. Show that run.
- And/or assert directly on the hook-created workspace that it is not a promisor: `git -C "$beta_repo" config remote.origin.promisor` absent, and `git rev-list --objects --missing=print --all` reporting nothing missing. Note that Git set `promisor=true` even when it ignored the filter, so the config check alone fails under my mutation.

Either way, the evidence must be an actual run in which the hook mutation fails the scenario.

### Dependency decision
- **#940 is now `blocked_by #966` and goes to Backlog.** It can't sit in Ready while its blocker is open. I've expanded #966's scope to "the standalone runner passes end to end", covering rows 1–3 above, because shipping a jq-only fix would move the same silent death to line 101.
- **I considered accepting this PR on unit evidence and rejected that.** Its only deliverable is E2E coverage, `default-hooks.test.ts` already existed before this PR, and finding 2 is a defect in the new assertion itself, not a missing environment.
- When #966 lands, #940 returns to Ready. Rebase, fix finding 2, and record a green run plus the red mutation run.

Non-blocking: the "Validation" section calls the 127 a "documented exception". It is a defect in the runner, not an environment exception like #938 or #956, and it hid rows 2 and 3.

moncher-dev (commented):
Steward re-test at head `7d6a1da9`. Both blocking findings from my 13:16 review are resolved.

**Rebased onto the repaired runner.** The branch contains `be5f7dfc` (#968) and GitHub reports it MERGEABLE. The two blockers #914 and #966 are closed, and CI is green (Test, Container Smoke).

**Scratch worktree, one step at a time:**
- `pnpm install --frozen-lockfile && pnpm build && pnpm test` is green: cli 589, core 340, orchestrator 357, control-plane 38, dashboard 21, extension 9.
- `./e2e/run-standalone-project-e2e.sh` with the seed hook as shipped (plain `git clone`) exits **0** and prints `standalone-project Docker E2E passed`.

**The mutation that previously stayed green now fails the scenario.** I changed only `e2e/seed/hooks/after_create.sh` back to `git clone --filter=blob:none "$SYMPHONY_REPOSITORY_CLONE_URL" …`, which is the regression #940 exists to catch, and re-ran the same script. It exits **128**:
```
remote: warning: lazy fetching disabled; some objects may not be available
remote: fatal: could not fetch 130d49de… from promisor remote
fatal: protocol error: bad pack header
```
The failure is in the positive host fetch from the **hook-created beta workspace** (`git -C "$host_repo" fetch … "$beta_repo"`). It is not in the self-built control, whose output goes to `/tmp/unsafe-publish.log`. The `file://` clone URL together with `uploadpack.allowFilter` before dispatch means the hook mutation now produces a real partial workspace, and the scenario sees it. The `remote.origin.promisor`, `partialclonefilter` and missing-object assertions right after the fetch are a second line behind it.

Non-blocking: the "unsafe control" block still builds its own filtered clone. That proves Git's failure signature, not the hook's behaviour. It is harmless as documentation now that the hook-workspace assertions are mutation-sensitive.

e2e/run-standalone-project-e2e.sh
```diff
@@ -84,6 +84,7 @@ printf "STUB_SCENARIO=happy\nSYMPHONY_ALLOW_WORKFLOW_HOOKS=1\n" > "$policy_proje
 printf "\n# newer committed policy\n" >> /e2e/repos/test-owner/test-repo/WORKFLOW.md
 git -C /e2e/repos/test-owner/test-repo add WORKFLOW.md
 git -C /e2e/repos/test-owner/test-repo commit -m "Update committed workflow policy" >/dev/null
+git -C /e2e/repos/test-owner/test-repo config uploadpack.allowFilter true
```
hojinzs, line 87: **P2 — this line is load-bearing for AC 2's failure signature, but it is unlabeled and sits inside the #955 stale-policy setup block.**

I reproduced the git mechanism locally (git 2.43.0) with the mutated hook (`git clone --filter=blob:none`) under each combination of the two fixture settings this PR introduces:

| fixture | `uploadpack.allowFilter` | clone URL | host fetch | `promisor` | missing objects |
| --- | --- | --- | --- | --- | --- |
| A (this PR) | `true` | `file://` | **FAIL** (`lazy fetching disabled` / `bad pack header`) | `true` | 1 |
| B | `false` | `file://` | PASS | `true` | 0 |
| C | `true` | plain path | PASS | `true` | 0 |
| D | `false` | plain path | PASS | `true` | 0 |

So line 87 and the `file://` switch at line 59 are **jointly** required; drop either and the positive host fetch goes green again under the exact regression #940 exists to catch. Line 59 is self-documenting (it sits in the `repository:` block and the PR body explains it); line 87 is a bare `git config` wedged between the WORKFLOW.md policy-supersede commit and the fixture reset, with nothing tying it to the assertion ~180 lines later that depends on it.

Non-blocking because the `remote.origin.promisor` assertion at line 269 is the real backstop — it flips to `true` in **all four** rows above, so the scenario still fails the mutation even if this line is later removed. But AC 2 specifically requires the `lazy fetching disabled` / `bad pack header` class of failure, and that is what silently disappears.

Suggest moving it adjacent to the `clone_url` fixture setup (or keeping it here) with a comment such as:

```bash
# The hook-created workspace must be able to *become* a partial clone for the
# mutation-sensitivity assertion at L262-L271 to mean anything; together with the
# file:// clone_url (L59) this makes `--filter` real instead of silently ignored.
git -C /e2e/repos/test-owner/test-repo config uploadpack.allowFilter true
```

---

e2e/run-standalone-project-e2e.sh
```diff
@@ -250,6 +251,50 @@ beta_repo=$(find "$PROJECT_ROOT/project-beta/.runtime/workspaces" -path "*/repos
...
+git -C "$beta_repo" commit -m "test: publish worker branch" >/dev/null
+host_repo=/tmp/standalone-host.git
+git init --bare "$host_repo" >/dev/null
+git -C "$host_repo" fetch --no-tags "$beta_repo" \
+  "refs/heads/$beta_branch:refs/heads/$beta_branch"
+test "$(git -C "$host_repo" show "$beta_branch:host-publish.txt")" = \
+  "published from hook-created workspace"
+test -z "$(git -C "$beta_repo" config --get remote.origin.promisor || true)"
+test -z "$(git -C "$beta_repo" config --get remote.origin.partialclonefilter || true)"
+test -z "$(git -C "$beta_repo" rev-list --objects --missing=print --all | grep "^?" || true)"
```
hojinzs, line 271: **P3 — `|| true` makes this assertion silently pass if `rev-list` itself fails.**

The script runs under `set -euo pipefail`. Inside the command substitution, `|| true` is needed to absorb `grep`'s exit 1 on no-match, but it also absorbs a failure of `git rev-list --objects --missing=print --all`. If that command ever errors (unsupported option on an older git in a rebuilt image, a corrupt workspace, a bad `$beta_repo`), the substitution yields empty output and `test -z` passes — the assertion reports "no missing objects" for a repository it never successfully traversed.

Splitting the two effects keeps the no-match case green while letting a real `rev-list` failure fail the scenario:

```bash
missing_objects=$(git -C "$beta_repo" rev-list --objects --missing=print --all)
test -z "$(printf '%s\n' "$missing_objects" | grep "^?" || true)"
```

Same class of concern does not apply to lines 269-270 — for `git config --get`, a non-zero exit *is* the passing condition.

---

e2e/run-standalone-project-e2e.sh
```diff
@@ -250,6 +251,50 @@ beta_repo=$(find "$PROJECT_ROOT/project-beta/.runtime/workspaces" -path "*/repos
...
+printf "unsafe publish\n" > "$unsafe_repo/unsafe-publish.txt"
+git -C "$unsafe_repo" add unsafe-publish.txt
+git -C "$unsafe_repo" commit -m "test: expose missing promisor blob" >/dev/null
+unsafe_host=/tmp/unsafe-host.git
+git init --bare "$unsafe_host" >/dev/null
+if git -C "$unsafe_host" fetch --no-tags "$unsafe_repo" \
+  refs/heads/unsafe-publish:refs/heads/unsafe-publish \
+  > /tmp/unsafe-publish.log 2>&1; then
+  echo "promisor workspace unexpectedly served the worker branch" >&2
+  exit 1
```
hojinzs, line 292: **nit — the stated rationale for this control is not what actually makes it fail, and the block is now redundant.**

Two separate points.

1. **The comment is inaccurate.** It claims the control works because "the policy setup above supersedes the original `WORKFLOW.md` blob", implying the missing object is specifically that superseded blob. It isn't — a `--filter=blob:none` clone has *no* blobs locally, so its `upload-pack` cannot serve **any** fetch that needs a blob, superseded or not. In my local reproduction the object git failed on was simply the first blob the packer reached. The WORKFLOW.md supersede commit is irrelevant to this block. As written the comment will send the next reader looking for a coupling that does not exist, and invites someone to "simplify" the policy setup on the belief it is load-bearing here (it is — but for line 87's `allowFilter`, not for the superseded blob).

2. **It is now redundant coverage.** @moncher-dev already flagged this as non-blocking in the re-test, and I agree: this block builds its *own* filtered clone, so it proves Git's behaviour, not the hook's. Since 7d6a1da the hook-created beta workspace is genuinely mutation-sensitive (lines 262-271), which is the assertion #940 AC 2 actually asks for. I'd read #940's Scope bullet "Ensure the scenario fails for a promisor clone missing a required blob" as satisfied by the beta assertions, making these ~20 lines optional.

Either drop the block, or keep it as a documentation fixture with a corrected comment — e.g. "A blob-filtered clone holds no blobs locally, so its upload-pack cannot serve a fetch at all; this pins the exact production failure signature the beta assertions above rely on." No change required to merge.

---

AGENT_TEST.md
```diff
@@ -248,7 +248,7 @@ Control worker behavior with the `STUB_SCENARIO` environment variable:
...
...re creates only directories and preserves reused contents; `packages/orchestrator/src/service.test.ts` covers fresh-hook cleanup, one-shot reuse, and missing-hook degradation without changing an issue retry count; `packages/cli/src/commands/{start,doctor}.test.ts` cover absent and non-executable paths.                                                                                                                                                                                                                                                                                                              | `./e2e/run-standalone-project-e2e.sh` first rejects a packaged standalone project with a missing hook, then dispatches two valid projects whose installed `after_create` hooks populate serviceable workspaces. It commits a worker blob in the hook-created beta workspace, performs the host-equivalent fetch into an external bare repository, and reads the fetched blob. A filtered-clone control missing the superseded policy blob must fail with the `lazy fetching disabled` / `bad pack header` signature.                                                                                            |
```
hojinzs, line 253: **P3 — the rewrite dropped a clause describing coverage that still exists.**

The previous text read "first rejects a packaged standalone project with a missing hook **and names its absolute path**". The new text keeps the rejection but drops the absolute-path clause, yet `e2e/run-standalone-project-e2e.sh:137` still asserts exactly that:

```bash
grep -q "$PROJECT_ROOT/project-broken/hooks/after_create.sh" /tmp/project-broken.log
```

That assertion is the operator-facing half of the fault path (a missing-hook error is only actionable if it names the path), and this table is where that mapping is recorded. Nothing in this PR touches line 137, so the removal looks incidental to making room for the new sentences rather than intentional.

Suggest restoring the clause: "…first rejects a packaged standalone project with a missing hook and names its absolute path, then dispatches two valid projects…".

Second, smaller point in the same cell: "A filtered-clone control missing the superseded policy blob must fail…" carries over the inaccurate rationale I noted on the runner. It also describes only the *self-built* control, and so under-sells what this PR actually delivers — the mutation-sensitivity now lives in the **hook-created beta workspace** assertions, which is the part #940 AC 2 asks for. Worth rewording to say the beta workspace is asserted non-promisor and complete, with the filtered-clone control pinning the failure signature.

---

hojinzs (approved):
## Approve — requirements met, mutation sensitivity independently verified

Reviewed at head `7d6a1da9`. All findings below are non-blocking (P2/P3/nit); none need to land before merge.

### Independent verification

Docker was not available in my environment, so rather than take the re-test on faith I reproduced the underlying git mechanism directly (git 2.43.0): a seed repo with a superseding `WORKFLOW.md` commit, the shipped `after_create.sh` verbatim, a worker commit, and a host-equivalent bare fetch.

- **Shipped hook (`git clone`)** → host fetch succeeds, `host-publish.txt` reads back, `remote.origin.promisor` and `partialclonefilter` unset, 0 missing objects.
- **Mutated hook (`git clone --filter=blob:none`)** → host fetch fails with exactly `remote: warning: lazy fetching disabled` … `fatal: protocol error: bad pack header`, `promisor=true`, 1 missing object.

That is #940 AC 2's required signature, produced by the positive assertion against the hook-created workspace rather than by the self-built control. The blocking finding from the 13:16 review is genuinely resolved, not just relocated.

I also ran the counterfactual matrix (mutated hook under each combination of the two fixture settings this PR introduces) — full table on the `L87` comment. The headline: `file://` (L59) and `uploadpack.allowFilter` (L87) are **jointly** load-bearing for the fetch failure, but `remote.origin.promisor` flips to `true` in all four configurations. So the config assertion at L269 is a stronger backstop than the PR body credits it for — the scenario would still catch this regression even if the transport setup regressed. That is a real point in the design's favour.

### Requirements traceability (#940)

| #940 requirement | Status |
| --- | --- |
| Worker branch created in the real hook-created workspace | ✅ beta workspace, `L262-268` |
| Host-equivalent fetch into an external host repository | ✅ `L266-267` |
| Fetched branch and blob readable | ✅ `L268` |
| Scenario fails for a promisor clone missing a required blob | ✅ verified above |
| `AGENT_TEST.md` scenario table updated | ✅ `L251` (one clause regressed, see comment) |
| **AC 1** — Docker suite proves host publication from a hook-produced workspace | ✅ |
| **AC 2** — fails with `lazy fetching disabled` / `bad pack header` when reverted to an unsafe partial clone | ✅ |

No changeset — correct, this is test-only with no released-package behavior change. CI green (Test, Container Smoke); `mergeable_state: clean`.

One candidate: the self-built filtered-clone control (`L273-292`). Now that the beta assertions are mutation-sensitive it re-proves Git's behavior rather than the hook's. I do **not** count it as scope creep, since #940's Scope explicitly lists "Ensure the scenario fails for a promisor clone missing a required blob" — but see the nit for a rationale-comment correction. Nothing else in the diff exceeds what the issue asked for; the `file://` switch and `allowFilter` are the minimum needed to make AC 2 achievable at all.

### Findings

| # | Sev | Location | Summary |
| --- | --- | --- | --- |
| 1 | P2 | `runner.sh:87` | `uploadpack.allowFilter` is load-bearing for AC 2's signature but unlabeled inside the #955 stale-policy block |
| 2 | P2 | — | No parity assertion between `e2e/seed/hooks/after_create.sh` and `DEFAULT_AFTER_CREATE_HOOK_CONTENT` (below) |
| 3 | P3 | `runner.sh:271` | `\|\| true` can silently no-op the missing-object assertion |
| 4 | P3 | `runner.sh:157` | alpha recovery clone left on plain-path transport (below) |
| 5 | P3 | `AGENT_TEST.md:251` | dropped "and names its absolute path", still asserted at `runner.sh:137` |
| 6 | nit | `runner.sh:273-292` | control block's stated rationale is inaccurate; block now redundant |

Findings 2 and 4 land on lines outside the diff, so they are here rather than inline:

**#2 (P2) — seed-hook parity is unasserted.** I diffed `e2e/seed/hooks/after_create.sh` against `DEFAULT_AFTER_CREATE_HOOK_CONTENT` in `packages/cli/src/workflow/default-hooks.ts`: identical today. But nothing enforces that — a grep across the repo finds no test comparing them, and the seed file is a hand-maintained copy consumed by `init-repo.sh`, `init-local.sh`, and this runner. This matters specifically *here* because #940's Problem statement names this exact risk ("divergence in the packaged/seeded runtime path could otherwise leave the integration scenario green while host publication fails"). The new coverage proves host publication from the **seed** hook; if the packaged hook later gains a `--filter` or `--depth`, this scenario stays green. Pre-existing and out of scope for a test-only PR — worth a follow-up issue asserting the seed file equals the exported constant, which would close the loop #940 opened.

**#4 (P3) — `runner.sh:157`.** Lines 59 and both fixture `cloneUrl` values moved to `file://`, but the pre-seeded alpha recovery workspace still clones via the plain path, which hardlinks rather than negotiating a pack. Harmless today (alpha is never host-published) and arguably correct for cheap fixture setup, but the divergence is now silent: extending the publish assertions to alpha later — a natural follow-up, since alpha is the recovery/reuse path — would land in row C of the matrix, where the mutation passes undetected. Either switch it for consistency or add a line noting the plain path is deliberate.

Good work turning the 13:16 blockers around with a real red-mutation run. Approving; take findings 1, 3, 5 and the nit on a convenience pass or a follow-up, and file #2 separately.

---

PR #968, opened 2026-09-14
test(e2e): repair standalone project runner

### LINKED ISSUE #966: test(e2e): repair the standalone scenario runner end to end (jq, stale-policy doctor, broken-project overlap)

moncher-dev:
## Issues

- Closes #966

## Summary

- TL;DR: make the standalone-project Docker scenario runnable and assertion-complete end to end.
- Install `jq`, supply the stale-policy doctor fixture environment, and isolate the intentionally broken project registration before concurrent dispatch.
- Convert silent checks into named failures, including the alpha issue-identity log assertion that previously could not fail under Bash `set -e`.

## Change-point diagram

- E2E runtime image → installs `jq` → stale-policy JSON assertions execute
- stale-policy fixture → valid file-tracker workflow → doctor emits `workflow_source_identity`
- missing-hook fixture → assertion completes → broken registration removed → alpha/beta dispatch stays isolated
- alpha worker logs → explicit forbidden-message conditional → diagnostic output and exit 1

## Start here

- `e2e/run-standalone-project-e2e.sh:93` — stale-policy status and doctor assertions with diagnostics
- `e2e/run-standalone-project-e2e.sh:128` — broken-project validation and scoped cleanup
- `e2e/run-standalone-project-e2e.sh:259` — issue-identity log assertion that now fails explicitly
- `docker/orchestrator-e2e.Dockerfile:22` — runtime executable dependencies

- `pnpm e2e:standalone-project` now reaches every scenario and cannot report success when the alpha worker log contains an issue-identity preflight failure.

## Validation

- `pnpm lint && pnpm test && pnpm typecheck && pnpm build` — pass on merged-base head `45e87e2`
- `./e2e/run-standalone-project-e2e.sh` — pass; exit 0 and ended with `standalone-project Docker E2E passed`
- Alpha preflight-log negative control — expected fail, exit 1 with `alpha worker hit the issue identity preflight failure`
- Missing-hook negative control — expected fail, exit 1 with `missing hook unexpectedly passed project start`
- Stale-policy status negative control — expected fail, exit 1 with `stale-policy status did not report the expected diverged workflow copy`
- Stale-policy doctor negative control — expected fail, exit 1 with `stale-policy doctor did not report the expected workflow identity warning`

## Changeset

- Not needed because #966 has no changeset label and changes only the repository E2E harness/image.

## Risks & rollback

- Risk is confined to the Docker E2E image and standalone scenario sequencing/assertions. Revert the two issue commits if the extra runtime package or repaired checks cause an unexpected harness regression.

## Changed files

- `docker/orchestrator-e2e.Dockerfile` — install `jq` in the E2E runtime stage.
- `e2e/run-standalone-project-e2e.sh` — repair doctor setup, isolate the broken fixture, and make silent assertions diagnostic and enforceable.

- [ ] Decide separately whether the standalone runner should be added to CI; that policy choice is outside #966.

## Security


moncher-dev (commented):
**Steward smoke test at `49f831a9`: the runner now completes, but one assertion after line 93 still cannot fail. Blocking. The fix is one line. Returning #966 to Ready.**

### What passes (run for real, not taken from the PR body)

- `pnpm install --frozen-lockfile && pnpm build && pnpm test`: pass (cli 40 files / 589 tests).
- `./e2e/run-standalone-project-e2e.sh` with Docker 29.6.2: **exit 0**, ends with `standalone-project Docker E2E passed`. The inner script runs under `set -euo pipefail`, is linear, and has no early success exit, so reaching the final `echo` means every line after 93 ran: the stale-policy status `jq`, the doctor call (exit 1, as expected) and its `workflow_source_identity` `jq`, the missing-hook check, the broken-registration cleanup, both project registrations, the duplicate-owner refusal, the two completed stub workers, and the per-project branch, skill, MCP and foreign-work assertions.

### Blocking: line 259 cannot fail

```bash
! grep -q "Issue identity preflight failed" $alpha_logs
```

Bash exempts a pipeline that starts with `!` from `set -e`. When the grep matches, the line returns 1, the script moves on, and it still prints `passed`.

**Repro (negative control, real container run):** I made a scratch copy of this runner. Immediately before line 259 it appends `Issue identity preflight failed` to alpha's `worker.log`:

```text
MUTATION_INJECTED
1                                   <- grep -c: the string is in the log
mut_exit=0
```

The same thing happens outside Docker: `bash -c 'set -euo pipefail; ! grep -q x <<<"x"; echo reached'` prints `reached`.

The line came from #935 (`ebe18de3`) and has never been able to fail. It is in scope here because #966 asks for the runner to be repaired end to end, and this PR already converts the other silent assertions into named failures. Suggested fix, in the same style as the rest of the PR:

```bash
if grep -q "Issue identity preflight failed" $alpha_logs; then
  echo "alpha worker hit the issue identity preflight failure" >&2
  cat $alpha_logs >&2
  exit 1
fi
```

Show the negative control after the fix: inject the string, and confirm the run exits 1 with that message. `grep -rnE '^[[:space:]]*! ' e2e` finds no other bare negated command; the `!` on line 186 is inside an `if` condition and is fine.

### Non-blocking notes

- The doctor step requires exit status **exactly 1** "from unrelated environment checks" (`gh_installation`, `gh_authentication`, and so on, which fail in the container). If the image ever satisfies those checks, this step fails even though nothing regressed. It fails loudly, so it is acceptable, but the comment should name the checks it depends on.
- `project-stale-policy/.env` is written without `chmod 600` (the runner warns about it in the log). Cosmetic.

moncher-dev (commented):
Re-review at head `45e87e25`. My blocking finding from `49f831a9` is resolved. Tested in a scratch worktree with Docker.

**Clean run.** `./e2e/run-standalone-project-e2e.sh` exits 0 and prints `standalone-project Docker E2E passed`.

**Negative control, line 259 (the original finding).** I appended `Issue identity preflight failed: INJECTED-BY-REVIEW` to each alpha `worker.log` just before the guard and re-ran. The runner now exits 1 with `alpha worker hit the issue identity preflight failure` and dumps the log. On `49f831a9` the same injection printed `passed`. The guarded string is real: the worker emits it verbatim at `packages/worker/src/index.ts:538`.

**Other `!` negations.** A grep for `^\s*!`, `&& !`, `|| !` and `; !` finds one more, at line 186. It sits inside the `if` condition of the registration check, so `set -e` exemption does not apply. Negative control: I deleted beta's `project.json` right after the wait loop, and the runner exits 1 with `standalone projects did not both register within 40 seconds`. The remaining `test ... && break` lines are inside polling loops, and each is followed by an unconditional assertion (lines 198 and 215).

Minor, non-blocking: the `doctor_status -ne 1` check pins doctor's exit code to the container missing `gh` and a workspace root. If a future image ships `gh`, this will fail loudly rather than silently. That is the right direction, but worth knowing.

Passing. Requesting hojinzs review.

e2e/run-standalone-project-e2e.sh
```diff
@@ -90,23 +90,40 @@ printf "[]\n" > "$FIXTURE"
...
+GH_SYMPHONY_FILE_TRACKER_ISSUES_PATH="$FIXTURE" \
+  node /app/packages/cli/dist/index.js --config "$CONFIG_DIR" --json doctor \
+    --project-dir "$policy_project" > /tmp/stale-policy-doctor.json
+doctor_status=$?
+set -e
+if [ "$doctor_status" -ne 1 ]; then
+  echo "stale-policy doctor unexpectedly exited with status $doctor_status (expected 1 from gh_installation, gh_authentication, gh_scopes, and workspace_root checks)" >&2
+  cat /tmp/stale-policy-doctor.json >&2
+  exit 1
+fi
```
hojinzs, line 114: **P2 — this gate couples the scenario to four checks it does not guard.**

`doctor` exits `0` when every check passes, `1` when any check fails, and `2` only when doctor itself errors (`packages/cli/src/commands/doctor.ts:3105,3112`). Pinning the expectation to exactly `1` makes this block depend on `gh_installation`, `gh_authentication`, `gh_scopes` and `workspace_root` *continuing to fail* inside the container — none of which is what the stale-policy scenario is about. The error message having to enumerate them is itself the tell.

I simulated the gate locally:

```
doctor exit=0   -> FAIL: "unexpectedly exited with status 0"
doctor exit=1   -> proceeds to jq predicate
doctor exit=2   -> FAIL
doctor exit=127 -> FAIL
```

So the day the E2E image gets a GH token, or `workspace_root` starts passing, this fails spuriously while `workflow_source_identity` is still warning exactly as intended.

Widening to the crash case keeps the exit-2 detection you wanted without the coupling:

```suggestion
if [ "$doctor_status" -gt 1 ]; then
  echo "stale-policy doctor errored with status $doctor_status (expected 0 or 1)" >&2
  cat /tmp/stale-policy-doctor.json >&2
  exit 1
fi
```

The `jq -e` predicate below is the real assertion and keeps its diagnostics either way. Non-blocking.

---

e2e/run-standalone-project-e2e.sh
```diff
@@ -117,6 +134,8 @@ if (cd "$PROJECT_ROOT/project-broken" && \
 fi
 grep -q "Project configuration fault" /tmp/project-broken.log
 grep -q "$PROJECT_ROOT/project-broken/hooks/after_create.sh" /tmp/project-broken.log
+broken_id=$(node -e "const {createHash}=require(\"crypto\");const d=\"$PROJECT_ROOT/project-broken\";console.log(\"project-broken-\"+createHash(\"sha256\").update(d).digest(\"hex\").slice(0,8))")
+rm -rf "$CONFIG_DIR/projects/$broken_id"
```
hojinzs, line 138: **P3 — guard the cleanup, and record why it exists.**

The fix is correct, and I verified both halves of it. The derivation matches the product today:

```
script inline        : project-broken-02fe9080
standaloneProjectId(): project-broken-02fe9080   ✅
```

and removing the registration is the right lever: `findOverlappingProjects` (`packages/cli/src/commands/project.ts:222`) iterates `listStandaloneProjects(configDir)`, so the `$CONFIG_DIR` entry is the only thing making `project-broken` visible to beta. #936's missing-hook assertions run before the removal, so that coverage stays intact.

Two small things:

**1. Assert it did something.** This is the third hand-rolled copy of `standaloneProjectId` in this file, and the first one used as an argument to `rm -rf`. If the id scheme changes, the cleanup silently becomes a no-op and the overlap regression returns — landing on the 40-second wait you just made diagnostic, which reports the symptom two hundred lines away from this cause. One line fails it here instead (and rules out the `rm -rf "$dir/"` shape should `broken_id` ever come back empty):

```suggestion
broken_id=$(node -e "const {createHash}=require(\"crypto\");const d=\"$PROJECT_ROOT/project-broken\";console.log(\"project-broken-\"+createHash(\"sha256\").update(d).digest(\"hex\").slice(0,8))")
# project-broken stays registered after its configuration fault, and its
# mapping is not provably disjoint from beta's, so beta's start would be
# refused. alpha escapes only because broken excludes the alpha label.
test -d "$CONFIG_DIR/projects/$broken_id"
rm -rf "$CONFIG_DIR/projects/$broken_id"
```

**2. Why the comment matters:** alpha surviving is incidental, not designed. `pickupLabelsMayOverlap` (`project.ts:329`) proves disjointness only through *exclude* lists — broken excludes `alpha`, courtesy of the `test "$label" = alpha && printf beta || printf alpha` ternary on line 47, so alpha is clear; beta's include `[beta]` against broken's `[broken]` is **not** provably disjoint, since a single issue can carry both labels. Without a note here, the next person to tidy up that ternary reintroduces the bug you just fixed.

A less state-dependent alternative — give `project-broken` its own `tracker.project_id` so no overlap exists at all — would also work, but what you have is correct and I would not hold the PR for it.

---

hojinzs (approved):
## Review — requirement coverage, smoke test, code quality

**Verdict: Approve.** Everything #966 asked for is here, the Cycle 1 finding is genuinely fixed, and the two inline notes are non-blocking (P2 / P3). No further ping-pong needed — land it, and fold the P2 in whenever this file is next touched.

### #966 acceptance coverage

| Acceptance item | Status |
| --- | --- |
| `jq` present so the stale-policy assertions can run | ✅ `docker/orchestrator-e2e.Dockerfile:23` |
| Defect 2 — establish whether the missing `workflow_source_identity` check was test setup or a product defect | ✅ Correctly resolved as **test setup**: the fixture now supplies `GH_SYMPHONY_FILE_TRACKER_ISSUES_PATH` to `doctor`. No product change was warranted, and none was made — right call. |
| Defect 3 — broken fixture no longer blocks the two-project dispatch | ✅ Registration removed after the assertion; #936's missing-hook coverage runs first and stays intact |
| Silent exits named where cheap (`project.json` wait, doctor predicate) | ✅ Both, plus the status predicate |
| Negative controls for #936 and #955 status/doctor | ✅ Recorded (see nit below) |
| CI adoption left out of scope | ✅ |

### Smoke test

Docker was unavailable to me, so rather than re-run the container scenario I exercised the changed logic directly:

- `bash -n` on the outer script **and** on the extracted inner `-lc` body (12,030 bytes) — both clean.
- Both `jq` predicates against synthetic fixtures: each passes on the expected shape and **fails** when the guarded behavior is removed (status → `relationship: repository`, equal revisions, no warning; doctor → `workflow_source_identity` absent). They discriminate, they aren't tautologies.
- `broken_id` derivation vs `standaloneProjectId` — identical (`project-broken-02fe9080`).
- The Cycle 1 finding, reproduced to confirm the repair was necessary:

  ```
  # old form, dirty log, under set -e
  ! grep -q "Issue identity preflight failed" dirty.log
  -> reached the next line, exit 0    (SILENT PASS)
  ```

  errexit is ignored for a `!`-negated command, so that assertion could never have failed. The new conditional exits 1 and dumps the logs. Confirmed fixed.
- Both check runs green on `45e87e2` (Test, Container Smoke); Codex review clean.

One, and it's small: the `doctor_status -ne 1` gate is a constraint #966 never asked for, and it is the single thing in this diff that can fail for reasons unrelated to what it guards. Details inline. Otherwise the change is properly minimal — 2 files, 37 added lines, one apt package, no product code touched. Choosing `jq` over the issue's "or use Node" alternative was the cheaper option and preserves the existing #955 assertions verbatim; agreed.

No changeset is correct — nothing under `packages/` changed, so `@gh-symphony/cli` behavior is untouched.

### nit — evidence

#966 asked you to *paste* the negative-control runs. The PR body summarizes each with its exit code and failure message, which I'm accepting as sufficient, but pasted output was the literal ask.

### The CI question — my call, since #966 deferred it to me

**Yes, this runner should join CI.** `ci.yml` runs only `e2e:compose-project-test` and `e2e:fixture-replacement-test` — which is precisely why #936 and #955 shipped assertions that had never executed green anywhere and sat rotting for six days. This PR restores the assertions but not the thing that would have caught their rot, so the same failure mode is still open.

Please open a follow-up issue for adding the standalone runner to CI. Out of scope here, as #966 states, and I don't want it holding this PR.

---

PR #970, opened 2026-09-14
fix(tracker-linear): isolate malformed state-list records

### LINKED ISSUE #893: fix(tracker-linear): isolate malformed state-list records
Parent Epic: #890. Approved registration plan: [Q3: #893](https://github.com/hojinzs/github-symphony/issues/893).

## Background

One invalid required state rejects the entire Linear state-list result. The same rejection is correct for a requested-ID refresh.

## Proposed Changes

Distinguish operation purpose; skip only recognized record-validation failures in state lists; preserve bounded `skippedItems` and rate-limit metadata through label filters. Keep auth, transport, and paging failures atomic and ID refresh strict. Do not change label-routing semantics.

## Affected Files

| File | Change | Scope |
| --- | --- | --- |
| `packages/tracker-linear/src/orchestrator-adapter.ts` | Typed validation and operation-specific policy | Integration |
| `packages/tracker-linear/src/tracker-linear.test.ts` | Mixed/all-invalid lists, strict refresh, page failures, metadata | Tests |
| `docs/trackers/linear.md`, `docs/architecture.md` | Describe provider semantics and coverage | Documentation |

Mixed lists return valid candidates with diagnostics; all-invalid lists return no candidates with diagnostics; malformed requested IDs reject; empty inputs avoid requests; optional-field fallback still works. Run the Linear suite and full required checks. No live provider call is required to prove malformed fixtures.

Integration, Observability. Keep provider-specific behavior in adapters and preserve current persistence, retry, metric, and unpublished-work contracts. Do not edit `docs/symphony-spec.md`. No new spec divergence is intended.

## Planning

- Priority: P2
- Size: M

## Dependencies

## Required Checks

```bash
pnpm lint
pnpm test
pnpm typecheck
pnpm build
```

## References

- #890
- Symphony §11.1 permits skipping malformed state-list records but requires malformed requested-ID refresh to fail.

moncher-dev:
## Issues

- Closes #893

## Summary

- Isolate recognized required-field validation failures in Linear state-list reads while preserving strict requested-ID refresh and atomic unexpected failures.
- Preserve bounded skipped-record and rate-limit metadata through pickup-label filtering.
- Keep valid candidate dispatch running when a strict supplemental active-run refresh fails, preserving unresolved runs by both provider ID and stable identifier.

## Change-point diagram

- Linear GraphQL pages → operation-aware normalization → valid issues plus bounded diagnostics
- Candidate label filter → retains `rateLimits` and `skippedItems`
- Missing active issue → strict requested-ID refresh → failure logged → ID/identifier identities retained → other candidates continue without suppressing the unresolved run
- Startup terminal lookup → logs skipped diagnostics → reconciles valid terminal issues

## Start here

- `packages/tracker-linear/src/orchestrator-adapter.ts:612` — operation-specific malformed-record isolation
- `packages/orchestrator/src/service.ts:1488` — cycle-local isolation of supplemental active-run refresh failures
- `packages/orchestrator/src/service.ts:2028` — ID-or-identifier preservation guard
- `packages/orchestrator/src/service.test.ts:5568` — diverged-ID orchestration regression

- Malformed Linear records no longer suppress valid state-list candidates, including when a malformed record already has an active run whose stored provider ID differs from its orchestration record. Requested-ID refresh remains strict; the orchestrator logs a failed supplemental refresh and avoids destructive reconciliation without current provider state.

## Validation

- `pnpm --filter @gh-symphony/orchestrator exec vitest run src/service.test.ts -t "continues dispatch when an active-run state refresh fails"` — pass
- `pnpm --filter @gh-symphony/orchestrator test` — pass (358/358)
- `pnpm lint` — pass
- `pnpm test` — pass (all 14 participating workspace projects)
- `pnpm typecheck` — pass
- `pnpm build` — pass
- `git diff --check` — pass
- `./e2e/run-standalone-project-e2e.sh` — documented exception #972; image build and orchestrator startup succeeded, then the unchanged image exited 127 because `jq` is absent

## Changeset

- Not needed because issue #893 has no changeset label.

## Risks & rollback

- Risk is limited to supplemental active-run reconciliation after a tracker refresh failure: matching runs are deliberately retained until a later successful cycle. Revert commits `1afbeea`, `d1343da`, `c483379`, `6c82322`, and `6e3dbd4` to restore prior behavior.

## Changed files

- `packages/tracker-linear/src/orchestrator-adapter.ts` — typed validation, purpose-specific policy, metadata preservation, and optional blocker fallback
- `packages/tracker-linear/src/tracker-linear.test.ts` — malformed-list boundaries, strict refresh, paging, fallback, metadata, and fixture intent
- `packages/orchestrator/src/service.ts` — startup diagnostics and ID-or-identifier active-run refresh isolation
- `packages/orchestrator/src/service.test.ts` — startup, active-run, and diverged-ID reconciliation regressions
- `docs/trackers/linear.md` — provider and supplemental-refresh semantics
- `docs/architecture.md` — layer behavior and conformance coverage
- `AGENT_TEST.md` — deterministic coverage mapping

- [ ] Standard CI/re-review; Docker image `jq` follow-up is tracked in #972.
- [ ] Bounded skip totals and rendered diagnostic length are tracked separately in #974.

## Security


moncher-dev (commented):
Blocking, but small: one required boundary has no test. The behavior itself is correct.

Scratch worktree at `6e3dbd49`. `pnpm install --frozen-lockfile && pnpm build && pnpm test` all pass (Linear 50/50).

**Exercised on the built dist** (`linearTrackerAdapter` with a fake `fetchImpl`):
- A. `[ENG-1, "dev_54 (copy)", ENG-3]` with pickup label `agent` returns `ENG-1, ENG-3`. `skippedItems` holds one entry with the identifier-pattern reason, and `rateLimits.remaining` = 1400 survives the label filter. This is the DEV-54 incident shape, and it no longer takes down the list.
- D. `fetchIssueStatesByIds` with a null state throws `LinearRecordValidationError`. Refresh stays strict.

**Mutation testing** (`packages/tracker-linear`, `vitest run`):
- M1: `purpose === "id-refresh" ||` → `false ||`. Caught by "fails when a requested Linear issue has malformed required state data".
- **M2: `!(error instanceof LinearRecordValidationError)` → `false`, so every error thrown during normalization is skipped. All 50 tests still pass.** #893 says "skip only *recognized* record-validation failures", and nothing enforces that. A real input that crosses the boundary: a node with `labels: { nodes: [null] }` throws `TypeError: Cannot read properties of null (reading 'name')` and today correctly rejects the whole list. Under M2 it would be silently skipped as a diagnostic, hiding a programming or schema error.
  Requested: a state-list regression test in which a non-validation failure inside normalization still rejects atomically. The null-label node above works as the fixture. The test must fail under M2.

**Non-blocking, worth a look while you are there:**
- C. Given a valid `ENG-4` with an inverse `blocks` relation whose blocker identifier is `"bad id"`, `ENG-4` itself is skipped. Its diagnostic reads `{identifier: "ENG-4", reason: "Linear issue identifier \"bad id\" must match ..."}`, which an operator will read as ENG-4's identifier being malformed. This is strictly better than main, where the whole list rejected. Still, either make the reason say it is a blocker relation identifier, or degrade the blocker entry (`identifier: null`) rather than dropping a dependent issue that may be dispatchable once its blocker is done. Your call. If you keep the skip, a clearer reason is enough.

packages/tracker-linear/src/orchestrator-adapter.ts
```diff
@@ -1113,21 +1157,41 @@ function normalizeLinearUserId(value: unknown): string | null {
...
+    throw new LinearRecordValidationError(`${label} is required.`);
   }
   return value;
 }
 
 function sanitizeLinearIdentifier(identifier: string): string {
   const sanitized = identifier.trim().toUpperCase();
   if (!LINEAR_IDENTIFIER_PATTERN.test(sanitized)) {
-    throw new Error(
+    throw new LinearRecordValidationError(
```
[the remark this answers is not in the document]
moncher-dev: Resolved in 6c82322. Optional blocker identifiers now use a fallback sanitizer: recognized identifier-validation failures produce `identifier: null` while the valid top-level issue remains dispatchable. Added a regression covering this exact malformed inverse-relation shape.

packages/tracker-linear/src/orchestrator-adapter.ts
```diff
@@ -587,13 +609,29 @@ async function listLinearIssues(
...
+      );
+    } catch (error) {
+      if (
+        purpose === "id-refresh" ||
+        !(error instanceof LinearRecordValidationError)
+      ) {
+        throw error;
+      }
+      if (skippedItems.length < MAX_LINEAR_SKIPPED_ITEMS) {
+        skippedItems.push(linearSkippedItem(node, index, error.message));
```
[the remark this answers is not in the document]
moncher-dev: Resolved in 6c82322. Startup cleanup now consumes `TrackedIssueList.skippedItems` and emits the bounded identifiers and reasons before reconciling valid terminal records. The orchestrator service regression verifies the diagnostic is written.

moncher-dev (commented):
Steward re-test at head `6c82322d`: the previous block is resolved.

`pnpm install --frozen-lockfile && pnpm build && pnpm test` is green in every package (tracker-linear 52/52, orchestrator 357/357, CI green). The branch merges cleanly onto current main `be5f7dfc`.

**Mutations, one at a time:**
- **M2 (the blocking one):** `!(error instanceof LinearRecordValidationError)` → `false`, so every normalization error is skipped. **Now caught** by "rejects a state list atomically for unexpected normalization failures" (the null-label fixture). "Skip only recognized failures" is now enforced.
- M3: `sanitizeOptionalLinearIdentifier` stops degrading and rethrows → caught by "falls back when optional blocker identifiers are malformed". This covers the non-blocking item C and Codex's P2: a malformed blocker identifier now becomes `identifier: null` instead of dropping ENG-4.
- M5: the startup-cleanup `skippedItems` diagnostic is disabled → caught by "uses listIssuesByStates for startup cleanup terminal lookups". This covers Codex's second P2.

packages/tracker-linear/src/orchestrator-adapter.ts
```diff
@@ -352,9 +368,14 @@ export const linearTrackerAdapter: OrchestratorTrackerAdapter = {
...
-    return listLinearIssues(project, undefined, dependencies, issueIds, {
-      applyPickupLabels: true,
-    });
+    return listLinearIssues(
+      project,
+      undefined,
+      dependencies,
+      issueIds,
+      { applyPickupLabels: true },
+      "id-refresh"
```
hojinzs, line 377: **P2 — the isolation is defeated when the malformed issue has an active run**

Strict `id-refresh` here is correct per §11.1, and I am not asking to weaken it. The problem is what the orchestrator does with a record that `listIssues` just skipped.

`reconcileProject` treats any active run whose identifier is absent from the candidate list as "missing from poll" and re-fetches it through this exact strict path:

- `packages/orchestrator/src/service.ts:1479` — `missingActiveIssueIds` = active runs not in `trackedIssuesByIdentifier`
- `packages/orchestrator/src/service.ts:1489-1494` — `await trackerAdapter.fetchIssueStatesByIds(...)`, **not** wrapped in try/catch

Failure scenario: ENG-2 is dispatched and running, then its `state` goes null (or its identifier is renamed to the DEV-54 shape). `listIssues` now correctly skips ENG-2 and returns ENG-1/ENG-3 — but ENG-2's active run makes it a `missingActiveIssueId`, the strict refresh throws `LinearRecordValidationError`, the outer handler at `service.ts:2202` catches it into `lastError`, and the tenant cycle dispatches nothing. That is the same whole-list suppression #893 set out to remove, relocated 50 lines later.

Note the asymmetry that makes this look unintentional rather than deliberate: the very similar workspace refresh immediately below (`service.ts:1518-1531`) *is* wrapped in try/catch with a "continuing" stderr message. The supplemental active-run refresh is the only one that is not.

Suggested fix is small and matches the existing neighbour — wrap the call, log, and leave those runs unresolved for this cycle:

```ts
let supplementalIssues: TrackedIssue[] = [];
if (missingActiveIssueIds.length > 0) {
  try {
    supplementalIssues = await trackerAdapter.fetchIssueStatesByIds(
      tenant,
      missingActiveIssueIds,
      trackerDependencies
    );
  } catch (error) {
    this.writeStderr(
      `[orchestrator] Active-run state refresh failed for ${tenant.projectId}; continuing: ${this.formatErrorMessage(error)}`
    );
  }
}
```

I recognise this is orchestrator-side and outside #893's stated Affected Files. If you would rather keep this PR's blast radius as scoped, a follow-up issue is a fine answer — but please say so on the thread rather than leaving it implicit, because as it stands the acceptance criterion "mixed lists return valid candidates" does not hold end-to-end for any issue that is already running.

---

packages/tracker-linear/src/orchestrator-adapter.ts
hojinzs, line 631: **P3 — the bound silently under-reports the skip count into an operator-facing metric**

Bounding the array is right and is what #893 asked for. The issue is that both consumers derive the *count* from the bounded array rather than from a real total:

- `packages/orchestrator/src/service.ts:1437` — `skipped = skippedItems.length`, which lands in the status snapshot `summary: { dispatched, suppressed, recovered, skipped }` (`service.ts:2269`)
- `packages/orchestrator/src/service.ts:2373` — `startup cleanup skipped ${skippedItems.length} malformed tracker item(s)`

Verified against the built `dist` with a 120-record all-invalid page: the adapter returns 0 issues, `skippedItems.length === 50`, and an operator is told exactly 50 records were skipped. There is no signal that 70 more were dropped. For an incident like DEV-54 the count is the first number someone looks at, and "50" is indistinguishable from a genuinely-50 case.

Worth noting the GitHub adapter (`packages/tracker-github/src/adapter.ts:555`) does not cap at all, so this is Linear-only behavior and the shared consumers were written assuming the length is the total.

Cheapest fix that keeps the memory bound: track the total separately and expose it, or append one synthetic final diagnostic when truncating:

```ts
let skippedCount = 0;
// ...
skippedCount += 1;
if (skippedItems.length < MAX_LINEAR_SKIPPED_ITEMS) {
  skippedItems.push(linearSkippedItem(node, index, error.message));
}
```

then either widen `TrackedIssueList` with an optional `skippedTotal`, or push a trailing `{ id: "linear-skipped-truncated", identifier: "…", reason: \`${skippedCount - MAX_LINEAR_SKIPPED_ITEMS} additional record(s) omitted from diagnostics\` }`. The second needs no core contract change.

---

packages/orchestrator/src/service.ts
```diff
@@ -2367,6 +2367,13 @@ export class OrchestratorService {
       return;
     }
 
+    const skippedItems = (issues as TrackedIssueList).skippedItems ?? [];
+    if (skippedItems.length > 0) {
+      this.writeStderr(
+        `[orchestrator] startup cleanup skipped ${skippedItems.length} malformed tracker item(s) for ${tenant.projectId}: ${[...new Set(skippedItems.map((item) => item.identifier))].join(", ")} (${[...new Set(skippedItems.map((item) => item.reason))].join(", ")})`
```
hojinzs, line 2393: **P3 — the `new Set` reason dedup does not dedup for Linear, so this line can reach ~19 KB**

This mirrors the existing candidate-poll line at `service.ts:1440`, so the shape is consistent — but the dedup assumption behind it only holds for the GitHub adapter. There, reasons are constant strings (`missing ${stateFieldName}`, `adapter.ts:1703`), so `new Set` collapses 50 skips to one reason.

Linear's reasons embed the offending value verbatim:

```
Linear issue identifier "dev_54 (copy)" must match ^[A-Z][A-Z0-9]*-\d+$.
```

so identifier-pattern failures produce one distinct reason per record and `new Set` collapses nothing. Worst case at the 50-record bound is 50 × 128 identifier chars + 50 × 256 reason chars ≈ 19 KB emitted as a single unbroken `writeStderr` line. That is unpleasant in a terminal and in whatever log pipeline sits downstream.

Suggest truncating the rendered list independently of the diagnostic bound, e.g. first 5 identifiers plus `and N more`, and capping the reason list the same way. Same treatment would apply to `service.ts:1440` — happy for that to be a follow-up since it is pre-existing, but this new line inherits it.

---

packages/tracker-linear/src/tracker-linear.test.ts
```diff
@@ -1202,6 +1202,104 @@ Prompt`,
...
+        },
+      })
+    );
+
+    await expect(
+      linearTrackerAdapter.listIssuesByStates(makeProject(), ["Todo"], {
+        fetchImpl,
+        token: "linear-token",
+      })
+    ).rejects.toThrow(TypeError);
```
hojinzs, line 1303: nit — this test is the M2 guard, but it is pinned to an implementation detail

This is the regression that enforces "skip only *recognized* failures", so it is the most load-bearing test in the PR. Right now it proves that by relying on `labels.nodes[0]` being `null` causing a property access to throw `TypeError`.

The risk: someone hardens label normalization with an ordinary `?.` or `.filter(Boolean)` — a change that looks entirely harmless — and the fixture silently becomes a *valid* record. The test then fails with "expected TypeError, got nothing", which reads like a stale test rather than a lost invariant, and the likely response is to delete it. The `LinearRecordValidationError`-only catch would then be unguarded again.

Pinning the intent rather than the mechanism keeps it honest, e.g. assert the error is not a validation error:

```ts
await expect(
  linearTrackerAdapter.listIssuesByStates(makeProject(), ["Todo"], {
    fetchImpl,
    token: "linear-token",
  })
).rejects.toSatisfy((error) => !(error instanceof LinearRecordValidationError));
```

which needs `LinearRecordValidationError` exported. Or keep the `TypeError` assertion and add a comment naming what the fixture is standing in for, so a future reader knows to replace the fixture rather than drop the test. Either is fine — the behavior itself is correct and I verified it rejects atomically against the built `dist`.

---

hojinzs (commented):
Reviewed at head `6c82322d`. The implementation is sound and the acceptance criteria in #893 are met. One P2 about a gap that survives outside this diff, two P3s, and a nit — none of them are rework of what you built.

## Verification

`pnpm install --frozen-lockfile && pnpm build` clean. `tracker-linear` 52/52, `orchestrator` 357/357. CI green (Test, Container Smoke).

Smoke-tested **against the built `dist`** (not the source) with a fake `fetchImpl`, one scenario per acceptance criterion:

| # | Scenario | Result |
| --- | --- | --- |
| A | Mixed list, DEV-54 shape (`dev_54 (copy)` between two valid), pickup label `agent` | returns `ENG-1, ENG-3`; one diagnostic carrying the identifier-pattern reason; `rateLimits.remaining` 1400 survives the label filter |
| B | 120 all-invalid records | 0 returned, 50 diagnostics — see P3 below |
| C | `fetchIssueStatesByIds` with a null state | rejects `LinearRecordValidationError`; refresh stays strict |
| D | Non-validation failure mid-list (`labels.nodes: [null]`) | rejects `TypeError` atomically |
| E | Malformed blocker identifier, no blocker states configured | `ENG-4` retained, `blockedBy[0].identifier: null` |
| E2 | Same, **with** `blockerCheckStates` enabled | still `dispatchable: false`, reason falls back to the blocker's id |
| F | Empty state list / empty ID list | zero provider calls |
| G | Page 1 has a malformed record, page 2 returns HTTP 500 | rejects atomically; partial pages are not returned |

E2 is the one I most wanted to disprove and could not: `sanitizeOptionalLinearIdentifier` degrading a blocker identifier to `null` does **not** weaken blocking, because `applyBlockerDispatchability` gates on `blocker.state` and only uses the identifier for the reason string, where `blocker.id` is already the fallback. The fix for Codex's P2 does not open a dispatch hole.

## Requirements vs. #893

Every acceptance criterion is covered, and I did not find scope beyond the issue. Operation-purpose plumbing (`LinearListPurpose`) is threaded through one parameter with a safe default rather than a new abstraction, and the diagnostics reuse the existing `TrackedIssueList.skippedItems` contract and the `tracker-github` skipped-item precedent instead of inventing a parallel one. The orchestrator change consumes what the poll path at `service.ts:1436` already consumes. No over-engineering to flag.

The "no changeset" call is correct — WORKFLOW.md:234 gates changesets on the `changeset:*` labels and #893 carries only `integration` / `error-handling`. (Unrelated to this PR: CLAUDE.md:89 states the rule as "behavior-changing PRs must add a changeset" with no mention of the label gate, so the two documents read differently. Worth reconciling separately, not here.)

## Prior review follow-through

Both Codex P2 threads have a substantive reply naming the resolving commit plus a regression, and your own blocking M2 item is now genuinely enforced by the atomic-rejection test. Nothing left hanging.

## Findings

- **P2** — strict `id-refresh` routes skipped records straight back into an unguarded refresh, so the isolation does not hold for an issue that already has an active run (inline on `orchestrator-adapter.ts:377`). This is the one I would like an explicit answer on — fix here or a follow-up issue, either is fine, just not silence.
- **P3** — the 50-record bound under-reports the skip count into `summary.skipped` and the operator logs (inline on `orchestrator-adapter.ts:630`).
- **P3** — reason dedup in the log line is a no-op for Linear, so the line can reach ~19 KB (inline on `service.ts:2373`).
- **nit** — the M2 regression test is pinned to `TypeError` rather than to the invariant (inline on `tracker-linear.test.ts:1300`).
- **nit** — `listIssuesByStates(project, [])` and `fetchIssueStatesByIds(project, [])` return a bare `[]` with no `skippedItems`. Both consumers use `?? []` so nothing breaks; contract consistency only, ignore if you disagree.

Address the P2 (or say why not) and I am ready to approve. The P3s and nits are yours to take or decline with a reason.

---

packages/tracker-linear/src/orchestrator-adapter.ts
moncher-dev: Resolved in c483379. The supplemental active-run refresh remains strict at the adapter boundary, but the orchestrator now catches a failed batch, logs it, continues dispatching valid list candidates, and explicitly leaves those active issue IDs unresolved for the cycle instead of suppressing their runs. The new service regression proves both valid dispatch and active-run preservation.

packages/tracker-linear/src/orchestrator-adapter.ts
moncher-dev: Agreed this is an observability gap, but it is independent of the malformed-record isolation contract and needs a deliberate shared metadata shape. I opened #974 to preserve bounded diagnostics while reporting the true total and added it to the workpad delegation rather than widening #893.

packages/orchestrator/src/service.ts
moncher-dev: Agreed. This affects the existing candidate-poll log as well as the new startup-cleanup log, so I kept the treatment consistent here and opened #974 to bound both rendered summaries together with the skip-total correction.

packages/tracker-linear/src/tracker-linear.test.ts
moncher-dev: Addressed in d1343da with a comment that explicitly identifies the null-label fixture as a stand-in for any unexpected non-validation normalization failure and instructs future maintainers to replace the fixture, not remove the boundary assertion, if label normalization becomes null-safe.

moncher-dev:
Review follow-up on the empty-input nit: I am keeping `listIssuesByStates(project, [])` and `fetchIssueStatesByIds(project, [])` as bare empty arrays. The shared `skippedItems` metadata is optional, both consumers intentionally use `?? []`, and allocating metadata on the zero-request fast path would not change observable behavior. The TypeError-fixture nit was addressed in `d1343da`; both P3 observability findings are tracked in #974.

packages/orchestrator/src/service.ts
```diff
@@ -2014,6 +2024,9 @@ export class OrchestratorService {
           ) ?? persistedRun;
         const issue = trackedIssuesByIdentifier.get(issueRecord.identifier);
         if (!issue) {
+          if (unresolvedActiveIssueIds.has(issueRecord.issueId)) {
```
hojinzs: **P2 — this guard keys on `issueId` only, but the run lookup two lines below deliberately does not**

The fix itself is right and I verified it holds (mutation notes in the summary). This is a hole in the new guard, not a disagreement with it.

`unresolvedActiveIssueIds` is populated from `currentActiveRuns.map((run) => run.issueId)` (`service.ts:1481-1492`), and this guard looks up `issueRecord.issueId`. But the very next statement resolves the run with the identifier-tolerant matcher:

- `service.ts:2023` — `isMatchingIssueRun(run, issueRecord.issueId, issueRecord.identifier)`
- `packages/core/src/observability/status-assembler.ts:10` — `run.issueId === issueId || run.issueIdentifier === issueIdentifier`

That `||` exists because the two ids can diverge. `currentActiveRuns` comes from `store.loadAllRuns()` (`service.ts:1422`) while `issueRecords` comes from the persisted orchestration records — two independent stores. When they disagree on `issueId` but agree on `identifier`, the guard misses and `isMatchingIssueRun` still finds the run, so control falls straight into the suppression path: `publishAssignedBranchForRun` → `SIGTERM` → `status: "suppressed"` → `releaseRunIssueOrchestration`.

Demonstrated, not theoretical. I cloned your new regression verbatim and changed one field — the orchestration record's `issueId` to `"issue-2-legacy"`, identifier untouched at `acme/platform#2`:

```
- "status": "running",     "lastError": null
+ "status": "suppressed",  "lastError": "Run suppressed because the tracker issue is no longer tracked."
```

The run is killed and marked *"no longer tracked"* on a cycle where the tracker was never successfully read — the exact invariant this PR establishes. (Probe discarded; the tree is clean.)

One-line shape, mirroring the matcher:

```ts
const unresolvedActiveIssueIds = new Set<string>();
const unresolvedActiveIssueIdentifiers = new Set<string>();
// ...in the catch:
for (const run of currentActiveRuns) {
  if (missingActiveIssueIds.includes(run.issueId)) {
    unresolvedActiveIssueIds.add(run.issueId);
    unresolvedActiveIssueIdentifiers.add(run.issueIdentifier);
  }
}
```

```ts
if (
  unresolvedActiveIssueIds.has(issueRecord.issueId) ||
  unresolvedActiveIssueIdentifiers.has(issueRecord.identifier)
) {
  continue;
}
```

I could not show that these two ids actually diverge in live data, so if you can establish that `run.issueId` and `issueRecord.issueId` are always equal for a matched pair, say so on this thread and I will take that — but then `isMatchingIssueRun`'s `||` is the thing that is wrong, and this guard is only accidentally safe. Either answer closes it for me.

---

packages/orchestrator/src/service.ts
```diff
@@ -1485,14 +1485,24 @@ export class OrchestratorService {
...
+      if (missingActiveIssueIds.length > 0) {
+        try {
+          supplementalIssues = await trackerAdapter.fetchIssueStatesByIds(
+            tenant,
+            missingActiveIssueIds,
+            trackerDependencies
+          );
+        } catch (error) {
+          for (const issueId of missingActiveIssueIds) {
+            unresolvedActiveIssueIds.add(issueId);
```
hojinzs: **P3 — one malformed record holds every other missing active run hostage**

Not a regression, and I am not asking you to change it in this PR — recording it so the tradeoff is on the record.

`fetchIssueStatesByIds` issues a single batched request under `purpose: "id-refresh"` (`orchestrator-adapter.ts:366-378`), which rejects atomically by design. Confirmed against the built `dist`:

```
P3 batch threw: LinearRecordValidationError -> whole batch lost, incl. the valid ENG-7
```

So this loop marks the whole batch unresolved. If ENG-7's issue was genuinely deleted and ENG-8 is merely malformed, ENG-7's run is never suppressed for as long as ENG-8 stays malformed — and since the malformed record is skipped from every subsequent `listIssues`, that repeats each cycle rather than self-correcting.

It is strictly better than `main` (which dispatched nothing at all on that cycle) and it fails in the conservative direction, which is the right one. It is also self-healing once the Linear record is corrected. Worth a line in the follow-up if you touch this area — splitting the refresh per-id would fix it but multiplies provider calls, so I would not do it on spec.

---

hojinzs (commented):
Round 2 at head `d1343da`. Every item from my last review has a real answer, and the P2 fix is genuinely enforced — I verified it by mutation, not by reading. One new P2 that is a hole *in* that fix, and it is a one-line change. Nothing else is outstanding.

## Verification

`pnpm install --frozen-lockfile && pnpm build` clean. `tracker-linear` 52/52, `orchestrator` 358/358. CI green (Test, Container Smoke).

**Smoke test against the built `dist`** (fake `fetchImpl`), one scenario per #893 acceptance criterion:

| # | Scenario | Result |
| --- | --- | --- |
| A | Mixed list, DEV-54 shape between two valid, pickup label `agent` | `ENG-1, ENG-3`; 1 diagnostic; `rateLimits.remaining` 1400 survives the label filter |
| B | 120 all-invalid records | 0 returned, 50 diagnostics (bound holds) |
| C | `fetchIssueStatesByIds` with a null state | rejects `LinearRecordValidationError` — refresh stays strict |
| D | Non-validation failure mid-list (`labels.nodes: [null]`) | rejects `TypeError` atomically |
| E | Malformed blocker identifier | `ENG-4` retained, `blockedBy[0].identifier: null` |
| F | Empty state list / empty ID list | 0 provider calls |

**Mutation testing on the new orchestrator guard** — both caught, so the regression is load-bearing rather than decorative:

- **M-A:** `if (unresolvedActiveIssueIds.has(...))` → `if (false)`. Caught — the run comes back `suppressed`, so the guard is what keeps it alive.
- **M-B:** `catch` logs but skips populating `unresolvedActiveIssueIds` (the partial fix). Caught — proves the test pins *both* halves of the behavior, continued dispatch **and** run preservation, not just the stderr line.

## Prior review follow-through

- **P2 (active run defeats the isolation)** — fixed in `c483379`, and fixed the way I hoped: strictness stays at the adapter boundary, the orchestrator catches the batch failure and declines to reconcile without provider state. The regression at `service.test.ts:5568` asserts the valid candidate still dispatches, the active run stays `running` with `lastError: null`, and the diagnostic is written. Resolved.
- **Both P3s** — declined with reasons I accept, and #974 is filed and accurate rather than gestured at. The rendered-log bound genuinely belongs with the skip-total correction; splitting them would have been worse. Resolved.
- **nit (M2 test pinned to `TypeError`)** — `d1343da` adds the comment naming the fixture as a stand-in and telling the next maintainer to replace it rather than delete the assertion. That was one of the two options I offered. Resolved.

## Requirements vs. #893

All acceptance criteria hold end-to-end, including the one that did not hold last round. Still no scope beyond the issue: the orchestrator change in `c483379` is outside #893's Affected Files, but it is there because I asked for it and it is 13 lines that mirror the existing neighbour at `service.ts:1518`. No new abstraction, no speculative generality. The "no changeset" call remains correct under WORKFLOW.md:234.

## Findings

- **P2** — the new guard keys on `issueId` alone while `isMatchingIssueRun` two lines below matches on `issueId || identifier`, so a diverged-id pair falls through to SIGTERM + suppress on a cycle where the tracker was never read (inline on `service.ts:2027`). I reproduced it by changing one field in your own regression. One-line fix suggested inline.
- **P3** — a single malformed record marks the entire batched refresh unresolved, so a genuinely-deleted issue's run is never suppressed while a sibling stays malformed (inline on `service.ts:1499`). Not a regression, conservative direction, recording only.

Push the P2 fix — or reply on that thread establishing the two ids cannot diverge — and I will approve on sight. The P3 needs nothing from you here.

---

packages/orchestrator/src/service.ts
moncher-dev: Resolved in 1afbeea. Failed supplemental refreshes now retain both the active run's issue ID and identifier, matching `isMatchingIssueRun`. The existing regression now intentionally persists `issue-2-legacy` while the live run remains `issue-2`; it verifies the valid candidate still dispatches and the identifier-matched active run remains running with no error.

packages/orchestrator/src/service.ts
moncher-dev: Acknowledged as a conservative tradeoff, with no change in this PR. Strict requested-ID refresh remains atomic by specification; if one record is malformed, keeping sibling active runs unresolved avoids unsafe suppression. Per-ID retries would multiply provider requests and are outside #893's malformed state-list isolation scope.

packages/orchestrator/src/service.test.ts
```diff
@@ -5565,6 +5565,119 @@ Test hook failures.
...
+      tempRoot,
+      "acme",
+      "platform"
+    );
+    const store = new OrchestratorFsStore(tempRoot);
+    const projectConfig = createProjectConfig(tempRoot, repository);
+    await store.saveProjectConfig(projectConfig);
+    await store.saveProjectIssueOrchestrations("tenant-1", [
+      {
+        issueId: "issue-2-legacy",
```
hojinzs, line 5583: nit — editing this fixture in place moved the coverage instead of widening it

Not blocking, and not a request to change the fix — the fix is the shape I asked for and I verified it holds. This is about what the suite now pins.

Changing `issue-2` → `issue-2-legacy` converted the one regression from covering the `issueId` operand to covering the `identifier` operand. Mutation results at this head, run one at a time on `service.test.ts` (225 tests):

- `unresolvedActiveIssueIdentifiers.has(issueRecord.identifier)` → `false`: **caught**, by this test. The new half is load-bearing, not decorative.
- `unresolvedActiveIssueIds.has(issueRecord.issueId)` → `false`: **survives**, all 225 pass. Nothing pins the first operand any more.

So a future refactor could delete `unresolvedActiveIssueIds` entirely and CI stays green.

I tried to show that operand is actually required before raising it, and could not. I cloned this test with the divergence inverted — record `issueId: "issue-2"` kept, run `issueIdentifier: "acme/platform#2-renamed"` — and instrumented the guard:

```
DBG record issue-2 acme/platform#2-renamed  ids ['issue-2']  idents ['acme/platform#2-renamed']
```

The orchestration record's `identifier` comes back already synced to the run's, so the identifier operand fires and the probe passes with the `issueId` operand disabled. I could not construct a case where the first operand is the only thing preserving the run. (Probe discarded; tree clean.)

That makes keeping it the right call — it mirrors `isMatchingIssueRun`'s `||`, and dropping it because today's persistence happens to keep the two in sync would be relying on a coincidence. I'd just rather the suite say so. If you touch this area again, duplicating the test (aligned ids in one, diverged in the other) rather than editing it pins both operands. Entirely your call whether that is worth a second fixture now.

---

hojinzs (approved):
Approving at head `1afbeea`. The round-2 P2 is fixed exactly as proposed, and I confirmed by mutation that the fix is enforced rather than merely present. Everything I raised across three rounds is now either fixed or declined with a reason I accept. One nit inline, which does not hold up the merge.

## Verification

`pnpm install --frozen-lockfile && pnpm build` clean. `pnpm lint`, `pnpm typecheck` clean. `tracker-linear` 52/52, `orchestrator` 358/358. CI green (Test, Container Smoke). Branch merges cleanly onto current main `8f823a7`, which has moved ahead since round 2.

**Smoke test against the built `dist`** (fake `fetchImpl`), one scenario per #893 acceptance criterion — 7/7:

| # | Scenario | Result |
| --- | --- | --- |
| A | Mixed list, DEV-54 shape between two valid, pickup label `agent` | `ENG-1, ENG-3`; 1 diagnostic with the identifier-pattern reason; `rateLimits.remaining` 1400 survives the label filter |
| B | 120 all-invalid records | 0 returned, 50 diagnostics (bound holds) |
| C | `fetchIssueStatesByIds` with a null state | `LinearRecordValidationError` — refresh stays strict |
| D | Non-validation failure mid-list (`labels.nodes: [null]`) | `TypeError`, atomic reject |
| E | Malformed blocker identifier | `ENG-4` retained, `blockedBy[0].identifier: null` |
| F | Empty state list / empty ID list | 0 provider calls |
| G | Malformed page 1 + HTTP 500 page 2 | atomic reject, no partial page returned |

## Prior review follow-through

- **P2 (guard keyed on `issueId` while the matcher keys on either)** — fixed in `1afbeea`, and fixed at both ends: the catch now records `run.issueIdentifier` alongside `run.issueId`, and the guard tests both. Mutating the new identifier operand to `false` fails the regression, so the half that closes the hole is genuinely pinned. Resolved.
- **P3 (one malformed record holds the whole batch unresolved)** — acknowledged as a deliberate conservative tradeoff, no change. I agree; per-ID retries would multiply provider calls to fix a self-healing condition. Resolved.
- Everything from rounds 1 and 2 — both Codex P2s, my two P3s, and both nits — has a substantive reply naming a commit or a reason, with #974 filed for the observability pair. Nothing is left implicit.

## Requirements vs. #893

All acceptance criteria hold end-to-end, including for an issue that already has an active run, which is the case that did not hold in round 1. Still no scope beyond the issue: `1afbeea` is 13 lines and reuses the existing `isMatchingIssueRun` identity rule rather than introducing a new one. No new abstraction, no speculative generality. The "no changeset" call remains correct under WORKFLOW.md:234, since #893 carries only `integration` / `error-handling`.

The CLAUDE.md:89 vs. WORKFLOW.md:234 wording mismatch I noted in round 1 is still worth reconciling, but it is unrelated to this PR and should not ride it.

## Findings

- **nit** — the round-2 regression was edited in place rather than duplicated, so the guard's `issueId` operand no longer has a test; mutating it to `false` leaves all 225 tests passing (inline on `service.test.ts:5583`). I tried to build a case proving that operand is required and could not, so keeping it is correct as a mirror of `isMatchingIssueRun` — this is about the suite, not the code. Take it or leave it.

Good to merge.

---

PR #964, opened 2026-09-14
test(ci): require package-aware tests on pull requests

moncher-dev:
## Issues

- Closes #891

## Summary

- Adds `pnpm test` as the authoritative package-aware pull-request CI gate.
- Retains the existing root coverage report and all worker, package, helper, and container checks.
- Documents a clean-checkout-compatible local sequence that builds workspace exports before package tests and explains the remaining root coverage behavior.

## Change-point diagram

- Pull request → format/build/lint/typecheck/worker startup → `pnpm test` package gates → legacy root coverage reporting → existing helper/package/container checks
- Clean local checkout → install dependencies → build workspace exports → `pnpm test`

## Start here

- `.github/workflows/ci.yml:50` — authoritative package-aware test step
- `AGENT_TEST.md:19` — matching clean-checkout local verification order
- `AGENT_TEST.md:31` — retained root coverage command semantics

## User-Visible Behavior / Operational Impact

- Pull requests now fail when any workspace package test fails, including the three control-plane TSX test files and package-specific setup/configuration.
- Local verification builds workspace package exports before invoking the authoritative test gate.

## Validation

- `pnpm format` — pass
- `pnpm lint` — pass
- `pnpm build` — pass
- `pnpm test` — pass; all 14 workspace package scripts completed
- `pnpm typecheck` — pass
- `pnpm --filter @gh-symphony/control-plane test -- --reporter=verbose` — pass; 5 files/38 tests, including all 3 TSX files
- `git diff --check` — pass
- Prior GitHub PR jobs on `72260cd` — Test and Container Smoke passed; refreshed checks will run on `6f9e14a`
- Docker E2E / blackbox evidence: not applicable; CI configuration and documentation changed, while existing container jobs remain unchanged

## Changeset

- Not needed because this is CI and documentation behavior with no changeset label or released package change.

## Risks & rollback

- Risk is limited to additional CI duration and surfacing package failures previously missed by root discovery. Roll back the package-aware step and accompanying documentation if emergency CI restoration is required.

## Changed files

- `.github/workflows/ci.yml` — requires package-aware tests and labels legacy coverage as report generation
- `AGENT_TEST.md` — identifies `pnpm test` as authoritative, orders build before tests, and explains coverage limitations
- `CLAUDE.md` — mirrors the clean-checkout-compatible before-shipping command order

## Post-merge / human validation

- [ ] Confirm refreshed pull-request jobs complete successfully on GitHub.

## Security

[...]

PR #971, opened 2026-09-14
test(ci): consolidate package-aware coverage

### LINKED ISSUE #892: test(ci): consolidate coverage with package test configuration
Parent Epic: #890. Approved registration plan: [Q2: #892](https://github.com/hojinzs/github-symphony/issues/892).

## Background

After [Q1: #891](https://github.com/hojinzs/github-symphony/issues/891), coverage still uses a separate configuration and may duplicate execution or miss package semantics.

## Proposed Changes

Adopt one validated package-aware coverage entry point, preserving package roots, aliases, setup, defines, and file serialization. Prevent report overwrites and define shared-source attribution. Remove duplicate coverage execution only after parity is proven.

## Affected Files

| File | Change | Scope |
| --- | --- | --- |
| `vitest.config.ts`, relevant `packages/*/vitest.config.ts` | Package-aware coverage discovery and setup | Test configuration |
| `.github/workflows/ci.yml`, `package.json`, `AGENT_TEST.md` | Reporting command, artifact, and usage | CI / documentation |

Normalized test-file sets match the authoritative package gate; frontend source appears in a nonempty aggregate report; built-artifact checks remain. Full required checks plus coverage artifact inspection.

Observability; cross-layer verification infrastructure. Keep provider-specific behavior in adapters and preserve current persistence, retry, metric, and unpublished-work contracts. Do not edit `docs/symphony-spec.md`. No new spec divergence is intended.

## Planning

- Priority: P2
- Size: L
- Estimate: 12 net implementation hours (excludes review, testing, and CI waiting).
- L-size split reviewed: retain the coverage entry point and aggregate output in one deployable change; the immediate gate is isolated in [Q1: #891](https://github.com/hojinzs/github-symphony/issues/891). Split again before Ready if scope expands.

## Dependencies

## Required Checks

```bash
pnpm lint
pnpm test
pnpm typecheck
pnpm build
```

## References

- #890

moncher-dev:
## Issues

- Closes #892

## Summary

- Consolidates pull-request test coverage into one package-project Vitest execution while preserving package roots, aliases, setup files, defines, and orchestrator file serialization.
- Proves normalized test discovery parity, scheduling parity, and control-plane frontend attribution before CI relies on the aggregate entry point.
- Publishes one non-overwriting `coverage/` artifact and retains built-artifact checks.

## Change-point diagram

- package Vitest configs → root project aggregation → root serialization constraint → one V8 coverage report → PR report and artifact
- package discovery + two-file lock probe → parity guard → aggregate execution permitted

## Start here

- `vitest.config.ts:13` — loads package projects and preserves required root-level file serialization.
- `scripts/verify-coverage-discovery.mjs:8` — compares normalized test sets and executes the serialization probe.
- `packages/orchestrator/test-fixtures/file-serialization-a.test.ts:1` — one half of the behavioral overlap guard.
- `.github/workflows/ci.yml:50` — invokes the consolidated CI test and coverage entry point.

- Pull-request CI executes the complete suite once with coverage, preserves orchestrator test isolation, and publishes one `coverage-report` artifact. Runtime and CLI behavior are unchanged.

## Validation

- `pnpm lint` — pass; all 14 workspace package lint scripts.
- `pnpm test` — pass; all 14 workspace package test scripts, including 137 Vitest files and 1,996 tests.
- `pnpm typecheck` — pass; all 14 workspace packages.
- `pnpm build` — pass, including control-plane frontend and built CLI artifacts.
- `pnpm test:coverage` — pass; discovery matches 137 files, behavioral serialization probe passes, 1,996 tests pass, and the aggregate report contains 17 frontend files with 721 covered statements.
- Docker E2E: not applicable; this changes test/CI configuration, not runtime integration behavior.

## Changeset

- Not needed because issue #892 has no changeset label and does not change released CLI/runtime behavior.

## Risks & rollback

- Risk: Vitest project scheduling options can move to root scope without an obvious discovery difference. The committed lock probe fails CI if orchestrator files overlap.
- Rollback: restore the separate `pnpm test` and root coverage CI steps and remove the aggregate project configuration and parity scripts.

## Changed files

- `vitest.config.ts` — package-project aggregation and root serialization constraint.
- `packages/orchestrator/vitest.config.ts` — isolates behavioral probe fixtures from ordinary package runs.
- `packages/orchestrator/test-fixtures/file-serialization-{a,b}.test.ts` — deterministic overlap probe.
- `scripts/verify-coverage-discovery.mjs` — normalized discovery and scheduling parity checks.
- `scripts/verify-coverage-report.mjs` — frontend coverage attribution check.
- `package.json` — consolidated `test:coverage` command.
- `.github/workflows/ci.yml` — single coverage execution and artifact upload.
- `AGENT_TEST.md` — updated verification procedure.

- [ ] Inspect the uploaded `coverage-report` artifact in GitHub Actions.

## Security


moncher-dev (commented):
Blocking. The aggregate run does not preserve orchestrator's file serialization, but this PR still makes it the only test gate in CI.

Scratch worktree at `a2549eaf` (Vitest 3.2.4). `pnpm install --frozen-lockfile && pnpm build` pass. `pnpm test` passes (137 files). `pnpm test:coverage` passes (discovery 137 files, 1,996 tests, 17 frontend files / 721 covered statements).

**Finding: `fileParallelism: false` is dropped in projects mode.** #892 requires the entry point to preserve "package roots, aliases, setup, defines, and file serialization", and to remove duplicate execution "only after parity is proven". `packages/orchestrator/vitest.config.ts` sets `fileParallelism: false` because its Git-heavy files mutate process-level Git resources. With `projects: ["packages/*"]`, that option is not applied: Vitest treats worker scheduling as a root-level setting.

Repro: add two throwaway files, `packages/orchestrator/src/zz-probe-{a,b}.test.ts`. Each one appends `start <ms> pid=<pid> cfg=$GH_SYMPHONY_CONFIG_DIR` to a log, sleeps 3 s, then appends `end <ms>`.

```
# package-local: cd packages/orchestrator && vitest run src/zz-probe
a start ...905569 pid=99892 cfg=orchestrator-config-*
a end   ...908572
b start ...908693 pid=529   cfg=orchestrator-config-*     <- serial
# aggregate: vitest run packages/orchestrator/src/zz-probe (same with --coverage.enabled)
b start ...912258 pid=550   cfg=orchestrator-config-*
a start ...912272 pid=549   cfg=orchestrator-config-*     <- overlapping, separate forks
```

`setupFiles` *is* honored (the config sandbox is set), so this is specific to scheduling. The comment in `vitest.config.ts`, "same roots, aliases, setup, defines, and scheduling as tests", is therefore not true for scheduling. `verify-coverage-discovery.mjs` compares file sets only, so it cannot catch this. I confirmed it does catch file drift: with `"!packages/core"` added to `projects` it exits 1 and lists the missing files.

What would resolve it (implementer's choice):
- Keep orchestrator serial in the aggregate run. For example, run it as its own serial invocation or project group with coverage merged, or use a root-level mechanism that serializes only that project. Or keep `pnpm test` in CI until parity is proven.
- Add a parity check that would fail if serialization were lost, not just file discovery. The two-file overlap probe above is a working template.

Also verify against the retained CI shape from #964: nothing else in CI now runs orchestrator tests serially.

moncher-dev:
Addressed the serialization blocker in `2151f0b`:

- the repository-level projects run now sets root-scoped `fileParallelism: false`, which preserves the orchestrator package's required file isolation;
- `verify-coverage-discovery.mjs` runs a committed two-file exclusive-lock probe and fails if aggregate scheduling overlaps those files;
- the probe fixtures are excluded from normal package discovery, which remains 137 files.

Fresh verification passed: `pnpm lint`, `pnpm test`, `pnpm typecheck`, `pnpm build`, and `pnpm test:coverage` (137 files, 1,996 tests, 17 frontend files, 721 covered statements).

moncher-dev (commented):
Steward re-test at head `2151f0b0`: the previous block is resolved.

`pnpm install --frozen-lockfile && pnpm build && pnpm test` is green in every package. `verify-coverage-discovery.mjs` passes (137 files, probe preserved). CI's new `pnpm test:coverage` step passed in 95 s and uploaded the artifact. The branch merges cleanly onto current main `be5f7dfc`.

**Re-ran the original two-file timing probe** (two throwaway `packages/orchestrator/src/zz-probe-{a,b}.test.ts` files, each sleeping 3 s):
- package-local: a 275460–278462, then b 278584 → serial
- aggregate `vitest run`: a 282123–285124, then b 285263 → **serial now (overlapped before)**
- aggregate `--coverage.enabled`: b 289492–292494, then a 292651 → **serial now**

**The guard detects the regression:** with the root `fileParallelism: false` deleted, `node scripts/verify-coverage-discovery.mjs` exits 1 with `EEXIST … active.lock` / "Aggregate coverage does not preserve file serialization". With it restored, the script exits 0. So the probe runs in CI through `test:coverage`, and it fails when the serialization it guards is removed.

Non-blocking trade-off for the reviewer: root `fileParallelism: false` serializes all 137 files in the aggregate run, not only orchestrator's. That is correct for #892's parity requirement, but it costs CI wall time (the coverage step is currently 95 s). Passing to hojinzs for review.

vitest.config.ts
```diff
@@ -1,12 +1,31 @@
...
+    ...(isRepositoryRun
+      ? {
+          // Load every package as a Vitest project so aggregate coverage uses
+          // the same roots, aliases, setup, and defines as tests. Vitest treats
+          // fileParallelism as a root scheduling option in projects mode, so
+          // preserve the orchestrator package's serialization requirement here.
+          projects: ["packages/*"],
+          fileParallelism: false,
+          coverage: {
+            reportsDirectory: "coverage",
```
hojinzs, line 22: **P2 — the published coverage headline is now misleading (36.12% reported vs 81.04% actual source coverage).**

`coverage.exclude` is not set here, so built output is instrumented. In the report this PR produced, `**/dist/**` accounts for **156 of 381 files and 60,058 of 104,073 statements (58% of the denominator) at 3.19% covered**:

```
dist : 156 files /  60058 stmts /  3.19%
src  : 225 files /  44015 stmts / 81.04%
TOTAL: 381 files / 104073 stmts / 36.12%   <- the number posted on every PR
```

`dist/` inclusion is pre-existing, so this is not a regression you introduced — but this PR *amplifies* it, and the interaction is worth a look because it works against the issue's "define shared-source attribution" goal.

I ran the previous root-config coverage command against this same build for comparison:

| | dist stmts | src stmts | reported |
| --- | --- | --- | --- |
| baseline (root config) | 60058 | 44024 | **41.05%** |
| this PR (projects mode) | 60058 | 44015 | **36.12%** |

The reported number *drops* even though the PR runs 3 more files and 23 more tests. The per-file delta shows exactly why, and it's entirely to the PR's credit: package `resolve.alias` entries now apply, so coverage moved off built bundles onto real sources.

```
LOSSES (all dist/)                 GAINS (all src/)
-777  packages/core/dist/workflow/parser.js        +303  control-plane/client/src/routes/issues/$identifier.tsx
-641  tracker-github/dist/adapter.js               +196  extension-github-workflow/src/approval-workflow.ts
-410  runtime-codex/dist/runtime.js                +189  control-plane/client/src/routes/index.tsx
-359  runtime-claude/dist/adapter.js               +118  core/src/contracts/orchestrator-channel.ts
```

So attribution got strictly better while the headline got worse: `dist/` went from 42,725 covered statements (masking the problem) to ~1,900, and is now pure noise in the denominator.

Excluding it is a 3-line change that I verified end to end on this head:

```ts
import { configDefaults, coverageConfigDefaults, defineConfig } from "vitest/config";
// ...
coverage: {
  reportsDirectory: "coverage",
  exclude: [...coverageConfigDefaults.exclude, "**/dist/**"],
},
```

Result: `lines/statements 81.04% (35670/44015)`, 225 files, 0 dist files. `pnpm test:coverage` still exits 0 and `verify-coverage-report.mjs` still passes.

Not blocking — the acceptance criteria in #892 don't mention `dist/`, and the behavior predates you. But since this PR's whole purpose is a trustworthy consolidated coverage report, landing it here rather than as a follow-up seems worth it. Your call.

---

vitest.config.ts
```diff
@@ -1,12 +1,31 @@
...
+          },
+        }
+      : {
+          include: ["**/*.test.ts"],
+          exclude: [...configDefaults.exclude, ".runtime/**"],
```
hojinzs, line 27: **P3 — this fallback branch picks up the probe fixtures and hard-fails.**

The `test-fixtures/**` exclusion lives only in `packages/orchestrator/vitest.config.ts`, so this branch's `exclude` doesn't cover it. The fixtures throw unconditionally when `VITEST_SERIALIZATION_PROBE_LOCK` is unset, so any discovery path that reaches them fails with an error that points at the fixture rather than at the misconfiguration.

Confirmed the branch is live code (not dead) and that it matches the fixtures:

```
$ cd packages/orchestrator && npx vitest list --filesOnly --config ../../vitest.config.ts
test-fixtures/file-serialization-a.test.ts
test-fixtures/file-serialization-b.test.ts
```

And the resulting failure, reproduced by running the pre-PR root config shape against this head:

```
FAIL  packages/orchestrator/test-fixtures/file-serialization-b.test.ts
Error: VITEST_SERIALIZATION_PROBE_LOCK is required
Test Files  2 failed | 134 passed (136)
```

That last one is the part I'd weigh: the PR's stated rollback is "restore the separate `pnpm test` and root coverage CI steps and remove the aggregate project configuration and parity scripts" — it doesn't mention the fixtures, and restoring the old root config without also deleting them reproduces exactly the failure above.

Cheapest fix is adding `"**/test-fixtures/**"` to this branch's `exclude`; alternatively have the fixtures `test.skip()` instead of throwing when the env var is absent, so they're inert wherever they get discovered. Either also covers the rollback path.

---

.github/workflows/ci.yml
```diff
@@ -68,6 +65,14 @@ jobs:
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: Upload coverage artifact
+        uses: actions/upload-artifact@v4
+        if: always()
+        with:
+          name: coverage-report
+          path: coverage/
+          if-no-files-found: error
```
hojinzs, line 74: **P3 — `if: always()` + `if-no-files-found: error` turns a discovery failure into two red steps.**

`test:coverage` is `verify-coverage-discovery.mjs && vitest … && verify-coverage-report.mjs`. When the *first* command fails — which is precisely the guard this PR added, and the case where the CI signal most needs to be legible — Vitest never runs, `coverage/` is never created, and this step then fails a second time with "no files found" on a fresh runner checkout.

The result is that the real diagnostic ("Aggregate coverage does not preserve file serialization") is followed by an unrelated artifact error, and the job summary shows the artifact step as the failure. `if-no-files-found: warn` keeps the genuine failure the only red step while still flagging a missing report on otherwise-green runs.

nit, same step: the `davelosert/vitest-coverage-report-action@v2` step above has the same shape and will also report against a missing `coverage-summary.json` in that scenario.

---

scripts/verify-coverage-report.mjs
```diff
@@ -0,0 +1,25 @@
+import { readFile } from "node:fs/promises";
+
+const report = JSON.parse(
+  await readFile(
+    new URL("../coverage/coverage-final.json", import.meta.url),
+    "utf8"
+  )
+);
+const frontendEntries = Object.entries(report).filter(([file]) =>
+  file.includes("/packages/control-plane/client/src/")
```
hojinzs, line 10: nit — this guard is narrower than the property it's protecting.

The failure mode this PR fixes is "a package's own roots/aliases/setup were not applied in the aggregate run," and control-plane frontend `.tsx` is one instance of it. The check hardcodes that one instance, so if a different package silently dropped out of the aggregate report the script would still exit 0.

A cheap generalization that stays in the spirit of #892: assert that every workspace package contributes at least one covered source file to `coverage-final.json`. That reuses the package list `verify-coverage-discovery.mjs` already builds and catches the whole class rather than the one known case.

Also `"/packages/control-plane/client/src/"` hardcodes the POSIX separator. Not an issue for CI, but it makes the script silently pass (`frontendEntries.length === 0` → exit 1, so actually silently *fail*) for anyone running `pnpm test:coverage` on Windows.

---

hojinzs (approved):
## Approve

Verified at head `2151f0b0` in a clean worktree: `pnpm install --frozen-lockfile`, `pnpm build`, `pnpm test`, `pnpm test:coverage` all pass. CI is green on both jobs and the branch is mergeable.

### Requirements traceability (#892)

| Acceptance criterion | Status | Evidence |
| --- | --- | --- |
| Normalized test-file sets match the authoritative package gate | ✅ | 137 files both sides; script exits 0 |
| Preserve package roots, aliases, setup, defines | ✅ | Coverage attribution moved from `dist/` bundles onto real sources — see the per-file delta in the `vitest.config.ts` thread |
| Preserve file serialization | ✅ | Independently reproduced, below |
| Frontend source appears in a nonempty aggregate report | ✅ | 17 frontend files / 721 covered statements; `routes/index.tsx` and `issues/$identifier.tsx` go 0 → 189 / 0 → 303 |
| Prevent report overwrites | ✅ | Single `reportsDirectory`, one non-overwriting artifact |
| Built-artifact checks remain | ✅ | `test:worker-startup` and `cli test:package` retained; the former has its own scoped config and is unaffected |
| Remove duplicate execution *only after parity is proven* | ✅ | Parity script gates the run and precedes it in `test:coverage` |
| Define shared-source attribution | ⚠️ | Source attribution is now correct, but `dist/` is still in the denominator — P2 thread |

### Re-verifying the previous blocker

I re-ran the guard independently rather than taking the re-test at face value. Deleting the root `fileParallelism: false` and running the parity script:

```
EEXIST … active.lock
Aggregate coverage does not preserve file serialization.
```

Restoring it exits 0. The probe is genuinely wired into `test:coverage` and genuinely fails when the property it guards is removed. The earlier finding is resolved, and the committed fixtures are proportionate to it — this is not over-engineering.

### The open trade-off from the last round

The re-test flagged global `fileParallelism: false` as an unresolved CI wall-clock cost. I measured it on a 4-core box, and I think it resolves in this PR's favor:

| | before (two steps) | after (one step) |
| --- | --- | --- |
| `pnpm test` | 1m20s | — |
| root coverage | ~55s | — |
| `test:coverage` | — | 2m11s (20s discovery + 1m51s serial run) |
| **total** | **~2m15s** | **~2m11s** |

Consolidation gives back roughly what serialization costs, so it's net neutral-to-better. Not a concern.

### Findings

| # | Sev | Summary |
| --- | --- | --- |
| 1 | **P2** | Published coverage headline reads 36.12% vs 81.04% actual source coverage — `dist/` is 58% of the denominator at 3.19% covered. Pre-existing, but amplified by this PR's (correct) attribution fix. Verified 3-line fix included. |
| 2 | P3 | Probe fixtures are discoverable outside the orchestrator project and throw there; the documented rollback path reproduces the failure. |
| 3 | P3 | `if: always()` + `if-no-files-found: error` masks a discovery failure behind a second, unrelated red step. |
| 4 | nit | `verify-coverage-report.mjs` hardcodes one package instead of asserting the general property. |

None of these block. Finding 1 is the one I'd most like to see land here rather than as a follow-up — it's small, verified, and this PR's value is a coverage report people trust — but it predates the change and isn't in the acceptance criteria, so I'm not gating on it.

No over-engineering or unrequested scope found; the changeset omission is correct for a test/CI-only change per CLAUDE.md.

Approving. Post-merge, the remaining manual item is inspecting the uploaded `coverage-report` artifact, as the PR notes.

---

Issue #890, opened 2026-09-05
[Epic] Address reliability and maintainability review findings


moncher-dev:
## Background

The [2026-09-05 reliability and maintainability review](https://github.com/hojinzs/github-symphony/blob/59b0516b8b0c5ba137544b58e15ccbb490f967da/docs/reports/2026-09-05-maintainability-and-reliability-review.md) identified four bounded improvement areas:

1. The root CI test path excludes three React test files (17 tests) and does not inherit package-specific setup or serialization.
2. The orchestrator service owns scheduling, preparation, processes, recovery, and observability in one large module.
3. Normal candidate polling contains five direct full-history run reads before helper-dependent reads. Production latency has not been measured.
4. A malformed Linear state-list record rejects otherwise valid candidates. This is an availability improvement, not by itself a MUST-level spec violation; malformed requested-ID refreshes must continue to fail.

This Epic coordinates resolution. It is a tracking item, not one worker-sized implementation assignment. The report is a point-in-time snapshot; revalidate affected source against the implementation base.

## Proposed Changes

- Close the CI verification gap with the existing package-aware test command, then consolidate coverage without losing package isolation or built-artifact checks.
- Isolate recognized Linear record-validation failures only for state-list operations, preserve diagnostics through filters, and retain atomic transport/pagination failure and strict ID refresh.
- Measure run-history cost, introduce a project-scoped storage query, and reduce repeated reads only at documented freshness boundaries. Preserve JSON compatibility, historical token totals, retries, and unpublished-work evidence.
- Reconcile structural extraction with #847 before moving orchestration responsibilities. Extract retained pure decisions first; do not build a new wrapper around responsibilities already scheduled for removal.

### Scope and classification

- **Symphony Layers:** Configuration, Coordination, Execution, Integration, Observability; cross-layer verification infrastructure.
- **Priority:** P1 — coordinates core contracts and verification boundaries. Individual work has its own priority.
- **Size:** L — tracking scope explicitly decomposed below; do not dispatch the Epic as one implementation.
- **Estimate:** 61 net implementation hours across the proposed children, preliminary. Per PROJECT_MANAGE.md this excludes review, testing, and CI waiting. Not a calendar commitment; re-estimate after the storage baseline and extraction-scope decision.
- **Initial status:** Backlog.
- **Type / Workstream:** Epic / GitHub Symphony.
- The project has no Estimate field; estimates are kept in issue bodies rather than adding a new project field.

## Affected Files

[...]

PR #973, opened 2026-09-14
refactor(orchestrator): extract retained retry and finalization decisions

### LINKED ISSUE #898: refactor(orchestrator): extract retained retry and finalization decisions
Parent Epic: #890. Approved registration plan: [Q8: #898](https://github.com/hojinzs/github-symphony/issues/898).

## Background

Retained retry/finalization decisions currently require navigating the service and its effectful state.

## Proposed Changes

Extract only the pure decisions selected in [Q7: #897](https://github.com/hojinzs/github-symphony/issues/897), using explicit inputs/outputs. Keep queue ownership and side effects in the existing façade; avoid passing the whole service to helpers. If the selected slice cannot fit M, narrow it to one decision family and propose another child separately.

## Affected Files

| File | Change | Scope |
| --- | --- | --- |
| `packages/orchestrator/src/service.ts`, proposed focused decision module | Retained pure logic and delegation | Coordination |
| Existing service/dispatch tests and focused decision tests | Preserve behavior through the façade | Tests |
| `docs/architecture.md` | Update the package map | Documentation |

Retry budget, reservation order, capacity, finalization outcome, event names, and persisted schemas are unchanged. Reconciliation and tracker-state queues remain separate. Full required checks plus relevant CI integration checks.

Coordination. Keep provider-specific behavior in adapters and preserve current persistence, retry, metric, and unpublished-work contracts. Do not edit `docs/symphony-spec.md`. No new spec divergence is intended.

## Planning

- Priority: P2
- Size: M

## Dependencies

Blocked by: [Q1: #891](https://github.com/hojinzs/github-symphony/issues/891), [Q7: #897](https://github.com/hojinzs/github-symphony/issues/897), #872.

Wait for #872 to establish the CI container black-box environment regression baseline before changing orchestration decision boundaries. Its existing #871 dependency remains in place. Schedule after [Q6: #896](https://github.com/hojinzs/github-symphony/issues/896) if the final patches overlap; that conditional sequencing is not an extra hard dependency.

## Required Checks

```bash
pnpm lint
pnpm test
pnpm typecheck
pnpm build
```

## References

- #890

moncher-dev:
## Issues

- Closes #898

## Summary

- Extract retained bounded-finalization and retry calculations into a focused pure decision module while the service retains effects and queue ownership.
- Use plan→resolve decisions so policy and poll loaders are driven by explicit outputs instead of duplicated caller predicates.
- Cover every retry-requeue outcome directly and preserve retry budgets, attempt advancement, capacity postponement, finalization outcomes, events, and schemas.

## Change-point diagram

- `OrchestratorService` gathers durable facts → pure planner selects suppression/accounting and required delay class → façade loads only the requested policy/poll input → pure resolver computes the record → façade performs store, queue, event, metric, cleanup, and worker effects.

## Start here

- `packages/orchestrator/src/retained-decisions.ts:27` — explicit finalization actions.
- `packages/orchestrator/src/retained-decisions.ts:82` — completed-run plan and resolver.
- `packages/orchestrator/src/retained-decisions.ts:165` — requeue plan and resolver.
- `packages/orchestrator/src/service.ts:3802` — effect-owning façade delegation.

- None. This is a behavior-preserving Coordination Layer refactor. Retry timing/budgets, reservation order/capacity, persistence and event schemas, queue separation, and provider boundaries are unchanged.

## Validation

- `pnpm lint` — pass across all workspace packages.
- `pnpm test` — one shared-host timeout in untouched `packages/core/src/workspace/hooks.test.ts`; required isolated rerun passed 12/12. Orchestrator façade run passed 224/224.
- `pnpm --filter @gh-symphony/orchestrator exec vitest run src/retained-decisions.test.ts` — pass, 14/14.
- Focused coverage — 89.79% branches for `retained-decisions.ts`; all five retry-requeue delay outcomes covered directly.
- `pnpm typecheck` — pass across all workspace packages.
- `pnpm build` — pass across all workspace packages.
- `git diff --check origin/main...HEAD` — pass.
- Upstream `docs/symphony-spec.md` unchanged.
- Docker E2E / blackbox — not applicable: no integration behavior, process lifecycle, tracker adapter, status API, or CLI runtime command changed.

## Changeset

- Not needed because the issue has only the `refactoring` label and no user-visible CLI/runtime behavior changes.

## Risks & rollback

- Reviewer focus: the plan delay class is the sole loader-selection contract, and exhausted unknown finalization records its event before retry handling without releasing the claim.
- Rollback is a revert of commits `3d6ef9e` and `1f3dfe9`; no data, configuration, or schema migration is involved.

## Changed files

- `packages/orchestrator/src/retained-decisions.ts` — pure finalization and plan→resolve retry decisions.
- `packages/orchestrator/src/retained-decisions.test.ts` — complete requeue decision table and invariants.
- `packages/orchestrator/src/service.ts` — façade delegation driven by explicit plans.
- `docs/architecture.md` — Coordination Layer package map and ownership boundary.

## Security


moncher-dev (commented):
Steward smoke test at head `1f3dfe9d` (base `41afaad9`): `pnpm install --frozen-lockfile && pnpm build && pnpm test` green; all packages pass, orchestrator 365/365.

**Key invariant from the #965 design review: deferral keeps the claim.** Verified by mutation:
- M1, in the extracted decision: `decideFinalizationDisposition` returns `complete` (which releases the claim) instead of `defer-finalization` below the bound → **8 fail** (7 in `service.test.ts`, including "persists unknown finalization deferrals…", "retains an exited deferred run when candidate reconciliation loses its item", and "recovers a transient unknown finalization read…", plus the new unit test).
- M2, in the service: the defer branch calls `releaseRunIssueOrchestration` before returning → **5 failed / 219 passed**, the same count as the same mutation on main before this PR. `service.test.ts` is not touched by this PR; the guards survived rather than being moved or weakened.

Other extracted decisions, mutated one at a time (all caught):
- completed-run suppression `>=` → `>`: 5 fail
- requeue retained due time dropped: 4 fail
- continuation no longer resets attempt to 1: 5 fail

Behaviour-equivalence read: the retry policy is still loaded only for non-recovery failure retries, and suppression is still decided before any policy load. The only difference: the requeue path now loads the project poll interval even when the retry is about to be suppressed. That is a harmless extra read with no behaviour change. Passing to hojinzs for review.

packages/orchestrator/src/retained-decisions.ts
```diff
@@ -0,0 +1,204 @@
...
+  | {
+      action: "defer-finalization";
+      consecutiveDeferrals: number;
+      exhausted: false;
+    }
+  | {
+      action: "enter-failure/continuation-retry";
+      consecutiveDeferrals: number;
+      exhausted: boolean;
+    };
```
hojinzs: **P2 — the disposition union under-specifies what the façade needs, so the façade still branches on raw tracker state.**

`enter-failure/continuation-retry` collapses two cases that need *different* effects:

- tracker `active` → fall straight through to retry classification, persist nothing;
- tracker `unknown` **and** deferrals exhausted → must first persist the deferral record and emit `[run-finalization-deferred]`, *then* fall through.

Because the returned action can't tell those apart, `service.ts:3807` has to re-test `currentTrackerProgress?.state === "unknown"` itself. The decision module owns the bound but not the disposition, which is the part the issue's "explicit inputs/outputs" was meant to fix.

Suggest a third action so the caller drives entirely off the return value:

```ts
| { action: "complete" }
| { action: "defer-finalization"; consecutiveDeferrals: number }
| { action: "record-exhausted-deferral"; consecutiveDeferrals: number }  // persist, then retry
| { action: "retry" }                                                     // active/no progress
```

That also removes the need for the `exhausted` flag, which is meaningless on the `active` variant (it's hardcoded `false` there).

---

packages/orchestrator/src/service.ts
```diff
@@ -3793,10 +3799,16 @@ export class OrchestratorService {
...
-      const exhausted = consecutiveDeferrals >= MAX_FINALIZATION_DEFERRALS;
+    const finalizationDisposition = decideFinalizationDisposition({
+      trackerProgress: currentTrackerProgress,
+      currentDeferralCount: runWithTokens.finalizationDeferralCount ?? 0,
+      maxDeferrals: MAX_FINALIZATION_DEFERRALS,
+    });
+    if (
+      currentTrackerProgress?.state === "unknown" &&
+      finalizationDisposition.action !== "complete"
+    ) {
```
hojinzs, line 3839: **P3 — the second conjunct is dead code.**

`decideFinalizationDisposition` returns `complete` only for `state === "non-actionable"`, so whenever the first conjunct holds the second is necessarily true. Verified by mutation: reducing this to `if (currentTrackerProgress?.state === "unknown") {` leaves `service.test.ts` at **224/224 passing** — nothing tests the conjunct because nothing can.

As written it reads as if the disposition were guarding something here, which is exactly the confusion the extraction is supposed to remove. Either drop it, or (preferred) adopt the explicit `record-exhausted-deferral` action suggested on `retained-decisions.ts` and gate purely on the disposition.

---

packages/orchestrator/src/service.ts
```diff
@@ -3915,17 +3927,37 @@ export class OrchestratorService {
...
+    });
+    const retryDisposition =
+      initialRetryDisposition.action === "retry" &&
+      !recovery &&
+      retryKind === "failure"
+        ? decideCompletedRunRetry({
+            ...retryDecisionInput,
+            retryPolicy: await this.loadRetryPolicy(tenant, run.repository),
+          })
+        : initialRetryDisposition;
```
hojinzs: **P2 — the speculative double-call duplicates the decision rule in the façade, where it can drift silently.**

To avoid loading the retry policy on paths that don't need it, the call site computes a throwaway disposition with `retryPolicy: null`, then re-derives the decision module's *own* branch predicate — `!recovery && retryKind === "failure"` — to decide whether to recompute. That predicate now lives in two places:

- `retained-decisions.ts:111` — `if (input.recoveryPresent || input.retryKind === "continuation")` picks the continuation delay;
- here — `!recovery && retryKind === "failure"` picks whether a policy is even fetched.

They agree today. But if the decision module is ever changed so another path consults `retryPolicy` (say, recovery retries gain a policy-driven delay), this gate keeps passing `null` and the delay silently degrades to `defaultFailureBackoffMs`. No type error, and no test failure — the pure tests pass a policy in directly, and the service tests would still see *a* plausible timestamp.

Two ways out, either is fine:

1. **Plan → resolve.** Have the module return the intent first, so the façade can't guess wrong:
   ```ts
   const plan = planCompletedRunRetry(retryDecisionInput); // "suppress" | "continuation" | "failure"
   const retryDisposition = resolveCompletedRunRetry(plan, {
     retryPolicy: plan.needsRetryPolicy
       ? await this.loadRetryPolicy(tenant, run.repository)
       : null,
   });
   ```
2. **Keep the delay in the façade.** Return only `attempt`, `failureRetryCount`, `persistedRetryKind` and `delayClass`, and let the façade — which already owns every other effect — turn `delayClass` into `nextRetryAt`. The accounting is the part worth extracting and testing; the `Date` arithmetic is not.

The same pattern repeats at `requeueRetryingRun` (line 5563), so whichever shape you pick, please apply it to both.

Also worth noting for the equivalence argument: the ordering guarantee that `loadRetryPolicy` is only reached after the suppression check is preserved here (`initialRetryDisposition.action === "retry"`), which is correct — just fragile for the same reason.

---

packages/orchestrator/src/service.ts
```diff
@@ -5532,45 +5537,51 @@ export class OrchestratorService {
...
+      retryPolicy: null,
+    });
+    const requeueDisposition =
+      initialRequeueDisposition.suppressed || !advanceAttempt
+        ? initialRequeueDisposition
+        : decideRetryRequeue({
+            ...requeueDecisionInput,
+            pollIntervalMs: DEFAULT_POLL_INTERVAL_MS,
+            retryPolicy: await this.loadRetryPolicy(tenant, run.repository),
+          });
```
hojinzs: **P2 — `DEFAULT_POLL_INTERVAL_MS` used as an inert placeholder, and one latent ordering change.**

Two separate things in this block:

**1. The sentinel (P2).** When `advanceAttempt` is true, `pollIntervalMs: DEFAULT_POLL_INTERVAL_MS` is passed as a value that is never read — the function returns from the `failure-policy`/`failure-fallback` branch first. But it isn't an obviously-inert value: it's the real 30 s default. If the branch order in `decideRetryRequeue` ever changes, a wrong-but-plausible 30 s delay silently replaces the project's configured poll interval, and nothing fails. Prefer `pollIntervalMs: number | null` with the function treating `null` as "must not reach the poll branch", so a reordering surfaces as a crash or a type error rather than a subtly wrong `dueAt`.

**2. The eager read (P3, latent).** `loadProjectPollInterval` is now awaited on `!advanceAttempt && retainedDueAt === null` — i.e. *before* `suppressed` is known. The original only called it inside `if (!suppressed)`.

Unreachable today: both `advanceAttempt: false` callers (lines 3768 and 5281) also pass `countFailure: false`, so `suppressed` is always `false` on that path. But it's worth guarding, because `loadProjectPollInterval` (line 5673) is the only one of the three loaders here with **no** `try`/`catch` — compare `loadRetryPolicy` (5710) and `loadMaxFailureRetries` (6282), which both swallow and fall back. So the day a caller passes `advanceAttempt: false` while letting `countFailure` default to `true`, a `loadProjectWorkflow` throw stops `requeueRetryingRun` from ever writing the max-failure-retry suppression record — the run silently keeps its old state instead of being marked suppressed.

Folding the read into whichever shape you pick from the `decideCompletedRunRetry` comment resolves both points at once.

---

packages/orchestrator/src/retained-decisions.test.ts
```diff
@@ -0,0 +1,131 @@
...
+      })
+    ).toMatchObject({
+      suppressed: false,
+      attempt: 2,
+      failureRetryCount: 1,
+      nextRetryAt: "2026-09-13T23:59:00.000Z",
+      delayClass: "retained",
+    });
+  });
+});
```
hojinzs: **P2 — `decideRetryRequeue` is the least-covered decision, and it's the one with the most outcomes.**

It returns five `delayClass` values; exactly one (`retained`) is tested directly. `none` (suppressed), `poll`, `failure-policy` and `failure-fallback` have no direct row.

Measured with only this file in scope:

```
npx vitest run packages/orchestrator/src/retained-decisions.test.ts \
  --coverage --coverage.include='packages/orchestrator/src/retained-decisions.ts'

 ...d-decisions.ts |   77.5 |    67.74 |     100 |    77.5 | ...166-173,183-200
```

67.74% branch — uncovered lines 166–173 (suppression) and 183–200 (poll / policy / fallback). The 100% in the coverage bot comment is the aggregate run, where `service.test.ts` drives those branches indirectly; that's the façade contract you already had before this PR, not the new direct coverage the description claims ("Add direct decision-table coverage", "8 focused retained-decision tests").

Since directly testable decision tables are the stated point of the extraction — and the issue asks for "focused regression cases for the acceptance criteria" — please fill in the remaining rows. An `it.each` over `(countFailure, advanceAttempt, retainedDueAt, retryPolicy) → (suppressed, attempt, failureRetryCount, delayClass)` would cover all five in roughly the space of the existing single case.

Two assertions worth pinning explicitly while you're there, since both encode invariants the façade depends on:
- `suppressed` requires `countFailure` — a non-counting requeue must never suppress, which is what keeps the capacity-postponement path at lines 3768/5281 safe;
- `scheduleRetryAt` receives the *advanced* attempt (`currentAttempt + 1`), matching the pre-refactor `scheduleRetryAt(now, attempt, retryPolicy)`.

---

hojinzs (commented):
Reviewed at head `1f3dfe9d` against #898. **No correctness defect found — I agree this is behavior-preserving.** My findings are all design/maintainability and test-coverage; none of them block on risk of a wrong retry or a lost run.

## Requirements vs. #898

| --- | --- |
| Retry budget unchanged | ✅ accounting is a faithful transcription; `scheduleRetryAt` still receives `run.attempt + 1` (`runWithTokens` spreads `run` without overriding `attempt`, so the old `run.attempt` / `runWithTokens.attempt` split was already equivalent) |
| Reservation order / capacity unchanged | ✅ retained-due-time path preserved; `retainedDueAt` is non-null only when `advanceAttempt === false`, so hoisting it behind `!advanceAttempt` is equivalent |
| Finalization outcome unchanged | ✅ `action === "complete"` ⟺ `state === "non-actionable"` |
| Event names / persisted schemas unchanged | ✅ no event or record shape touched |
| Reconciliation and tracker-state queues separate | ✅ untouched |
| `docs/architecture.md` package map updated | ✅ |
| `docs/symphony-spec.md` untouched | ✅ |
| Required checks | ✅ Test + Container Smoke green |

Scope discipline is good: three decisions, one module, effects left in the façade, no speculative interfaces. No over-engineering, and no unrequested behavior changes. Omitting the changeset is right — `refactoring` label, no user-visible surface.

## Smoke test

No preview environment applies to a Coordination-layer refactor, so I verified locally at head:

- `pnpm build` — clean.
- `retained-decisions.test.ts` + `service.test.ts` — **232/232 pass** (51 s).
- Mutation check on the guard at `service.ts:3807` — removing the `&& action !== "complete"` conjunct leaves **224/224** green, confirming it is dead code (P3 below).
- Direct-only coverage of the new module — **67.74% branch**, not the 100% the bot comment shows (that figure is aggregate, driven by the pre-existing service suite). Details in the inline comment.

I'll also confirm the author's own note: the requeue path now reads the project poll interval before suppression is known. That is unreachable with today's callers, so it is indeed harmless right now — but `loadProjectPollInterval` is the only loader in that block without a `try`/`catch`, which is why I've flagged it rather than waved it through.

## Findings

| # | Sev | Where | What |
| --- | --- | --- | --- |
| 1 | **P2** | `retained-decisions.ts:14-25` | Disposition union collapses "tracker active" and "unknown + exhausted", so the façade must still branch on raw tracker state |
| 2 | **P2** | `service.ts:3946-3958` | Speculative `retryPolicy: null` double-call re-derives the decision's own predicate in the caller — two copies that can drift silently |
| 3 | **P2** | `service.ts:5563-5578` | Same pattern, plus `DEFAULT_POLL_INTERVAL_MS` passed as an inert-but-plausible sentinel |
| 4 | **P2** | `retained-decisions.test.ts:108-131` | `decideRetryRequeue` has 5 outcomes, 1 directly tested |
| 5 | **P3** | `service.ts:3807-3810` | Provably dead conjunct (mutation-verified) |
| 6 | P3 | `service.ts:5565` | Poll-interval read hoisted above the suppression check — latent, see inline |
| 7 | nit | `retained-decisions.ts:22` | `"enter-failure/continuation-retry"` as a discriminant — the `/` reads as two actions; `exhausted: false` is meaningless on the active variant |

Findings 2 and 3 are one theme: the extraction moved the *arithmetic* out but left the *branch predicates* in the façade, duplicated. That's the main thing I'd like addressed, since it's the failure mode this refactor exists to prevent.

**To approve I need:** finding 4 (the missing decision-table rows — small and mechanical), and finding 2/3 either fixed or answered with a reasoned "not doing this, because…". Findings 1, 5, 6, 7 are yours to take or leave; a one-line reply on each is enough. If you'd rather keep this PR to the pure transcription and land the caller-shape cleanup as a follow-up child of #890, say so and I'll take that as the answer.

Nice work on the mutation evidence in the PR description — verifying that deferral keeps the claim, rather than asserting it, is what made this reviewable.

---

packages/orchestrator/src/retained-decisions.ts
moncher-dev: Addressed in `3d6ef9e`. The disposition now has explicit `retry` and `record-exhausted-deferral` actions. Both unknown-state actions carry the diagnostic reason/error, so the façade selects persistence and retry effects entirely from the disposition and no longer re-tests raw tracker state.

packages/orchestrator/src/service.ts
moncher-dev: Addressed in `3d6ef9e`. The dead conjunct is gone; the façade now records deferrals only for the two explicit decision actions, with `record-exhausted-deferral` selecting the persist-then-retry path.

packages/orchestrator/src/service.ts
moncher-dev: Addressed in `3d6ef9e` with the suggested plan→resolve shape. `planCompletedRunRetry` owns suppression, accounting, and the `continuation`/ `failure` policy requirement; the façade loads policy only when the plan explicitly requests `failure`, then `resolveCompletedRunRetry` computes the timestamp. The same shape is applied to requeue decisions.

packages/orchestrator/src/service.ts
moncher-dev: Addressed in `3d6ef9e`. `planRetryRequeue` resolves suppression before effects and returns an explicit `poll` or `failure` requirement. The façade passes `pollIntervalMs: null` for non-poll paths and loads the poll interval only for a `poll` plan, removing both the plausible default sentinel and the eager loader call.

packages/orchestrator/src/retained-decisions.test.ts
moncher-dev: Addressed in `3d6ef9e`. The focused table now covers `none`, `retained`, `poll`, `failure-policy`, and `failure-fallback`. Separate assertions pin that suppression requires `countFailure` and that policy scheduling uses the advanced attempt. The focused file passes 14 tests with 89.79% direct branch coverage; the remaining uncovered branches belong to other decision families.

hojinzs (approved):
## Approve — round 2 at head `3d6ef9e`

All seven findings from my round-1 review are addressed. I verified each rather than taking the thread replies at face value; the two I gated approval on (finding 4, and findings 2/3) are fixed in the stronger of the two shapes I offered.

### Findings — verified state

| # | Sev | Claim | Verified |
| --- | --- | --- | --- |
| 1 | P2 | explicit `retry` / `record-exhausted-deferral` actions | ✅ façade no longer re-tests raw tracker state; `reason`/`error` now flow from the disposition |
| 2 | P2 | plan→resolve for completed-run retry | ✅ predicate exists once, in `planCompletedRunRetry` |
| 3 | P2 | plan→resolve for requeue; `pollIntervalMs: number \| null` | ✅ sentinel gone; null now throws rather than yielding a plausible 30 s |
| 4 | P2 | all five `delayClass` rows covered | ✅ **89.79% direct branch**, measured below |
| 5 | P3 | dead conjunct removed | ✅ gone |
| 6 | P3 | eager poll-interval read | ✅ `delayClass === "poll"` implies `!suppressed` structurally — `planRetryRequeue` returns on suppression first |
| 7 | nit | `/` discriminant, meaningless `exhausted` | ✅ both gone; `exhausted` is derived in the façade from the action |

### Equivalence re-check of the restructure

The plan→resolve split is itself a behavior-affecting change, so I re-derived the loader gates rather than assuming:

- **Completed-run retry.** Old gate `action === "retry" && !recovery && retryKind === "failure"`; new gate `delayClass === "failure"`, which holds iff `!recoveryPresent && retryKind === "failure"`. Equivalent.
- **`scheduleRetryAt`** moved from `currentAttempt + 1` to `plan.attempt`. On the `failure` delay class `persistedRetryKind` is necessarily `"failure"`, so `attempt === currentAttempt + 1`. Equivalent.
- **Requeue.** Policy gate `delayClass === "failure"` ⟺ old `!suppressed && advanceAttempt`. Equivalent. Poll gate is now *strictly narrower* than the old `!advanceAttempt && retainedDueAt === null` — that narrowing is finding 6's fix, not a regression.
- **Deferral branch.** Both persisting actions are returned only from the `state === "unknown"` arm, so the new action test matches the old `state === "unknown" && action !== "complete"`.

### Smoke test (local, head `3d6ef9e`)

No preview environment applies to a Coordination-layer refactor, so I ran it locally:

- `pnpm install --frozen-lockfile`, `pnpm build`, `pnpm typecheck`, `pnpm lint`, `pnpm format` — all clean.
- `retained-decisions.test.ts` + `service.test.ts` — **238/238 pass** (was 232; six new).
- Direct-only coverage of the new module — **89.79% branch**, up from 67.74%. Matches the description exactly. Remaining uncovered: the `failure-policy`/`failure-fallback` ternary in `resolveCompletedRunRetry` and the poll-interval throw — both exercised indirectly by `service.test.ts`.
- CI on this head: Test + Container Smoke green.

**Mutation checks on what changed this round** (the guards have to be load-bearing, not just present):

- Drop the `record-exhausted-deferral` arm from the façade → **1 fail**. Guarded, if thinly.
- Force `retryPolicy: null` at the completed-run call site — i.e. exactly the silent-drift scenario finding 2 was about → **13 fail**. Previously this drift was invisible; it now breaks loudly. That is the substantive win of this round.

### Scope

Still three decisions, one module, effects in the façade. The round-2 changes stayed inside the findings — no speculative interfaces, no widening. `delayClass` is consumed only by the two loader gates and never reaches an event or persisted record, so the extra plan fields are inert at the boundary. Omitting the changeset remains correct.

Good work on the plan→resolve shape — turning the loader requirement into an explicit output is what makes the duplicated-predicate failure mode unrepresentable rather than merely absent.

---

moncher-dev (commented):
Steward re-test at head `3d6ef9e3`. hojinzs's two approval conditions both hold.

The branch merges cleanly (MERGEABLE) onto current main `be5f7dfc`. `pnpm install --frozen-lockfile && pnpm build && pnpm test` is green in every package, with orchestrator at 371/371.

**Finding 4: the `decideRetryRequeue` table is now pinned.** I mutated `retained-decisions.ts` one change at a time and ran `retained-decisions.test.ts` + `service.test.ts` (238 tests). Every mutation was caught:

| Mutation | Result |
| --- | --- |
| R1 suppression without `countFailure` | 1 fail (new "requires failure accounting before suppressing") |
| R2 requeue `>=` → `>` | 2 fail (new table row + service "retry refresh exhaustion suppresses") |
| R3 retained due time dropped (`retained` → `poll`) | 4 fail (new row + 3 service tests) |
| R4 `poll` outcome becomes `failure` | 1 fail (new row) |
| R5 suppressed plan still schedules (`none` → `failure`) | 2 fail |
| R6 requeue policy backoff uses the unadvanced attempt | 2 fail (new "schedules policy backoff with the advanced attempt") |
| R7 requeue fallback ignores the default backoff | 1 fail (new `failure-fallback` row) |
| R8 poll interval ignored | 2 fail |
| C2 completed-run policy backoff uses the unadvanced attempt | 2 fail (service) |

At `1f3dfe9d`, R1, R4, R6 and R7 had no direct test. They are all caught now.

**Findings 2/3: fixed, not deferred.** Both call sites now branch only on the plan's own output. The completed-run site loads policy only when `retryPlan.delayClass === "failure"`, and the requeue site loads the poll interval only when `requeuePlan.delayClass === "poll"`. Neither re-derives a predicate from raw inputs. The `DEFAULT_POLL_INTERVAL_MS` sentinel is gone: a `poll` plan with a null interval now throws rather than scheduling at a plausible default, and the poll-interval read is no longer above the suppression check (finding 6).

**Non-blocking. Neither is a regression, since both behave the same on main:**
- Two completed-run branches remain unguarded. C1 routes a `recoveryPresent` failure retry to failure-policy scheduling instead of the continuation delay. C3 drops the default backoff for the no-policy fallback. Both stay 238/238 green. I applied the same two mutations to main's inline `service.ts`, and both stay 224/224 green there too. So these are pre-existing gaps, but the new module makes each one a single table row. Worth adding while you are here, or as a follow-up.
- `resolveCompletedRunRetry` (lines 147–154) and `resolveRetryRequeue` (lines 239–248) repeat the same failure-policy/fallback timestamp block. This is the same drift theme as findings 2/3, one level down, and a small shared helper would close it.

This is a Coordination-layer refactor with no preview surface. The evidence is the mutation runs above, not a live dispatch.

Passing back to hojinzs for review.

packages/orchestrator/src/retained-decisions.ts
```diff
@@ -0,0 +1,249 @@
...
+  return {
+    ...plan,
+    nextRetryAt: (input.retryPolicy
+      ? scheduleRetryAt(input.now, plan.attempt, input.retryPolicy)
+      : new Date(input.now.getTime() + input.defaultFailureBackoffMs)
+    ).toISOString(),
+    delayClass: input.retryPolicy
+      ? ("failure-policy" as const)
+      : ("failure-fallback" as const),
+  };
```
hojinzs, line 155: **nit — non-blocking, do not respin this PR for it.**

This block and the tail of `resolveRetryRequeue` are the same five lines: pick `scheduleRetryAt` vs. `now + defaultFailureBackoffMs`, then stamp `failure-policy` / `failure-fallback`. Extracting a shared `resolveFailureDelay(now, attempt, retryPolicy, defaultFailureBackoffMs)` would close it.

I'm flagging it because it's the findings 2/3 theme one level down — the same rule in two places, able to drift silently. It's materially smaller though: both copies are pure, in one file, adjacent, and directly testable, so a drift shows up as a failing table row rather than a wrong `dueAt` in production. That's why it's a nit and not a P3.

Worth folding into the same follow-up as the two uncovered branches below, not into this PR.

---

hojinzs (approved):
## Approve stands — round 3 at head `3d6ef9e`

Head is unchanged since my round-2 approval (`3d6ef9e`, tree clean locally at that exact SHA), all five round-1 threads are resolved, and CI is green (Test + Container Smoke). The only new input is the steward's re-test, which raised two items I had not seen. I verified both myself rather than accepting them, and neither changes the verdict.

### Verifying the steward's two new items

**Claim: completed-run branches C1 and C3 are unguarded, but pre-existing.** Confirmed on both halves — the second half is the one that matters, and it is the one a re-test could most easily have gotten wrong.

| Mutation | This PR (`retained-decisions.ts`) | Base `41afaad` (inline `service.ts`) |
| --- | --- | --- |
| C1 — `recoveryPresent` failure retry drops to failure-policy instead of the continuation delay | 238/238 green | 224/224 green |
| C3 — no-policy fallback ignores `defaultFailureBackoffMs` | 238/238 green | 224/224 green |

Both survive here *and* survive the equivalent mutation on the pre-refactor inline code. So these are inherited gaps in `service.test.ts`, not guards this refactor weakened — which is the distinction that decides blocking vs. not. Baseline 238/238 reproduced, matching the steward's count.

**Claim: duplicated failure-policy/fallback block.** Confirmed; flagged inline as a nit with my reasoning for why it ranks below the findings 2/3 it resembles.

### Verdict

Approving, and I am **not** asking for another round. Concretely:

- **Nothing blocks merge.** The two uncovered branches are pre-existing and equally uncovered on `main`; merging does not make them worse. The duplication is a nit.
- **Do not add them to this PR.** #898 scoped this to extracting the retained decisions, and it did exactly that — three decisions, one module, effects in the façade. Pulling in test gaps this refactor inherited would widen a PR that has already met its acceptance criteria.
- **Follow-up instead.** C1, C3 and the shared-helper nit are worth one small child of #890. The extraction is what makes them cheap — each is now a table row against a pure function rather than a service-level scenario. That is the refactor paying off, not a debt it created.

To be explicit about round count: this is round 3 of a maximum 6, and I'm closing the loop here deliberately. The acceptance criteria in #898 are met, the approval conditions I set in round 1 were met in round 2 and re-verified above, and continuing would be exactly the unbounded ping-pong the process is meant to prevent. Please merge; no reply to the inline nit is needed.

Requirements re-check against #898 is unchanged from my round-2 review — retry budget, reservation order/capacity, finalization outcome, event names and persisted schemas all preserved; queues still separate; `docs/architecture.md` updated; `docs/symphony-spec.md` untouched; changeset correctly omitted for a `refactoring`-labelled change with no user-visible surface.

---

Issue #896, opened 2026-09-05
perf(orchestrator): reduce run scans at explicit freshness boundaries


moncher-dev:
Parent Epic: #890. Approved registration plan: [Q6: #896](https://github.com/hojinzs/github-symphony/issues/896).

## Background

Merely exposing scoped reads leaves repeated service-level scans in place.

## Proposed Changes

Adopt [Q5: #895](https://github.com/hojinzs/github-symphony/issues/895) and remove redundant reads at explicitly documented freshness boundaries. Retain fresh active-run observations and historical publication/metric evidence; use measurements from [Q4: #894](https://github.com/hojinzs/github-symphony/issues/894). Do not reuse one immutable snapshot blindly across asynchronous updates.

## Affected Files

| File | Change | Scope |
| --- | --- | --- |
| `packages/orchestrator/src/service.ts`, `service.test.ts` | Query adoption and update-freshness regressions | Coordination |
| Existing snapshot tests and `docs/architecture.md` | Verify stable aggregates and document query ownership | Observability / documentation |

## Acceptance Criteria and Verification

Same dispatch/retry choices, cumulative tokens, legacy recovery, and unpublished-work protection; mid-tick events remain visible; benchmark improvement is reported against [Q4: #894](https://github.com/hojinzs/github-symphony/issues/894). Full required checks and relevant Docker integration scenarios in an available CI environment.

## Symphony Layers and Compatibility

Coordination, Observability. Keep provider-specific behavior in adapters and preserve current persistence, retry, metric, and unpublished-work contracts. Do not edit `docs/symphony-spec.md`. No new spec divergence is intended.

## Planning

- Priority: P2
- Size: M
- Estimate: 8 net implementation hours (excludes review, testing, and CI waiting).
- Initial project status: Backlog.

## Dependencies

Blocked by: [Q5: #895](https://github.com/hojinzs/github-symphony/issues/895).

## Required Checks

For implementation changes, run:

```bash
pnpm lint
pnpm test
pnpm typecheck
pnpm build
```

Add focused regression cases for the acceptance criteria. For applicable integration changes, use the existing CI container checks; #872 records that Docker is intentionally unavailable on the orchestration host. Do not provision Docker or duplicate #887/#889. Update applicable living documents in the same change.

## References

- #890
[...]

PR #978, opened 2026-09-14
refactor(store): add project-scoped bounded run inventory

### LINKED ISSUE #895: refactor(store): add project-scoped bounded run inventory
Parent Epic: #890. Approved registration plan: [Q5: #895](https://github.com/hojinzs/github-symphony/issues/895).

## Background

`loadAllRuns()` reads history outside the target project and fans out reads without an explicit bound.

## Proposed Changes

Add a compatible project-scoped run-query contract and filesystem implementation with bounded read concurrency. Preserve legacy layout lookup and existing methods for callers not migrated yet. Keep authoritative JSON storage.

## Affected Files

| File | Change | Scope |
| --- | --- | --- |
| `packages/core/src/contracts/state-store.ts` | Compatible query contract | Core contract |
| `packages/orchestrator/src/fs-store.ts`, `fs-store.test.ts` | Scoped reads, bounded I/O, legacy tests | Persistence |
| `docs/architecture.md` | Record storage boundary | Documentation |

Only target-project records are returned while legacy compatibility remains; deterministic tests establish bounded read concurrency; no run deletion, schema migration, or token-total change. Existing callers remain functional before [Q6: #896](https://github.com/hojinzs/github-symphony/issues/896). Full required checks.

Coordination (core contract and filesystem implementation). Keep provider-specific behavior in adapters and preserve current persistence, retry, metric, and unpublished-work contracts. Do not edit `docs/symphony-spec.md`. No new spec divergence is intended.

## Planning

- Priority: P1
- Size: L
- Estimate: 12 net implementation hours (excludes review, testing, and CI waiting).
- L-size split reviewed: keep the compatible query contract and filesystem implementation atomic; adoption is isolated in [Q6: #896](https://github.com/hojinzs/github-symphony/issues/896). Split again before Ready if scope expands.

## Dependencies

Blocked by: [Q4: #894](https://github.com/hojinzs/github-symphony/issues/894).

## Required Checks

```bash
pnpm lint
pnpm test
pnpm typecheck
pnpm build
```

## References

- #890

moncher-dev:
## Issues

- Closes #895

## Summary

- Adds a compatible project-scoped run inventory contract for callers migrating away from global scans.
- Reads only authoritative records for the requested project while retaining legacy flat-layout lookup and scoped-record precedence.
- Bounds filesystem `run.json` reads to eight concurrent operations for scoped and global inventory paths, with mutation-sensitive regression coverage for each guarantee.

## Change-point diagram

- `@gh-symphony/core` state-store contract → `loadRuns({ projectId })`
- `@gh-symphony/orchestrator` filesystem store → requested scoped + legacy paths → bounded reads → project filter → scoped precedence
- Coordination architecture → records the compatibility and persistence boundary

## Start here

- `packages/core/src/contracts/state-store.ts:11` — project-scoped query contract
- `packages/orchestrator/src/fs-store.ts:263` — bounded scoped and legacy inventory implementation
- `packages/orchestrator/src/fs-store.test.ts:612` — path isolation, compatibility, precedence, and concurrency regressions
- `docs/architecture.md:148` — storage ownership and compatibility boundary

- No CLI surface changes. Existing `loadAllRuns()` callers remain functional until adoption in #896; authoritative JSON storage, token accounting, retry behavior, metrics, and unpublished-work handling are unchanged.

## Validation

- `pnpm exec vitest run packages/orchestrator/src/fs-store.test.ts` — pass (30 tests)
- `pnpm lint` — pass
- `pnpm test` — pass (all 14 tested workspace projects; orchestrator 380 tests)
- `pnpm typecheck` — pass
- `pnpm build` — pass
- `git diff --check` — pass
- Mutation M3 (remove legacy `projectId` content filter) — scoped inventory regression fails by returning `legacy-other`.
- Mutation M4 (remove scoped-record de-duplication) — precedence regression fails because the legacy duplicate replaces the scoped copy.
- Mutation M2 (replace bounded global mapping with `Promise.all`) — global concurrency regression fails with peak 12 instead of 8.
- Mutation M5 (read every project's scoped directory before filtering) — observed-path regression fails on the unrelated `project-2` path.
- Docker E2E / blackbox: not applicable to this persistence-internal contract addition; dispatch, worker lifecycle, tracker adapters, status API, and CLI runtime behavior are unchanged.

## Changeset

- Not needed because the issue has no changeset label and does not change published CLI behavior.

## Risks & rollback

- Reviewer focus: legacy flat records are filtered by stored `projectId`, scoped records take precedence for duplicate run IDs, and both inventory APIs enforce the deterministic read cap.
- Rollback is the three issue commits; no migration, deletion, or schema rewrite occurs.

## Changed files

- `packages/core/src/contracts/state-store.ts` — adds the project-scoped run query.
- `packages/orchestrator/src/fs-store.ts` — adds bounded scoped and legacy inventory reads and bounds the retained global method.
- `packages/orchestrator/src/fs-store.test.ts` — verifies read-path isolation, legacy filtering, scoped precedence, and both concurrency caps.
- `docs/architecture.md` — documents the run-storage contract and compatibility boundary.

- [ ] Confirm caller adoption remains scoped to #896.

## Security


moncher-dev (commented):
Steward review at `4cc55d20`. **Changes needed: three documented guarantees are unguarded by tests.**

**What passes**
- `pnpm install --frozen-lockfile && pnpm build && pnpm test`: all 14 packages pass (orchestrator 379, cli 589, worker 242, core 340).
- **Real data.** I ran `loadRuns` against a read-only copy of this host's live daemon state: 242 run directories, all in the legacy flat `runs/` layout. `loadRuns({ projectId: "github-symphony-7f2244c6" })` returns 236 records, the same set as `loadAllRuns()` filtered by project. An unrelated projectId returns 0. The 242 `run.json` reads peak at **8** concurrent, and every read goes through the counted `readJsonFile` wrapper from #969.
- **Mutation M1.** Replacing `loadRuns`' `mapWithConcurrency` with `Promise.all` fails `bounds concurrent run record reads`. The loadRuns bound is pinned.

**What no test catches.** For each mutation below, `fs-store.test.ts` still passes 29/29. For M2, the whole orchestrator package still passes 379/379.

1. **M3: drop the `run?.projectId === query.projectId` content filter.** This filter is the entire scoping mechanism on this host, because every live run is in the legacy flat layout. With the filter removed, `loadRuns({ projectId: "some-other-project" })` against the live copy returns **236** records from a different project, and no test fails. The scoping test puts only a project-1 record in the legacy directory, so the filter is never exercised. Add a legacy-layout record owned by another project and assert that it is excluded.
2. **M4: drop `!projectRuns.has(run.runId)`, so the legacy record overrides the scoped one.** `docs/architecture.md` says the implementation "prefers the scoped record when a run exists in both layouts". No test has the same runId in both layouts. Add one where the two copies have different contents, and assert that the scoped copy wins.
3. **M2: revert `loadAllRuns` to unbounded `Promise.all`.** The docs say reads are "capped at eight concurrent reads for both APIs", but only `loadRuns` has a concurrency test. Apply the same observer-based assertion to `loadAllRuns`.
4. **M5: also read every other project's `runs` directory, then content-filter.** The results are identical, so nothing fails. Yet the issue's stated problem is that `loadAllRuns()` "reads history outside the target project". Use `observeRunRecordReads` to assert the set of read paths, meaning no path under another project's `projects/<id>/runs`.

Each new or changed test should fail when the guarded line is removed. Please state the mutation you checked for each one in the PR.

moncher-dev:
Steward re-review at `5dd04fc5`. **Passes. Requesting review from @hojinzs.**

**Build and suite:** `pnpm install --frozen-lockfile && pnpm build && pnpm test` passes on all 14 packages (orchestrator 380, cli 589, worker 242, core 340). CI is green and the branch merges cleanly with `main` (`89d82563`).

**All four findings from the last review now fail a test when the guarded line is removed.** I applied each mutation to `fs-store.ts` and ran `fs-store.test.ts` (30 tests):
- **M3, drop the `run?.projectId === query.projectId` filter:** `loads only the requested project's scoped and legacy run records` fails because the result gains the `legacy-other` record.
- **M4, let the legacy copy override the scoped copy:** the same test fails because the `scoped copy` record is replaced by the `legacy copy`.
- **M2, revert `loadAllRuns` to unbounded `Promise.all`:** `bounds concurrent legacy-compatible global run record reads` fails (peak read count 12, expected 8).
- **M5, also read every other project's `projects/<id>/runs` before filtering:** the scoped test fails because 5 read paths are observed instead of 4 (the extra one is under `project-2`).

**The earlier passing checks still hold on this head.** I made a read-only copy of this host's live `run.json` files: 244 run directories, all in the legacy flat layout, 238 with a record. Against that copy, `loadRuns({ projectId: "github-symphony-7f2244c6" })` returns 238 records, the same set with the same contents as `loadAllRuns()` filtered by project. An unrelated projectId returns 0 records. Both APIs peak at **8** concurrent reads, and all 244 reads are observed through #969's counted `readJsonFile` wrapper, which is the only `run.json` read path in `fs-store.ts`.

**Non-blocking:** the scoped test asserts the exact order of `observedPaths`, and that order comes from unsorted `readdir` output for the legacy directory. It passed 8 of 8 runs here and passes in CI, but on filesystems that return directory entries in hash order it could flake. Comparing the paths as a sorted list would remove that risk.

packages/orchestrator/src/fs-store.test.ts
```diff
@@ -608,6 +609,99 @@ describe("OrchestratorFsStore.loadRecentRunEvents", () => {
...
+      () => store.loadRuns({ projectId: "project-1" })
+    );
+
+    expect(runs).toEqual([records[0], records[2]]);
+    expect(observedPaths).toEqual([
+      join(store.runDir("scoped", "project-1"), "run.json"),
+      join(store.runDir("legacy"), "run.json"),
+      join(store.runDir("legacy-other"), "run.json"),
+      join(store.runDir("scoped"), "run.json"),
+    ]);
```
hojinzs, line 643: **P2 — this exact-order assertion passes by coincidence, not by construction.**

`safeReadDir` is a bare `readdir` with no sort ([`packages/core/src/observability/fs-reader.ts:16-18`](https://github.com/hojinzs/github-symphony/blob/5dd04fc58f2769ba4ee8ecb1a3847a9884d75c19/packages/core/src/observability/fs-reader.ts#L16)), so the order of the three legacy entries here is whatever the filesystem returns.

I measured it rather than assuming. Creating entries in the order `scoped, legacy, legacy-other, zz-run, aa-run, mm-run` and reading the directory back gives `aa-run, legacy, legacy-other, mm-run, scoped, zz-run` — i.e. this filesystem returns entries **sorted**, not in creation order. The asserted order (`legacy`, `legacy-other`, `scoped`) is exactly the alphabetical order of those three names, which is why it is green here and in CI. On ext4 with `dir_index`, `readdir` returns entries in filename-hash order, and this assertion would fail for reasons unrelated to the guarantee it is protecting.

This matters more than a normal flake because this is the assertion carrying the M5 guarantee. I confirmed M5 does fail correctly today (patched `loadRuns` to read every project's scoped directory → this test fails, 5 observed paths instead of 4), so the coverage is genuinely there — it is only the ordering that is fragile.

The set-membership property is the one you actually want, and it survives any `readdir` order:

```suggestion
    expect([...observedPaths].sort()).toEqual(
      [
        join(store.runDir("scoped", "project-1"), "run.json"),
        join(store.runDir("legacy"), "run.json"),
        join(store.runDir("legacy-other"), "run.json"),
        join(store.runDir("scoped"), "run.json"),
      ].sort()
    );
```

Note `expect(runs).toEqual([records[0], records[2]])` on line 637 is *not* affected — the scoped-before-legacy grouping in `runPaths` is deterministic, and the two survivors are unambiguous regardless of intra-directory order. Only `observedPaths` needs this.

---

packages/orchestrator/src/fs-store.test.ts
```diff
@@ -608,6 +609,99 @@ describe("OrchestratorFsStore.loadRecentRunEvents", () => {
...
+    expect(
+      observedPaths.some((path) => path.startsWith(store.runsDir("project-2")))
```
hojinzs, line 645: **P3 — this reads a `private` member, and no repo check can catch it.**

`runsDir` is declared `private` ([`fs-store.ts:84`](https://github.com/hojinzs/github-symphony/blob/5dd04fc58f2769ba4ee8ecb1a3847a9884d75c19/packages/orchestrator/src/fs-store.ts#L84)); the neighbouring `runDir` is public, which is why every pre-existing test in this file uses `runDir` only.

Compiling this file directly confirms it is a real type error:

```
packages/orchestrator/src/fs-store.test.ts(645,58): error TS2341:
  Property 'runsDir' is private and only accessible within class 'OrchestratorFsStore'.
```

`pnpm typecheck` stays green because `packages/orchestrator/tsconfig.json` sets `"exclude": ["src/**/*.test.ts"]`, and Vitest transpiles without type checking — so this compiles away silently and will keep doing so. Not a runtime problem, but it is an unchecked type violation that the next person touching `runsDir`'s visibility will discover the hard way.

Cheapest fix, using only the public accessor:

```suggestion
      observedPaths.some((path) =>
        path.startsWith(store.runDir("", "project-2"))
      )
```

Alternatively promote `runsDir` to public alongside `runDir`, if exposing the directory shape to tests is intended.

---

packages/orchestrator/src/fs-store.test.ts
```diff
@@ -608,6 +609,99 @@ describe("OrchestratorFsStore.loadRecentRunEvents", () => {
 });
 
 describe("history benchmark fixture", () => {
+  it("loads only the requested project's scoped and legacy run records", async () => {
```
hojinzs, line 612: **P3 — the three new tests are nested in the wrong `describe`.**

These land inside `describe("history benchmark fixture")` (line 611), but none of them touch the benchmark fixture — they construct their own `OrchestratorFsStore` and exercise `loadRuns` / `loadAllRuns` directly.

This is visible in the output. When I ran the M5 mutation, the failure reported as:

```
FAIL  src/fs-store.test.ts > history benchmark fixture > loads only the requested project's scoped and legacy run records
```

which points a future debugger at the benchmark fixture rather than at scoped inventory.

The file's own convention is one `describe` per store method — `OrchestratorFsStore.loadRecentRunEvents`, `OrchestratorFsStore.loadProjectIssueOrchestrations`. Closing the preceding `describe` and opening `describe("OrchestratorFsStore.loadRuns", ...)` for the two scoped tests would match it; the `loadAllRuns` concurrency test arguably belongs with the other `loadAllRuns` coverage near line 437.

---

packages/orchestrator/src/fs-store.ts
```diff
@@ -253,12 +257,39 @@ export class OrchestratorFsStore implements OrchestratorStateStore {
...
+      (runPath) => readJsonFile<OrchestratorRunRecord>(runPath)
     );
     return runs.filter((run): run is OrchestratorRunRecord => Boolean(run));
   }
 
+  async loadRuns(
+    query: OrchestratorRunQuery
+  ): Promise<OrchestratorRunRecord[]> {
+    const projectRunIds = await safeReadDir(this.runsDir(query.projectId));
+    const legacyRunIds = await safeReadDir(this.runsDir());
```
hojinzs, line 272: **P2 (not blocking this PR — flagging so #896 sets the right expectation).**

The scoping here is real for the *returned* set, but the legacy branch reads the entire flat `runs/` directory for every project, so it does not reduce *reads* on a host whose history is still flat.

Measured on a synthetic store — 50 legacy records owned by `p-other`, 1 scoped record owned by `p-mine`:

```
loadRuns({ projectId: "p-mine" })  →  1 record returned, 51 run.json files read
```

Your own live-host check in the re-review comment says all 244 run directories are in the legacy flat layout. So on that host today, `loadRuns` reads exactly the same 244 files as `loadAllRuns`; the only difference is that filtering moved inside the store.

That is the correct trade-off for this issue — #895 explicitly forbids migration or deletion, and `docs/architecture.md:152` documents the compatibility lookup honestly. The acceptance criterion ("only target-project records are returned") is met.

The risk is only that #896 lands expecting a read-count win and does not get one, because `saveRun` writes scoped but nothing ever moves existing records. Worth stating in #896 that the amplification win arrives as legacy records age out, not at adoption — otherwise the epic's baseline in `docs/reports/2026-09-14-run-history-polling-cost.md` (5 inventories/tick → 10,002 reads) will look unchanged after the migration and read as a regression in the follow-up.

---

packages/orchestrator/src/fs-store.ts
```diff
@@ -23,19 +24,22 @@ import { appendFileDurably, writeFileAtomically } from "./durable-file.js";
...
+const runRecordReadObserver = new AsyncLocalStorage<
+  (path: string) => void | Promise<void>
+>();
 
 // Every run.json read in this module must route through this wrapper. The
 // history benchmark's file-boundary read accounting depends on that invariant.
 async function readJsonFile<T>(path: string): Promise<T | null> {
   if (basename(path) === "run.json") {
-    runRecordReadObserver.getStore()?.(path);
+    await runRecordReadObserver.getStore()?.(path);
```
hojinzs, line 36: **nit —** widening the observer to `void | Promise<void>` puts an `await` on the production `run.json` read path purely so the concurrency tests can suspend inside the observer.

In production the store is unset, so this is `await undefined` — one extra microtask per run record, which I confirmed costs nothing measurable (see the benchmark in my summary). No objection to the approach; the async observer is what makes the bound observable at all.

Worth one line in the comment above `readJsonFile` recording *why* the await is load-bearing, though. As written, a future reader sees `await` on an optional callback that is `undefined` at runtime and will be tempted to drop it — which would silently defeat both concurrency tests rather than failing them.

---

packages/orchestrator/src/fs-store.ts
```diff
@@ -253,12 +257,39 @@ export class OrchestratorFsStore implements OrchestratorStateStore {
...
+      ...legacyRunIds.map((runId) => join(this.runDir(runId), "run.json")),
+    ];
+    const runs = await mapWithConcurrency(
+      runPaths,
+      RUN_RECORD_LOAD_CONCURRENCY,
+      (runPath) => readJsonFile<OrchestratorRunRecord>(runPath)
+    );
+    const projectRuns = new Map<string, OrchestratorRunRecord>();
+    for (const run of runs) {
+      if (run?.projectId === query.projectId && !projectRuns.has(run.runId)) {
```
hojinzs, line 286: **nit — two edge cases this predicate decides silently.** Both are fine as written; noting them so the choices are on record rather than accidental.

1. **A record with no `projectId` is dropped from the scoped view but still returned by `loadAllRuns`.** Verified: a legacy `run.json` written without the field yields `loadRuns → 0`, `loadAllRuns → 1`. `OrchestratorRunRecord.projectId` is a required `string` (`status-surface.ts:191`) and your live scan found the field on all 238 records, so this is only reachable for hypothetical pre-namespacing data — and dropping an unattributable record from a project-scoped query is the right call. Just means #896 loses it with no diagnostic.

2. **De-duplication keys on `record.runId`, not the directory name.** Two *different* directories whose records claim the same `runId` collapse to one under `loadRuns` while `loadAllRuns` returns both (verified: scoped 1, global 2). Directory name and `runId` agree everywhere `saveRun` writes, so this is unreachable in practice; it is the correct key for the scoped-vs-legacy precedence you are implementing.

---

hojinzs (approved):
## Approve — round 2 of 6. All four prior findings independently confirmed fixed; remaining items are P2/P3/nit and none block merge.

I re-ran everything from scratch rather than taking the PR's validation table on trust, and I verified the previous review's four mutations myself.

### Requirements vs. issue #895

Every acceptance criterion is met:

| Criterion | Status |
| --- | --- |
| Compatible project-scoped query contract | ✅ `loadRuns({ projectId })` added alongside the retained `loadAllRuns()` |
| Only target-project records returned | ✅ content filter on `projectId`; verified with a foreign-project legacy record |
| Legacy flat-layout compatibility preserved | ✅ legacy `runs/` still read; scoped copy wins on duplicate `runId` |
| Deterministic bounded-concurrency tests | ✅ both APIs pinned at 8, exact-peak assertions |
| No deletion / migration / schema or token change | ✅ read-only addition; `saveRun` untouched |
| Existing callers functional before #896 | ✅ all 10 `loadAllRuns()` call sites in `service.ts` unchanged, orchestrator 380/380 green |

No over-engineering found. `mapWithConcurrency` is ~20 lines with no new dependency, and applying the cap to `loadAllRuns` as well is squarely within the issue's stated problem ("fans out reads without an explicit bound") rather than scope creep. No changeset is correct — nothing reaches published CLI behavior.

### Previous review follow-up — all four mutations reproduced

I applied each mutation to `fs-store.ts` myself and ran `fs-store.test.ts` (30 tests). Each one fails exactly one test, and only one:

| Mutation | Result |
| --- | --- |
| **M2** — revert `loadAllRuns` to unbounded `Promise.all` | ❌ 1 failed / 29 passed |
| **M3** — drop the `projectId` content filter | ❌ 1 failed / 29 passed |
| **M4** — drop `!projectRuns.has(runId)` precedence | ❌ 1 failed / 29 passed |
| **M5** — read every project's scoped dir before filtering | ❌ 1 failed / 29 passed (5 observed paths, expected 4) |

The guarantees documented in `docs/architecture.md` are genuinely guarded.

### Checks, run locally on `5dd04fc5`

`pnpm lint` ✅ · `pnpm build` ✅ · `pnpm test` ✅ (14 packages; orchestrator 380, cli 589, worker 242, core 340) · `pnpm typecheck` ✅

### Smoke test

No API or UI surface here, and no preview environment applies — this is a persistence-internal contract addition, so I exercised the new code path directly against real on-disk stores instead. Behaviour matched the documented contract in every case: scoped/legacy isolation, legacy filtering by project, scoped-over-legacy precedence, and both concurrency caps holding at a peak of exactly 8.

**I also benchmarked the one thing the PR changes but does not measure** — capping the *existing* `loadAllRuns` at 8 concurrent reads, which sits on the reconciliation hot path called five times per tick per `docs/reports/2026-09-14-run-history-polling-cost.md`. At the report's 10,000-record fixture:

```
bounded(8) = 1339 / 1442 / 1392 ms
unbounded  = 1353 / 1346 / 1648 ms
ratio      = 1.00x
```

The cap is free at the size that matters. This was my main merge concern and it is resolved — worth adding to the PR's validation list, since "we bounded a hot path and it cost nothing" is the claim a future reader will want evidence for.

### Findings (all non-blocking, left inline)

- **P2** `fs-store.test.ts:638` — the exact-order `observedPaths` assertion. You flagged this yourself as non-blocking; I'd nudge it to P2 with evidence. `readdir` returned entries **sorted** on my host, and the asserted order happens to *be* the alphabetical order of those three names — so it is green by coincidence, not construction. On ext4 with `dir_index` (hash order) it breaks. Suggested a sorted comparison inline; it costs one line and this is the assertion carrying M5.
- **P2** `fs-store.ts:272` — scoping does not reduce *reads* while history is flat: 51 files read to return 1 record in my test, and your own live host is 244/244 legacy. Correct per the issue (no migration allowed) and honestly documented — but #896 should not be scoped expecting a read-count win at adoption.
- **P3** `fs-store.test.ts:645` — `store.runsDir(...)` is `private`; `tsc` reports TS2341, but `tsconfig.json` excludes `src/**/*.test.ts` so no repo check catches it.
- **P3** `fs-store.test.ts:612` — three `loadRuns`/`loadAllRuns` tests nested inside `describe("history benchmark fixture")`; failures misreport their origin.
- **nit** `fs-store.ts:36` — the load-bearing `await` on the observer deserves a comment. **nit** `fs-store.ts:286` — two silent edge cases recorded for the record.

None of these change behaviour. The P2 test-ordering fix is the only one I'd genuinely like to see land, and it can ride this PR or a follow-up — your call, not a condition of the approval.

---

PR #980, opened 2026-09-15
ci: run standalone Docker E2E on pull requests

moncher-dev:
## Issues

- Closes #979

## Summary

- Adds a dedicated `Standalone Docker E2E` job to every pull request targeting `main`.
- Makes every non-zero runner status—including Docker preflight `69`—fail CI, and also rejects a successful exit that never reaches the runner's final confirmation.
- Repairs the runner's remaining standalone `!` assertion and documents the CI/unattended-worker distinction and image-build boundary.

## Change-point diagram

- pull request → `Standalone Docker E2E` → runner output captured with `tee` → non-zero status or missing final marker fails the job

## Start here

- `.github/workflows/ci.yml:130` — dedicated pull-request CI job and final-marker gate
- `e2e/run-standalone-project-e2e.sh:295` — explicit unsafe-host negative assertion
- `AGENT_TEST.md:193` — documented CI contract and image-build isolation

## User-Visible Behavior / Operational Impact

- Pull requests now surface standalone Docker E2E regressions as a distinct CI job. Exit `69` is a failure, not a skip, and exit 0 is accepted only after the final `standalone-project Docker E2E passed` line.

## Validation

- `pnpm format` — pass
- `pnpm lint` — pass
- `pnpm test` — pass
- `pnpm typecheck` — pass
- `pnpm build` — pass
- `bash -n e2e/run-standalone-project-e2e.sh` — pass
- `./e2e/run-standalone-project-e2e.sh` — pass; reached `standalone-project Docker E2E passed`
- Final-head CI at `be59b3b` — [`Test`, `Container Smoke`, and `Standalone Docker E2E` pass](https://github.com/hojinzs/github-symphony/actions/runs/34949924746)
- Final-head `Standalone Docker E2E` — [passed in 60 seconds](https://github.com/hojinzs/github-symphony/actions/runs/34949924746/job/104318274077) and reached its final marker
- Assertion audit: `rg -n '^[[:space:]]*![[:space:]]+' e2e/run-standalone-project-e2e.sh` — only the expected `! test` inside an `if` condition remains
- Mutation evidence: identity failure string planted → [`Standalone Docker E2E` run 34946704338 failed](https://github.com/hojinzs/github-symphony/actions/runs/34946704338/job/104307787054)
[...]

PR #981, opened 2026-09-15
perf(orchestrator): reduce run scans at freshness boundaries

### LINKED ISSUE #896: perf(orchestrator): reduce run scans at explicit freshness boundaries

moncher-dev:
## Issues

- Closes #896

## Summary

- TL;DR: replace repeated global run inventories with one project-scoped historical snapshot plus bounded current-run refreshes at explicit asynchronous boundaries.
- Merge #977's terminal cleanup without restoring its global scan: reuse scoped evidence, refresh only live candidates, and retain identity-checked process liveness.
- Preserve dispatch/retry decisions, cumulative token metrics, legacy recovery, mid-tick visibility, and unpublished-work protection while reducing the 10,000-history fixture from 50,006 to 10,006 reads (79.99%).

## Change-point diagram

- project-scoped run inventory → stable historical evidence for the tick
- active/current run IDs → bounded refreshes after reconciliation, before candidates, after suppression, and before the final snapshot
- terminal cleanup → scoped snapshot or scoped fallback → fresh candidate read → identity-checked process liveness

## Start here

- `packages/orchestrator/src/service.ts:1419` — project-scoped inventory and freshness ownership
- `packages/orchestrator/src/service.ts:2450` — bounded current-run refresh and legacy compatibility
- `packages/orchestrator/src/service.ts:6236` — scoped terminal-cleanup liveness
- `packages/orchestrator/src/service.test.ts:14748` — combined #977 and unpublished-work reconciliation matrix

- Polling performs one history-sized run inventory per project tick and bounded active/current reads at four explicit freshness boundaries.
- Terminal cleanup avoids global scans, rechecks only live candidates, and continues to defer when same-tick branch publication records a transport failure or unpublished worktree.
- The #977 merge-order hazard is resolved in this PR; its terminal cleanup and lifecycle tests are preserved.

## Validation

- `pnpm lint` — pass
- `pnpm test` — pass across all workspace projects, including 392 orchestrator tests
- `pnpm typecheck` — pass
- `pnpm build` — pass
- `./e2e/run-standalone-project-e2e.sh` — pass in Docker
- `pnpm --silent --filter @gh-symphony/orchestrator benchmark:history > /tmp/symphony-896-cycle3-benchmark.json` — pass; 10,006 reads at 10,000 history in both layouts (legacy 71,889.65 ms; shared 72,544.59 ms)
- Mutation evidence: reintroduce `loadAllRuns()` in terminal cleanup → three terminal-reconciliation rows fail on the no-global-scan assertion; scoped implementation restored
- Mutation evidence retained from Cycle 2: remove the post-suppression `refreshCurrentRunSnapshot()` call → both unpublished-work rows fail because the sentinel workspace is deleted
- Mutation evidence retained from Cycle 1: replace the project inventory with `loadAllRuns()` → scoped-query tests fail; bypass current-run refreshes → token-freshness and polling-cost tests fail

## Changeset

- Not needed because issue #896 has no changeset label.

## Risks & rollback

- Risk: terminal cleanup adds one bounded read per candidate with a non-null process ID; candidate cardinality is proportional to project concurrency, not history.
- Reviewer focus: scoped cleanup evidence, legacy fallback retention, and process identity checks after merging #977.
- Rollback: revert merge commit `b984a88` and the three issue commits together to restore the prior inventory behavior.

## Changed files

- `packages/orchestrator/src/service.ts` — scoped inventory ownership, bounded freshness refreshes, and scoped terminal-cleanup liveness
- `packages/orchestrator/src/service.test.ts` — query, freshness, unpublished-work, and merged #977 lifecycle regressions
- `packages/orchestrator/src/fs-store.test.ts` — deterministic polling read-count guard
- `packages/orchestrator/src/history-benchmark.ts` — project-scoped benchmark instrumentation
- `docs/architecture.md` — query ownership, legacy compatibility, and terminal-cleanup freshness
- `docs/reports/2026-09-14-run-history-polling-cost.md` — benchmark comparison and matrix

- [ ] Observe production polling latency separately; the warm-cache benchmark is not a production latency threshold.

## Security


moncher-dev:
Steward review at `52bed14f`. **Fails. Moving #896 to Ready.** The Codex P1 at `service.ts:2253` is real: I reproduced it. On this head, terminal cleanup deletes a workspace that holds unpublished work. This breaks one of #896's acceptance criteria ("unpublished-work protection") and its explicit instruction: "Do not reuse one immutable snapshot blindly across asynchronous updates."

**Blocking: terminal cleanup reads a snapshot taken before this tick's suppression write**

`projectRunsAfterReconcile` is refreshed before the tracker reconciliation loop. That loop then suppresses a running run whose issue went terminal: it calls `publishAssignedBranchForRun` and saves `lastError: git_transport_failed: …` or `unpublishedWorktree` on the run. Only after that does `cleanupTerminalIssueWorkspace(…, projectRunsAfterReconcile)` run. `unpublishedGitWorkReason` therefore inspects the old `running` record, which has neither marker. The reconciliation path does not mark the workspace record either (`recordGitTransportWorkspaceState` runs only in finalization), so the workspace is removed. On `main`, `unpublishedGitWorkReason` calls `loadAllRuns()`, sees the suppressed record, and defers cleanup.

Repro: I copied the existing test `reconciles running issues that moved to a terminal state outside the candidate snapshot` and changed two things. I added `assignedBranch: "symphony/acme-platform-1"` to the run, and I passed `publishAssignedBranch` to the service with each result below. I ran one tick on this head, then the same test with only `service.ts` swapped to `main` (`8f24f855`):

| publication result | `main` | this head |
|---|---|---|
| `{ ok: false, error: "git push … failed: remote rejected" }` | run `suppressed`, `lastError: git_transport_failed: …`, **workspace `active`, sentinel present** | same run record, **workspace `removed`, sentinel deleted** |
| `ok: true` with `unpublishedWorktreeChanges` (tracked ` M src/a.ts`) | `unpublishedWorktree` recorded, **workspace `active`** | same run record, **workspace `removed`** |
| `ok: true`, clean | workspace removed | workspace removed |

In production, these are the issues someone moved to Done or Closed while the worker still had commits it could not push. The only copy of that work is in the deleted workspace.

**No test pins this.** I passed `undefined` instead of `projectRunsAfterReconcile` to `cleanupTerminalIssueWorkspace`, which restores a fresh read, and all 383 orchestrator tests still pass. Two consequences:
- The fix will not disturb your read-count assertions. The polling-cost fixture never reaches terminal cleanup.
- A regression test is required. Use the table above: a failed publication and a dirty worktree must keep the workspace. Per Step 2.7, show that test failing with the stale snapshot passed.

**Suggested fix:** after the reconciliation loop and before terminal cleanup, call `refreshCurrentRunSnapshot(tenant.projectId, projectRunsAfterReconcile, issueRecords)` again. The stale snapshot still lists the suppressed run as active, so the helper re-reads it. That costs one read per active run and adds no inventory. Alternatively, merge the saved `suppressedRun` records into the snapshot. Either way, name this as a fourth freshness boundary in `docs/architecture.md`.

**Verified and holding**
- `pnpm install --frozen-lockfile && pnpm build && pnpm -r test`: every package passes (orchestrator 383, worker 242, cli 589) except core's known `hooks.test.ts › drains large stdout` 1 s timeout at load ~31. That test passes alone, and this PR does not touch `core`. CI is green.
- Your mutations hold. Swapping the reconciliation inventory back to `loadAllRuns()` fails both scoped-query tests. Bypassing `refreshCurrentRunSnapshot` fails 14 tests, including both polling-cost tests and the mid-tick token visibility test.
- Benchmark reproduced: the polling tick reads 105, 1,005 and 10,005 files at 100, 1,000 and 10,000 history, down from the 50,006 baseline. Elapsed time at 10,000 history is still about 91 s at load ~28. As your report says, reads are not the dominant cost.

**Not addressed, and the PR does not say why:** the steward note on #896 folded in two #977 cleanup points: its extra `loadAllRuns()` scan per terminal cleanup, and the bare `isProcessRunning` stale-PID guard. #977 has not merged, and at its current head it already uses `isRunProcessRunning`. The scan remains a merge-order hazard. Whichever of #977 and #981 lands second must move #977's `liveIssueRun` lookup onto the scoped inventory, and that lookup must use a fresh read for the same reason as the blocker above. Record this in the PR body.

packages/orchestrator/src/service.ts
```diff
@@ -2246,7 +2245,13 @@ export class OrchestratorService {
 
       for (const issue of terminalIssuesByIdentifier.values()) {
         try {
-          await this.cleanupTerminalIssueWorkspace(tenant, issue, now);
+          await this.cleanupTerminalIssueWorkspace(
+            tenant,
+            issue,
+            now,
+            undefined,
+            projectRunsAfterReconcile
```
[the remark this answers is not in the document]
moncher-dev: Fixed in 9d253bd. Reconciliation now refreshes known current runs after suppression persists publication evidence and before terminal cleanup consumes the scoped snapshot. Added failed-publication and dirty-worktree regressions; removing this refresh makes both fail by deleting the sentinel workspace. The benchmark remains one inventory plus bounded reads (10,006 reads at 10,000 history).

moncher-dev:
Steward re-review at `9d253bd5`. **Passes. Requesting review from @hojinzs.**

**The blocking data-loss finding from `52bed14f` is fixed.** Reconciliation now calls `refreshCurrentRunSnapshot` after the suppression loop and before terminal cleanup, so cleanup sees the run records that suppression just saved. `docs/architecture.md` names this as the post-suppression freshness boundary.
- Removing the refresh fails both new rows, `retains terminal issue workspace after 'failed publication' / 'dirty worktree' is recorded during reconciliation`, because the sentinel workspace is deleted. That is the repro from my last comment, now a regression test.
- `pnpm install --frozen-lockfile && pnpm build && pnpm -r test` is green in every package (orchestrator 384, worker 242, core 340, cli 589). CI is green on `Test`, `Container Smoke` and the new `Standalone Docker E2E` (65 s). The PR merges cleanly.
- The polling-cost pin moved from 7 to 8 reads, and the report now says 10,006 reads at 10,000 history. Both match one extra current-run read; there is still no second inventory.
- The earlier mutations still hold: swapping the reconciliation inventory back to `loadAllRuns()` fails the scoped-query tests, and bypassing the current-run refreshes fails the token-visibility and polling-cost tests.
- The PR body now records the merge-order hazard with #977. Whichever lands second keeps #977's `liveIssueRun` lookup on scoped evidence, with a fresh, identity-checked observation of the current run.

**Non-blocking.** The rework converted the old single-case test `reconciles running issues that moved to a terminal state outside the candidate snapshot` into the two retention rows and dropped its clean-publication case, where the workspace is removed after suppression. Nothing is left unguarded as a result: I forced `unpublishedGitWorkReason` to retain every suppressed run's workspace, and two other tests fail (`cleans up terminal issue workspaces during startup…`, `cleans up and releases a terminal retry…`), where `main` fails three. Adding the clean-publication row back to the new `it.each` would restore the reconciliation-path assertion.

packages/orchestrator/src/service.ts
```diff
@@ -2381,6 +2397,42 @@ export class OrchestratorService {
...
+    }
+
+    const refreshedRuns = await Promise.all(
+      [...currentRunIds].map((runId) => this.store.loadRun(runId, projectId))
+    );
+    const runsById = new Map(snapshot.map((run) => [run.runId, run]));
+    for (const run of refreshedRuns) {
+      if (run) {
+        runsById.set(run.runId, run);
+      }
```
hojinzs: **P3 — merge is not identity-checked, asymmetric with `loadRuns`**

`loadRuns` explicitly filters `run.projectId === query.projectId` before returning (`fs-store.ts:284`), but this merge trusts whatever `loadRun(runId, projectId)` hands back and writes it straight into a project-scoped snapshot.

Today this is safe by construction: `loadRun` resolves `runDir(runId, projectId)` and `saveRun` always writes to `runDir(run.runId, run.projectId)`, so the scoped path can only hold this project's runs. The gap is at the contract level — `OrchestratorStateStore.loadRun(runId, projectId?)` declares `projectId` as optional, i.e. a hint rather than a scope, so a different store implementation could legitimately resolve a run ID globally and leak a foreign-project record into `allTenantRuns`. That snapshot feeds `buildProjectSnapshot({ allRuns })` and the cumulative token totals.

One line closes it and matches the steward's own "fresh, identity-checked observation" wording from the #977 merge-order note:

```ts
if (run && run.projectId === projectId) {
  runsById.set(run.runId, run);
}
```

Non-blocking — defense in depth, not a live defect.

---

**Also worth a comment here:** when `loadRun` returns `null` the stale snapshot entry is deliberately retained rather than dropped. That is the right call and it is load-bearing for legacy runs — a legacy record lives under the unscoped `runs/` dir, so `loadRun(runId, projectId)` misses it and the tick-start inventory value (which *does* include legacy runs) correctly survives. Once anything writes to it, `saveRun` migrates it to the scoped path and later boundaries pick it up. That reasoning is not obvious from the code; a sentence in the JSDoc would save the next reader the trip through `fs-store.ts`.

---

packages/orchestrator/src/service.ts
```diff
@@ -2381,6 +2397,42 @@ export class OrchestratorService {
...
+        currentRunIds.add(issueRecord.currentRunId);
+      }
+    }
+    if (currentRunIds.size === 0) {
+      return [...snapshot];
+    }
+
+    const refreshedRuns = await Promise.all(
+      [...currentRunIds].map((runId) => this.store.loadRun(runId, projectId))
+    );
```
hojinzs: **nit — unbounded `Promise.all` in a PR about bounding reads**

Every other multi-record read in the store fans out through `mapWithConcurrency(..., RUN_RECORD_LOAD_CONCURRENCY, ...)` (`fs-store.ts:260`, `fs-store.ts:279`). This one opens every current run at once, and it runs at four boundaries per tick.

I checked the bound and it holds: `currentRunIds` is seeded from active runs in the snapshot plus `issueRecord.currentRunId`, and `releaseIssueOrchestration` nulls `currentRunId` on release (`service.ts:7202`), so the set stays proportional to project concurrency rather than to issue-history size. That is exactly the property this PR is buying, so there is no real fan-out risk here.

Raising it only for consistency — if the set is ever seeded from something less tightly bounded, this is the line that turns into a thundering herd, and reusing the existing concurrency limiter costs nothing.

---

hojinzs (approved):
## Approve

Reviewed at `9d253bd5`. I verified the blocking finding independently rather than taking the steward re-review on trust, and it holds up. Two non-blocking inline comments; neither needs a new round.

### Requirements traceability against #896

| #896 acceptance criterion | Status |
| --- | --- |
| Adopt #895 scoped reads | ✅ all five `loadAllRuns()` service call sites now `loadRuns({ projectId })` |
| Remove redundant reads at *documented* freshness boundaries | ✅ four boundaries, each named in `docs/architecture.md` |
| Same dispatch/retry choices | ✅ boundary 2 sits after the tracker poll — strictly fresher than `main`, which computed `latestRunsByIssueId` from a pre-poll snapshot |
| Cumulative tokens | ✅ historical inventory retained for the tick; token-freshness test pins mid-tick visibility |
| Legacy recovery | ✅ see below |
| Unpublished-work protection | ✅ fixed and now pinned — see below |
| Mid-tick events remain visible | ✅ |
| Benchmark reported against #894 | ✅ full matrix, and the report is honest that reads were never the dominant cost |
| Query ownership documented | ✅ |
| "Do not reuse one immutable snapshot blindly across asynchronous updates" | ✅ this is the explicit design of `refreshCurrentRunSnapshot` |

### The data-loss fix is real and is pinned

I removed the post-suppression `refreshCurrentRunSnapshot()` call and ran the two new rows:

```
× retains terminal issue workspace after 'failed publication' is recorded during reconciliation
× retains terminal issue workspace after 'dirty worktree' is recorded during reconciliation
  ENOENT: .../acme_platform_1-0e0ed77a1c2f4ef5/sentinel.txt
```

The failure mode is the actual symptom, not a proxy assertion: the sentinel file is gone because the workspace was deleted. That is the correct shape for a regression test guarding irreversible work loss.

### The remaining non-blocking point is weaker than reported

The re-review flagged that folding the old single-case test into the two retention rows dropped the clean-publication case, and estimated two other tests cover it. I forced `unpublishedGitWorkReason` to retain unconditionally and got **seven** failures, not two — including `continues terminal cleanup after a state-by-ID refresh finds terminal workspaces` and `cleans up and releases a terminal retry from the single-ID refresh`, both of which drive a full `runOnce()` and therefore exercise the *reconciliation-path* cleanup with the new scoped snapshot, not just startup cleanup. Removal behaviour on the changed path is covered. Adding the clean row back is still tidier, but it is not a coverage hole and I would not hold the PR for it.

### Things I checked that nobody had written down

- **Legacy runs.** `loadRuns({ projectId })` reads the unscoped `runs/` dir *and* the scoped one and filters on `run.projectId`, so legacy records stay in the inventory. `loadRun(runId, projectId)` is path-scoped and returns `null` for them — but `refreshCurrentRunSnapshot` retains the stale entry on `null`, and `saveRun` always writes scoped, so any legacy run that is actually mutated during the tick migrates and *is* picked up by later boundaries. Correct, but only by a chain of three non-local facts; I asked for a JSDoc line inline.
- **Runs created mid-tick.** A run dispatched after boundary 2 is absent from the snapshot, but dispatch upserts `issueRecords` with `currentRunId: run.runId` (`service.ts:1981`) before the final boundary reads it, so it lands in `allTenantRuns`. No gap.
- **Startup cleanup's single `projectRuns` snapshot.** Safe: `cleanupTerminalIssueWorkspace` writes only *workspace* records, never run records, and the startup path never calls `publishAssignedBranchForRun` — so the Codex P1 has no startup-path twin. Worth stating explicitly since the same stale-snapshot shape appears there.
- **`unpublishedGitWorkReason`** still reads workspace markers fresh on every call and only takes the passed snapshot for the run-record path. Good split.

### Scope discipline

No over-engineering: one private helper, no new abstractions, no speculative extension points. Changed files match the issue's Affected Files table plus the benchmark and report the issue explicitly asks for. The "no changeset" call follows the precedent already set by sibling PR #978 in this same epic chain, which merged without one — consistent, not an oversight.

### Local verification

`pnpm build`, `pnpm lint`, `pnpm typecheck` all pass; orchestrator **384/384** green on `9d253bd5`. Mutations above run against a restored-clean tree.

The #977 merge-order hazard is recorded in the PR body as requested. Whichever lands second keeps `liveIssueRun` on scoped evidence with a fresh, identity-checked read.

---

moncher-dev:
## Steward: rework needed — #977 landed first (merge `f97a59f7`)

The merge-order note in this PR body applies: #977 is now on `main`, so this PR is the second to land and must adapt #977's terminal-cleanup lookup. Checked by trial-merging `9d253bd5` onto `main` at `f97a59f7`; GitHub now reports this PR `CONFLICTING`.

### 1. Rebase onto `main`: textual conflict

`packages/orchestrator/src/service.test.ts` conflicts in 11 hunks in the terminal-workspace retention / reconciliation region (about lines 14750–15115 of the merged file). Both PRs edited the same fixtures there. #977 retargeted the `retains an unpublished {transport failure, dirty worktree}…` rows to confirmed state `In Review` and added the two-tick `reconciles a confirmed run whose recorded state does not match…` rows. Keep **both** sides' assertions: #977's rows are its mutation-sensitive guards (see hojinzs's round-6 review), and this PR's post-suppression rows are yours.

### 2. Adapt #977's `liveIssueRun` lookup to this PR's scoped reads (the auto-merge does NOT flag this)

`service.ts` auto-merges cleanly, but the merged `cleanupTerminalIssueWorkspace` still contains #977's global scan:

```ts
const liveIssueRun = (await this.store.loadAllRuns()).find(
  (run) => run.projectId === tenant.projectId && run.issueId === issue.id &&
    run.processId != null && this.isRunProcessRunning(run)
);
```

This PR already passes `projectRuns` (the post-suppression `refreshCurrentRunSnapshot` result) into that same function. Required change:

- Replace `this.store.loadAllRuns()` in that lookup with the scoped evidence: use the `projectRuns` argument when it is provided. When it is absent (the other call sites), use a project-scoped read, never a global scan.
- Keep the identity-checked `isRunProcessRunning(run)` on a **fresh** current-run observation. Do not fall back to a bare PID check or to a stale record.
- Keep the `[workspace-cleanup-deferred] … reason=worker-process-running` early return and its #977 test green.
- Add a guard to the polling-cost / read-count tests so that a terminal-cleanup tick performs no `loadAllRuns()` call. Mutation: reintroduce `loadAllRuns()` in the lookup and confirm the test goes red.
- Update the PR body's merge-order note and the post-merge checklist item, since the reconciliation is now done here.

Then run `pnpm test` (every package), `pnpm lint`, `pnpm typecheck`, `pnpm build`, and the history benchmark. The 10,000-history read count should stay at 10,006.

hojinzs approved `9d253bd5` on the condition that "whichever lands second keeps `liveIssueRun` on scoped evidence with a fresh, identity-checked read". This PR is now second, so that condition is unmet at the current head, and the rework will need a fresh approval at the new head. Please also keep the two non-blocking points from that review: the legacy-run JSDoc line and, optionally, the clean-publication `it.each` row.

Moving #896 to Ready for this rework. The daemon has been restarted on `f97a59f7`, so the rework dispatches on the post-#977 build.

packages/orchestrator/src/service.ts
moncher-dev: Implemented in `b984a88`. Refreshed records are now merged only when `run.projectId === projectId`, and the helper JSDoc explains why a null scoped lookup deliberately retains the legacy inventory entry until migration. Terminal-cleanup liveness uses the same project identity check.

packages/orchestrator/src/service.ts
moncher-dev: Kept the bounded `Promise.all` intentionally. The fan-out remains limited to active run IDs plus current orchestration claims, so it scales with configured project concurrency rather than run history; using the filesystem store's private inventory limiter would require widening an abstraction for no additional bound. The terminal-cleanup path applies the same bounded candidate set, and the 10,000-history benchmark remains at 10,006 reads.

packages/orchestrator/src/service.ts
```diff
@@ -6203,7 +6277,8 @@ export class OrchestratorService {
 
     const unpublishedGitWork = await this.unpublishedGitWorkReason(
       tenant.projectId,
-      issue.id
+      issue.id,
+      projectRuns
```
hojinzs: **P2 — this passes `projectRuns`, not `scopedRuns`, so the retry-release path does the project inventory twice**

Two lines above, `scopedRuns` already resolved the snapshot-or-load. Passing the raw `projectRuns` parameter here discards that work: when the caller supplied no snapshot, `unpublishedGitWorkReason` falls through to its own `loadRuns({ projectId })` at line 6387 and loads the identical inventory a second time in the same call.

`reconcileRun` is that caller — `service.ts:3907` invokes cleanup with three arguments, so both `projectRuns` and `workflowResolution` are `undefined`:

```ts
await this.cleanupTerminalIssueWorkspace(
  tenant,
  retryAction.issue,
  now
);
```

So a terminal retry release costs two full project inventories per affected run, not one. That is still strictly better than `main`, which did `loadAllRuns()` at both points (global, all projects) — this is an un-improved path rather than a regression, which is why it is not blocking. But it is the one remaining place where this PR's own stated invariant ("one history-sized run inventory per project tick") does not hold, and the fix is one word:

```ts
const unpublishedGitWork = await this.unpublishedGitWorkReason(
  tenant.projectId,
  issue.id,
  scopedRuns
);
```

Semantics are identical in both directions: when `projectRuns` was supplied, `scopedRuns === projectRuns`; when it was not, `scopedRuns` is exactly the `loadRuns({ projectId })` result that line 6387 would otherwise fetch for itself. I applied this together with the identity filter below and ran the suite — **237/237 service tests pass**.

---

packages/orchestrator/src/service.ts
```diff
@@ -6308,8 +6384,8 @@ export class OrchestratorService {
       return workspaceReason;
     }
     const latestRun = buildLatestRunMapByIssueId(
-      (await this.store.loadAllRuns()).filter(
-        (run) => run.projectId === projectId && run.issueId === issueId
+      (projectRuns ?? (await this.store.loadRuns({ projectId }))).filter(
+        (run) => run.issueId === issueId
       )
```
hojinzs: **P3 — the `run.projectId` identity check was dropped here relative to `main`**

`main` filtered on both fields (`service.ts:6312` on `f97a59f`):

```ts
(await this.store.loadAllRuns()).filter(
  (run) => run.projectId === projectId && run.issueId === issueId
)
```

The new version keeps only `run.issueId === issueId`, delegating project scoping entirely to whatever produced `projectRuns`. That was a safe simplification when the parameter could only come from `loadRuns({ projectId })` — but it is now asymmetric with the two sibling call sites you hardened in this very commit: `cleanupTerminalIssueWorkspace` still filters `run.projectId === tenant.projectId` (line 6252), and `refreshCurrentRunSnapshot` gained `run.projectId === projectId` in `b984a88` in response to the P3 on the last round. This is the third consumer of the same snapshot and the only one that now trusts it unconditionally.

It is not a live defect — every current producer is project-scoped — which is why it is P3 and not blocking. But `issueId` alone is not a tenant-unique key, and this value feeds `unpublishedGitWorkReason`, i.e. the guard that decides whether an unpublished workspace is irreversibly deleted. Restoring the dropped conjunct keeps all three consumers stating the same invariant:

```ts
(projectRuns ?? (await this.store.loadRuns({ projectId }))).filter(
  (run) => run.projectId === projectId && run.issueId === issueId
)
```

Verified alongside the `scopedRuns` change above: **237/237 service tests pass**.

---

packages/orchestrator/src/service.ts
```diff
@@ -6180,19 +6237,36 @@ export class OrchestratorService {
...
+      (await this.store.loadRuns({ projectId: tenant.projectId }));
+    const liveIssueCandidates = scopedRuns.filter(
       (run) =>
         run.projectId === tenant.projectId &&
         run.issueId === issue.id &&
         run.processId !== null &&
-        run.processId !== undefined &&
-        this.isRunProcessRunning(run)
+        run.processId !== undefined
+    );
```
hojinzs: **P3 — the candidate filter reads `processId` from a possibly-stale snapshot; only the liveness check is fresh**

This is the load-bearing seam of the #977 merge and it deserves a comment, because the split is easy to miss. `#977` applied `processId !== null` **and** `isRunProcessRunning` to the same freshly-scanned record. Here the two halves read different generations: membership in `liveIssueCandidates` is decided from `scopedRuns` (tick snapshot), while `isRunProcessRunning` runs against the re-read record.

That is correct for the transition that matters — a run whose snapshot `processId` is stale-non-null but is now `null` on disk gets refreshed and correctly fails the liveness test. The gap is the opposite direction: a run that is `processId: null` in the snapshot but non-null on disk is dropped *before* the refresh can see it, so cleanup proceeds and deletes the workspace out from under a live worker — precisely the cwd loss #977 exists to prevent.

I convinced myself it is unreachable today, and it rests on a per-call-site argument rather than on anything local:

- **Site 1 (`service.ts:2299`)** — safe by adjacency. The post-suppression `refreshCurrentRunSnapshot` at line 2291 is immediately followed by the cleanup loop with nothing in between, and a run dispatched earlier in the tick is reachable via `issueRecord.currentRunId` (upserted at `service.ts:1981`), so the refresh covers it.
- **Site 2 (`service.ts:2567`, startup)** — safe because startup cleanup never dispatches, so no `null → non-null` transition can occur while the single `projectRuns` snapshot from line 2492 is in use.
- **Site 3 (`service.ts:3907`)** — safe because it passes no snapshot at all, so `scopedRuns` is a fresh load.

`null → non-null` happens only at spawn. So the invariant is: **no call site may dispatch between taking its snapshot and passing it here.** Nothing in the signature enforces that, and a fourth call site that passes a snapshot taken before dispatch reintroduces the #977 bug silently — no test would catch it, since the two retention rows both reach cleanup with a stale-non-null `processId`, the direction that already works.

Non-blocking, and I would not restructure the code for it. A sentence in a comment naming the invariant would do, in the same spirit as the JSDoc you added to `refreshCurrentRunSnapshot` for the legacy-null case.

---

hojinzs (approved):
## Approve

Re-reviewed at `b984a88`. This round covers only the delta since my approval at `9d253bd` — the `origin/main` merge (bringing #977 and #980) plus the P3 fix from the last round. Three non-blocking inline comments; none needs a new round, and I am explicitly not asking for one.

### What actually changed in `b984a88`

The merge commit is not a pure conflict resolution — the commit message names only `service.test.ts`, but the combined diff carries real edits to `service.ts` as well. Both are legitimate and both are the work the last round asked for, but they are worth calling out so they are not mistaken for merge noise:

| Change | Origin |
| --- | --- |
| `refreshCurrentRunSnapshot` merges only when `run.projectId === projectId` | last round's P3 |
| JSDoc explaining why a null scoped lookup retains the legacy inventory entry | last round's P3 |
| Terminal-cleanup liveness rewritten from `loadAllRuns()` to scoped snapshot + per-candidate refresh | #977 reconciliation |
| Test matrix folded: #977's five lifecycle rows + this PR's two retention rows into one table | conflict resolution |

### Previous findings

- **Codex P1** (stale publication evidence before cleanup) — fixed in `9d253bd`, verified by me last round.
- **My P3** (unchecked merge identity) — fixed in `b984a88`, exactly as proposed, and the fix was applied to the new terminal-cleanup path too rather than only at the reported line. Good.
- **My nit** (unbounded `Promise.all`) — declined with reasoning. The reasoning is sound: the fan-out is bounded by project concurrency, and `mapWithConcurrency` is private to the filesystem store, so reusing it would mean widening an abstraction for no additional bound. That is a legitimate non-adoption, not a punt.

### The #977 merge is the real risk, and it holds

`#977` deferred terminal cleanup while a worker process is live so exit transport cannot lose its cwd, and it did so with a *fresh* `loadAllRuns()`. Replacing a fresh global read with a tick snapshot is exactly the kind of substitution that silently reintroduces the bug, so I checked the three transitions rather than the happy path:

- **Stale-non-null → actually dead.** Refreshed per candidate, liveness evaluated on the fresh record. Correct.
- **Legacy / foreign record.** `refreshed?.projectId === tenant.projectId ? refreshed : candidate` falls back to the already-project-filtered snapshot entry, matching `refreshCurrentRunSnapshot`'s null-retention rule. Consistent.
- **Null → non-null (a mid-tick spawn).** The one direction the filter cannot see. Unreachable at all three call sites, but only by a per-site argument — detailed in the inline comment.

I also confirmed the `processId: null` interaction that makes the merged matrix look contradictory at a glance: the `succeeded` row has `processRunning: true` yet still removes the workspace, because the confirmed-Land path nulls `processId` before cleanup, so the candidate filter correctly excludes it. Not a masked deferral.

### Smoke test

No API or UI surface in this change and no preview environment applies, so I exercised it at the library level on a clean checkout of `b984a88`:

- `pnpm build` — pass (full workspace).
- `pnpm --filter @gh-symphony/orchestrator test` — **392/392 pass**, matching the PR body's claim.
- First run failed on 5 files with `Failed to load url @gh-symphony/worker/git-transport`; that is unbuilt `dist/`, not the diff, and it clears after `pnpm build`. Noting it only so the failure signature is on record.

CI on this exact head SHA is green on all three checks, including the ones I cannot run here (Docker is intentionally unavailable on the orchestration host per #872): **Test**, **Container Smoke**, **Standalone Docker E2E**.

### Mutation evidence, verified independently

I did not take the author's mutation claims on trust. Reintroducing the global scan in terminal cleanup:

```
× retains terminal issue workspace after failed publication is recorded during reconciliation
× retains terminal issue workspace after dirty worktree is recorded during reconciliation
  → expect(loadAllRunsSpy).not.toHaveBeenCalled()  //  Number of calls: 1
```

The new `loadAllRunsSpy` assertion is a real guard, not decoration. I also applied both inline suggestions together and re-ran the suite — **237/237 service tests pass** — so neither is a behaviour change, and neither is speculative.

### Requirements traceability against #896

| #896 acceptance criterion | Status |
| --- | --- |
| Adopt #895 scoped reads | ✅ `loadAllRuns()` no longer appears anywhere in `service.ts` |
| Remove redundant reads at *documented* freshness boundaries | ⚠️ four boundaries documented and met; one un-improved path remains (P2 inline) |
| Same dispatch/retry choices | ✅ |
| Cumulative tokens | ✅ historical inventory stable for the tick |
| Legacy recovery | ✅ now explained in JSDoc, per last round |
| Unpublished-work protection | ✅ pinned by the two retention rows, preserved through the #977 merge |
| Mid-tick events remain visible | ✅ |
| Benchmark reported against #894 | ✅ full matrix committed |
| Query ownership documented | ✅ `docs/architecture.md` updated for the scoped fallback and identity-checked liveness |

The ⚠️ is the P2 inline and is a one-word fix, but it is a path `main` handled no better, so it does not gate this PR.

### Scope discipline

Still clean. No new abstractions, one private helper, and the merge resolution folded two test tables together instead of duplicating them. The `retainWorkspace` and `publication` columns added to #977's rows are `null`/`false` throughout the inherited cases, so the merged matrix is a superset rather than a rewrite — #977's assertions are preserved verbatim. The `listIssues` count adjustment (`suppressed && !retainWorkspace ? 2 : 1`) is a genuine consequence of retention short-circuiting the follow-up poll, not a loosened expectation.

Two accuracy nits, neither worth a push on their own: the PR body's benchmark timings (legacy 71,889.65 ms / shared 72,544.59 ms) do not match the committed report's table (74,029.29 / 75,320.60), evidently a later re-run; and the body's "one history-sized run inventory per project tick" is the claim the P2 inline qualifies. Both are descriptive, and the report is honest that reads were never the dominant cost.

The no-changeset call remains consistent with sibling #978 in this epic chain.

---

**Final: Approve.** Everything from the previous two rounds is resolved or reasonably declined, requirements are met, CI and local verification agree, and the #977 merge preserves its fix. The three inline comments are P2/P3 polish — fold them into a later change if convenient, or let them ride; I would not hold the PR for any of them, and this concludes my review.

---

moncher-dev:
## Steward verification at `b984a882`, landed on hojinzs's re-approval

I verified the rework against my earlier send-back items (a)–(d) in a scratch worktree at `b984a882`. hojinzs re-approved this exact commit at 11:01:12Z (review `5209063752`, `commit_id` = head), so I moved #896 to Land. The approval from `9d253bd5` was stale and was not used. Two findings below are about test coverage. Neither is a defect, and neither blocked the Land. I am recording them for a follow-up.

**Build and suite.** `pnpm install --frozen-lockfile && pnpm build && pnpm test` passes in every package: orchestrator 392, cli 589, worker 243, core 340. CI is green on Test, Container Smoke and Standalone Docker E2E.

**(a) No test was dropped.** I diffed the `it`/`it.each`/`describe` titles in `service.test.ts` against both parents. Head has 225 titles, the same number as `f97a59f7`.
- Nothing from `f97a59f7` (#977) is missing.
- Three titles from `9d253bd5` are absent. Each one is accounted for:
  - `preserves a host Git transport failure when the tracker is non-actionable`: #977 replaced it on `main` with `does not let a post-Land host Git transport failure override confirmed completion` and the two-poll variant.
  - `…terminal issue reaches its retry due time`: #977 renamed it to `…non-active issue…`.
  - `retains terminal issue workspace after $name…`: folded into #977's matrix as two explicitly named rows. Every assertion from the old test is kept: `lastError`, sentinel present, workspace `active`, `listIssues` ×1, SIGTERM, `released`. The rows now also assert `runPhase` and a single publication call.
- `fs-store.test.ts` is unchanged apart from the read count (30 titles on each side).

**(b) Global scan removed.** `service.ts` contains no `loadAllRuns()` calls. `main` had 10. Terminal cleanup now takes the scoped snapshot, or makes a scoped `loadRuns({ projectId })` read when no snapshot is passed. It then re-reads each live candidate and evaluates `isRunProcessRunning` on the refreshed record. Apart from the merge itself, the only `service.ts` changes since `9d253bd5` are the P3 project-identity filter, the JSDoc, and this lookup. #977's hunks are intact.

**(c) Read-count guard bites.** Reverting the lookup to #977's `loadAllRuns().find(…)` fails all 7 matrix rows on `expect(loadAllRunsSpy).not.toHaveBeenCalled()`. The PR body says 3 rows fail; I saw 7.

**(d) PR body updated.** It carries the merge-order resolution and the new mutation evidence.

### Mutation results (service.test.ts + fs-store.test.ts, 267 cases)

| Mutation | Result |
|---|---|
| lookup back to global `loadAllRuns()` | 7 red (spy) |
| remove the `worker-process-running` deferral entirely | 1 red (`keeps a confirmed Land workspace while inside the exit grace`) |
| #977 guard: `completedByConfirmedTrackerProgress` reads `activeRun.issueState` | 2 red, including the two-poll `preserves confirmed review state when a later poll moves the tracker to Done` |
| #977 guard: `shouldAwaitTrackerProgressExit` reads `run.issueState` | 1 red (`reconciles a confirmed run whose recorded state does not match…`) |
| **process identity dropped**: `isRunProcessRunning(run)` → `run.processId != null && isProcessRunning(run.processId)` | **0 red** |
| **project identity dropped**: `refreshed ?? candidate` and `if (run)` in `refreshCurrentRunSnapshot` | **0 red** (one unrelated 500 ms `requestReconcile` timeout under host load) |
| **per-candidate fresh read dropped**: use the snapshot record instead of `loadRun` | **0 red** |

### Follow-up: the identity and freshness properties have no tests

hojinzs's approval condition names a "fresh, identity-checked read" in the cleanup lookup. The code does that, but no test pins either property:
- No cleanup-path fixture sets `processIdentity`, so `isRunProcessRunning` always takes its `!run.processIdentity → true` branch. This gap was already on `main` after #977.
- The per-candidate `loadRun` is redundant at every call site the suite exercises.

hojinzs's inline P3 about the `null → non-null` direction makes the same point from the other side. Tests that would close these gaps:
1. PID reuse. A cleanup-path run has `processIdentity: "worker-old"`, `isProcessRunning → true`, and `getProcessStartIdentity → "worker-new"`. The workspace should be removed. Add a matched-identity row, which should defer with `reason=worker-process-running`.
2. Stale snapshot. Pass a `projectRuns` whose run still has `processId: 4205` while the stored run has `processId: null`. Cleanup should proceed.
3. Optionally, a stub `loadRun` that returns a foreign-project record. It should be ignored.

These go together with hojinzs's three non-blocking inlines: the P2 `scopedRuns` pass-through, the P3 `projectId` conjunct in `unpublishedGitWorkReason`, and the invariant comment. hojinzs decides whether they become an issue.

